PR #8, opened 2015-03-02
Morestuff

sydneyjd:
Here are a few tiles. Please merge them, and if we have issues we can swap them out later.
For some reasong it look like it wants to re-merge the icefreezetest level from here, even though i did not change anything in it.


tobbi:
Can you please leave the 'S' sign out while we think of a better replacement for it (as discussed on IRC). Also, you accidentally committed icefreezetest.

sydneyjd:
Well, can we put the S sign in because i really don't want to have to 
undo all that and i darkened the S in it so it doesn't look like blood. 
When we have a better replacement we can swap it out easily, vs having 
to undo that all.
I know i accidentally committed the icefreezetestlevel and i don’t know 
how i did. Can you some how not merge that?

-Sydney

tobbi:
Alright, alright. I'll do that when I'm back home.

sydneyjd:
Im not trying to be hard on this, sorry if i seem rude.

Are you ok with the 2 other signs i did?

tobbi:
Yup, those are fine :)

HybridDog:
you could squash the commits l think

PR #62, opened 2015-04-12
Make it easier to build SuperTux with Visual Studio

Flakebi:
What is needed, is a `dependencies` folder in the root directory. Maybe this can be added to the download repository?
CMake somehow decides to put the binary into a `Debug` or `Release` directory so SuperTux fails to find its data directory, adding a symlink (which requires admin rights on windows...) or copying them works.

I took most of the CMake code and some dependency libraries from @supertuxkart :-)


tobbi:
Can you please squash the commits?

PR #63, opened 2015-04-13
Include AppData file in tarball

amigadave:
The install phase fails if the AppData file is not present.


tobbi:
I will merge your pull request with the 0.3.5 branch, and release an updated version soon.

PR #88, opened 2015-06-14
Updated translations

Savsish:
Changed the egg-0.png to egg.png in the string "Eggs The egg makes Tux grow larger. Tux can then smash wooden blocks with his head."


tobbi:
Remove unrelated commit d2035a7 from this PR, please.

PR #89, opened 2015-06-14
Update translations

Savsish:
Changed the egg-0.png to egg.png in the string "Eggs The egg makes Tux grow larger. Tux can then smash wooden blocks with his head."


tobbi:
Thank you very much!

maxteufel:
In the future, please try to squash all these commits into one, it's just pretty much adding junk to the commit history (`l10n: update all translations` would do as well as 45 times `Update *.po`)

giby:
mmmmm… you skipt Tx for that?

PR #122, opened 2015-08-30
Add git submodule update --init --recursive to INSTALL.md

sydneyjd:


maxteufel:
Submodules **should** be included in the source tarballs (without the `.git` directory, which is removed when building one, git-submodule won't work anyway). The reference to cloning the git repository has been updated in db1dc90f.

tobbi:
If you clone from source, the submodules are not included. @sydneyjd Please update the comment to reflect that?

Karkus476:
This **needs** to be added. I was extremely confused when I could not build from source because of tinygettext. Not easy on beginners ;P

Issue #127, opened 2015-09-06
Level Intro Should Show Relevant Tux Animation


Karkus476:
When you start any level, the Tux animation "small-walk-right" is shown. It would be nice if this animation could be replaced with the walk right animation for the current bonus that Tux has. For example, if you start the level as Fire Tux, the "fire-walk-right" animation should display.
-Karkus

Karkus476:
For clarification, I am talking about the level intro (where the level creator and best statistics are shown).

Karkus476:
I've made a fix for this on my fork, how should I go about submitting my changes?

maxteufel:
[Open a pull request.](https://help.github.com/articles/using-pull-requests/)

PR #128, opened 2015-09-07
Fixed Bug #127 Concerning the Level Introductions.

Karkus476:
The correct Tux animation should now show on the level intro screen.

### DIFF
--- a/src/supertux/game_session.cpp
+++ b/src/supertux/game_session.cpp
@@ -439,7 +439,7 @@ GameSession::setup()
   if ((!levelintro_shown) && (total_stats_to_be_collected > 0)) {
     levelintro_shown = true;
     active = false;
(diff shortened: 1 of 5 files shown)


tobbi:
Not sure if you can do anything about this, but: Tux walks during the jump animation as well (which shouldn't be the case). Try changing this to the jump animation.

Karkus476:
Done

tobbi:
Please check the travis build errors (they're actually warnings with -Weff). Also, I'd rather not make Tux jump in the level intro when pressing space. I'd rather start the level when pressing space.

Karkus476:
Yeah that's probably better

tobbi:
Oh, right, you created another pull request. Closing this.

Karkus476:
Yeah I should've said. I was stupid enough to create this pull request from my master branch :P

PR #133, opened 2015-09-07
Explosion animation fps fix

Savsish:
Explosions now last 1 second again


tobbi:
When did this regress?

Savsish:
There was a change to timing system of Mr Bomb and the timers were dropped, but someone forgot to change the fps of explosion graphics and it broke things (somewhat), so if there are 8 frames in animation, setting the fps to 8 makes the explosions last 1 second again.

tobbi:
Please rebase your branch with latest changes from master.

PR #137, opened 2015-09-08
Explosion animation fps fix

Savsish:
Explosions now last 1 second again


Savsish:
There was a change to behavior of Mr Bomb and the timers were dropped, but someone forgot to change the fps of explosion graphics and it broke things (somewhat), so if there are 8 frames in animation, setting the fps to 8 makes explosions last 1 second again.

PR #144, opened 2015-09-18
Menu item virtualization

Hume2:
Menu item is now a virtual class. There are now multiple class for menu items, one per menu item kind. Each one has separate methods. It's much nicer than all the switches.

There should be no visible changes.


tobbi:
Please check the classes derived from the "MenuItem" base. Methods of classes that are not bases themselves don't need "virtual" in front of their signature.

tobbi:
Second: Please make methods const that don't alter the classes data in any way.

Issue #108, opened 2015-08-14
New home screen - Menu layout


maceleiro:
It would be nice to make a new start screen, like the SuperTuxKart application. Furthermore, it would have to remove the buttons Start game and Levels provided, and merge them in the button Play. By pressing this button, show all worlds. Thanks!

maceleiro:
@tobbi Any news?

The current display is quite ugly.

tobbi:
We're debating it internally, I guess. I *could* possibly do something about this, but I have no idea whether it'd look good or not. Plus, you still should be able to watch the Tux animation in the background. If you can code something together to showcase, we can decide whether it looks good or not.

Hume2:
So what you're suggesting is putting the main world into contrib levels and delete the "start game" button. Am I right? And how will the people recognize which world is main?

MarkkusCraft123:
I think this is somewhat of what he means about the play button.
![this](https://cloud.githubusercontent.com/assets/11138048/9345123/8f1477c6-45d3-11e5-8e42-148ee2b0f0ad.png)
I got kind of lazy after the menu thing. :)

maceleiro:
Exactly! The design that sent @MarkkusCraft123 is perfect! I just want to download levels function directly in the list of add-ons

Hume2:
Now it looks like a short crappy menu...
So when I want to play levels in a levelset, I'd have to press enter once more per level. (Annoying)

MarkkusCraft123:
It may be a tiny bit longer but it is more organized. I'm sure most people won't mind. How about temporarily implement it and see how it goes, but have an option to go back to the classic menus.

Hume2:
It's only my opinion, so we should wait for the other developers to decide that.

MarkkusCraft123:
Yeah, that is probably best.

mahboi:
I believe this is what they mean: https://youtu.be/XW_nLfSbBYk (yes, I wrote this myself)

sydneyjd:
@mahboi Sorry for the late response, but if you would like to polish up that menu then open a pull request, we would love to at least test it out.

Thanks!

Also, @MarkkusCraft123 @maceleiro @Hume2 @tobbi What do you guys think of his video?

Hume2:
Firstly, there's the one-extra-enter annoyonce.
Secoundly, the forest world should be available after the icy island is done.

maceleiro:
I really love the target distribution that appears in the video.

[...]

PR #155, opened 2015-09-23
CMakeLists: exclude test levels from Release builds

maxteufel:
As being mentioned in #108, exclude `test` and `test_old` level directories when built with `CMAKE_BUILD_TYPE=Release`.


tobbi:
Merging, but next time use [ci skip], please. This shouldn't affect compilation at all. :-)

Issue #143, opened 2015-09-13
Be able to throw Mr Bomb


Karkus476:
At the moment you can only drop Mr Bomb. I think it would be better if you could throw him, as dropping a bomb is useless as it just blows you up.

Karkus476:
Interestingly, in the code this is intentionally disabled. See bomb.cpp Bomb::ungrab():
"// portable objects are usually pushed away from Tux when dropped, but we
  // don't want that, so we set the position
  //FIXME: why don't we want that? shouldn't behavior be consistent?"
Do we want it or not.
Note: commenting out the line below has next to no effect.

brmbrmcar:
It would be good to be able to throw the bomb. However, it could annoy users who are used to the old dropping method.

Hume2:
Is there anyone who used carrying bombs frequently?
No, because of this thing.

brmbrmcar:
I don't do it either.

sydneyjd:
I would carry bombs more if i could throw them. i always get killed.

brmbrmcar:
Cheeseman is a modification with interesting bomb-carrying behaviours.

Karkus476:
Assign me to this, I'm gonna try to get it done.

sydneyjd:
@Karkus476 No need to assign anyone, just fix it and open a pull request. We don't have much competion about who fixes stuff around here. :smile:

Karkus476:
Can I change this issue to include frozen enemies? They could slide along the ground...

sydneyjd:
@Karkus476 I would leave the frozen bombs alone. They should not do what icecubes do. ;)

sydneyjd:
Also, i cannot assign people who are not in the Supertux organization to a issue.

sydneyjd:
Merged in #159

tobbi:
@Karkus476 Next time, you can add "closes ##" to the commit message to automatically close this :)

PR #159, opened 2015-09-23
Bombs can now be thrown

Karkus476:
I just copied the gold bomb code over, as gold bombs could already be thrown. Currently the goldbomb and mrbomb are not connected in any way, therefore the code I could have reused was instead copied. It just feels wrong that the gold bomb doesn't inherit the code of the bomb... any thoughts? See supertux/supertux#143


Karkus476:
There is no reason for this code to be any less stable than the code we had before.

PR #167, opened 2015-10-01
Submodule build improvements; add support for out-of-tree builds

maths22:
This pull request improves the build error messages if a submodule is not checked out, and builds tinygettext and squirrel separately, and then links them in to the SuperTux build.  It also fixes the directories used to properly support out-of-tree builds (which should never touch the source directory).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -211,11 +211,27 @@ INCLUDE(ConfigureChecks)
 
 ## Also execute instructions in external/squirrel/CMakeLists.txt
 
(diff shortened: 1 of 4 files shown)


tobbi:
I keep getting this warning with your changes:
  MACOSX_RPATH is not specified for the following targets:

   tinygettext

This warning is for project developers.  Use -Wno-dev to suppress it.

Not sure if there's anything you can do about it.

maths22:
I think that comes a a result of activating the tinygettext CMakeLists.txt. I'll see if I can check this out on an Apple machine later today (I don't have one, but the CS labs at my school do.)  How did the rest of it look?

PR #180, opened 2015-10-10
New Bonus Block sprites

165your4:
Red with semicolon - run a script on hit
Orange - contains badguys

The functionality for these already exists, I'm just giving them sprites.


tobbi:
So, you're saying that these bonus blocks work correctly without any need for C++ code to be added?

In any case: Can you add these bonus blocks to a test level?

Karkus476:
I like the semicolon though.

Karkus476:
This would have to be an object, rather than a block, of course.

maxteufel:
> This would have to be an object, rather than a block, of course.

IIRC there's already script-running and badguy-spawning objects, but with the normal `?` sign.

Karkus476:
@maxteufel Are these used in any levels? Can we confirm this?

Hume2:
Please don't mess up with merges. Do "git pull --rebase" instead.
So basically, when I see an orange bonusblock, I would get rid of him, because it always contains an enemy.

165your4:
@Hume2 Thanks for the advice. I'm still new to GitHub. Also, sometimes enemies can be useful (especially Mr. Iceblocks).
@Tobbi: I've added a test level in the latest commit

Issue #112, opened 2015-08-16
Update the ground blocks in Ghostly World (Forest World)


marc280492:
The current ground is very ugly:

![screenshot011](https://cloud.githubusercontent.com/assets/11487343/9295509/9657c114-4470-11e5-8dfb-204871dea0fb.png)

Could you please tell if this is the web where to translate SuperTux? https://www.transifex.com/arctic-games/supertux/

Thanks! :smile:

tobbi:
Agreed, we definitely need better blocks for this. We could actually need more tiles for Ghost Forest overall.

That above URL is correct, by the way.

Hume2:
I know that I have already failed with it, but I'll try it once again.
Would you be for this? http://supertux.lethargik.org/wiki/File:Ghostwood.png

Karkus476:
@Hume2 can you make a pull request (with [ci skip] in the commit message) so I can pull and see what they look like in-game?

marc280492:
The new design is beautiful! Thanks.

PR #181, opened 2015-10-10
Ghostwood tiles overhaul

Hume2:
Fixing bug #112 .


maxteufel:
Can you put the `[ci skip]` in the third line of the commit message so it's less annoying when reading the log?

maxteufel:
Also, if you add `Fixes #112` (oslt) to the commit message, GitHub will auto-close the issue. (See <https://help.github.com/articles/closing-issues-via-commit-messages/>).

I propose a commit message like this:

```
Ghostwood tiles overhaul (fixes #112)

[ci skip]
```

maxteufel:
LGTM, but looks like it'll break add-on compatibility.

Hume2:
Which levels, exactly?

maxteufel:
Well, why did you change the `ghostly.stl` then? (Sorry, github doesn't diff them because it thinks it's a 3D-model)

Hume2:
I changed it because it was broken.
And is there any other level, that it broke? I don't know any, because these tiles seemed too ugly, so noone used them.

maxteufel:
No I don't know/remember any others using these tiles.

Oh well, and if something breaks, add-on maintainers are responsible for fixing it anyway.

PR #206, opened 2015-10-27
try to add new cloud tile

sydneyjd:
Im stumped. I appear to be doing it properly, but i cannot get the cloud to show up in the editor. Ive never had an issue before. Can you all see what im doing wrong?


Hume2:
I got that trere are three problems:
1) The storm-cloud.png seems to be damaged. Can you open it anyway?
2) The image is 4 tiles width, but you have only 3 tiles here.
3) You have 6 tiles, but only 4 attributes.

sydneyjd:
I will try it and see what i can do.

sydneyjd:
@Hume2 Is that better?

Also, why is it telling me that this branch has conflicts?

I still don't see the clouds in my editor... wonder if its a local issue.

tobbi:
You checked the "Background" category, didn't you? And the file works in an image editor, doesn't it?

sydneyjd:
Yes, i checked both. It opens fine in a image editor, and i did check under tha background category. This makes no sense!

tobbi:
Regarding the capability to merge: Please fetch master and rebase it against origin/master

sydneyjd:
I squashed the last commit and pushed to my master branch... aparently that wasn't right acording to you. So i built ran cmake .. and make and now the clouds show up in the editor... o_O

tobbi:
So, if it works now, I will merge manually...

Mathnerd314:
I'm not sure why we need tiles for this; adding it as a decal should work just fine.

Hume2:
And for the solid cloud: Why is whole cloud solid? Make the bottom half unisolid and the upper half shouldn't be solid at all.
  (tiles
    (width 4)
    (height 2)
    (ids
            0       0       0       0
      3427 3428 3429 3430
    )
    (attributes
       0 0 0 0
       3 3 3 3
    )
    (image "tiles/background/storm-cloud.png")
  )

sydneyjd:
Ok, im working on it.

PR #211, opened 2015-10-28
Add invincibleone() function to Console.

brmbrmcar:
invincible() ran it an infinite number of times. This does it once.


Hume2:
It's good. May I just ask why is it called invincibleone and not make_invincible?

brmbrmcar:
To make it more like the other script. I could make that change though if you think it would be better.

Hume2:
You don't have to change it. I was just wonder.
Anyway, when you make a commit that doesn't change the source code, please put "[ci skip]" to the commit message. This will skip the Travis CI build.

tobbi:
I don't like the name. Why not change the parameters on the invincible function? Or add a paremeter (bool once = false) but can be overridden.

tobbi:
I believe invincible(bool continuous = true) would be good.  If continuous is set to false, this translates to only one invincible thing.

brmbrmcar:
But there is no script for invincible().

brmbrmcar:
But bool continuous =true would work.

brmbrmcar:
Your idea does not work, unfortunately.

Karkus476:
A script which gives tux invincibility and another which toggles it

brmbrmcar:
Error with updating it.

brmbrmcar:
How to fix?

maxteufel:
ref: #220

PR #223, opened 2015-11-10
Configs now per profile

### LINKED ISSUE #184: Keyboard Layout Settings Per Profile
Rather than for all users.
In fact, all settings could be included in this scope.
[This is the config file I am talking about.](http://pastebin.com/5nT7HMDL)

Karkus476:
Fixes #184
Please quickly have a look at the code for any formatting mistakes.
And this needs very thorough checking, as this change is quite big, I guess.
old 'config' file has been moved into profile1/config and will update from old config automatically. It will keep the old one to use as a fix if anything goes wrong or a new profile is created. You should not lose any data. addon-list now holds the addons, because they are not per-profile.

### DIFF
--- a/src/lisp/writer.cpp
+++ b/src/lisp/writer.cpp
@@ -14,10 +14,15 @@
 //  You should have received a copy of the GNU General Public License
 //  along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+#include <string.h>
+#include <stdexcept>
+#include "util/log.hpp"
+
 #include "lisp/writer.hpp"
 
 #include "physfs/ofile_stream.hpp"
 #include "util/log.hpp"
+#include "lisp/lisp.hpp"
 
 namespace lisp {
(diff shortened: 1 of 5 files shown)


brmbrmcar:
Should be Post-0.4.0 (bugs may be present).

Karkus476:
Closing for quick fix

Karkus476:
Please don't try to merge this; it doesn't work!

brmbrmcar:
Also, make a separate file for the Console to be enabled. WE don't want that to be affected.

Karkus476:
We don't? Surely you wanting the console doesn't make your brother or sister or friend want it.

- Karkus

brmbrmcar:
I mean, the keyboard shortcut wouldn't work for it, but it would still be on.

Karkus476:
No longer valid. I will create a new pull request

PR #230, opened 2015-11-17
Added unordered path mode

Hume2:
When running: Goes randomly between the nodes.
"goto_node" causes the platform to go dirrectly to the given node.


src/object/path_walker.cpp
```diff
@@ -108,6 +113,11 @@ PathWalker::advance_node()
   current_node_nr = next_node_nr;
   if (static_cast<int>(current_node_nr) == stop_at_node_nr) running = false;
 
+  if (path->mode == Path::UNORDERED) {
+    next_node_nr = graphicsRandom.rand( path->nodes.size() );
```
Mathnerd314: This should be gameRandom.

Karkus476:
[unordered_path_demo.txt](https://github.com/SuperTux/supertux/files/36946/unordered_path_demo.txt)
Run with ./supertux2 <path to test level> --play-demo <path to the file above>

Hume2:
@Karkus476 I have no idea what the platform should do when the command "goto_node" is typed and the platform is already running. Well, there is one option: refuse the platform to going to another node when it's running. May I apply that?

Karkus476:
@Hume2 It's up to you. Another question, does this 1)randomly go from node to (random) node, or 2) traverse the path in a random manner. It looks like the answer would be 1, but I can't tell.

Hume2:
@Karkus476 Option 1.

Karkus476:
@Hume2 I see. 2 might be more useful, but I see no reason why this cannot be merged once you've made those changes.

PR #242, opened 2015-11-30
Sprites from pre_git_bugtracker_worldmap

Hume2:


tobbi:
Yeah, okay, I guess these can be added to the repository, in order to get the add-on levels working. However, we should possibly encourage others not to use the sprites.

Would anyone know the license these graphics are under? Most likely the same as the levels, right?

Hume2:
My thoughts about these sprites:
* crystals: Looking good, could be used for some random reasons
* snowman: This could be used as a friendly NPC, just could use some cleanup.
* big ball: I guess, this could be implemented as a new object in the future, just could use some cleantup. The red one looks weird.
* thunder: Idea is good, but the sprite needs an overhaul.
* ice square: I guess, this was used just because the 0.3.0 editor didn't allow to resize tilemaps, so they could be substituted by tilemaps.
* antena: Not much fitting to the theme, but OK.
* pipes: They aren't used anyway, so no need for them.

tobbi:
Besides, s_man sprites shouldn't be inside the crystals directory.

PR #247, opened 2015-12-02
Fixed console command 'functions(...)'

joserebelo:
vargc should be vargv.len() in Squirrel 3.0 and obj.parent isn't supported anymore.


brmbrmcar:
Thanks.

brmbrmcar:
Does not work for me.

PR #257, opened 2015-12-09
Fix up some errors I caused

brmbrmcar:
Yep. You'd better execute me.


Karkus476:
I am seeing two different files, how could one be a fix of the other?

brmbrmcar:
I changed thw wrong file before.

Hume2:
Could you all please pay more attention when merging brmbrmcar's pull-requests? There's now the level "Two Tiny Towers" instead of "Tree Fortress". At least brmbrmcar fixed that what he caused, but this would be a terrible bug.

PR #263, opened 2015-12-12
Added warp function to console scripts

pedromlcosta:
Made a console script, similar to existing ones, called warp. It moves Tux X and Y blocks relative to his current position. It's usage, on the console is: warp(offset_x, offset_y)


tobbi:
Works fine. Thanks.

PR #278, opened 2015-12-21
Added tiles for the new worlds

Hume2:
Licence: DO NOT ASK FOR LICENCE, IT'S FREE!

@maxteufel you wanted to make a level contest based on these tiles. I haven't edited them so far, most likely I will in the future. So it's yours.


maxteufel:
Looks good to merge as they won't be used, but please wait with merging unless someone else of @SuperTux/developers approves.

Regarding the contest, let's discuss that on IRC or in a separate issue. However, it'd be nice if you'd keep your branch or in future master updated with your edits, if they're improving the graphics.

That said, it'd be also nice to have the XCF files or whatever program you're using in https://github.com/SuperTux/media.

Karkus476:
@Hume2 Okay, that's a lot of tiles... Maybe we should have some separate tilesets?

maxteufel:
@Karkus476: I'm a bit worried about not being able to mix tiles anymore then.

Hume2:
This is no longer valid.

PR #280, opened 2015-12-21
Changed maximum coins to INT_MAX

brmbrmcar:


Hume2:
Just some numbers:
10 000 coins = 45 lives
100 000 coins = 66 lives
1 000 000 coins = 88 lives
10 000 000 coins = 110 lives
100 000 000 coins = 132 lives
1 000 000 000 coins = 154 lives

INT_MAX coins = 161 lives
10× more coins = 21,854345327 extra lives (considering that the base amount of coins is more than 250)

Considering that dying without the use of reset point doensn't cost any coins, 45 lives is now far enough. Maybe we'll incerese it if we implement the shops, but then we'll set it ourselves to the most fiting value.

brmbrmcar:
I disagree. It should be the highest possible, because it is doing no harm being there and each time the user reaching the maximum, well...

PR #294, opened 2015-12-25
Improved userdir handling; force windows logging to default userdir

maths22:


src/supertux/main.cpp
```diff
@@ -205,18 +205,27 @@ class PhysfsSubsystem
...
+      
+	  std::string physfs_userdir = PHYSFS_getUserDir();
+#ifdef _WIN32
+	  std::string olduserdir = FileSystem::join(physfs_userdir, PACKAGE_NAME);
+#else
+	  std::string olduserdir = FileSystem::join(physfs_userdir, "." PACKAGE_NAME);
+#endif
+	  if (FileSystem::is_directory(userdir)) {
+		  log_info << "Moved old config dir " << olduserdir << " to " << userdir << std::endl;
+		  rename(olduserdir.c_str(), userdir.c_str());
```
maxteufel: What happens when this fails (e.g. no permission, target directory already exists)?

maths22: Are you referring to the rename?  I forgot to add error handling code.
I'll do that shortly.

Hume2:
Shall I rebase this to recent master to check out whether the CI output is still valid?

tobbi:
Please do so.

src/supertux/main.cpp
```diff
@@ -207,18 +208,55 @@ class PhysfsSubsystem
     }
     else
     {
-      std::string physfs_userdir = PHYSFS_getUserDir();
+		userdir = PHYSFS_getPrefDir("SuperTux","supertux2");
```
Hume2: This doesn't work for some reason.

tobbi: This should probably be https://wiki.libsdl.org/SDL_GetPrefPath

maths22: That's right-I now remember why I didn't merge it: the physfs version is
better, but it is not in the latest release (it is only in trunk).  Also,
let's still wait to merge it-the SDL version does not quite do what we want
either.

src/supertux/main.cpp
```diff
@@ -209,18 +210,55 @@ class PhysfsSubsystem
...
+		  boost::filesystem::rename(itr->path().string().c_str(), userpath / itr->path().filename());
+		}
+		catch (const boost::filesystem::filesystem_error& err)
+		{
+		  success = false;
+		  log_warning << "Failed to move contents of config directory: " << err.what();
+		}
+	  }
+	  if (success) {
+	    try
```
maxteufel: Indentation is not consistent with that of the block above.

tobbi:
What's still needed for this despite fixing the obvious code comment?

PR #297, opened 2015-12-27
getting a null string exception when the data dir is unable to be set

devnexen:
tested under OpenBSD, launched supertux executable from the compiled directory, data dir is null as SDL_GetBasePath returns NULL. Small proposition is to set data dir to the current directory in this case.

### DIFF
--- a/src/supertux/main.cpp
+++ b/src/supertux/main.cpp
@@ -160,7 +160,7 @@ public:
     {
       // check if we run from source dir
       char* basepath_c = SDL_GetBasePath();
-      std::string basepath = basepath_c;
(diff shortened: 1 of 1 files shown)


maxteufel:
I wonder if this works on Windows as well. That's why I'm not ok with merging this until that's clear.

Hume2:
I think, this must work. I've seen the code change and I don't see anything that shouldn't work.

tobbi:
According to https://en.wikipedia.org/wiki/Path_(computing)#Representations_of_paths_by_operating_system_and_shell "." works on Unix and Windows shells.

PR #314, opened 2015-12-30
Option to enter ghost mode from cheat menu

Karkus476:
Made a pull request, as I'm not sure if this is wanted or not.

### DIFF
--- a/src/supertux/menu/cheat_menu.cpp
+++ b/src/supertux/menu/cheat_menu.cpp
@@ -22,6 +22,7 @@
 #include "supertux/game_session.hpp"
 #include "supertux/player_status.hpp"
 #include "supertux/sector.hpp"
+#include "scripting/functions.hpp"
 #include "util/gettext.hpp"
 
 CheatMenu::CheatMenu()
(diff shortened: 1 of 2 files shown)


brmbrmcar:
It should be called edit mode. Otherwise people might think its the ghost forest transition.

maxteufel:
@brmbrmcar edit mode is already something different. And people that don't know what it does probably shouldn't have this enabled anyway.

Hume2:
:+1: This could be useful.

brmbrmcar:
What is the difference between ghost and edit mode?

maxteufel:
Ghost mode is one-time, edit mode is valid during the full time you play the level and will automatically put you in ghost mode when you'd die.

brmbrmcar:
Is that preferred over edit mode?

Karkus476:
@maxteufel So, 0.5.0?

maxteufel:
@Karkus476 I think we can actually put this in 0.4.1.

@SuperTux/core, @SuperTux/developers: objections?

LMH0013:
Seems like a logical change, I doubt you'll find any objections.

MiroslavR:
This could be useful for translators who want to test their translations without having to go through the level by normal means.

Grumbel:
Merged.

Issue #303, opened 2015-12-28
Occasional Glitch/Bug when Tux is hurt.


NikkOfTime:
So, I was playing 0.4.0 (Main levels), and I encountered a bug/glitch. I hit an enemy and this happened:
![glitch1](https://cloud.githubusercontent.com/assets/16466020/12022192/6ad47e48-ad42-11e5-8679-b751cf10e315.png)
Enemies still could walk just fine, and fireballs hit the ground just fine too.

Karkus476:
If possible, please provide a [recording]() to showcase the bug, or provide clearer steps to reproduce. (What do you mean by "hit" an enemy?)

LMH0013:
Almost looks like the wrong sprite is being displayed: like Tux is actually small, with the small Tux hitbox but the graphic being displayed is large Tux.  More info is definitely needed.  What powerup did you have prior to being hit?  Were you doing anything unique when it happened (such as backfliping)?  How were you hit?  Did the bug persist, or did it disappear once you moved?  Has the bug reoccurred or was it a one-time deal?  If it does repeat, a recording would answer all of these.  Most likely someone will have to reproduce the bug prior to fixing it.

NikkOfTime:
I had a fire helmet. What I meant by "hit" is got hurt by.

Hume2:
I think, it's the same bug as #321 .

Karkus476:
I agree it's similar, but not similar enough to close without conclusive evidence

- Karkus

LMH0013:
It's only similar in the fact that they both appear to be glitches in which sprite is displayed.  #321 is caused by the collision code still being run after a badguy has been neutralized but before the death animation is complete.  Whatever is happening here is something weird-- at least in the sense that it is difficult to reproduce.  I've had no luck, and would be interested if anyone has seen it other than this one case.

@NikkOfTime am I correct in assuming that Tux's movement was unhindered while half "submerged" in the ice (you could still move normally)?  Also, did the bug ever fix itself such as when you got another fire flower or after finishing/restarting the level?  I'd really be interested if you happened to try to backflip while the bug was occurring, if so how did that pan out?  I have a vague idea what might be happening, but no idea why.

Hume2:
It seems like O11c from the IRC has experianced this or similar bug, http://picpaste.com/supertux-bug-nMsoKZZ7.png .

[...]

PR #332, opened 2016-01-16
Fix a case of grown Tux having smaller height than he appears

MiroslavR:
If you place Tux inside a wall (e.g. via ghost mode) while he is 'grown', he gets hit but retains his 'grown' state (not 100% reproducible, sometimes he just dies immediately), while his height (the value) is lowered to that of small/'nobonus' Tux. This results in behavior seen in #303.

I've found that in this case adjust_height() is called twice for some reason:
* in kill(), before calling set_state(): https://github.com/SuperTux/supertux/blob/c3edde698049d00dad851e62ae67b7108d0976fd/src/object/player.cpp#L1534
* in set_state(): https://github.com/SuperTux/supertux/blob/c3edde698049d00dad851e62ae67b7108d0976fd/src/object/player.cpp#L1138

The first call is superfluous, because set_state() already performs height adjustment via adjust_height(). Now, if adjust_height() finds out that Tux's height would increase AND the new height would cause Tux to be stuck inside a wall, it fails (returns false) without changing his height. If adjust_height() fails in set_state(), Tux's state won't be changed, either.

What happens is that kill() sets Tux's height from big to small, and then set_state() attempts to set Tux's height from small to small. This does not seem like a problem. However, due to float precision issues (arising from RectF not being defined by its height -- the height needs to be calculated), the second call of adjust_height() thinks that Tux's height is to be increased (in my case from 30.7999878 to 30.8), therefore performs stuck check and fails (because obviously I'm in a wall). Therefore, Tux remains in 'grown' state, but his height is small.

I'm not sure whether or how this is related to #303, but there is a possibility this PR could fix it. Also, this PR removes some code that is never executed.

### DIFF
--- a/src/object/player.cpp
+++ b/src/object/player.cpp
@@ -1531,7 +1531,6 @@ Player::kill(bool completely)
       set_bonus(GROWUP_BONUS, true);
     } else if(player_status->bonus == GROWUP_BONUS) {
       safe_timer.start(TUX_SAFE_TIME /* + GROWING_TIME */);
-      adjust_height(SMALL_TUX_HEIGHT);
       duck = false;
       backflipping = false;
       sprite->set_angle(0.0f);
--- a/src/object/player.cpp
+++ b/src/object/player.cpp
@@ -1537,10 +1537,6 @@ Player::kill(bool completely)
       powersprite->set_angle(0.0f);
       lightsprite->set_angle(0.0f);
       set_bonus(NO_BONUS, true);
-    } else if(player_status->bonus == NO_BONUS) {
-      safe_timer.start(TUX_SAFE_TIME);
-      adjust_height(SMALL_TUX_HEIGHT);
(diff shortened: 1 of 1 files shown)


Hume2:
Merged.

MiroslavR:
@Hume2 Sure. Actually I meant to create a separate branch, but forgot to do so. Out of curiosity though, would you mind elaborating on why using my "master" branch is an issue for you guys? Thank you.

Hume2:
@MiroslavR Actually you don't need your own master branch. I have set my master to track origin/master and I deleted the hume2/master. When I want to update my build, I just pull from master and when I want to make a pull request, I branch master and set the branch to track hume2/<branch>. And I don't need to update hume2/master anymore.
It's also because I can't have two branches the same name, even when both track another remote, so I had to put your master another name.

PR #341, opened 2016-01-22
Fix Haywire's exploding state not being removed completely when frozen

MiroslavR:
If you freeze an exploding Haywire and then thaw it, it's expected to return back to its normal state. However, currently this isn't done completely, as the Haywire retains fast speed and exploding animation (but otherwise acts normally and won't explode).


Hume2:
Merged, Thanks!

PR #346, opened 2016-01-27
Fix tiles with images larger than 32x32px not being visible

### LINKED ISSUE #317: Igloo becomes invisible when Tux is a couple of tiles below it
Seen in version 0.4.0:
If Tux is lower than an igloo, the igloo turns invisible. As Tux approaches closer to the igloo, it becomes visible again. I did not have to use the scrolling feature to notice it, but with the scrolling feature it becomes more obvious, of course.
My screen has a resolution of 1920×1080 (but I played in a maximized window which was slightly smaller).

I have seen this in the custom world “The Ice Continent”, level “Always Higher”. Tux has to climb upwards a lot and an igloo is located at the top, so Tux has to approach the igloo from below. That's how I noticed the problem.

MiroslavR:
Fixes #317.

This checks tiles surrounding the draw rect from top and left (as far as defined by EXTENDING_TILES) to see if their image overlaps with the draw rect.

### DIFF
--- a/src/object/tilemap.cpp
+++ b/src/object/tilemap.cpp
@@ -244,6 +244,57 @@ TileMap::draw(DrawingContext& context)
     } /* for (pos y) */
   } /* for (pos x) */
 
+  /* Make sure that tiles with images larger than 32x32 that overlap
+   * the draw rect will be drawn, even if their tile position does
+   * not fall within the draw rect. */
+  static const int EXTENDING_TILES = 32;
+  int ex_left = std::max(0, t_draw_rect.left-EXTENDING_TILES);
+  int ex_top = std::max(0, t_draw_rect.top-EXTENDING_TILES);
+  Vector ex_start = get_tile_position(ex_left, ex_top);
+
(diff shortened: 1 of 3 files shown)


Hume2:
Please don't merge this before we get whether there can't be any better sollution.

tobbi:
I'd actually be in favor of merging this (just to fix the bug). We can look for a better solution any time.

Hume2:
OK, but the conflicts should be resolved.

MiroslavR:
I've resolved the conflicts and added tint to extending tiles.

Hume2:
I'll merge it. I appreciate your git skills.

Hume2:
Merged.

tobbi:
I'm sorry, but I needed to backout the fix. I'm getting this:

Assertion failed: (index < (width * height)), function draw, file /Users/tobiasmarkus/workspace/supertux/src/object/tilemap.cpp, line 259.

I did try it with the level "shaft racing" from Yeti's Revenge add-on. Once you kick the ice block into the wooden blocks on the right, it always crashes with the above assertion.

MiroslavR:
Should be fixed in PR #363.

PR #356, opened 2016-01-31
Create info_block.sprite

brmbrmcar:
Makes this thing usable.

### DIFF
--- a/data/images/objects/bonus_block/info_block.sprite
+++ b/data/images/objects/bonus_block/info_block.sprite
@@ -1,0 +1,6 @@
+(supertux-sprite
+    (action
+      (name "default")
+      (images "info_block.png")
+    )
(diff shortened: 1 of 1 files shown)


Hume2:
Do we still need that thing? Can't we just delete that?

brmbrmcar:
Well, it was added for something...

tobbi:
Are we talking about the telephone? This is ugly, IMO.

brmbrmcar:
I like it.

Hume2:
Yes, this is the telephone.

Hume2:
Refused.

brmbrmcar:
Then it should be deleted as soon as possible.

Hume2:
Now it's deleted.

PR #363, opened 2016-02-08
Fix assertion failures introduced by PR #346

MiroslavR:
Fixes the fix for #317. :-)

Quoting @tobbi:
> I'm sorry, but I needed to backout the fix. I'm getting this:
> 
> Assertion failed: (index < (width * height)), function draw, file /Users/tobiasmarkus/workspace/supertux/src/object/tilemap.cpp, line 259.
> 
> I did try it with the level "shaft racing" from Yeti's Revenge add-on. Once you kick the ice block into the wooden blocks on the right, it always crashes with the above assertion.

The problem is that _TileMap::get_tiles_overlapping_ returns a deformed rectangle (one whose top > bottom or left > right) when a tilemap is out of view. My code relies on the draw rectangle being valid. It didn't cause crashes previously, because the original 2 nested for loops ruled out these deformities. I've added a check, which should have been there before.

Anyway, in this particular case, the assertion failed when the cart (which is a tilemap with a defined path) got out of view.

### DIFF
--- a/src/math/rect.hpp
+++ b/src/math/rect.hpp
@@ -51,6 +51,7 @@ public:
 
   int get_width()  const { return right - left; }
   int get_height() const { return bottom - top; }
+  bool is_valid() const { return right > left && bottom > top; }
 };
 
 #endif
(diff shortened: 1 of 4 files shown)


tobbi:
Thanks, tested and works fine. :+1:

PR #365, opened 2016-02-13
Display localized name instead of English translation

### LINKED ISSUE #291: Language names should be displayed in their language instead of in English
Instead of showing German, English, French, I think it would make more sense to show

Deutsch, English, Français (?) so that people recognize their language.

Also, finding the language selection menu can get tedious when you selected a different language accidentally.

tobbi:
SuperTux changes. Commit hash of tinygettext submodule will update periodically as more translations get added.

Unfortunately, I couldn't test it due to tinygettext refusing to build.

Depends on supertux/tinygettext#3
Fixes #291


tobbi:
I'm going to merge this. The not-displaying chars will need to be dealt with in a followup PR.

PR #367, opened 2016-02-18
New badguy: HangingBall

Hume2:
It's a forest equivalent of stalactite.


Savsish:
Great graphics

tobbi:
What about having a different sound for this instead of the cracking sound used for stalactites?

Hume2:
@tobbi Do I look as someone who can hear properly?

Hume2:
Sound taken from: http://freesound.org/people/juskiddink/sounds/140867/

Hume2:
OK, the sound was Tobbi's joke. So it's a placeholder for now.

maxteufel:
How about <https://freesound.org/people/kvgarlic/sounds/184547/> as sound?

src/badguy/stalactite.cpp
```diff
@@ -36,7 +37,7 @@ Stalactite::Stalactite(const ReaderMapping& lisp) :
 {
   countMe = false;
   set_colgroup_active(COLGROUP_TOUCHABLE);
-  SoundManager::current()->preload("sounds/cracking.wav");
+  SoundManager::current()->preload(get_cracking_sound_path());
   SoundManager::current()->preload("sounds/sizzle.ogg");
   SoundManager::current()->preload("sounds/icecrash.ogg");
```
maxteufel: The two other sounds should also get an option so they can be changed for Hanging Ball.

src/badguy/stalactite.cpp
```diff
@@ -165,4 +166,91 @@ Stalactite::deactivate()
     remove_me();
 }
 
+std::string
+Stalactite::get_cracking_sound_path() const
+{
+  return "sounds/cracking.wav";
+}
+
+HangingBall::HangingBall(const ReaderMapping& reader) :
```
maxteufel: Please use a file per class.

src/badguy/stalactite.hpp
```diff
@@ -50,12 +50,38 @@ class Stalactite : public BadGuy
...
 
+  virtual std::string get_cracking_sound_path() const;
+
 protected:
   Timer timer;
   StalactiteState state;
   Vector shake_delta;
 };
 
+class HangingBall : public Stalactite
```
maxteufel: Please use a file per class.

src/badguy/stalactite.cpp
Hume2: This is already in some other cases. This is not the only one.

src/badguy/stalactite.cpp
Hume2: The icecrash seems reasonable, but the sizzle is generic for melting process.

src/badguy/stalactite.cpp
maxteufel: That does not mean we should continue to do so.

weluvgoatz:
a good concept, but hanging ball is not the most creative thing we could use.

weluvgoatz:
Obviously, we are not going to use this.  Hume2 has not been active for a very long time.  Can someone close this?

PR #378, opened 2016-03-06
Fix #377

### LINKED ISSUE #377: It is possible to finish "knee-deep in the depth" without taking the key
...making all the work useless. ;)

To reproduce easily, just use ghost mode to get to the right, press Action to disable it and jump over the key instead of taking it.

The goal is not even needed, as touching the key will end the level with a backflip. Finishing the level without touching the key does not award the key and the level has to be played again. The other key levels do not even have goals, just this one unnecessarily has. Removing the goal will fix the bug.

brmbrmcar:
This should fix it, but could be improved.


maxteufel:
Please find a more descriptive commit message, one that doesn't rely on GitHub. That means, describe the issue in the commit message.

tobbi:
@brmbrmcar You can change the commit message with git commit --amend

brmbrmcar:
Why does it need amending anyway?

maxteufel:
> Why does it need amending anyway?

Because `commit --amend` is way easier than `rebase --interactive` and rewording.

brmbrmcar:
I meant why the actual comment needed amending, not why commit --amend was needed.

maxteufel:
> Please find a more descriptive commit message, one that doesn't rely on GitHub. That means, describe the issue in the commit message.

Relying on GitHub means that to understand this issue, knowing what #377 is about is important. However, when GitHub would close down services at some point, it's impossible to know what #377 is. As such, please **describe** the **issue** and the **way** you used to **resolve** it in the commit message, following <https://github.com/SuperTux/supertux/blob/master/CONTRIBUTING.md#pull-requests>. This link is also shown when you create a pull request.

brmbrmcar:
Ah, OK.

Hume2:
Merged.
I've edited the commit log instead of you.

PR #383, opened 2016-03-07
Added back "Up we go..." as default message whenever climbing

Savsish:


tobbi:
Please compile-test your changes. This doesn't compile. Also, the "Up we go..." message was removed deliberately.

Issue #305, opened 2015-12-29
Problems with non-ASCII characters


Wuzzy2:
SuperTux 0.4.0 seems to have problems with some non-ASCII characters. It just does not show them correctly.

For example, the German translation of the level name “'... or is it just me?'” uses the German quotation marks („ and “) but in the game those characters look like “T”s.
Note that this is just an example, there are other characters where this is happening as well (e.g. ellipsis  (“…”), n-dash (“–”), m-dash (“—”), etc.)

Maybe is there some problem with the font or somethiing? I have no idea.

Karkus476:
Relevant: #130

Hume2:
Check out whether they use UTF-8 correctly.

Hume2:
It's because those chars aren't in SuperTux fonts.

giby:
… is, 
– and — Never noticed... 
„ and “ I'll cheack :p

maxteufel:
This will be fixed by #386, which is schedule for post-0.5.0 release.

tobbi:
As SuperTux now uses vector fonts, this should be fixed.

Issue #130, opened 2015-09-07
Discussion: Use SDL_ttf for font rendering


tobbi:
I *think* it might be more efficient, and we can ditch a lot of code in the process if we were to ditch our own font rendering code for SDL_ttf. 

I think this would require a rewrite of our rendering code and converting the font files to ttf.

Karkus476:
May it fix https://github.com/SuperTux/supertux/issues/97 
?

tobbi:
It might. It might fix #94 as well.

tobbi:
Karkus476: If you want to be helpful, you may try to get fontforge to load and create ttf files out of the data/engine/fonts directory... I couldn't get it to run on my Mac.

Karkus476:
Yeah, that makes it sound like a pretty good idea. If it's done we need to *update the website* so people know to get that as a dependancy

tobbi:
Yes, but that will be done afterwards. After you created the ttf fonts, I can try to play around with SDL_ttf in order to get things going. I'm not sure how much of a trouble this is.

Karkus476:
Okay, I'll see if I can create the font files

Karkus476:
[Font forge](http://fontforge.github.io/en-US/) can't open the .png or the .stf files.

maxteufel:
> can't open the .png

Well, TTF and all other common font file types (OTF, WOFF) use [vector graphics](https://en.wikipedia.org/wiki/Vector_graphics) to ensure the font looks good even when using a very big size.

> .stf

These files are used by SuperTux's custom font rendering system (and are afaik not used by any other software).

This means one would have to re-draw the letters as vector graphics (or use other open-source fonts). Furthermore, TTF doesn't support (multiple) colors inside one single letter ([OTF does](https://en.wikipedia.org/wiki/OpenType#Color)).

Karkus476:
Yeah I thought TTF might use vector graphics

tobbi:
You can find the glyphs in PNG format here:
data/images/engine/fonts/

Karkus476:
You can't convert png to ttf!

tobbi:
I was hoping that FontForge would be able to open the glyphs, vector-trace them and then create a ttf out of this.

maxteufel:
You could use Inkscape to [trace the PNGs](https://inkscape.org/en/doc/tracing/tutorial-tracing.html) and then try to import it in FontForge.

Karkus476:
[...]

Issue #97, opened 2015-06-27
Arabic Not Displayed Correctly


Fahad-Alsaidi:
After reading in changelog about adding Arabic font to supertux in this commit:
https://github.com/supertux/supertux/commit/54c3cb298de0e1bfce6d512764e08ccfdf72c22f

I tested the dev version in Ubuntu 14.04. what I found is that arabic letters are not connected together and have wrong order (from left to right). They are useless in current status.

tobbi:
Ah, thanks for the bug report. So the order of the letters is incorrect, but the orientation is correct, right?

Fahad-Alsaidi:
the orientation is incorrect also.

tobbi:
So, basically it needs to be mirrored on the x-axis and the order of the letters reversed?

Fahad-Alsaidi:
Yes, but I am not sure if the letter will get connected by these step only. 
However here is a project you may find it interested
https://github.com/lxnt/ex-sdl-freetype-harfbuzz

tobbi:
Can you please check out the latest master, and see whether at least the orientation is correct? I will deal with connecting the letters in a separate commit.

Fahad-Alsaidi:
I didn't install required packages to build it from the master. I have to wait the next daily build for ubuntu to test it. 
If you can just take a screenshot and post here, I can tell you whether it is correct or not.

tobbi:
See http://screencast.com/t/DHZEQAt9

Fahad-Alsaidi:
actually no thing change !

tobbi:
Interesting. apparently, the rtl detection code doesn't work... I'll look into it.

tobbi:
The problem is that we set check the rtl flag globally, that means: If a non-arabic font is loaded after the arabic one, the flag gets overridden.

I hope you're still there to test any changes I make to fix this problem?

Fahad-Alsaidi:
Is there any update regarding this issue?

tobbi:
Yes, the update is that I have "kind of" found the cause. And we are debating switching to a different font because of this. Currently, the font glyphs are inside image files, which is a pain to handle. In order to reverse the order of a string, I tried reversing it using methods from std namespace, which didn't work due to the unicode nature of the text string.

So, I will probably try thinking through SDL_ttf or harfbuzz, as you mentioned above, before the next release.

Fahad-Alsaidi:
[...]

Issue #94, opened 2015-06-24
Improve font loading time, don't load font files multiple times


tobbi:
I checked what fonts are loaded at startup using a debug statement. Result: We're loading each font file multiple times and creating surfaces for each of these. We should probably find a better way to deal with font files...

sydneyjd:
Yes we should, but its not *completely* neccessary.
What we have works. ;)

giby:
Yes, especially if translators of some asian languages become active…

tobbi:
@maxteufel This mostly depended on SDL_ttf, right? So, since we use that as the default now, this can be closed right?

PR #386, opened 2016-03-08
(WIP) Use SDL_ttf for font rendering 

tobbi:
With this PR, our ancient and ugly font rendering code will be replaced with SDL_ttf. 

xref #130 #305 #94 #97


Hume2:
I don't see any text. At least I remember the order of the buttons. However, the size of the menu seems to be correct.

tobbi:
Please make sure that you have SDL_ttf installed. Also, check both renderers please.

Hume2:
It works now, we just need a font that supports latin-extended.

tobbi:
Note: Misrenderings with Japanese font in info boxes are not font problems but po file problems. Maybe, before merging this PR we should update the po files from transifex.

tobbi:
https://github.com/lxnt/ex-sdl-freetype-harfbuzz

tobbi:
Of interest is also: https://github.com/wutipong/drawtext-sdl2-freetype2-harfbuzz (which might seem to be easier)

tobbi:
Japanese font to use: https://osdn.jp/downloads/users/8/8636/genjyuugothic-20150607.7z (Genjyu Gothic Normal).

tobbi:
Since SDL_ttf is a direct wrapper on top of freetype, I could as well copy the code over from the relevant TTF_* methods.

Savsish:
How about fixing Windows MSVC builds

tobbi:
If you're referring to the MSVC builds with fretype and all that: The dependency package needs to be updated, and we will only do that if the modern fonts branch is saved.

maxteufel:
@Savsish Please build the dependencies for Windows yourself, if you want to test this now. Otherwise, you'll have to wait. However, we won't tell you how to compile the dependencies, you can figure that out using Google. In the future, please open a separate issue for such problems.

tobbi:
Video of the current font: https://www.youtube.com/watch?v=0oyzlA7w2I4&feature=youtu.be

PR #390, opened 2016-03-20
Boxes are now Schroedinger's boxes

tobbi:
The boxes now contain cats in an undefined state. Only by hitting the box, Tux can find out whether it's still alive or not (and sometimes the cat is still alive but doesn't say *meow*).


HybridDog:
You could make this an easter-egg which is enabled only on [Cat Day](https://www.bbc.com/news/world-asia-35628239).

PR #397, opened 2016-03-28
Fix Ice in the Hole path being too easy

brmbrmcar:
Yes...


Hume2:
Sorry, but I've already done it a better way.

PR #398, opened 2016-03-28
Add help() function to Console

brmbrmcar:
The content could probably be improved.

### DIFF
--- a/data/scripts/console.nut
+++ b/data/scripts/console.nut
@@ -87,8 +87,22 @@ function none()
 function help()
 {
 	println("This is the console, enabled with developer mode. You can run the same scripts as you can in-level, but make sure to put 'sector.' before a command for a sector command. You can also run scripts defined in scripts/console.nut."); 
+println("Commands:"); 
+println("flip(): Flip the level vertically."); 
+println("finish(): Complete the level."); 
+println("edit(): Place the game into an edit mode, so when Tux dies or reaches a point he will be in ghost mode."); 
+println("play(): Take the game out of the edit mode (will restart the level)."); 
+println("worldmapfinish(): Whilst viewing the worldmap, this command will complete every level in that worldmap.");
+println("grow(): Grows Tux as if he had taken an egg.");
+println("fire(): Makes Tux Fire Tux as if he had taken a fireflower."); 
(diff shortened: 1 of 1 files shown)


Hume2:
It's good, but the help function usually writes a list of all commands those can be used and says what a specific command does.

brmbrmcar:
I will modify.

brmbrmcar:
I believe the console now needs some sort of scrolling thing.

maxteufel:
People who would need this aren't supposed to use the console.

Furthermore, that is not a very complete overview of commands - there's quite a bunch of classes with methods you can find instances of in the console.

brmbrmcar:
> People who would need this aren't supposed to use the console.

I feel that is not true. Most people actually use it to cheat, or try something out. I did it for a long time, not knowing what a lot of the tricks were.

> that is not a very complete overview of commands

I only included working commands, and ones defined in console.nut.

maxteufel:
> Most people actually use it to cheat,

Cheat menu exists for that purpose.

Console is mostly for level designers and developers, both of whom should be able to read the relevant documentation in [the scripting reference](https://github.com/SuperTux/wiki/blob/mediawiki/Scripting_reference.mw) or in the code.

> I only included working commands, and ones defined in console.nut.

`console.nut` only defines some functions that make access to others easier. What also works are functions like `println()`, but there's also several instances of classes in the console with callable methods (see scripting reference linked above) - such as an instance of `Sector` called `sector.settings`, an instance of `Player` called `sector.Tux`, and some more.

brmbrmcar:
> console.nut only defines some functions that make access to others easier. What also works are functions like println(), but there's also several instances of classes in the console with callable methods (see scripting reference linked above) - such as an instance of Sector called sector.settings, an instance of Player called sector.Tux, and some more.

Yes, but if you happened to need the help() you would only need the preset commands anyway, most likely.

maxteufel:
Yeah, but if you need `help()` you don't need the console but the cheat menu.

maxteufel:
In fact, we should focus on making high quality levels that people will be able to play. Even the cheat menu isn't designed for end users. I've yet to see a commercial game saying **CHEAT HERE**.

brmbrmcar:
I guess so... maybe when running the Console on the first run it should be stating that.

PR #404, opened 2016-04-25
New object: Enemy Blocker

Savsish:
It's like Inivisble Wall but the player can pass it.

### DIFF
--- a/src/object/enemy_blocker.cpp
+++ b/src/object/enemy_blocker.cpp
@@ -18,7 +18,7 @@
 
 #include "badguy/badguy.hpp"
 #include "supertux/object_factory.hpp"
-#include "util/reader.hpp"
+#include "util/reader_mapping.hpp"
 
 EnemyBlocker::EnemyBlocker(const ReaderMapping& lisp) :
   MovingSprite(lisp, "images/objects/invisible/invisible.sprite", LAYER_TILES, COLGROUP_STATIC),
--- a/src/object/enemy_blocker.cpp
+++ b/src/object/enemy_blocker.cpp
@@ -1,0 +1,44 @@
+//  SuperTux
+//  Copyright (C) 2007 Kovago Zoltan <DirtY.iCE.hu@gmail.com>
+//
+//  This program is free software: you can redistribute it and/or modify
+//  it under the terms of the GNU General Public License as published by
+//  the Free Software Foundation, either version 3 of the License, or
+//  (at your option) any later version.
+//
+//  This program is distributed in the hope that it will be useful,
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//  GNU General Public License for more details.
+//
+//  You should have received a copy of the GNU General Public License
+//  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#include "object/enemy_blocker.hpp"
+
+#include "badguy/badguy.hpp"
+#include "supertux/object_factory.hpp"
+#include "util/reader.hpp"
+
+EnemyBlocker::EnemyBlocker(const Reader& lisp) :
+  MovingSprite(lisp, "images/objects/invisible/invisible.sprite", LAYER_TILES, COLGROUP_STATIC),
+  physic(),
+  width(32),
+  height(32)
+{
+  lisp.get("width", width);
+  lisp.get("height", height);
+  bbox.set_size(width, height);
+}
+
+HitResponse
+EnemyBlocker::collision(GameObject& other, const CollisionHit& )
+{
+  BadGuy* badguy = dynamic_cast<BadGuy*> (&other);
+  
+  if (badguy == 0)
+    return ABORT_MOVE;
+  return FORCE_MOVE;
+}
+
+/* EOF */
--- a/src/object/enemy_blocker.cpp
+++ b/src/object/enemy_blocker.cpp
@@ -20,7 +20,7 @@
 #include "supertux/object_factory.hpp"
 #include "util/reader.hpp"
 
-EnemyBlocker::EnemyBlocker(const Reader& lisp) :
+EnemyBlocker::EnemyBlocker(const ReaderMapping& lisp) :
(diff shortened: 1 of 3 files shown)


tobbi:
@Savsish Can you squash the commits together so that you only have one commit?

Hume2:
Well, no need to inherit MovingSprite, MovingObject is enough. And also no need to store the width/height separately, when it's stored in the bbox.

src/object/enemy_blocker.cpp
```diff
@@ -0,0 +1,44 @@
+//  SuperTux
+//  Copyright (C) 2007 Kovago Zoltan <DirtY.iCE.hu@gmail.com>
```
maxteufel: There's something wrong with this copyright (same for the header). Please correct that.

Additionally, I'd recommend dedicate such small changes to [Public Domain](https://en.wikipedia.org/wiki/Public_domain) or using a license such as [CC0](https://creativecommons.org/about/cc0/) if you didn't steal the changes from other files (if that's the case, mark the original authors and use GPL, please).

src/object/enemy_blocker.hpp
```diff
@@ -0,0 +1,40 @@
...
+#ifndef HEADER_SUPERTUX_OBJECT_ENEMY_BLOCKER_HPP
+#define HEADER_SUPERTUX_OBJECT_ENEMY_BLOCKER_HPP
+
+#include "object/moving_sprite.hpp"
+#include "supertux/physic.hpp"
+
+class Player;
+
+/** An object that only solid for badguys. */
+class EnemyBlocker : public MovingSprite
```
Hume2: No need to inherit MovingSprite, use MovingObject instead.

src/object/enemy_blocker.hpp
```diff
@@ -0,0 +1,40 @@
...
+{
+public:
+  EnemyBlocker(const ReaderMapping& lisp);
+
+  HitResponse collision(GameObject& other, const CollisionHit& hit);
+
+private:
+  Physic physic;
+  float width, height;
```
Hume2: No need for this, because it's never used.

src/object/enemy_blocker.cpp
```diff
@@ -0,0 +1,44 @@
...
+  lisp.get("height", height);
+  bbox.set_size(width, height);
+}
+
+HitResponse
+EnemyBlocker::collision(GameObject& other, const CollisionHit& )
+{
+  BadGuy* badguy = dynamic_cast<BadGuy*> (&other);
+  
+  if (badguy == 0)
```
Hume2: Please, do this via the function collides(GameObject*)

src/object/enemy_blocker.cpp
```diff
@@ -0,0 +1,44 @@
...
+#include "util/reader_mapping.hpp"
+
+EnemyBlocker::EnemyBlocker(const ReaderMapping& lisp) :
+  MovingSprite(lisp, "images/objects/invisible/invisible.sprite", LAYER_TILES, COLGROUP_STATIC),
+  physic(),
+  width(32),
+  height(32)
+{
+  lisp.get("width", width);
+  lisp.get("height", height);
```
Hume2: "width" and "height" should be local variables there, not class members.

src/object/enemy_blocker.cpp
```diff
@@ -0,0 +1,44 @@
...
+//  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#include "object/enemy_blocker.hpp"
+
+#include "badguy/badguy.hpp"
+#include "supertux/object_factory.hpp"
+#include "util/reader_mapping.hpp"
+
+EnemyBlocker::EnemyBlocker(const ReaderMapping& lisp) :
+  MovingSprite(lisp, "images/objects/invisible/invisible.sprite", LAYER_TILES, COLGROUP_STATIC),
```
Hume2: Do you really need to draw an empty sprite?

src/object/enemy_blocker.hpp
```diff
@@ -0,0 +1,40 @@
+//  SuperTux
+//  Copyright (C) 2007 Kovago Zoltan <DirtY.iCE.hu@gmail.com>
```
Hume2: When you're at it, correct this as well.

src/object/enemy_blocker.hpp
```diff
@@ -0,0 +1,40 @@
...
+/** An object that only solid for badguys. */
+class EnemyBlocker : public MovingSprite
+{
+public:
+  EnemyBlocker(const ReaderMapping& lisp);
+
+  HitResponse collision(GameObject& other, const CollisionHit& hit);
+
+private:
+  Physic physic;
```
Hume2: This variable is never used.

src/object/enemy_blocker.hpp
```diff
@@ -0,0 +1,40 @@
...
+//  You should have received a copy of the GNU General Public License
+//  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#ifndef HEADER_SUPERTUX_OBJECT_ENEMY_BLOCKER_HPP
+#define HEADER_SUPERTUX_OBJECT_ENEMY_BLOCKER_HPP
+
+#include "object/moving_sprite.hpp"
+#include "supertux/physic.hpp"
+
+class Player;
```
Hume2: This is useless.

src/object/enemy_blocker.hpp
```diff
@@ -0,0 +1,40 @@
...
+//  GNU General Public License for more details.
+//
+//  You should have received a copy of the GNU General Public License
+//  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#ifndef HEADER_SUPERTUX_OBJECT_ENEMY_BLOCKER_HPP
+#define HEADER_SUPERTUX_OBJECT_ENEMY_BLOCKER_HPP
+
+#include "object/moving_sprite.hpp"
+#include "supertux/physic.hpp"
```
Hume2: After you delete the variable "Physic", you don't need this include.

tobbi:
@Savsish The idea is not bad. Can you please restore your branch and reopen this PR or open another PR? We just told you some code review steps.

Savsish:
No longer vaild. I will create new pull request.

PR #407, opened 2016-05-02
add custom physfs as dependency.

devnexen:
with cmake 3.5.2 the physics library did not build (with make 3.0.x it does), this tiny change makes sure it does in all cases.


maths22:
I thought I fixed that :(

PR #420, opened 2016-05-14
Make desktop resolution string translatable

sebras:


tobbi:
@SuperTux/core Can this be merged?

src/supertux/console.cpp
```diff
@@ -384,7 +385,7 @@ Console::autocomplete()
   // depending on number of hits, show matches or autocomplete
   if (cmds.empty())
   {
-    m_buffer.addLines("No known command starts with \"" + prefix + "\"");
+    m_buffer.addLines(_("No known command starts with \"") + prefix + _("\""));
```
tobbi: Please use boost::format for this, as seen here:
https://github.com/SuperTux/supertux/blob/ed17b8153a084a1d34845414a0bbc4bf7c669ff8/src/supertux/menu/profile_menu.cpp#L37

(otherwise, translators will have to translate " which is silly).

src/supertux/menu/joystick_menu.cpp
```diff
@@ -187,31 +187,31 @@ JoystickMenu::refresh_menu_item(Controller::Control id)
...
         break;
 
       case SDL_HAT_RIGHT:
-        name = "Hat Right";
+        name = _("Hat Right");
         break;
 
       default:
-        name = "Unknown hat_dir";
+        name = _("Unknown hat_dir");
```
tobbi: I wonder whether this is a user-visible string, after all it has _hat_dir_, which appears to be a variable. Seems weird. Can anyone shed some light onto this?

Mathnerd314: There are other hat directions:
https://www.libsdl.org/tmp/docs-1.2/_s_d_l__joystick_8h.html
But on the other hand, SuperTux doesn't support them:
https://github.com/SuperTux/supertux/blob/ed17b8153a084a1d34845414a0bbc4bf7c669ff8/src/control/joystick_config.cpp#L203
So really this should be an error.

src/supertux/console.cpp
maxteufel: I think we should refrain from translating anything console.

sebras: why is that? is this string not user-visible?

src/supertux/menu/joystick_menu.cpp
sebras: so it should throw some kind of exception. I'll see what I can whip up. what about the other strings above? are they fine as being translatable?

src/supertux/console.cpp
maxteufel: Console output is visible in the debug console and/or stdout. The feature isn't intended to be used by users but for debugging. Thanks in advance for Chinese debug information which no one in the team understands.

sebras: Ah, this is a hidden feature or similar. I get it ok. I'll drop this from the patch. How about the rest, do you feel that is makes sense?

maxteufel: Looks okay, not sure about that hat_dir part though.

uafr: Hidden feature? Not really. You get his text if you try to
tab-complete a non-existing command in the ingame console.
 
    > foo[tab]
    No known command starts with "foo"
    > pla[tab]
    play()
    play_sound()
    play_demo()
    play_music()
    > play

maxteufel: I realize it's used there; however I don't think we need translations in that place. If we do, all other console output (including warnings/errors) should be translated as well.

src/supertux/menu/joystick_menu.cpp
Mathnerd314: Yeah, they should show up in the settings menu, which is user-visible. (I don't have a joystick handy to test though).

maxteufel:
Looks good to merge in general.

I'm concerned that translators won't know what to put for the symbols and variables that are now marked as translatable. This could be solved using comments/translator hints. We could tell `xgettext(1)` to include comments preceding gettext function calls (maybe filter for a special l10n tag), but that is most likely out of the extent for this pull request.

tobbi:
r+

PR #424, opened 2016-05-25
tools: add .editorconfig

maxteufel:
Allows consistent configuration of indentation and encoding rules across
different editors (http://editorconfig.org/).

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,13 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 2
+indent_style = space
+insert_final_newline = true
+max_line_length = 80
+trim_trailing_whitespace = true
+
+[*.md]
+trim_trailing_whitespace = false


maxteufel:
Rebased. Also added `max_line_length = 80` (see also [Google's styleguide](https://google.github.io/styleguide/cppguide.html#Line_Length) for reasons to do this).

brmbrmcar:
I believe that this may scare away new contributors. Perfect indentation and encoding is not necessary.

maxteufel:
@brmbrmcar this is more about making it easier for new contributors. If your editor supports [EditorConfig](http://editorconfig.org/), it will be set up automatically.

Besides that, indentation and encoding are actually necessary (the former for readability and consistency, the latter because you don't want a file with some strange Russian or Asian encoding in your repository).

brmbrmcar:
I don't use a special editor, and I wouldn't, but if it works on those things, then fine. I do disagree about consistency, but you seem to control the place anyway.

maxteufel:
> I don't use a special editor, and I wouldn't, [...]

A reasonable code editor seems to be common even for new contributors (no one wants to edit this with Word, or LibreOffice).

> I do disagree about consistency, but you seem to control the place anyway.

Feel free to tell us about your concerns, especially if you have good arguments why we shouldn't care about code consistency.

Hume2:
I think, this is good enough. What I know sure, it can't mess anything.

brmbrmcar:
I do just use gedit for anything really.

Consistency just holds development back, with more time spent looking through and seeing if it is perfect than implementing the great new feature.

tobbi:
r+, a+

PR #426, opened 2016-05-30
Stop player movement and hide player when door opened

### LINKED ISSUE #229: Stop doors fom blacking out the screen when they aren't being used
It is extremely annoying in some levels where there is a door near a jump, and you cannot see what you are doing, making the jump very difficult.

Karkus476:
We need a door animation with tux walking through the door
fixes #229

### DIFF
--- a/src/object/player.cpp
+++ b/src/object/player.cpp
@@ -408,7 +408,21 @@ Player::update(float elapsed_time)
     }
   }
 
+  // handle door waiting
+  if (door_waiting && !dying) {
+    if (door_timer.check()) {
+      door_timer.stop();
+      this->visible = true;
+      door_waiting = false;
+    } else if (!door_timer.started()) {
+      door_timer.start(DOOR_OPEN_TIME);
+      this->visible = false;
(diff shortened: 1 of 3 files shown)


brmbrmcar:
This is a terrible idea and DOES NOT fix my issue. It would make it worse, by forcing the player to go through, when it is only actually a problem if the player doesn't want to use the door.

Karkus476:
@brmbrmcar Surely if you don't want to go through the door, you don't press up?

brmbrmcar:
No, because you have jump with up!

Karkus476:
@brmbrmcar Why would you want jump with up? You should set the Up key to another key.

brmbrmcar:
Because it becomes a pain then, as you need to use a different key to navigate the worldmap.

Hume2:
Exactly, since we have this "jump with up" option, we should expect that the people may use it.

Karkus476:
@brmbrmcar. That is most certainly a good point.

Karkus476:
I'm going to close this, because this is obviously not a solution

PR #427, opened 2016-05-30
Fix Stop doors from blacking out the screen when they aren't being used

brmbrmcar:
This should do the trick...


tobbi:
IMO, a better solution would be to abort screenfade if tux moves during transition, because we can assume the player wants to abort screenfade.

Hume2:
This doesn't fix it IMO.

maxteufel:
It seems like people don't want this change.

PR #433, opened 2016-06-04
InvisibleWall: inherit from MovingObject, remove unnecessary code

maxteufel:
As per @Hume2's suggestion.


Hume2:
The content is OK, but I didn't tested it. I'd say that it's able to merge, because it doesn't come with any new features.

tobbi:
This can be squashed and merged as well!

PR #436, opened 2016-06-05
SSpiky: be invincible to flaming while asleep

### LINKED ISSUE #315: Sleeping Mr Spikey's need thier own melting sprite
It seems to use the melting sprite for the normal one when melting a sleeping spikey, which makes it look like it has to stand up as it's being melted. I can verify with the one in the beginning of "The Journey Begins"

maxteufel:
This has no effect on frozen SSpikys, since freezing them wakes them up.

Fixes #315


Hume2:
Seems good to merge.

tobbi:
Looks good.

PR #438, opened 2016-06-05
CMakeLists: don't ship Incubator Island in Release builds

maxteufel:
Incubator Island doesn't have a story - which we might want to avoid in the official release. The levels should be either incorporated into a worldmap with an actual story or archived (maybe in a separate worldmap available as unofficial addon).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -722,7 +722,8 @@ IF(CMAKE_BUILD_TYPE MATCHES Release)
   INSTALL(DIRECTORY ${CMAKE_BINARY_DIR}/data/levels
                     DESTINATION ${INSTALL_SUBDIR_SHARE}
                     PATTERN "data/levels/test" EXCLUDE
-                    PATTERN "data/levels/test_old" EXCLUDE)
+                    PATTERN "data/levels/test_old" EXCLUDE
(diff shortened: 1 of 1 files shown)


Hume2:
It seems good, but I haven't tested it.

brmbrmcar:
I don't like the idea. There are much worse unofficial things like the test collection, and even worse that scary automatic console thing that fade over and block the view of other game objects.

maxteufel:
> the test collection

We've already excluded both from the release builds before 0.4.

> that scary automatic console thing that fade over and block the view of other game objects

If you're talking about the debugging console that displays warnings: as far as I know, that only shows up if you're running in developer or debugging mode or have the console already enabled.

maxteufel:
From IRC:

```
18:47:29 <@maxteufel> I suggest not shipping it in 0.5
18:47:37 <@maxteufel> if people complain, we release an addon
```

brmbrmcar:
It seems you are right about the developer text thing. But anyway, it should remain in there as standard. Reason being that there may not be an internet connection, or permission to install the addon, and the levels are to meant to be eventually official.

maxteufel:
```
18:28:23 <@tobbi> can be merged, however, Incubator should become an add-on.
```

PR #441, opened 2016-06-12
AddonManager: make add-on repository URL configurable

maxteufel:


src/supertux/command_line_arguments.cpp
```diff
@@ -341,6 +342,17 @@ CommandLineArguments::parse_args(int argc, char** argv)
     {
       enable_script_debugger = true;
     }
+    else if (arg == "--repository-url")
+    {
+      if (i+1 >= argc)
```
tobbi: nit: i + 1

PR #447, opened 2016-06-19
Dispenser: Unfreeze when squished

### LINKED ISSUE #445: Dispenser Bug
SuperTux 0.4.0 Nightly Build
Windows 7 64-bit
After doing these in order; freezing a working squishable canon, stomping on it and running in to it with the icehat will cause weird behaviors to happen. The behaviors are: Tux cannot jump if he is on the  canon (only move left and right), the canon has the wrong animation and freezing it again will make the canon become unsquishable.

Hume2:
fixes #445

### DIFF
--- a/src/badguy/dispenser.cpp
+++ b/src/badguy/dispenser.cpp
@@ -138,6 +138,10 @@ Dispenser::collision_squished(GameObject& object)
     return false;
   }
 
+  if (frozen) {
+    unfreeze();
+  }
+
   sprite->set_action(dir == LEFT ? "broken-left" : "broken-right");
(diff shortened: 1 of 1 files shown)


maxteufel:
`difficulty:easy` is intended for issues, not for pull requests. Removing.

Shouldn't the frozen dispenser become unsquishable?

Hume2:
No, because then Tux needs a fire flower to kill it. Flowers can't be switched so far, so it might be too hard to get a fire flower.

maxteufel:
Looks good to merge then.

PR #448, opened 2016-06-21
SoundManager: Set up listener correctly

### LINKED ISSUE #274: Positional audio exaggerated
As far as I can tell, sounds are positional, relative to the sound source and the player character position.

However the impression you get from listening is disassociated with what you see. This is best demonstrated when using headphones and melting an enemy while standing on its left side, then walking to the right of the enemy. it moves a sound source - that feels close to the right ear - so that it feels close to the left ear. I tried to illustrate the issue (see attachment).

Demo video: https://youtu.be/3ocUr1Z464Q

Again: this should be tested with headphones.

I'm not familiar with the code but if this could be solved by changing some values/formulas, I might be able to try to come up with a solution if you could point me to the right file(s) and expressions to look for.
hearsee

Hume2:
Set the listener z-pos to -300 and the orientation to {[0,0,1],[0,-1,0]}

fixes #274

I think that the sounds now sound less aggressive than before, but it's just my feeling.

### DIFF
--- a/src/audio/sound_manager.cpp
+++ b/src/audio/sound_manager.cpp
@@ -54,7 +54,7 @@ SoundManager::SoundManager() :
     sound_enabled = true;
     music_enabled = true;
 
-    set_listener_orientation(Vector(0.0f, 0.0f), Vector(0.0f, 0.0f));
+    set_listener_orientation(Vector(0.0f, 0.0f), Vector(0.0f, -1.0f));
   } catch(std::exception& e) {
     if(context != NULL) {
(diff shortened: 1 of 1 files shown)


tobbi:
Not sure I noticed a difference. but lgtm.

Hume2:
I advertise you to use headphones.

tobbi:
Yeah, I did. Doesn't matter. If it's better for our users, then it's okay with me.

maxteufel:
Looks good to merge.

PR #471, opened 2016-06-30
makepot.sh: Look for translator/l10n comments

maxteufel:
Also, make makepot.sh executable.


tobbi:
r+

Hume2:

PR #483, opened 2016-07-11
Added the source for the 'Autumn Leaves' particle system

J-128:
Added:
        /src/object/leaves_particle_system.cpp
        /src/object/leaves_particle_system.hpp

### DIFF
--- a/src/object/leaves_particle_system.cpp
+++ b/src/object/leaves_particle_system.cpp
@@ -1,0 +1,162 @@
+//  SuperTux
+//  Copyright (C) 2006 Matthias Braun <matze@braunis.de>
+//
+//  This program is free software: you can redistribute it and/or modify
+//  it under the terms of the GNU General Public License as published by
+//  the Free Software Foundation, either version 3 of the License, or
+//  (at your option) any later version.
+//
+//  This program is distributed in the hope that it will be useful,
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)


src/object/leaves_particle_system.cpp
```diff
@@ -0,0 +1,162 @@
...
+  parse(reader);
+}
+
+LeavesParticleSystem::~LeavesParticleSystem()
+{
+}
+
+void LeavesParticleSystem::init()
+{
+  leavesimages[0] = Surface::create("images/objects/particles/leaf5.png");
```
tobbi: for(int i = 0; i <= 5; i++)
leavesimages[i] = Surface::create()

maxteufel:
This is basically using the same code as snow particles do, so please deduplicate the code.

J-128:
@maxteufel Well, of course it uses code similar to the snow particles, because I copied the snow particle code, and adapted it to the leaves. But I don't know about de-duplicating it. I don't know how I'd do that. :confused:

maxteufel:
You inherit from a common class. This is a minimum amount of effort for such code changes, I really don't want to accept this without that change. Would be really bad code style otherwise.

Karkus476:
@J-128 Create a class for all particle systems and allow the programmer to create sub classes which define what images are displayed and other variables if they are possible. Then create two of these, one for snow and one for leaves.

J-128:
Closing this for now, not making any headway on this.

Issue #488, opened 2016-07-16
Burned fishes (and possibly snails) re-appear after some time


maxteufel:
**SuperTux version:** master, as of 2016-07-16

##### Expected behavior
Burned fishes and snails don't re-appear.

##### Actual behavior
Burned fishes (and possibly snails) re-appear.

##### Steps to reproduce actual behavior

1. Use an appropriate level (e.g. "Bouncy Coils" for the fishes)
2. Burn some fishes
3. Leave the area so that the place where they were disappears from screen
4. Wait a bit (dunno how long, but about 30 to 60 seconds)
5. Re-enter that area
6. Burned fishes should be visible again

maxteufel:
Working on this now.

tobbi:
Shouldn't the code rather be in one of the cases in Badguy::ignite()?

PR #490, opened 2016-07-17
Fish, Snail: call remove_me() when ignited

maxteufel:
I have no idea if this is in the right place, but it seems to fix the problem (#488).


tobbi:
Shouldn't the code rather be in one of the cases in Badguy::ignite()?

Hume2:
r+

PR #495, opened 2016-07-19
Coin: Change the sound pitch depending on its y-pos

Hume2:
I think, this wasn't such a bad idea. It was already there, just commented.

### DIFF
--- a/src/object/coin.cpp
+++ b/src/object/coin.cpp
@@ -17,6 +17,7 @@
 #include "object/coin.hpp"
 
 #include "audio/sound_manager.hpp"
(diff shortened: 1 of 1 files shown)


maxteufel:
Does this also affect coins in boxes?

Hume2:
No, it doesn't.

maxteufel:
Can that be changed as well?

Hume2:
This would be a bit more difficult.

maxteufel:
Looks good to merge. What I suggested can be added later. But do we want to merge this for 0.5? Isn't it 0.6?

PR #496, opened 2016-07-19
Display the current editor tool at the mouse cursor

### LINKED ISSUE #491: Editor: Show currently selected tool at mouse position
I often find myself confused which tool is currently selected and whether I am in tile select or object select mode.

I think it would be good if we could show the currently selected tool near the mouse pointer.

Hume2:
fixes #491

### DIFF
--- a/src/editor/tool_icon.cpp
+++ b/src/editor/tool_icon.cpp
@@ -56,4 +56,9 @@ ToolIcon::next_mode() {
   }
 }
 
+SurfacePtr
+ToolIcon::get_current_surface() const {
+  return surfaces[mode];
+}
+
 /* EOF */
--- a/src/gui/mousecursor.cpp
+++ b/src/gui/mousecursor.cpp
@@ -31,7 +31,8 @@ MouseCursor::MouseCursor(const std::string& cursor_file,
   m_mid_x(0),
   m_mid_y(0),
   m_state(MC_NORMAL),
-  m_cursor()
+  m_cursor(),
+  m_icon()
 {
   m_cursor.push_back(Surface::create(cursor_file));
   m_cursor.push_back(Surface::create(cursor_click_file));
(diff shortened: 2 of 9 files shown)


src/editor/tool_icon.cpp
```diff
@@ -56,4 +56,9 @@ ToolIcon::next_mode() {
   }
 }
 
+SurfacePtr
```
tobbi: const

Hume2: Done, it's on the line just bellow.

maxteufel:
In #491, @tobbi said he also wanted to see if he's in object select mode, this pull request doesn't really implement that part.

maxteufel:
Looks good to merge once the builds succeed. :+1:

src/gui/mousecursor.hpp
```diff
@@ -58,6 +58,9 @@ class MouseCursor
   /// Draw MouseCursor on screen.
   void draw(DrawingContext& context);
 
+  /// Set the icon
+  void set_icon(SurfacePtr icon_);
```
tobbi: const SurfacePtr& icon_

src/gui/mousecursor.cpp
```diff
@@ -53,6 +54,11 @@ void MouseCursor::set_mid(int x, int y)
   m_mid_y = y;
 }
 
+void MouseCursor::set_icon(SurfacePtr icon_)
```
tobbi: Same here:
const SurfacePtr& icon_

Hume2: This doesn't work, because set_icon(nullptr). SurfacePtr is std::shared_ptr<Surface>, so it is already a pointer.

tobbi:
I think a better position for this is in the following screenshot:
http://screencast.com/t/kQPVBVluf4Zm (red square).

What does the rest of the team think? IMO it's more natural that way.

Rusty-Box:
@tobbi: I think it looks good.

PR #501, opened 2016-07-22
Pass some parameters as const refs.

leper:
Also mark one function const.

The src/gui/item_* changes don't change a lot since the whole vector is still stored in quite a few places. Possibly only storing a reference in those places might be nice (though I'm not sure if that is easily possible).


tobbi:
lgtm. Let's wait for travis and appveyor to complete before merging, though.

PR #502, opened 2016-07-22
Make possible to switch tilesets in editor

Hume2:


src/editor/editor.cpp
```diff
@@ -425,4 +425,18 @@ Editor::delete_markers() {
   inputcenter.delete_markers();
 }
 
+void
+Editor::change_tileset() {
+  tileset = TileManager::current()->get_tileset(level->get_tileset());
+  tileselect.input_type = EditorInputGui::IP_NONE;
+  for(auto& sector : level->sectors) {
```
tobbi: I wonder whether this iterator and the iterator below that also work with const auto&

src/supertux/tile_set.cpp
```diff
@@ -108,4 +113,45 @@ TileSet::add_tile(int id, std::unique_ptr<Tile> tile)
...
+  } else {
+    tile = m_tiles[id].get();
+  }
+
+  if (tile) {
+    tile->load_images();
+    tile->draw(context, pos, z_pos, color);
+  } else if (Editor::is_active()) { // Draw a notile sign
+    context.draw_surface(notile_surface, pos, 0, color, Blend(), z_pos);
+    context.draw_text(Resources::small_font, std::to_string(id),
```
maxteufel: I don't think we should draw the text here. This should rather spit out a `log_warning`.

Hume2: @maxteufel It's for the case when the tileset is changed and there are already some tiles in the tilemap those are defined in the previous tileset but aren't defined in the new tileset. The number is there for the case so the people can distinguish them. A log warning each tick wouldn't do anything useful.

maxteufel: @Hume2 True. I noticed it because I tested using 99999 or so as tile ID which doesn't entirely fit into the tile and looks a bit ugly. Maybe in the future that could be changed, but it's good enough for now, I guess.

tobbi:
Can you please rebase this so I can merge?

PR #486, opened 2016-07-13
Handle string collisions in tinygettext properly

tobbi:
Depends on SuperTux/tinygettext#4

PR #504, opened 2016-07-22
Add addons with precedence to the tinygettext search path.

### LINKED ISSUE #309: Always prefer langpack strings over strings contained in packaged PO files
Langpack strings are most likely more recent than the ones contained in the PO files contained within the game.

However, currently, these are being ignored as it is. We should probably change tinygettext in that aspect.

leper:
Fixes #309, #486.

Depends on SuperTux/tinygettext#5


tobbi:
r+ I tested the change and it works even in those cases where my change didn't work.

maxteufel:
Closed by de93cc819a85d4288851f2b58e5ffc1c040c9dcd.

maxteufel:
Thanks for the pull request, @leper. :+1:

PR #508, opened 2016-07-25
Make possible to resize whole sector with given offset

Hume2:


src/object/tilemap.hpp
```diff
@@ -67,8 +67,9 @@ class TileMap : public GameObject,
   /** resizes the tilemap to a new width and height (tries to not destroy the
    * existing map)
    */
-  void resize(int newwidth, int newheight, int fill_id = 0);
-  void resize(Size newsize);
+  void resize(int newwidth, int newheight, int fill_id = 0,
+              int xoffset = 0, int yoffset = 0);
+  void resize(Size newsize, Size resize_offset);
```
tobbi: Can't the parameters of this be const Size& as well?

src/supertux/sector.hpp
```diff
@@ -195,7 +195,7 @@ class Sector
   /**
    * resize all tilemaps with given size
    */
-  void resize_sector(Size& old_size, Size& new_size);
+  void resize_sector(Size& old_size, Size& new_size, Size& resize_offset);
```
tobbi: Same here: parameters as const references.

src/object/tilemap.cpp
```diff
@@ -209,8 +213,12 @@ ObjectSettings
 TileMap::get_settings() {
   new_size_x = width;
   new_size_y = height;
+  new_offset_x = 0;
```
tobbi: I don't think you want to reset the new_offset_x and new_offset_y values to 0 here. Otherwise they don't stick.

maxteufel:
@Hume2 It needs to be clear what this is doing and how to use it.

tobbi:
I'm postponing this PR from 0.5.0 to 0.5.1, since this needs better UX and other things.

tobbi:
Since there has been no progress on this PR in a long time, I will close it. Please re-open when there's better UX for this lined up.

PR #511, opened 2016-07-28
Reorder layer objects when their object menu is closed

Hume2:


src/editor/layers_gui.cpp
```diff
@@ -255,4 +260,9 @@ EditorLayersGui::get_layer_pos(const Vector& coords){
   return (coords.x - Xpos - sector_text_width) / 35;
 }
 
+bool
+EditorLayersGui::less_z_pos(const std::unique_ptr<LayerIcon>& lhs, const std::unique_ptr<LayerIcon>& rhs) {
```
tobbi: This can be marked const afaict.

Hume2: @tobbi Static functions can't be marked const.

tobbi: Right, didn't see it was static. It's alright then.

tobbi:
lgtm

tobbi:
Actually, now that I think about it:
Can we reverse the order of the layers somehow. I think the minus layers should be left and the plus layers should be right.

PR #513, opened 2016-07-29
ParticleSystem_Interactive: Inherit ParticleSystem

Hume2:
instead of GameObject

### DIFF
--- a/src/object/particlesystem_interactive.cpp
+++ b/src/object/particlesystem_interactive.cpp
@@ -17,18 +17,18 @@
 #include "object/particlesystem_interactive.hpp"
 
 #include "math/aatriangle.hpp"
+#include "math/vector.hpp"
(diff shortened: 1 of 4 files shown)


src/object/particlesystem_interactive.cpp
```diff
@@ -17,18 +17,20 @@
...
 
 #include "math/aatriangle.hpp"
+#include "math/vector.hpp"
 #include "object/tilemap.hpp"
+#include "supertux/game_object.hpp"
 #include "supertux/collision.hpp"
+#include "supertux/sector.hpp"
 #include "supertux/tile.hpp"
+#include "util/reader.hpp"
+#include "video/surface_ptr.hpp"
```
tobbi: Does it need all those includes?

PR #520, opened 2016-08-01
More arcade-looking Yeti death sequence

pelya:
I didn't like Yeti's face stretched to whole screen after you defeat it, so this animation looks more like platform games used to be.
https://www.youtube.com/watch?v=A__k5N5oayg#t=54s

### DIFF
--- a/src/object/endsequence_fireworks.cpp
+++ b/src/object/endsequence_fireworks.cpp
@@ -23,8 +23,7 @@
 
 EndSequenceFireworks::EndSequenceFireworks() :
   EndSequence(),
-  endsequence_timer(),
-  bosskillsequence_timer()
+  endsequence_timer()
 {
 }
 
@@ -41,22 +40,13 @@ void
 EndSequenceFireworks::starting()
 {
   EndSequence::starting();
-  // Delay fireworks a bit, to see the boss falling off the screen
-  bosskillsequence_timer.start(3.0f * ScreenManager::current()->get_speed());
+  endsequence_timer.start(7.3f * ScreenManager::current()->get_speed());
+  Sector::current()->add_object(std::make_shared<Fireworks>());
 }
 
 void
 EndSequenceFireworks::running(float elapsed_time)
 {
-  if (bosskillsequence_timer.check()) {
-    if (!endsequence_timer.started() && !isdone) {
-      endsequence_timer.start(7.3f * ScreenManager::current()->get_speed());
-      Sector::current()->add_object(std::make_shared<Fireworks>());
-    }
-  } else {
-    return;
-  }
-
   EndSequence::running(elapsed_time);
   //Player& tux = *Sector::current()->player;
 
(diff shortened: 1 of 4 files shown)


Rusty-Box:
This is actually a nice idea.

pelya:
I wanted to do ground shaking and snowball 'explosions' flying out of Yeti,
but my animation skills are limited.

Or better, Tux smiting Yeti with Final Fantasy - style Ultimate Ice
Breaking Heaven Strike Of Eternity attack.

src/object/endsequence_fireworks.cpp
```diff
@@ -40,13 +41,22 @@ void
 EndSequenceFireworks::starting()
 {
   EndSequence::starting();
-  endsequence_timer.start(7.3f * ScreenManager::current()->get_speed());
-  Sector::current()->add_object(std::make_shared<Fireworks>());
+  // Delay fireworks a bit, to see the boss falling off the screen
```
maxteufel: I'm not happy with this solution. Fireworks might be used elsewhere too. What about a separate sequence for the bosses that will trigger the fireworks sequence automatically?

pelya:
I've reverted changes to Fireworks class, and added some snowballs flying out of Yeti as it is dying.

pelya:
And a new video: https://www.youtube.com/watch?v=cI2ip8l9_7g#t=27s

Rusty-Box:
I'm not so sure about the snowballs. The are too big, imo.

pelya:
Shooting out smaller fireballs would be okay for some other boss, but it
goes contrary to Yeti's snowy appearance.

What if I'll re-color the fireball image into blue, what do you think?

On Aug 3, 2016 12:11 AM, "Carsten Wirtz" <notifications@github.com> wrote:

I'm not so sure about the snowballs. The are too big, imo.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/SuperTux/supertux/pull/520#issuecomment-237045397>, or mute
the thread

pelya:
Here's how it looks with ice bullets instead of snowballs:
https://www.youtube.com/watch?v=wx8YL9YIrGk#t=41s

Rusty-Box:
This looks much better.

pelya:
Yay. Do you want me to squish all changes into a single commit?

maxteufel:
@pelya Yes, that would be good.

In my opinion it looks better than the current animation. Note that as this is a larger change, I'd like to hear more feedback from @SuperTux/core, @SuperTux/developers, and @SuperTux/level-designers. Additionally, this will be a feature for 0.6.0 as it's a larger change and currently we're only accepting bugfixes.

maxteufel:
Why did you close this?

pelya:
Because Github won't allow me to force-update a branch while keeping the same pull request. So I've tried to delete and re-create the branch to update the request, but Github just closed the request instead. See https://github.com/isaacs/github/issues/18

New pull request with all commits squashed is here: https://github.com/SuperTux/supertux/pull/524

maxteufel:
Git won't allow you to overwrite history of branches by default. You need to force-push it (`git push -f <remote> <branch>`). This works perfectly fine with pull requests, I often do that. Not that important, but remember it next time!

pelya:
I did not know about this, thanks.

Issue #281, opened 2015-12-21
Add more unit tests


maxteufel:
Add more unit tests to the `tests/` directory, like for the game engine, physics, badguy and object behavior, scripting, and much more...

PR #531, opened 2016-08-06
Add test for collision.cpp

mandaltapesh:
This test is for the intersects function that lies in supertux/collision.cpp
There are 5 testcases. 2 each for the conditions(seperated by ||) in each of
the if statements and one for the default case i.e. if it doesn't satisfy any
of the conditions. In default case it returns true else if it satisfies the
conditions in if statements it returns false.

Ref: #281


tobbi:
In any case: lgtm. So, let's merge it after the release.

maxteufel:
Thanks for the pull request! :+1:

PR #532, opened 2016-08-08
Scripting: new function set_next_worldmap for transitioning between worldmaps

pelya:


tobbi:
For code style:
const std::string& param_name

Keep the & to the param type, not the name.

tobbi:
Also, please take a look at the following efficiency warning: https://travis-ci.org/SuperTux/supertux/jobs/150751917#L1994

PR #457, opened 2016-06-25
New object: Enemy Blocker

Savsish:
It's like Inivisble Wall but the player can pass it. Original code by Kovago Zoltan.
Don't forget to merge also the branch in data submodule.
SuperTux/data#15

PR #536, opened 2016-08-09
New Object: Enemy Blocker (Updated)

Karkus476:
Blocks badguys, like invisible walls, but not Tux.
Original code by **Kovago Zoltan**. *(who?)*
Submitted by @Savsish originally in #457.
I have rebased his commits for him, and pulled from my fork.

**Disclaimer:** *I haven't tested this*, and this pull request is only to secure the code changes Savs submitted, in case anything happens at his end. Furthermore, he appears to have been banned from this repository, so he cannot contribute further.

@Hume2 You commented on his commit before. Please could you do the same for this one here, and if you wish to fix it, I can push my branch to the SuperTux repo, so that you can do so.

Please see SuperTux/data#15

Edit by @Karkus476: 9th August 2016: Correct mentions and emboldening/italics

### DIFF
--- a/src/object/enemy_blocker.cpp
+++ b/src/object/enemy_blocker.cpp
@@ -1,0 +1,80 @@
+//  SuperTux
+//  Copyright (C) 2007 Kovago Zoltan <DirtY.iCE.hu@gmail.com>
+//
+//  This program is free software: you can redistribute it and/or modify
+//  it under the terms of the GNU General Public License as published by
+//  the Free Software Foundation, either version 3 of the License, or
+//  (at your option) any later version.
+//
+//  This program is distributed in the hope that it will be useful,
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//  GNU General Public License for more details.
+//
+//  You should have received a copy of the GNU General Public License
+//  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#include "object/enemy_blocker.hpp"
+
+#include "editor/editor.hpp"
+#include "badguy/badguy.hpp"
+#include "supertux/object_factory.hpp"
+#include "util/gettext.hpp"
+#include "util/reader_mapping.hpp"
+#include "video/drawing_context.hpp"
+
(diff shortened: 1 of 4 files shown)


src/object/enemy_blocker.cpp
```diff
@@ -0,0 +1,80 @@
...
+EnemyBlocker::after_editor_set() {
+  bbox.set_size(width, height);
+}
+
+HitResponse
+EnemyBlocker::collision(GameObject& other, const CollisionHit& )
+{
+  BadGuy* badguy = dynamic_cast<BadGuy*> (&other);
+  
+  if (badguy == 0)
```
Hume2: This should be done via the function collides. Is it so hard to get it?

maxteufel:
Can this class an the InvisibleWall class be generalized to a common Blocker class from which both classes will inherit then? That could save us some lines of code.

maxteufel:
@SuperTux/level-designers is this feature useful for levels at all?

Rusty-Box:
@maxteufel I would say no and I'm not sure why we would need an enemy blocker at all.

maxteufel:
I suggest we close this then.

maxteufel:
If no one objects to closing this because this seems to be an unnecessary feature, I will close this tomorrow.

qwertychouskie:
I think this might be useful for some levels, currently the only enemy that can go back and forth in a fixed range is the [Crystallo](http://supertux.lethargik.org/wiki/Crystallo).

maxteufel:
10:53:22 <+mteufel[m]> RustyBox: do you think https://github.com/SuperTux/supertux/pull/536#issuecomment-244474582 is a valid reason to add enemy blocker?
10:54:50 <+mteufel[m]> I'm not exactly comfortable with enemy blocker if you say you don't need it. It also just adds more code to maintain. And I don't think that trading that off for add-on designers is a good idea.
10:55:09 <+mteufel[m]> Plus, it might confuse users
10:58:42 <RustyBox[m]> Max Teufel: I agree. The enemy blocker might be confusing, because for example: you would never expect a mr.snowball to turn around without running against a wall.
10:59:16 <+mteufel[m]> Closing then.
10:59:40 <RustyBox[m]> I would say so

PR #546, opened 2016-08-13
Editor: Add a full-brightness option

Hume2:
so the people can actually see how the dark levels look like.

### DIFF
--- a/src/video/drawing_context.cpp
+++ b/src/video/drawing_context.cpp
@@ -31,6 +31,8 @@
 #include "video/texture_manager.hpp"
 #include "video/video_system.hpp"
 
+bool DrawingContext::render_lighting = false;
+
 DrawingContext::DrawingContext(VideoSystem& video_system_) :
   video_system(video_system_),
   transformstack(),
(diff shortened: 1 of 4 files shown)


maxteufel:
Clarify the name of this. What does "Full Brightness" do? If you edit a level without darkness, you'd never notice a difference.

Hume2:
@maxteufel Full brightness means that you have something like night-vision, the lighting just isn't rendered. It doesn't save into the level in any way. When there is no darkness in the level, it doesn't do anything.

maxteufel:
@Hume2 yes. The option could be confusing though if one doesn't know what it does.

Also: make sure lightmap tiles are still rendered/displayed in some way.

src/video/drawing_context.cpp
```diff
@@ -373,11 +375,13 @@ DrawingContext::do_drawing()
     handle_drawing_requests(lightmap_requests);
     lightmap.end_draw();
 
-    DrawingRequest* request = new(obst) DrawingRequest();
-    request->target = NORMAL;
-    request->type = DRAW_LIGHTMAP;
-    request->layer = LAYER_HUD - 1;
-    drawing_requests.push_back(request);
+    if (!full_brightness) {
+      DrawingRequest* request = new(obst) DrawingRequest();
```
tobbi: While we're at it:
auto request = ...

C'mon, show some respect for C++ 11. :P

tobbi:
Anyone can think of a better name for this?

maxteufel:
"Lighting" as a checkbox? @tobbi @Hume2

qwertychouskie:
I like "Render lighting".

tobbi:
Render lighting sounds good. What's a good German translation for this? "Lichteffekte rendern"?

maxteufel:
@Hume2 Please change this to use "Render Lighting". If that's done, please remove your assignment.

src/video/drawing_context.cpp.orig
```diff
@@ -0,0 +1,593 @@
+//  SuperTux
```
maxteufel: I doubt you wanted to get this file into the repository.

PR #560, opened 2016-08-20
Better control when changing direction mid-jump

wansti:
Second attempt from a fresh fork


maxteufel:
Looks good to merge. :+1: But this should not be merged before the 0.5.0 release.

PR #572, opened 2016-08-28
Fix levelset / worldmap creation

### LINKED ISSUE #503: Editor: Editing worldmap should be possible from level select's level subset properties
Currently, you can only edit the worldmap if a level was loaded previously. This doesn't make sense. You should be able to edit a worldmap even from the Level subset properties.

tobbi:
This fixes #503 among other things, and removes some performance inefficiencies in the code, for example we don't need to load a savegame when we just want to display the levels.

### DIFF
--- a/src/supertux/menu/contrib_menu.cpp
+++ b/src/supertux/menu/contrib_menu.cpp
@@ -84,7 +84,8 @@ ContribMenu::ContribMenu() :
   {
     try
     {
-      auto levelset = std::unique_ptr<Levelset>(new Levelset(*it));
+      auto levelset =
+        std::unique_ptr<Levelset>(new Levelset(*it, /* recursively = */ true));
       if (levelset->get_num_levels() == 0)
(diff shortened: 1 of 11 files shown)


tobbi:
Please test thoroughly everyone!

src/supertux/menu/contrib_menu.cpp
```diff
@@ -84,7 +84,8 @@ ContribMenu::ContribMenu() :
   {
     try
     {
-      auto levelset = std::unique_ptr<Levelset>(new Levelset(*it));
+      auto levelset =
+        std::unique_ptr<Levelset>(new Levelset(*it, /* recursively = */ true));
```
christ2go: Shoudl the `/* recursively = */`be here?

tobbi: Yes, that's to indicate what the extra parameter means.

Issue #577, opened 2016-08-29
Editor: changing a worldmap's tileset does not work


maxteufel:
**SuperTux version:** master at the time of opening this issue

##### Expected behavior

Changing a worldmap's tileset works

##### Actual behavior

When entering the "Tile set" dialog, it displays nothing and immediate exits back to the origin dialog.

##### Steps to reproduce actual behavior

Try to change a worldmap's tile set.

maxteufel:
c8c330a93b89f0bf3b67bb3cd9138004aa7a3435 is the first bad commit.

maxteufel:
That commit changes a call from `add_back(...)` to an `add_entry(-1, ...)` where `-1` is the `id`. Then it adds a check that assumes `-1` is the id for clicking OK. However, src/gui/menu.cpp tells us other calls use `id = -1` too (in this case, `add_file(...)`).

PR #578, opened 2016-08-29
EditorLevelMenu: use -2 as value to check for OK

maxteufel:
This is an attempt to fix an issue where certain values could not be changed in
the level menu. It happened because c8c330a93b89f0bf3b67bb3cd9138004aa7a3435
introduced a check for certain values upon clicking "OK". However, it did that
by checking for clicks on menu items with id = -1, which was also used for
certain other menu items. That caused the pop up menu of those menu items to
disappear immediately upon attempting to enter them.

Closes SuperTux/supertux#577 ("Editor: changing a worldmap's tileset does not
work")

### DIFF
--- a/src/supertux/menu/editor_level_menu.cpp
+++ b/src/supertux/menu/editor_level_menu.cpp
@@ -48,7 +48,7 @@ EditorLevelMenu::EditorLevelMenu() :
   }
 
   add_hl();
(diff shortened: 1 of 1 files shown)


tobbi:
Looks good to me. I wasn't aware of the ambiguity.

PR #580, opened 2016-08-29
WeakBlock: Straw: play sound/fire.ogg when burning

maxteufel:
Change the behavior of the straw-type WeakBlock so that when it burns, it will
play the same sound also used for burning BadGuys. The sound may be changed to
use an own version of a burning sound, but it works for now.

Requested-by: @SuperTux/level-designers


maxteufel:
@SuperTux/core, @SuperTux/developers: Can someone review this please?

@SuperTux/level-designers: You can also review this, and maybe suggest a different (free) sound.

Issue #333, opened 2016-01-17
Make icecube eyes blink a moment before waking up


Grumbel:
Make icecube eyes blink a moment before waking up

maxteufel:
This is a feature that's been assigned to a bugfix release. Postpone for now.

maxteufel:
@Grumbel code exists now in pull request #581 (might need some tweaking for details, though), graphics are needed now.

PR #581, opened 2016-08-30
MrIceBlock: play a waking animation before turning normal

maxteufel:
Instead of switching directly from ICESTATE_FLAT to ICESTATE_NORMAL when
flat_timer has expired, we now go to an ICESTATE_WAKING state. When entering the
ICESTATE_WAKING state, the sprite action is set to "waking-left" or
"waking-right", depending on the direction of MrIceBlock. After the animation is
finished, we go to the ICESTATE_NORMAL state.

This might still need some tweaking with regards to the attributes of the waking
state (should it still be possible to kick it? will it hurt the player?) and
the animation loop.

Furthermore, it should be mentioned that this will break custom sprites that do
not have animations for "waking-left" and/or "waking-right" defined. If that is
the case, sprite->animation_done() will always return false, causing the
MrIceBlock object to stay in the ICESTATE_WAKING state forever. Graphics have
been added and animations defined for the standard ice block as well as its pink
version. Those graphics were contributed by BlasterMaster.

Link: <http://forum.freegamedev.net/viewtopic.php?f=68&t=7277>
Closes SuperTux/supertux#333 ("Make icecube eyes blink a moment before waking
up")


Hume2:

maxteufel:
Graphics are in place for this now.

PR #584, opened 2016-08-31
Add collect-script property for coins

tobbi:
Allows to run a script when a coin is collected.

Useful for things like the Yeti's Revenge levels where usually a script trigger was placed over the coin to achieve the required effect.

### DIFF
--- a/src/object/coin.cpp
+++ b/src/object/coin.cpp
@@ -34,7 +34,8 @@ Coin::Coin(const Vector& pos)
     offset(),
     from_tilemap(false),
     add_path(false),
-    physic()
+    physic(),
+    collect_script()
 {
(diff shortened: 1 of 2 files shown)


tobbi:
Depends on SuperTux/data#70

Hume2:

src/object/coin.cpp
```diff
@@ -74,6 +77,8 @@ Coin::Coin(const ReaderMapping& reader)
     set_pos(v);
   }
 
+  if(!reader.get("collect-script", collect_script)) collect_script = "";
```
tobbi: If I don't do this, then coverity will complain about unchecked return.

PR #595, opened 2016-09-15
Fix Squirrel Level Edit Bug

### LINKED ISSUE #207: Squirrel Level.edit bug
Running Level.edit(true); then Level.edit(false); closes the game.

christ2go:
Using the function Level.edit(false) through a switch when in
edit mode currently causes the game to crash. This is due to
the current Level object being destroyed when the edit mode is
left, but squirrel is still referencing / trying to use it, and
this leads to the seg fault.

The Fix is, that the GameSession object keeps a reference to the
last delteted level, as an extra member variable, thus squirrel
can still access the level object.

Fixes #207

### DIFF
--- a/src/supertux/game_session.cpp
+++ b/src/supertux/game_session.cpp
@@ -56,6 +56,7 @@
 GameSession::GameSession(const std::string& levelfile_, Savegame& savegame, Statistics* statistics) :
   reset_button(false),
   level(),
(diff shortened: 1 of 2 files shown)


tobbi:
Small nit: Can you change the variable to old_level ?

christ2go:
@tobbi Done

PR #597, opened 2016-09-15
Add ultra-violet lighting

Hume2:
Penguins see four colour channels, so we should have a four-channel lighting. I think, this could be an useful feature anyway.

### DIFF
--- a/src/video/color.cpp
+++ b/src/video/color.cpp
@@ -14,6 +14,7 @@
 //  You should have received a copy of the GNU General Public License
 //  along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+#include <algorithm>
 #include <config.h>
 
 #include "video/color.hpp"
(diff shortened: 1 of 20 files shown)


src/video/color.cpp
```diff
@@ -34,22 +34,25 @@ Color::Color() :
   alpha(1.0f)
 {}
```
maxteufel: You'll want to initialize the ultra_violet member here as well, otherwise the builds will fail.

Hume2: I'm just in it, thanks anyway.

tobbi:
Next up: Change test.

165your4:
It's a bit unnecessary/weird to be honest.

1. If your justification is "penguins see four colour channels" then surely we'd have to edit every texture to take this into account? It makes no sense to have four-colour light but have everything else be three-colour.

2. Actual ultraviolet light wouldn't look purple to a penguin anyway, it'd be some colour that we wouldn't be able to comprehend. That is, if penguins even see colour the same way we do.

3. Adding to the previous point, if this ultraviolet light is represented by purple light, then it doesn't exactly offer any new colour options. Every colour can already be made with the current RGB system.

Hume2:
The point is that penguins see it, but humans can see it just like a tiny purple tint.

165your4:
@Hume2 Humans see some UV light as a purple tint, but only for wavelengths very close to regular violet. UV light beyond that is invisible to us. And anyway, what I'm trying to say is that if we add UV light for realism, we'd also have to alter the textures for realism, seeing as colour is based the absorption/reflection of light. 

For example, a tree might look green to us humans, but if we could see ultraviolet light (like a penguin) it would look completely different. If we have UV light, it would make sense to change the appearance of the trees in the Forest World and make them look "completely different" too. The issue with that is that we have no idea what trees would look like with UV vision. In fact, we don't know what _anything_ would look like with UV vision because we are so used to seeing the world as RGB.

Hume2:
Yes, this is the close UV, so humans see it as a purple tint. And I think, the textures are already painted as "how the humans see it", so humans see trees as green, so they should stay as are. The "how the humans see it" doesn't depend on this being implemented.

165your4:
@hume2 The point is, if you're going to make the light "how the penguins see it", why would everything else be "how the humans see it"?

Hume2:
It's because SuperTux is a game about penguins played by humans. Therefore Tux sees the ultra-violet light, but we don't. If it was designed for penguins to play this game, we would have probably RGBU (red green blue ultra-violet) monitors and it would already contain the RGBU graphics.

maxteufel:
I think this needs more discussion. However, I don't see this as something to make the game more realistic or whatever. It's just a feature that could be, in some levels, used to do something useful.

maxteufel:
@Hume2 Please provide a test level.

Hume2:
Test level here: http://pastebin.com/VLCz3x1M

tobbi:
We decided against merging this. Sorry.

PR #605, opened 2016-09-20
(WIP?) Cannon no longer unfreezable by walking at it

### LINKED ISSUE #604: Walking into a frozen cannon unfreezes it
Walking into a cannon frozen with the Iceflower seems to unfreeze it as soon as Tux touches the sprite. I noticed this in 0.5.0 rc4 so I'd imagine the bug is still around.

christ2go:
This fixes #604. In my opinion at least for canons it is not expected behaviour for canons to behave that way. An open question is, if rocket launchers should break when running against them too.

### DIFF
--- a/src/badguy/dispenser.cpp
+++ b/src/badguy/dispenser.cpp
@@ -167,7 +167,10 @@ Dispenser::collision(GameObject& other, const CollisionHit& hit)
       return FORCE_MOVE;
     }
     if(frozen){
-      unfreeze();
+      if(type != DT_CANNON)
+      {
+        unfreeze();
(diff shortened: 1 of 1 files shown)


src/badguy/dispenser.cpp
```diff
@@ -167,7 +167,10 @@ Dispenser::collision(GameObject& other, const CollisionHit& hit)
       return FORCE_MOVE;
     }
     if(frozen){
-      unfreeze();
+      if(type != DT_CANNON)
```
maxteufel: Please merge these two if statements.

PR #607, opened 2016-09-22
Add Twitter button

tobbi:
This adds a twitter button to the README (might be nice for visibility)

[ci skip]


maxteufel:
This doesn't work with tracking protection tools. Given the fact that there is a link to Twitter already I don't agree with this.

maxteufel:
Please reopen if this should be discussed again.

PR #612, opened 2016-09-25
CMake: Accept RelWithDebInfo as a valid release type

akien-mga:
Distro packagers tend to use the `RelWithDebInfo` mode ("release with debug info") when building packages, as it will use optimisation but leave some debug symbols for the package builder to extract (and generate e.g. -debuginfo packages for RPM, and likely something similar for DEB).

Both types should be considered as "release" types as far as SuperTux is concerned.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -65,7 +65,7 @@ INCLUDE(TestBigEndian)
 TEST_BIG_ENDIAN(WORDS_BIGENDIAN)
 
 ## Add definitions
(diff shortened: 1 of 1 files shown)


akien-mga:
Note: I can amend the commit to remove the added `MESSAGE`s if you don't consider it a useful addition.

akien-mga:
Removed the status message as discussed on IRC with mt, as users can just read the CMakeLists.txt instead.

PR #617, opened 2016-09-26
Added Sound Volume and Music Volume options

### LINKED ISSUE #273: Allow changing volume
The music volume is too loud. Also this is very subjective.

Please make it possible to change volume of sounds and music individually in the settings (currently it is only possible to turn on/off).

This can be done using a <> menu like for resolution (0%, 25%, 50%, 75%, 100%).

I recommend 50% as default values.

paezao:
User can select between 0%, 25%, 50%, 75% and 100%
When sound or music are 0% they are disabled
Removed sound_enabled and music_enabled from config
Replaced those with sound_volume and music_volume

Closes #273

### DIFF
--- a/src/supertux/main.cpp
+++ b/src/supertux/main.cpp
@@ -218,7 +218,7 @@ public:
 	if (FileSystem::is_directory(olduserdir)) {
 	  boost::filesystem::path olduserpath(olduserdir);
 	  boost::filesystem::path userpath(userdir);
-	  
+
 	  boost::filesystem::directory_iterator end_itr;
 
 	  bool success = true;
@@ -254,7 +254,7 @@ public:
     if (!FileSystem::is_directory(userdir))
     {
 	  FileSystem::mkdir(userdir);
-	  log_info << "Created SuperTux userdir: " << userdir << std::endl;  
+	  log_info << "Created SuperTux userdir: " << userdir << std::endl;
     }
 
     if (!PHYSFS_setWriteDir(userdir.c_str()))
@@ -366,8 +366,10 @@ Main::launch_game()
 
   timelog("audio");
   SoundManager sound_manager;
-  sound_manager.enable_sound(g_config->sound_enabled);
-  sound_manager.enable_music(g_config->music_enabled);
+  sound_manager.enable_sound(g_config->sound_volume > 0 ? true : false);
+  sound_manager.enable_music(g_config->music_volume > 0 ? true : false);
+  sound_manager.set_sound_volume(g_config->sound_volume);
+  sound_manager.set_music_volume(g_config->music_volume);
 
   Console console(console_buffer);
 
@@ -458,7 +460,7 @@ Main::run(int argc, char** argv)
 	freopen((prefpath + "/console.out").c_str(), "a", stdout);
 	freopen((prefpath + "/console.err").c_str(), "a", stderr);
 #endif
- 
+
   int result = 0;
 
   try
--- a/src/supertux/menu/options_menu.cpp
+++ b/src/supertux/menu/options_menu.cpp
@@ -37,8 +37,8 @@ enum OptionsMenuIDs {
   MNID_FULLSCREEN_RESOLUTION,
   MNID_MAGNIFICATION,
   MNID_ASPECTRATIO,
-  MNID_SOUND,
-  MNID_MUSIC,
+  MNID_SOUND_VOLUME,
+  MNID_MUSIC_VOLUME,
   MNID_DEVELOPER_MODE,
   MNID_CHRISTMAS_MODE,
   MNID_TRANSITIONS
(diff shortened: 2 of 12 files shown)


src/supertux/menu/options_menu.cpp
```diff
@@ -179,6 +183,65 @@ OptionsMenu::OptionsMenu(bool complete) :
...
+          next_sound_volume = cnt_;
+          break;
+      }
+      ++cnt_;
+  }
+
+  if (!sound_volume.empty())
+  {
+      next_sound_volume = sound_volumes.size();
+      sound_volumes.push_back(sound_volume);
```
christ2go: Is this for adding a specific sound level to the options if the size isn't yet in the sound_volumes list?

paezao: This works like the other selects because the user can technically setup any volume he wants in the config.

christ2go: Might be better to sort the options then (it's just a small comment, but it might else irritate some users).

paezao: Actually I don't think that if makes sense at all. I'll change to something different.

paezao: Okay @christ2go I added sorting.

src/supertux/menu/options_menu.cpp
```diff
@@ -179,6 +183,77 @@ OptionsMenu::OptionsMenu(bool complete) :
     resolutions.push_back(fullscreen_size_str);
   }
 
+  // Sound Volume
+  sound_volumes.clear();
```
maxteufel: Make this allow an arbitrary float within a defined range. Limiting the user's options does not make any sense.

src/supertux/main.cpp
```diff
@@ -366,8 +366,10 @@ Main::launch_game()
 
   timelog("audio");
   SoundManager sound_manager;
-  sound_manager.enable_sound(g_config->sound_enabled);
-  sound_manager.enable_music(g_config->music_enabled);
+  sound_manager.enable_sound(g_config->sound_volume > 0 ? true : false);
+  sound_manager.enable_music(g_config->music_volume > 0 ? true : false);
```
maxteufel: Volume 0 is usually not exactly mute; don't know how OpenAL handles this.

paezao: I thought it would be better to disable sound if volume is 0 so we don't waste resources playing sounds with 0 gain. According to https://www.openal.org/documentation/openal-1.1-specification.pdf page 32 a gain of 0 is silence.

src/supertux/menu/options_menu.cpp
paezao: Do you mean in the interface or config file? Config-file-wise you can set any value you want and it will work but not in the interface because it uses the selector.

src/supertux/main.cpp
maxteufel: Then this is a fault in OpenAL, and indeed, we shouldn't waste resources. However allow the user to quickly toggle music and/or sound on/off by keeping the old button. Use OR here.

src/supertux/menu/options_menu.cpp
maxteufel: In the interface. Restricting the user to arbitrarily imposed options is no good idea.

maths22: If we want to support arbitrary volumes, then we first need some form of slider or integer spinner menu component.

src/supertux/main.cpp
maths22: How is it a fault in OpenAL that a volume of 0 is equivalent to mute?  This would seem to be the behavior of any rational program.  I also question if it is advisable to add even more options to the settings dialog-it already is pretty long as it is.  While allowing the user to short-circut the selector to 0/mute might be useful, having music on/off and music volume as separate line items seems redundant and a waste of space.

tobbi: We could, however, have checkboxes (or something) for immediate toggling of the settings.

christ2go:
An alternative might be to have two buttons left and right (currently: < 120 > , then  < < 120 > > ), and the first and last could be used for instant toggling to highest/lowest.
> Am 06.10.2016 um 13:51 schrieb Tobias Markus <notifications@github.com>:
> 
> We could, however, have checkboxes (or something) for immediate toggling of the settings. 
> 
>

src/supertux/main.cpp
christ2go: An alternative might be to have two buttons left and right (currently: < 120 > , then  < < 120 > > ), and the first and last could be used for instant toggling to highest/lowest.

src/supertux/menu/options_menu.cpp
Hume2: Maybe the colorchannel item could be used for that, well, it wouldn't be draggable and the values would go from 0 to 1 instead of 0% to 100%. I personally wouldn't mind choosing between fixed values, but I don't see why music_volumes should be different than sound_volumes. And when it's like this, the int values can be precalculated, so we don't need to extract them from the string on each change.

maxteufel:
@paezao Are you still interested in working on this?

The plan would be to make this in 1/10th steps, but allow different input via the configuration. If the configuration has a non-standard value the in-game box should probably be disabled because it can't yet handle undefined values, afaik. 0 values should be no amplification, as far as possible; mute options should stay.

tobbi:
Since @paezao is not responding, I assume we need to source those patches and re-submit a PR with things fixed.

Any volunteers? @maxteufel maybe?

PR #619, opened 2016-09-28
Calculate Text object box size correctly

tobbi:


maxteufel:
As I said on IRC, splitting the text in multiple lines so it fits the current box would be preferred.

maxteufel:
If this is supposed to be for 0.5.1, please close this PR and open one on the release/0.5 branch.

src/object/text_object.cpp
```diff
@@ -103,8 +101,9 @@ TextObject::draw(DrawingContext& context)
     return;
   }
 
-  float width  = 500;
-  float height = 70;
+  // 20 = twice the padding in the draw_text call below
+  float width  = font->get_text_width(text) + 20;
+  float height = font->get_text_height(text) + 20;
```
Hume2: Maybe it would be better if the size was precalculated on set_text.

tobbi:
Superseded.

PR #621, opened 2016-10-01
Tilemap: Add an option to change draw target

### LINKED ISSUE #600: Level Editor: Tilemap as Lightmap Option
I'm not sure, whether this is intended to be or not, but I noticed that in the external Editor you were able to choose, whether or not a tilemap should be a lightmap. But in the in game editor you can't.

Hume2:
fixes #600

### DIFF
--- a/src/object/tilemap.cpp
+++ b/src/object/tilemap.cpp
@@ -220,6 +220,11 @@ TileMap::get_settings() {
   result.options.push_back( ObjectOption(MN_COLOR, _("tint"), &tint));
   result.options.push_back( ObjectOption(MN_INTFIELD, _("Z-pos"), &z_pos));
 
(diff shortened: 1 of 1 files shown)


src/object/tilemap.cpp
```diff
@@ -220,6 +220,11 @@ TileMap::get_settings() {
   result.options.push_back( ObjectOption(MN_COLOR, _("tint"), &tint));
   result.options.push_back( ObjectOption(MN_INTFIELD, _("Z-pos"), &z_pos));
 
+  ObjectOption dto(MN_STRINGSELECT, _("Draw target"), &draw_target);
```
maxteufel: Now that's a truly descriptive variable name, isn't it?

PR #622, opened 2016-10-01
tools/build-addon-index.py: langpack support

maxteufel:
Add support for building language packs with `tools/build-addon-index.py`. Also rewrite and/or update some parts of that script.


tobbi:
Since the transifex tool is also a python lib, can't you also make it run that automatically and then use the path where the po files were downloaded?

PR #460, opened 2016-06-26
Add an option to snap objects to the grid

Hume2:

PR #626, opened 2016-10-02
Add an option to snap objects to the grid

Hume2:


src/editor/input_center.cpp
```diff
@@ -748,7 +763,7 @@ EditorInputCenter::draw_tile_tip(DrawingContext& context) {
 }
 
 void
-EditorInputCenter::draw_tile_grid(DrawingContext& context) {
+EditorInputCenter::draw_tile_grid(DrawingContext& context, const Color line_color, int tile_size) {
```
tobbi: Please pass color by ref:
const Color& line_color

src/editor/input_center.hpp
```diff
@@ -82,16 +87,17 @@ class EditorInputCenter
...
 
     void process_left_click();
     void process_right_click();
 
     // sp is sector pos, tp is pos on tilemap.
-    Vector tp_to_sp(const Vector& tp);
-    Vector sp_to_tp(Vector sp);
-    Vector tile_screen_pos(const Vector& tp);
+    Vector tp_to_sp(const Vector& tp, int tile_size = 32);
+    Vector sp_to_tp(Vector sp, int tile_size = 32);
```
tobbi: This should be "const Vector&" as well, please.

src/editor/point_marker.cpp
```diff
@@ -56,4 +56,8 @@ Vector PointMarker::get_point_vector() const {
   return Vector(0,0);
 }
 
+Vector PointMarker::get_offset() const {
+  return Vector(0,0);
```
tobbi: While you're at it: Space after comma.

maxteufel:
> What about an option to snap to a 16x16 grid? The external editor has this and it's really useful for badguy/coin/bonus block placement.

— @165your4 on #460 

@165your4: This has been added now.

src/editor/input_center.cpp
```diff
@@ -432,7 +434,18 @@ EditorInputCenter::move_object() {
       dragged_object = NULL;
       return;
     }
-    dragged_object->move_to(sector_pos - obj_mouse_desync);
+    Vector new_pos = sector_pos - obj_mouse_desync;
+    if (snap_to_grid) {
+      auto& snap_grid_size = snap_grid_sizes[selected_snap_grid_size];
+      new_pos.x = int(new_pos.x / snap_grid_size) * snap_grid_size;
```
tobbi: I think you can use
new_pos = (new_pos / snap_grid_size) * snap_grid_size;

We have operators defined for vectors.

src/editor/input_center.cpp
```diff
@@ -857,32 +886,32 @@ EditorInputCenter::draw(DrawingContext& context) {
...
 Vector
-EditorInputCenter::sp_to_tp(Vector sp) {
+EditorInputCenter::sp_to_tp(const Vector& sp, int tile_size) {
   auto tilemap = dynamic_cast<TileMap*>(Editor::current()->layerselect.selected_tilemap);
   if (tilemap) {
-    sp -= tilemap->get_offset();
-    int x = sp.x / 32;
-    int y = sp.y / 32;
+    Vector sp_ = sp - tilemap->get_offset();
+    int x = sp_.x / tile_size;
```
tobbi: Same here: Doesn't return sp_ / tile_size; work?

src/editor/input_center.cpp
Hume2: No, because vector counts with floats and I have to round it to int.

tobbi: Alright, then.

src/editor/input_center.cpp
Hume2: No, because it should be an int.

PR #627, opened 2016-10-02
Remove uneeded include for SDL.h in main.cpp

SuicSoft:


maths22:
This include is actually necessary for Windows, which uses SDL2_main to set things up properly on Windows.

PR #554, opened 2016-08-17
Store script variables

christ2go:
Currently scripts can only store data by modifying the savegamefile directly.
Now the functions store and load allow scripts to easily preserve data.
Syntax: store(varname,value), where varname is a string, and value is a
serializable object. A squirrel object is serializible iff it is of type
Array, Table, String, Int, Bool, Float. Arrays and Tables are also serializable,
iff they only contain serializible items (eg. Array of Array of Int is serializable).
To load an object the function load(varname) is provided.

Serialization
Is currently done in the form of S-Expressions.  The file is in the same
directory as the savegame file and has the extension .sss

Implementation
When a object is stored it is tranformed into a internal data structure
representing arrays (sqarray in sqarrayser.cpp) , tables (sqdict in sqdictser.cpp)
 and scalar values (stored using boost::any container). The dictionary class
provides a general storage of names and the associated item in the internal
data structure. When a worldmap is loaded a new dictionary is created,
when a worldmap is left a dictionary is destroyed, and the file is created.

The internal data structures for arrays and tables  are implemented as classes,
which all provide the following methods
- load(ReaderCollection& r)
  
  Used for loading from the file
- get(HSQUIRRELVM v)
  
  Used for pushing onto the stack
- serialize(Writer\* serialize)
  
  Used for serializing the object
- loadArray(HSQUIRRELVM v) (or loadDict)
  
  Used for loading from squirrel

Issues / Further Modifications

-Currently only basic types are storable. Extend it to "special" C++
 Classes which e.g. implement an Interface Storable or use an auto-
 matically generated wrapper for serializing some classes.

-Add option to store from the game and make such stored items "private".

Closes #530

PR #630, opened 2016-10-05
Remove Boost dependency

SuicSoft:
This will remove the Boost dependency of SuperTux. I have done play testing of this branch and there seems to be no bugs.

Uses some C++11.

Tested on Ubuntu 16.04.


Grumbel:
The format() patch looks completely messed up, lots of boost format strings are left in that printf() won't accept and worse, lots of passing std::string into format() and then into printf() which doesn't work either, that only accepts standard C types.

For getting rid of boost in is_christmas() use std::chrono.

I would stay away from std::experimental, boost can do that more portable until std:: absorbs those features properly.

christ2go:
Please also note, that boost is required for merging #554. Else I agree with @Grumbel.

SuicSoft:
@Grumbel I just changed the boost::format to std::format and removed the % and other stuff. And can you explain what you mean by "lots of passing std::string into format() and then into printf() which doesn't work either, that only accepts standard C types." since my code does work and acts the same as the normal SuperTux git

Grumbel:
Take this:
```
#include <iostream>
#include <string>
#include "format.hpp"

int main()
{
  std::string str = "Hello World";
  std::cout << format("%s\n", str) << std::endl;
  return 0;
}
```
And run it:
```
$ ./main 
Џ��
```

`format()` is handing the data over to `printf()` and `printf()` is an old C function that doesn't deal with C++ types like std::string and just produces gibberish.

maxteufel:
No. We didn't ask for this, and there does not seem to be any point in removing the boost dependency.

PR #635, opened 2016-10-08
Improve handling of package version

maths22:


src/supertux/main.cpp
```diff
@@ -202,7 +202,7 @@ class PhysfsSubsystem
     }
     else
     {
-		userdir = PHYSFS_getPrefDir("SuperTux","supertux2");
+		userdir = PHYSFS_getPrefDir("SuperTux", PACKAGE_NAME);
```
maxteufel: Can we get "SuperTux" a separately defined constant too? I imagine it could be useful, also in the future (e.g. for scripts that run via CMake).

src/supertux/main.cpp
```diff
@@ -311,7 +311,7 @@ class SDLSubsystem
 void
 Main::init_video()
 {
-  SDL_SetWindowTitle(VideoSystem::current()->get_renderer().get_window(), PACKAGE_NAME " " PACKAGE_VERSION);
+  SDL_SetWindowTitle(VideoSystem::current()->get_renderer().get_window(),  "SuperTux " PACKAGE_VERSION);
```
maxteufel: See the comment above.

CMakeLists.txt
```diff
@@ -267,6 +267,12 @@ IF(${VERSION_LIST_SIZE} GREATER 0)
 		SET(VERSION_STRING_GIT "${MAJOR_VERSION_GIT}.${MINOR_VERSION_GIT}.${PATCH_VERSION_GIT}")
 	ENDIF()
 
+        IF("${VERSION_STRING_GIT}" MATCHES "v[0-9]+\\.[0-9]+\\.[0-9]+-[A-Za-z0-9]+")
```
maxteufel: The `^v` should be optional in this string. Also, it could handle SemVer completely, who knows if someone might patch in a build/metadata suffix.

PR #636, opened 2016-10-12
makepot.sh: supply more information for POT generation

maxteufel:
Supply the following additional information when generating POT files:
- Define the package name as ~~"supertux"~~ "SuperTux"
- Define a package version via ~~`git describe --tags`~~ git
- ~~Define the copyright holder as "SuperTux"~~
- Define https://github.com/SuperTux/supertux as report address for
  msgid bugs

Furthermore, the xgettext calls have been made more readable by
splitting them up onto multiple lines.


tobbi:
Those `find <...>` calls are all the same, can we add a bash function for that so that we decrease code duplication?

maxteufel:
@tobbi I think we should actually use a POTFILES.in file or so

PR #637, opened 2016-10-19
build fixes for gcc-6 and uncommon environments

schnitzeltony:


Grumbel:
Removing the `SYSTEM` in `INCLUDE_DIRECTORIES(SYSTEM ${GLEW_INCLUDE_DIR})` and all the other includes will break compiling with warnings enabled, as gcc will start emit warnings for all the included files instead of just the SuperTux files.

schnitzeltony:
OK I just pushed V2:

* remove make log in first patch *@maxteufel*
* remove second patch.

Have not much experience regarding github workflow: Do I need to open
another PR for this?

tobbi:
You don't. It synchronized this one.

Issue #350, opened 2016-01-31
[RFC] Reformat code (using clang-format)


maxteufel:
This would make code more consistent and if spaces are used instead of tabs will make the code be displayed more consistently on different text editors and setups.
- [ ] How to handle doxygen comments?

Alasdairbugs:
Closed as per advice from Vankata, Allie

PR #639, opened 2016-10-21
Coding Standards Changes

maxteufel:
This pull request is intended to be used for tracking proposed changes to the "SuperTux Coding Standards" document, as well as changes to its realization.

This includes:
- [x] Reformatting the document to Markdown
- [ ] Collect proposals for changes
- [ ] Reformat the entire codebase according to it, see #350


CODINGSTYLE.md
```diff
@@ -0,0 +1,87 @@
...
+## Compiler options
+
+Compile with a maximum warning level and with `-Werror`:
+
+```
+-Werror -ansi -pedantic -Wall -Wextra -Wnon-virtual-dtor -Weffc++ -Wcast-qual
+-Winit-self -Wno-unused-parameter
+```
+
+Possible additional flags for the future are:
```
christ2go: Travis already uses -Wshadow => update this?

CODINGSTYLE.md
```diff
@@ -0,0 +1,87 @@
...
+
+```
+-Wconversion -Wshadow
+```
+
+## Comments
+
+Use the `//` syntax for regular comments, don't use `/* */`.
```
christ2go: Maybe clarify if `/* */` can still be used for multiline comments?

christ2go:
The coding standars should also include a section about using C++11/C++14 features (and maybe mention stuff like `const`-correctness. 

We should decide if we want to follow cpplint's suggestion to only use const references and to replace non-const references with pointers (cpplint marks this, but we use it rather often).

The guide should also have some suggestions where to put spaces (i.e. after keywords like `for, if, else` etc.)

## Modern C++
Whenever possible use modern c++ features. This includes 
-  using  `auto`when the type can be inferred by the compiler **and** the type is obvious. 
-  using C++11-style for-loops whenever possible (i.e. when iterang over a vector): `for (const auto& x : vector)`
- striving for `const`-correctness: If a method doesn't change member variables, mark it const. If a argument is not affected by the method: Mark it const. If you can mark something const, mark it const.
- do not reinvent the wheel: If there is a function in the standard library which solves a problem, use it. 
- initialize every class member in the initializer list (the order of members is also important)

Before submitting a pull request reformat your code using clang-format.

PR #645, opened 2016-10-25
Fix plural forms handling

siiptuo:
Plural form is needed in English and other languages with different kind of
rules. Also translation for level editor's levelset select menu was lacking
necessary context for some languages.

### DIFF
--- a/makepot.sh
+++ b/makepot.sh
@@ -3,7 +3,7 @@
 package_name="SuperTux"
 package_version="$(git describe --tags --match "?[0-9]*.[0-9]*.[0-9]*")"
 
-xgettext --keyword='_' -C -o data/locale/main.pot \
+xgettext --keyword='_' --keyword='__:1,2' -C -o data/locale/main.pot \
   $(find src -name "*.cpp" -or -name "*.hpp") \
   --add-comments=l10n \
   --package-name="${package_name}" --package-version="${package_version}" \
--- a/src/scripting/functions.cpp
+++ b/src/scripting/functions.cpp
@@ -105,6 +105,16 @@ std::string _(const std::string& text)
   return translate(text);
 }
 
+std::string translate_plural(const std::string& text, const std::string& text_plural, int num)
+{
+  return g_dictionary_manager->get_dictionary().translate_plural(text, text_plural, num);
+}
+
+std::string __(const std::string& text, const std::string& text_plural, int num)
+{
+  return translate_plural(text, text_plural, num);
+}
+
 void display_text_file(const std::string& filename)
 {
   ScreenManager::current()->push_screen(std::unique_ptr<Screen>(new TextScroller(filename)));
--- a/src/scripting/functions.hpp
+++ b/src/scripting/functions.hpp
@@ -107,6 +107,10 @@ void abort_screenfade();
 std::string translate(const std::string& text);
 std::string _(const std::string& text);
 
+std::string translate_plural(const std::string& text, const std::string&
+    text_plural, int num);
+std::string __(const std::string& text, const std::string& text_plural, int num);
+
 /**
  * Load a script file and executes it. This is typically used to import
  * functions from external files.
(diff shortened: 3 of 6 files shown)


maxteufel:
> Plural form is needed in English and other languages with different kind of rules.

Why is it needed?

MCMic:
It allows to translate plurals accordingly, including in languages which does work like english.
http://docs.translatehouse.org/projects/localization-guide/en/latest/l10n/pluralforms.html
For instance in english you say «0 cats, 1 cat, 2 cats», but in french we say «0 chat, 1 chat, 2 chats», 0 is not plural in french but it is in english.
Some languages have even more plural forms depending of how much stuff there is.

maxteufel:
Certainly if we intend to do this properly it needs to be done:

- at *all* existing parts of code
- in the scripting API too, so people can access this from scripts

Hume2:
All existing parts of code = editor levelset select menu. I don't think that this appears elsewhere. And for the scripting API, I don't see it as a huge problem.

It might not sound so bad in English (and in German), but it does sound really weird in languages those use different forms depending on the count. As I've seen from plural_forms.cpp, the main aspects are:
* Sometimes it depends only on last one or two digits of the count.
* Some languages use dual forms when the count is 2. They have singular, dual and plural.
* Some languages use genetive forms when the count is bigger.

maths22:
We almost certainly should aim to do this, and should do it properly.  I
would suggest splitting this PR into two separate pieces: 1 which adds the
tooling support (including for scripting) and 1 which replaces all of the
appropriate references in the string (i.e. the level editor change and
other changes would be separated from this PR)

maths22:
Originally posted by @Hume2 (and accidentally deleted by me)

All existing parts of code = editor levelset select menu. I don't think that this appears elsewhere. And for the scripting API, I don't see it as a huge problem.

It might not sound so bad in English (and in German), but it does sound really weird in languages those use different forms depending on the count. As I've seen from plural_forms.cpp, the main aspects are:

-     Sometimes it depends only on last one or two digits of the count.
-     Some languages use dual forms when the count is 2. They have singular, dual and plural.
-     Some languages use genetive forms when the count is bigger.

maxteufel:
> We almost certainly should aim to do this, and should do it properly.  I
> would suggest splitting this PR into two separate pieces: 1 which adds the
> tooling support (including for scripting) and 1 which replaces all of the
> appropriate references in the string (i.e. the level editor change and
> other changes would be separated from this PR)

I would suggest to simply split it up in two separate commits, separate PRs seems too much and creates more interdependence than necessary.

tobbi:
@siiptuo I assume you're not willing to work on pluralizing the rest of the strings in SuperTux, right? If you do, please comment on and update this PR. Otherwise, I'll close it.

siiptuo:
Sorry, I had completely forgot about this.

I've formatted the commits and added scripting API. I can't find any other strings than the level editor one that need pluralization. For example levels contain static strings and statistics like "Coins: 100" don't need pluralization.

tobbi:
Travis fails with:
```
/home/travis/build/SuperTux/supertux/src/supertux/menu/editor_levelset_select_menu.cpp: In constructor ‘EditorLevelsetSelectMenu::EditorLevelsetSelectMenu()’:
/home/travis/build/SuperTux/supertux/src/supertux/menu/editor_levelset_select_menu.cpp:80:9: error: ‘boost’ has not been declared
         boost::format(__("%d level", "%d levels", level_count)) % level_count <<
         ^
make[2]: *** [CMakeFiles/supertux2_lib.dir/src/supertux/menu/editor_levelset_select_menu.cpp.o] Error 1
make[2]: *** Waiting for unfinished jobs....
make[1]: *** [CMakeFiles/supertux2_lib.dir/all] Error 2
make: *** [all] Error 2
```

Did you manage to compile it localy?

siiptuo:
Now it's fixed.

PR #647, opened 2016-10-29
Check whether Tux carries a certain portable

### LINKED ISSUE #559: Scripting support for checking whether tux carries certain Portable
Basically, I think level could consist of a script trigger that checks whether tux carries a certain portable object and then only allows access if this is the case.

Therefore, I am proposing the following:

Portable objects have names (don't they?)

We could have a scripting interface, like:
sector.Tux.is_carrying("my-portable-name"), which returns a boolean true / false.

tobbi:
Fixes #559


tobbi:
This supports Rocks, and every GameObject class derived from `Rock`, so Rock, Trampolines, Lanterns.

src/scripting/player.cpp
```diff
@@ -173,6 +173,12 @@ Player::get_velocity_y() const
   return m_parent->physic.get_velocity_y();
 }
 
+bool
+Player::has_grabbed(const std::string& name) const
```
maxteufel: Write documentation for this scripting method.

christ2go: Isn't the documentation already in the hpp file?

Issue #558, opened 2016-08-17
Use bitfield for ObjectOption flags


tobbi:
I think we should use a bit field for object option flags, because stray boolean values get confusing.

I was thinking about something like 
OPTION_ALLOW_EMPTY and OPTION_VISIBLE.

Or rather:
OPTION_HIDDEN and OPTION_MANDATORY or something.

tobbi:
Fixed by cba9674

PR #653, opened 2016-11-11
Use a bitfield for ObjectOption flags, addresses issue #558

shawnanastasio:
ObjectOption flags are now kept using a bitfield as opposed to booleans.
To check the presence of an option, use the bitwise AND operator (&) with
the ObjectOption's `flags` variable and the desired flag (declared at the
top of src/editor/object_option.hpp).

All code that I found that uses the previous booleans has been updated to use the new bitfield,
and I have done some preliminary testing on the code and it appears to work fine.

### DIFF
--- a/src/editor/object_menu.cpp
+++ b/src/editor/object_menu.cpp
@@ -32,7 +32,7 @@ ObjectMenu::ObjectMenu(GameObject *go) :
   add_label(os.name);
   add_hl();
   for(auto& oo : os.options) {
-    if(!oo.visible) {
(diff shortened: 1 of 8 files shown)


shawnanastasio:
@maxteufel Yeah absolutely. I normally work with C so I didn't even consider using enums :P. Should I preserve the capital names, or is a different naming style more appropriate?

maxteufel:
The all-caps value names should probably be kept, if you choose to name the enum itself probably UpperCaseOnWordBeginning.

PR #656, opened 2016-11-20
Dispenser: Don't shoot on auto setting when there's something blocking

tobbi:
Previously, enemies were spawned into walls, I now check whether the target box is free of tiles so that we can spawn the enemies in. If it isn't, then we don't spawn at all.

From the docs: Dispenser "auto" setting follows Tux, so this "seems" to be correct behaviour. 

Please check this regardless.

### DIFF
--- a/src/badguy/dispenser.cpp
+++ b/src/badguy/dispenser.cpp
@@ -212,12 +212,21 @@ Dispenser::launch_badguy()
   //FIXME: Does is_offscreen() work right here?
   if (!is_offscreen() && !Editor::is_active()) {
     Direction launchdir = dir;
-    if( !autotarget && start_dir == AUTO ){
-      Player* player = get_nearest_player();
-      if( player ){
+    if( !autotarget && start_dir == AUTO ) {
+      auto player = get_nearest_player();
+      if( player ) {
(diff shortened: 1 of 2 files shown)


src/badguy/dispenser.cpp
```diff
@@ -212,13 +212,22 @@ Dispenser::launch_badguy()
...
     }
 
+    // First pass: Check whether target box is free of tiles. Spawning Enemies
+    // into walls is not fun.
+    if(is_blocked(launchdir))
+      launchdir = dir;
+
+    // Second pass with changed direction. Let's check again!
+    if(is_blocked(launchdir))
+      return;
```
Hume2: Please, make sure that we don't test for the same direction twice.

src/badguy/dispenser.cpp
```diff
@@ -371,6 +380,14 @@ Dispenser::set_correct_action()
   }
 }
 
+bool
+Dispenser::is_blocked(const Direction& direction) const
+{
+  auto target_box = get_bbox();
+  target_box.move(Vector((direction == RIGHT) ? 32 : -32, 0));
```
Hume2: This is not actually true. It depends on the type of dispenser.

tobbi:
I'm gonna close this, Hume's right, this was mostly a failed attempt.

PR #663, opened 2016-12-10
Multiple display support (WIP)

tobbi:
This is WIP because setting the window position doesn't actually work yet.


tobbi:
This works for OpenGL, not for SDL, yet, though. Furthermore, SuperTux needs to be restarted after a change.

tobbi:
Closing as per review.

Issue #216, opened 2015-11-06
Image recompression is overly agressive


Grumbel:
Haven't looked into the details, but this last recompression of images was overly aggressive and lead to image formats that are no longer easily modifiable. This one for example:

    ./data/images/background/halloween/fog.png

Will show up in Gimp as indexed image with 1bit alpha, instead RGB with 8bit alpha channel. It still renders correctly in other viewers. So this might be a bug in in Gimp, either way, I would suggest to not convert into formats that are outside the normal 24bit RGB or 32bit RGBA to avoid that kind of trouble or better yet just not convert anything at all in the source directory. Make it an optional part of the build process if you have to.

-------

commit 8be7d1bee94ee53bdba7942db0a374c3e0d6bd83
Author: Tobias Markus <tobbi@mozilla-uk.org>
Date:   Sat Sep 12 21:07:50 2015 +0200

    Re-compress image files with latest ImageOptim
    [ci skip]

brmbrmcar:
How about reversing the affected images for now, and deal with it after  0.4.0?

Grumbel:
Original Image:

    ./data/images/background/halloween/fog.png...
      Image Width: 1280 Image Length: 256
      Bitdepth (Bits/Sample): 8
      Channels (Samples/Pixel): 4
      Pixel depth (Pixel Depth): 32
      Colour Type (Photometric Interpretation): RGB with alpha channel 
      Image filter: Single row per byte filter 
      Interlacing: No interlacing 
      Compression Scheme: Deflate method 8, 32k window
      Resolution: 2835, 2835 (pixels per meter)
      FillOrder: msb-to-lsb
      Byte Order: Network (Big Endian)
      Number of text strings: 1 of 9
        Comment (xTXt deflate compressed): Created with GIMP

-----
Recompressed image:

    ./data/images/background/halloween/fog.png...
      Image Width: 1280 Image Length: 256
      Bitdepth (Bits/Sample): 8
      Channels (Samples/Pixel): 1
      Pixel depth (Pixel Depth): 8
      Colour Type (Photometric Interpretation): PALETTED COLOUR with alpha (201 colours, 201     transparent) 
      Image filter: Single row per byte filter 
      Interlacing: No interlacing 
      Compression Scheme: Deflate method 8, 32k window
      Resolution: 0, 0 (unit unknown)
      FillOrder: msb-to-lsb
      Byte Order: Network (Big Endian)
      Number of text strings: 0 of 0

[...]

PR #673, opened 2016-12-26
Optimize pngs using matthiaskrgr/woptipng

matthiaskrgr:
195 of 1992 files optimized, 8358891 bytes reduced to 6375354 bytes; -1983537 bytes, -23.72%
Optimization threshold was 10% (optimizations below 10% were discarded)

Script uses imagemagick, advdef and optipng.


maxteufel:
Also see #216

matthiaskrgr:
Note that optipng -np -nc avoids reduction to palette color space.

You can find palette images in the branch via
````
git diff  --name-only master | grep "\.png$" | xargs identify -verbose | grep "^Image\|PLTE" 
````
There were 4 palette images
````
data/images/tiles/auxiliary/black.png
data/images/objects/lightmap_light/lightmap_light-large.png
data/images/engine/console2.png
data/images/background/fog.png
````
however all of these were already palette in the first place on master branch so I think there won't be a problem with these.

maxteufel:
I think a compression script should be rather run when packaging distributed versions. As far as the git repository is concerned, this will not bring any advantage, it will only increase its size.

matthiaskrgr:
It will reduce size for final releases and shallow clones.
If a package maintainer runs image compression, bugs (like the sdl palette problem) might stay unnoticed until the final package is shipped to the user.

ildar:
Excuse me, IMHO optimizing for 2 megabytes is for nothing if it doesn't fix
any issues. And AFAIU it doesn't indeed.

matthiaskrgr:
I was talking to @tobbi on irc and he wanted me to make this pull request.

tobbi:
Bbut....doesn't anyone think of the users who still have dialup?

maxteufel:
I don't think we need this for now. Please bring it up here if anyone really thinks this is necessary.

PR #679, opened 2017-01-06
[Heavily WIP][Don't merge] Async load

tobbi:
I am just posting this here to let you guys see what I am working on and maybe even try it out if you want to.

I have moved two things into new threads:
- Listing files using PhysFS
- Parsing Lisp

Since this is experimental, I just tried out the listing files method, which works fine.


tobbi:
Speed for listing files has been improved in PhysFS 3.0. I don't think I'll work on this anymore. I am still in favour of moving the initialization code to other threads where possible but I'm clearly not good enough to plan something like that.

PR #685, opened 2017-01-28
Remove old worldmap tiles

Rusty-Box:
Remove the old worldmap tiles from icy island.

### DIFF
--- a/data/images/worldmap.strf
+++ b/data/images/worldmap.strf
@@ -285,7 +285,6 @@
     37 42 44 38
     70 69 47 45
     68 71 40 8
-  (deprecated #t)
   )
   (datas
     6  10 23 30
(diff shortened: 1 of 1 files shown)


tobbi:
I can't remember the status of this. @Rusty-Box @maxteufel Does anyone of you know more?

busybody42:
Why? What if somebody wants to use them? Also, why would you waste something the artist(S) worked hard on?

Rusty-Box:
@busybody42 There are two versions of the same sprites, but with a blue background which makes them pretty useless when using the animated water tiles. So if these get removed the tiles themselves still exist, but with a transparent background.

maxteufel:
@Rusty-Box good idea in theory, but in practice the pull request how it is right *now* would break most contrib levels.

Those files from `data/images/worldmap/antarctica/`, are *all* still being the sole source of Icy Island tiles in the `worldmaps.strf` file. While they are not necessary for `ice_world.strf`, these tiles are still essential.

I think this can't really be fixed, because worldmaps using these tiles would need to be updated in any case if transparent tiles are used (since special setup is necessary with transparent tiles when you want e.g. water to show up).

Grumbel:
.strf format supports a `(deprecated #t)` tag, that will hide tiles from the (C#) editor, but keep them available to levels already using them.

tobbi:
@Rusty-Box Can you implement the (deprecated #t) thing so this can get merged?

Rusty-Box:
@tobbi If you tell me how I can try doing that.

weluvgoatz:
Is this still going to be considered?

PR #692, opened 2017-03-19
Updating the Forest World (WIP)

Rusty-Box:
We want to bring the forest world into a stable state, so we can finally add it to the story mode and progress with the story of Tux's journey to rescue Penny.

For that we have to remove old levels, revamp/improve all the others and add a boss for the forest.
Also, depending on which levels are going to be in the forest world at the end, the worldmap will also get another update, so that all levels fit into it.


Rusty-Box:
@maxteufel I will. But shouldn't we wait until we know the final amount of levels?

weluvgoatz:
I think some of the newer non-ghost forest levels should be split up. "Ancient Ruins" is definitely too long.  It took me over 8 mins and I was going quickly!

weluvgoatz:
i can help redesign some levels. i have already done a few.

maxteufel:
@weluvgoatz I think we rather need folks to suggest more ideas for the story. This is also relevant in this area, but we don't think about it enough.

I don't know how far cutscene making has progressed, @SuperTux/level-designers any updates on this?

weluvgoatz:
@maxteufel : I think I can help with the cutscenes, and the story in general.

StephenPs:
Can I add my castle level

weluvgoatz:
when will this be merged? i would love to see these levels in-game!

serano01:
When it's done. There are still some ghost forest levels missing, I think.

Alzter:
Some of these levels rely too much on gimmicks. 'Tux the Builder' sounds like a good idea in concept, but moving blocks around sluggishly is far from fun, especially with sporadic crush deaths and frustrating physics glitches usually involving stacking the blocks that make the level that much more tedious.

Another thing that needs work is the presentation. Firstly, the backgrounds. The backgrounds of the forest world as a whole aren't that good. They look blurry, stretched, and the parallax backgrounds are badly cropped.

Secondly, the tiles. The forest tiles are like the ice tiles but with dirt instead of ice and grass instead of snow on top, which isn't that realistic, or diverse. Every level looks pretty similar, if not the same.

IcyIsland had a lot of similar looking levels but made up with the caves, bridge levels and sky high platforms. The forest world has basically the same tiles in every level. The caves are just use the same ground tile as the normal forest but darkened. The ghost forest tiles are also edited from the normal ones. The trees are all the same image (which has existed since milestone 1) albeit with horizontally shrinked variants.

Overall the levels need more diversity. What about Grumbel's concept of a level in the treetops? What about a more fleshed out cave? What about a rocky ravine with logs to jump across?

This concludes my not very well structured rant. I know you guys aren't a big team and it can be difficult to implement changes, suggestions or ideas of this magnitude, so I don't expect this to be immidieately adressed, rather put into some form of consideration.

weluvgoatz:
So.... will this PR be merged for 0.6.0?

Rusty-Box:
@weluvgoatz Yes. But without the ghost forest levels for now.

PR #704, opened 2017-06-02
Split off recording code from GameSession class

### LINKED ISSUE #703: Move capture code out of GameSession class
Currently, GameSession is a very huge class. Part of it is all the capturing code. I'd prefer if we could move that part of the code into its own class so it doesn't distract from the rest of the code.

If anyone wants to volunteer and submit pull requests, go ahead. Otherwise, I'd do it in a few days.

tobbi:
Fixes #703

This refactoring allows us to split off the recording code from the GameSession class and makes the whole thing much more readable.

Please, review it carefully.

### DIFF
--- a/src/supertux/game_session_recorder.hpp
+++ b/src/supertux/game_session_recorder.hpp
@@ -1,0 +1,50 @@
+//  SuperTux
+//  Copyright (C) 2017 Tobias Markus <tobbi.bugs@googlemail.com>
+//
+//  This program is free software: you can redistribute it and/or modify
+//  it under the terms of the GNU General Public License as published by
+//  the Free Software Foundation, either version 3 of the License, or
+//  (at your option) any later version.
+//
+//  This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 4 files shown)


tobbi:
This *appears* good to go. Can someone else please take a final look at this?

On a related note: Codacy is way too strict with formatting currently. I think we should do something about that as well.

src/supertux/game_session_recorder.hpp
```diff
@@ -0,0 +1,49 @@
...
+  std::string capture_file;
+  std::istream* playback_demo_stream;
+  CodeController* demo_controller;
+
+private:
+  GameSessionRecorder(const GameSessionRecorder&) = delete;
+  GameSessionRecorder& operator=(const GameSessionRecorder&) = delete;
+};
+
+#endif
```
christ2go: Could you add a newline at the end of the file? (git diff about this)

tobbi: done.

PR #711, opened 2017-06-06
Add submission of upload URLs

maths22:
This will allow me to replace http://download.supertuxproject.org/ with content that is actually up to date.

See https://download.supertuxproject.org/view.php for a preview of the new download page.

### DIFF
--- a/.travis/package.sh
+++ b/.travis/package.sh
@@ -14,8 +14,13 @@ if [ "$TRAVIS_OS_NAME" = "linux" ] && [ "$PACKAGE" = "ON" ]; then
     mv ~/out/* .
 fi
 
+declare -A urls
+
 for file in SuperTux*; do
     echo "Uploading $file";
-    curl --upload-file "$file" "https://transfer.sh/$file"
-    echo
+    url=$(curl --upload-file "$file" "https://transfer.sh/$file")
+    urls[$file]=$url
+    echo $url
 done
+
+declare -p urls > ~/urls.dat
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -3,6 +3,10 @@ os: Visual Studio 2015
 configuration: Release
 platform: x86
 
+environment:
+  DOWNLOAD_APIKEY:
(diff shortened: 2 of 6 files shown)


appveyor.yml
```diff
@@ -28,6 +32,25 @@ after_build:
 - cmd: '"C:\Program Files (x86)\CMake\bin\cpack.exe"'
 - ps: |
     Write-Output Get-FileHash SuperTux-*
+    
+    function transfer ($filename)
+    {
+        $file = Get-Item $filename;
+        invoke-webrequest -method put -infile $file.FullName https://transfer.sh/$($file.Name)
```
maxteufel: Do we need to upload this to transfer.sh or can we link to AppVeyor, which keeps the builds for longer? If we do need to use transfer.sh, please add notes to the concerned files on the page that they are only available for 2 weeks (or remove them).

.travis/package.sh
```diff
@@ -14,8 +14,13 @@ if [ "$TRAVIS_OS_NAME" = "linux" ] && [ "$PACKAGE" = "ON" ]; then
     mv ~/out/* .
 fi
 
+declare -A urls
+
 for file in SuperTux*; do
     echo "Uploading $file";
-    curl --upload-file "$file" "https://transfer.sh/$file"
-    echo
+    url=$(curl --upload-file "$file" "https://transfer.sh/$file")
```
maxteufel: Please add notes to the concerned files on the page that they'll only be available for 2 weeks. Since Travis CI builds every week (using cron), there should always be a new one available though.

maxteufel:
I don't see any specific code explaining how this will handle non-master branches or PRs. The files should probably include this information in the build metadata part (<http://semver.org/#spec-item-10>) of the version. I would prefer that over disabling updates for non-master branches and PRs since these URLs would be useful to hand out to people to test something quickly.

appveyor.yml
maths22: AppVeyor does not support any sort of direct linking to the artifacts, which is unfortunate

.travis/package.sh
maths22: The view.php page will only show files from the last two weeks, so I can add a note, but they will drop off anyways

PR #713, opened 2017-06-19
Add first level design contest worldmap (WIP)

tobbi:
These are the levels of the first level design contest. As far as I know, I haven't weeded out all the incompatible ones.

This is just to shut up a certain person ;)


maxteufel:
This should be in the SuperTux/addons-src repo.

PR #715, opened 2017-06-21
Fix bounding box for torch and add editor property

### LINKED ISSUE #714: Torch Problem
I have noticed that the torch graphics in the game (but not the candles) have a flaw.  _The sphere of light coming from the torch is not centered around the fire!_ 

torch issue

(Using SuperTux version 0.5.1 on a Windows 7 Home Premium SP 1 platform)

tobbi:
This fixes the following things:
* Selecting / hovering over torches in the editor
* Collision detection code with Tux (torch starts burning once Tux walks past)
* This adds an editor property to control "burning"

Fixes #714

**Note:** This is a breaking change that will alter / fix the position of torches in all levels. 

@serano01 @Rusty-Box Please check whether you have torches in any of your levels and readjust their position after this commit. We *might* have to e-mail donors a fixed package if this is the case.


tobbi:
Levels that need fixing in our data directory:
https://github.com/SuperTux/supertux/search?utf8=%E2%9C%93&q=%28torch+extension%3Astl&type=
(some of them are resetpoints which can be ignored)

Add-ons that will need fixing:
https://github.com/SuperTux/addons-src/search?utf8=%E2%9C%93&q=%28torch+extension%3Astl&type=

tobbi:
By the way: I also changed the default for burning from true to false. My rationale for this is that the torch will start burning once Tux goes past it. If this is bad and breaks too many things, I am happy to undo them.

tobbi:
The torch height is exactly 100 pixels so fixing levels was fairly easy.

PR #722, opened 2017-07-19
Add "Others" category with tiles that are not in any other category.

tobbi:
Problem with this is that it shows a lot of empty tiles because we have IDs assigned to empty tiles. We should fix that.

### DIFF
--- a/src/editor/input_gui.cpp
+++ b/src/editor/input_gui.cpp
@@ -28,12 +28,13 @@
 #include "supertux/menu/menu_storage.hpp"
 #include "supertux/menu/editor_tilegroup_menu.hpp"
 #include "supertux/colorscheme.hpp"
+#include "supertux/console.hpp"
+#include "supertux/gameconfig.hpp"
 #include "supertux/globals.hpp"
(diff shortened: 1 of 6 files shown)


tobbi:
I changed the code slightly to give me the tile IDs of those blank tiles. Resulting screenshot:
https://www.dropbox.com/s/axqcroq4x306fjv/Screenshot%202017-07-20%2000.21.39.png?dl=0

maxteufel:
@tobbi That's only for the new "Others" category then? If so, that seems like a good idea.

For what it's worth, we could put this whole "Others" category behind the developer toggle.

tobbi:
I didn't change *this* code. This was only locally.

maxteufel:
I say that if that only applies to the Others category, then it would be a good idea to push it to this pull request.

Moreover, we should put the entire Others category behind the developers toggle.

tobbi:
One more thing: We should check whether certain levels use the duplicate IDs and change them to the other ones.

tobbi:
I can't remember whether this still needs work or not. Any ideas?

PR #724, opened 2017-07-31
GameMenu: add a submenu for confirming level restarts

maxteufel:
**Work in progress**, since this is not yet an option. Also maybe we should design a general ConfirmActionMenu, but that might be a bit harder.

Closes SuperTux/supertux#599 (""Restart level" button really needs an "Are you
sure?" prompt")


tobbi:
Is this only missing a review?

Issue #650, opened 2016-11-03
Use SQRat


christ2go:
The use of a squirrel binding library like SQRat will lead to the exposed classes being first class squirrel citizens, e.g. you'll be able to change attributes values from squirrel. It will further reduce code complexity and less automatically generated code.

*Changed 3rd November 21:25: Use SQRat instead of SQPlus (it's not compatible with Squirrel version 3)*

maxteufel:
@christ2go Can you tell us more about the status of this? Even if you didn't do a lot about this (or nothing at all), just tell us.

tobbi:
Superseded by #2792

PR #732, opened 2017-08-24
[Heavily WIP] SQRat

tobbi:
This is, as the title says, heavily work in progress. Currently, the following works:

Calling functions of objects that have them exposed (currently: AmbientSound and Player).

For every object there is, we create a class type sqratclass_ + <name of the object> and an instance type: sqratinstance_ + <name of the object>. I know this is totally stupid, since we should create just one class type and then instances of it when the objects are created actually, but who cares. I'll fix it later, probably.

So, for example for the menu level, you can do:
`
sqratinstance_Tux.deactivate();
`

- [ ] Get sqratthread to work with internal scheduler
- [ ] Don't pass VM to each function (use Sqrat::DefaultVM() instead)
- [ ] Expose color constants to squirrel


tobbi:
The Travis build fails because of warnings in SQRat code, not ours. *phew*

maxteufel:
Reference: SuperTux/supertux#650 ("Use SQRat")

tobbi:
Threading is bound to give us some problems. If anyone could implement the threading scripting functions, feel free. I'm struggling with them.

tobbi:
This issue was abandoned.

PR #736, opened 2017-09-24
Add swimming sprites, enhance swimming behaviour, include swimming by default.

Narre:
Add swimming sprites, enhance swimming behaviour (no more 'walking' on water), include swimming by default.


tobbi:
I think the swimming should look more penguin-like (less human-like) but that might only be me.

What does everyone else think?

Rusty-Box:
@tobbi I agree with you, since it would make the swimming ability more "unique" than (for example) from mario.

Narre:
I agree too, but it wouldn't fit the swimming currently in SuperTux, so I tried to make the images fit the moves.

tobbi:
There is no swimming currently in SuperTux, so you're free to exert your creativity as long as it's in line with our narrow understanding of how we want it ;)

Narre:
I mean the swimming that is coded in the game but not included by default. And I don't know how much I would be able to code, because I don't know much C++. The coding part of this PR was mostly done by Hume2, but we will do some C in school this year, so I should hopefully get better at it. So far, I have only done one bigger project in JavaScript, which I wouldn't call 'programming'. :)

weluvgoatz:
There is already a new (and better) swimming PR, so might this one become closed?

Rusty-Box:
Closing this due to lack of activity and not fitting the attend for how swimming in game should work.

PR #743, opened 2017-10-20
Remove extension from icon name in desktop file

volkov0aa:
According to Icon Theme Specification
https://standards.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html
extensions should not be used in icon names (LookupIcon algorithm).


tobbi:
@maxteufel @maths22 I assume this is fairly trivial. I just wanted your final review on this, since you know that stuff better than me.

maxteufel:
I *think* this should be okay, assuming you have tested it, @volkov0aa.

volkov0aa:
Try 'grep --color Icon /usr/share/applications/*' on your system. Almost all desktop files contain icon names without extensions.

tobbi:
I reviewed the spec and its example and this seems to be the way to go.

PR #746, opened 2017-11-09
Add Community Island 2016

serano01:
-Add the winner levels of last years contest with an fitting Island to the main game


tobbi:
Do we make this an add-on or not?

maxteufel:
This should be an add-on.

PR #749, opened 2017-12-23
Torch: expose to scripting interface

maxteufel:
This exposes the Torch object to the in-game squirrel scripting interface. There
are two methods as of now:

- Torch::get_burning returns the state of the flame of the torch (true for
burning, false for extinguished)

- Torch::set_burning takes a bool to set the state of the torches flame (true to
set to burning, false to extinguish)


maxteufel:
Note: this does not change in any way the default behavior of a torch that does not burn, which is to light up once the player reaches it.

Issue #545, opened 2016-08-11
Ghost Forest Torch Checkpoint Should Give Out Light


Karkus476:
(since it's a torch)

PR #750, opened 2017-12-23
Firefly: make torch checkpoint glow

maxteufel:
Since this is a torch, it should be glowing. The glow is only enabled for
checkpoints using the torch sprite (it checks for "torch" in the file name of
the sprite).

~~**To discuss:** Should we keep using the huge light sprite from the Torch object?~~

Closes SuperTux/supertux#545 ("Ghost Forest Torch Checkpoint Should Give Out
Light")

### DIFF
--- a/src/object/firefly.cpp
+++ b/src/object/firefly.cpp
@@ -20,12 +20,15 @@
 
 #include "audio/sound_manager.hpp"
 #include "math/random_generator.hpp"
+#include "math/vector.hpp"
 #include "object/player.hpp"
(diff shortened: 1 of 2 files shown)


maxteufel:
This is how it currently looks in a level...

Preview of the change in a level

I have locally tried using other sprites, but those are in a whiter light and off-center (that could be fixed I think, but I don't know how).

Rusty-Box:
@maxteufel imo the light radius could be a little bit smaller.

PR #760, opened 2018-02-09
Automatically update translation templates

tobbi:
(I'm hereby testing the automatic translation template updater, using my own translations repo for now).


tobbi:
Well, this doesn't work. If anyone who knows travis well enough knows what's wrong, feel free to tell me.

christ2go:
You must add transifex-client, pip3, easy_install3 and all the other packages the added scripts are using as a dependency.

maxteufel:
This is the most terrible way of doing this. If anything, the translation templates should be (re)generated locally when changing some source code. If we put the translations/ repo content back into our main repo (I was waiting on @maths22's repo reorganization plans for this), then we can easily add this into CMake to automagically handle this when compiling to test the changes.

christ2go:
@maxteufel I only took a look at the travis run (https://travis-ci.org/SuperTux/supertux/jobs/339744411). It complains about those things, but you're right, doing this locally is better.

maxteufel:
Oh ­­- I was just looking at this pull request. That's a different branch.

tobbi:
This was a test. And you're right. It is horrible, that's why I closed it. I will work on the transifex client thingy and check that out.

PR #765, opened 2018-02-11
Spelling: Fractions for tiles

comradekingu:


tobbi:
Does this work with our font?

brmbrmcar:
It does not.

tobbi:
Okay, then this depends on our SDL_ttf branch and will be merged when that's ready.

HybridDog:
As far as I know, it is generally a bad idea to use Unicode characters for text which can be written in ASCII. Non-ASCII characters are often missing in some font files, whereas printable ASCII characters are usually always available. 
Furthermore, searching and replacing small pieces of text to make them look slightly better and then reviewing these changes is a waste of time I think.

⅛ is missing in the fonts:
2019-10-11-133347_1920x1080_scrot
2019-10-11-133350_1920x1080_scrot

comradekingu:
@HybridDog Not the biggest deal one way of the other. Generally it is a good idea, that is what they are for. Websafe colours are similarly no longer needed. I think the issue is fixing the font.

HybridDog:
> I think the issue is fixing the font.

It needs to work with all fonts in which the text may be shown. As long as ⅛ is a just a question mark, this cannot be merged.

busybody42:
We should add some characters, like the 1/8 character, to the fonts!

busybody42:
shadow-big-1slash8
white-big-1slash8

tobbi:
> shadow-big-1slash8
> white-big-1slash8

That is incredibly hard to read!!

busybody42:
... I used piskel, ok?
I used the numbers from the small font, since there's no tiny 8 yet...
I was gonna make a pull request to add in those sprites...

HybridDog:
> I was gonna make a pull request to add in those sprites...

If this should be merged, you also need to add the character to the ttf fonts (maybe with fontforge) or add a fallback functionality to the source code so that letters unknown to the SuperTux font are shown in e.g. dejavu sans mono.

busybody42:
I mean, adding in the missing character 1/8 that I made for this pull is a good start!

slackstone:
I will personally close this ticket when a $26.00 donation comes into the ST Dev donation queue. There needs to be chocolate involved for a fix like this.

weluvgoatz:
Seen as no progress has really been made on this for the longest time, I assume it's not going to, and this is an unnecessary feature anyway.  The fractions are also very hard to read.

comradekingu:
@weluvgoatz I see, the alternative would be "half tile", "quarter tile" and "half quarter tile", but that isn't as good. Small bigger biggest or "Small, medium, big" might work though?

tobbi:
> @weluvgoatz I see, the alternative would be "half tile", "quarter tile" and "half quarter tile", but that isn't as good. Small bigger biggest or "Small", "medium", "big" might work though?

That's actually a really good idea in my opinion. Can we do it like this?

comradekingu:
@tobbi Edits allowed for maintainers for then one you meant. I prefer the latter, since it is easily understood by kinds (?)

tobbi:
@comradekingu: I meant "Can you do it like this?" obviously.

I also agree with the other sentiment.

PR #766, opened 2018-02-11
Spelling: ellipsis …

comradekingu:


brmbrmcar:
It works, but it doesn't look as nice in my opinion.

HybridDog:
I think this message is no longer used. Can you close this?

PR #767, opened 2018-02-11
Spelling: Tux doll

comradekingu:


tobbi:
I actually think that all menu items should be capitalized here.

tobbi:
@comradekingu Would it be possible to capitalize all menu items in that menu? If not, I'll merge this as is in a few days.

comradekingu:
@tobbi The nature of menus is you can either do them
Like this
And this

Or

Like This
And This

The latter only allowed in English, translates badly to having correct translations.
How it pans out with cognition is the former is easier right away, and the latter more prominent long term, since the shape has more prominently different shapes given their bigger size.

Whether that really comes into play, I have never seen any data on, and it could be argued unlikely that the amount of use this menu sees ever takes it there.

Instead one can add pictograms to it, but then those are slower than just text, and even pictograms alone, losing still the fast results right away.

PR #769, opened 2018-02-15
[WIP] Add add-on wizard

### LINKED ISSUE #717: Export to add-on option for the editor
We desperately need an "Export to add-on" option for the editor.

This is, I think, the third time people tried to load a ZIP file with only levels by putting it in the add-ons directory.

What's missing is an NFO file. Creating that should be fairly trivial. I am not sure about creating a ZIP file from SuperTux, though. I'll investigate.

christ2go:
Users can now export add-ons as a zip directly from the level editor.

Adds 2 new dependencies:

1 . libzip (For opening and writing to zip archives).
2. zlib  (For zip compression)
Uses a fork of libzip which does not require gzip as a library. 
The add-on is saved in the root supertux directory.
**TODO **
- Set correct type of add-om

Closes #717


maxteufel:
Some concerns:

- How do you expect this to compile? There's no integration with the build system whatsoever.
- Please depend on them dynamically. I don't want to see the bazillionth SuperTux fork of whatever. If libzip depends on gzip, then it depends on gzip. That's not our problem.
- Yay, another *required* dependency. Nope. Please add an option to disable it. It's not necessary for everyone, and adding this few code behind some #if guard isn't hard.

Now to more SuperTux-related concerns:

- This should figure out what mixed in dependencies the add-on needs, and pack them. This is probably where physfs fails, because it's not necessarily what it was made for.

src/supertux/menu/editor_addon_wizard.cpp
```diff
@@ -0,0 +1,122 @@
...
+      PHYSFS_close(fhandle);
+      zip_source_t * zips = zip_source_buffer_create(buffer,stat.filesize,1,&er);
+      zip_file_add(arch,dateiname.c_str(),zips,ZIP_FL_ENC_UTF_8);
+    }
+    int e = zip_close(arch);
+    if(e != 0)
+    {
+      log_debug << "Error saving" << zip_error_strerror(zip_get_error(arch)) << std::endl;
+    }
+    // Create nfo file
```
maxteufel: ?

src/supertux/menu/editor_addon_wizard.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+    // Save the archive
```
maxteufel: ?

src/supertux/menu/editor_addon_wizard.cpp
```diff
@@ -0,0 +1,122 @@
...
+  add_entry(MNID_PACK,_("Pack Addon"));
+  add_back(_("Back"));
+}
+
+void EditorAddonWizard::menu_action(MenuItem* item)
+{
+  if(item->id == MNID_PACK)
+  {
+    // First write the NFO-File
+   std::stringstream   nfostream;
```
maxteufel: Some spacing got messed up here.

maxteufel:
Things that need work now:

- Depend on libzip optionally
- Integrate with Windows builds, too
- Code Style (Structure code properly, English variable names, …)
- Some more user feedback? The add-on zip is dumped into $PWD without any feedback to the user

maths22:
Once everything else is good, I can take care of windows build integration.

tobbi:
A couple remarks:

1) For some reason, the "Pack addon" button doesn't do anything. All I can see in the console is
```
Starting to create the zip file levels/My_Levelset
```

2) Can we pre-propagate the addon's ID and name from the levelset / worldmap name?

3) Not particularly an issue with your add-on but something to keep in mind: Should we change levelset's license selection from string entry to select menu?

src/supertux/menu/editor_addon_wizard.cpp
```diff
@@ -0,0 +1,128 @@
...
+      log_debug << "Error creating zip file" << std::endl << zip_error_strerror(&e) << std::endl;
+    }
+
+    // Create nfo file
+    std::string nfo = nfostream.str();
+    std::string adf = id+".nfo";
+    zip_source_t * zipl  = zip_source_buffer_create(nfo.c_str(),nfo.size(),0,&er);
+    zip_file_add(arch,adf.c_str(),zipl,ZIP_FL_ENC_UTF_8);
+
+    for (auto dateiname:f)
```
tobbi: Say '''no''' to unnecessary copying.

Please use 
```
for (const auto& filename : f)
```

src/supertux/menu/editor_addon_wizard.cpp
```diff
@@ -0,0 +1,128 @@
...
+#include "util/writer.hpp"
+#include <zip.h>
+
+static PHYSFS_EnumerateCallbackResult printDir(void *data, const char *origdir, const char *fname)
+{
+    PHYSFS_Stat sr;
+    std::vector<std::string>* liste = (std::vector<std::string>*) data;
+    std::string realf = FileSystem::join(origdir,fname);
+    PHYSFS_stat(realf.c_str(),&sr);
+    if (sr.filetype == PHYSFS_FILETYPE_DIRECTORY)
```
tobbi: Please use https://github.com/SuperTux/supertux/blob/master/src/physfs/physfs_file_system.cpp#L47

PR #772, opened 2018-02-16
ScriptTrigger: add oneshot option

maxteufel:
This allows the creation of a ScriptTrigger that only runs once. After running
for the first time, it is disabled.

Reference: <https://forum.freegamedev.net/viewtopic.php?f=67&t=6687&p=76020#p76017>
Suggested-by: @manuel-st

### DIFF
--- a/src/trigger/scripttrigger.cpp
+++ b/src/trigger/scripttrigger.cpp
@@ -31,7 +31,9 @@ ScriptTrigger::ScriptTrigger(const ReaderMapping& reader) :
   triggerevent(),
   script(),
   new_size(),
-  must_activate(false)
+  must_activate(false),
+  oneshot(false),
+  runcount(0)
 {
   reader.get("x", bbox.p1.x);
   reader.get("y", bbox.p1.y);
@@ -41,6 +43,7 @@ ScriptTrigger::ScriptTrigger(const ReaderMapping& reader) :
   bbox.set_size(w, h);
   reader.get("script", script);
   reader.get("button", must_activate);
+  reader.get("oneshot", oneshot);
   if(script.empty()) {
     log_warning << "No script set in script trigger" << std::endl;
   }
@@ -55,7 +58,9 @@ ScriptTrigger::ScriptTrigger(const Vector& pos, const std::string& script_) :
   triggerevent(EVENT_TOUCH),
   script(script_),
   new_size(),
-  must_activate()
+  must_activate(),
+  oneshot(false),
+  runcount(0)
 {
   bbox.set_pos(pos);
   bbox.set_size(32, 32);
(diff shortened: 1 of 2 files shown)


src/trigger/scripttrigger.cpp
```diff
@@ -55,7 +58,9 @@ ScriptTrigger::ScriptTrigger(const Vector& pos, const std::string& script_) :
   triggerevent(EVENT_TOUCH),
   script(script_),
   new_size(),
-  must_activate()
+  must_activate(),
+  oneshot(false),
+  runcount(0)
```
maxteufel: I am not sure if initializing these here to those values is the right thing to do. @tobbi, what can you say about this?

tobbi: Yes, this is th right thing to do. Why wouldn't it?

src/trigger/scripttrigger.cpp
```diff
@@ -94,7 +100,12 @@ ScriptTrigger::event(Player& , EventType type)
   if(type != triggerevent)
     return;
 
+  if (oneshot && runcount >= 1) {
```
tobbi: If we're only ever having two conditions (runcount 1 or 2), why not turn this into a bool?

src/trigger/scripttrigger.cpp
maxteufel: Because, compared to the first constructor, where there is `must_activate(false)`, `must_activate()` is used here. What does leaving out the argument do?

tobbi: Use the default value for the data type, which, coincidentally happens to be ```false```.

src/trigger/scripttrigger.cpp
maxteufel: Because, a) maybe in the future we could expose the run count to the script, even though I don't know why this would be useful, and b) because I found it easier to increment a counter. Also, the states are not 1 or 2, but essentially 0 (not triggered) or 1 (triggered, which now means disable).

tobbi:
I'm not sure whether the label "Oneshot" really conveys what this does. How about "Run only once" ? I don't know. Maybe someone else can think of a better label.

maxteufel:
This is what's used for platform path modes too. So out of consistency, I used that. Also, this is a proper English word describing exactly what we mean: <https://en.wiktionary.org/wiki/one-shot#Alternative_forms>, <https://en.wiktionary.org/wiki/one-shot#Adjective>.

src/trigger/scripttrigger.cpp
tobbi: Maybe you should move the 1 into a private max_runcount variable? Increases visibility and makes it easier to expose to scripting later on.

PR #773, opened 2018-02-16
Forest Overground and Underground Tile Improvements

jonfx1007:
This expands the forest underground tileset in a slightly new style; also fixes some small graphical and tiling errors in overground tiles, as well as adding upside-down slopes
Based on requests from the Supertux "Tileset Improvements" project guide


Rusty-Box:
These look amazing. Good job!

maxteufel:
I can only agree with @Rusty-Box.

Thanks @jonfx1007 for this amazing first-time contribution. I suppose we'll play around with this change for a few days and then merge it!

(To make sure, I hope you do agree with the licensing on those to match the default CC-BY-SA and GPLv2+ dual license? Feel free to add yourself to `data/AUTHORS`, too)

jonfx1007:
Thank you guys so much! And yes, I will license all my work under any license you need for it to be accepted. I plan to work on more tileset work in the future, following the requests in the "Tileset Improvements" project guide like I mentioned. I actually already have work for the snowcastle and browncastle tilesets, which I will upload as soon as this one goes through.
Just fyi, I'm user Jonathan on the forums.

maxteufel:
That sounds great. I've been playing around some Forest levels (Forest Island as currently in development releases, and some add-ons) and been seeing some problems. I'm not sure if *all* of them are caused by this change, but I'm pretty sure some of them are.



This one's from "Up and Down" in Forest Island. There are some tiling issues in the ground on the bottom left.

Furthermore, if you look a bit more right (the wall left to where Tux is standing, there is some pretty serious tiling issue).

And if you look to the wall left of the bomb, there is, at the bottom some grass: this seems either too vertical for grass (add some small curves), or it is another tiling issue.



Here ("Bye Bye Forest", Forest Island) you can see that last issue from above level a bit more.



Some level from Serano's Islands. The changed forest slopes don't match the old normal tiles anymore.

jonfx1007:
Ok, I've been investigating what you said. 

Those slopes in the third picture aren't the ones I changed; I changed the 45 degree slopes; these are the 30 degree slopes. These ones will probably also need redesigned a bit.

I'm _pretty_ sure the tiling issues in the top picture were just bad tiling on the part of the original author.

Now as for the grass ending abruptly in a wall. That's because the proper tiles for ending the grass didn't exist yet. 

This is an example of what a level would look like with the new tiles, before changing the level at all:
screenshot005


You would need to use these tiles (the ones to the left of the mouse):
screenshot007

to make it look like this:
screenshot008

This will just require some minor redesigns of the levels. I could do it if you want, but I figured I'd leave it to @Rusty-Box since the Forest World is kind of his project.

maxteufel:
> I'm pretty sure the tiling issues in the top picture were just bad tiling on the part of the original author.

I agree. That's what I suspected too in this case.

For the other cases, I see that now. I unfortunately didn't compare, but only looked at what could be wrong.

Also, it should be indeed the task for @SuperTux/level-designers to make the levels use the new tiles.

maxteufel:
Thanks again @jonfx1007 for the great work!

If you tell us what tiles exactly from "Tileset Improvements" this adds, we can remove them/mark them as done in the project.

jonfx1007:
It's under Projects -> Tileset Improvements -> Underground -> Items 1 and 4 ( "missing slopes" and "concave tiles").

Issue #287, opened 2015-12-22
Bad performance in Windows10


Grumbel:
Even so the game runs at around 66fps according to the FPS counter it feels pretty sluggish. I assume the issue is that the game isn't vsync'ing or something along the lines.

OronDF343:
I can confirm that VSync is needed. There is visible tearing which makes some levels like "Oh no! More snowballs" painful to play

tobbi:
@Grumbel I assume this can be closed, right?

PR #780, opened 2018-02-22
Remove vsync option

brmbrmcar:
It isn't consistent between systems, and can lead to unfair advantages such as in speedrunning.


maxteufel:
As far as I understand this, it it designed to improve the gameplay experience.

SuperTux is a game in the first place, not solely targeted to speedrunning. Hence, we want to improve gameplay for everyone. Thus I don't see a need to remove this.

brmbrmcar:
Vsync doesn't improve it though, especially when it doesn't work on half of all systems. Even with it off, there doesn't seem to be any tearing to me. It isn't just hardcore speedrunners affected; if anyone wanted to try something together on different systems it could cause problems because one runs at a different rate.

brmbrmcar:
It also varies difficulty, because it then runs at a different speed.

maxteufel:
Closing as per https://github.com/SuperTux/supertux/pull/780#issuecomment-368002272 (and citing https://github.com/SuperTux/supertux/issues/287#issue-123479644) after discussion with the team.

PR #697, opened 2017-04-12
Zeeklings screech and have curved flight path

wbrawner:
* Added new screech sound
* Track Zeekling's and player's positions before diving
* Modified Zeeklings y velocity on dive based on distance they've moved

I'm not so sure about the math here, so any suggestions on how to improve it would be very welcome. Also, the screech sound came from here: https://opengameart.org/content/15-monster-gruntpaindeath-sounds There's a link there to the publication of the sounds in which the author says the sounds are free to use but he requests attribution: http://cubeengine.com/forum.php4?action=display_thread&thread_id=2164 so I'm not sure if this can be used here or not but I figured I'd give it a shot anyways

PR #784, opened 2018-03-05
Zeekling: Add curved flight path and screaming

### LINKED ISSUE #533: Zeeklings: Curved flight path + screech
Zeeklings should curve down to be on the same level as the player, and then level off, to give the player more warning that they are coming. This should be accompanied by a screech as they attack, so that the player is not surprised.

christ2go:
Refinement of wbrawners pull request #697.
Adds curved flight path. 

The new flight path is a parabola f(x) = ax^2+bx+c.
The coefficients are calculated by solving a 3x3 linear system of equations using the gauss algorithm.

This probably changes the difficulty of many levels, so please test e.g. with Down the Rabbit Hole. 
Difficulty can be adjusted by changing the interval of player distance in which the zeekling will start flying down.

Closes #533


src/badguy/zeekling.cpp
```diff
@@ -182,18 +188,120 @@ Zeekling::should_we_dive() {
 
   return false;
 }
+/**
```
tobbi: I think the maths functions are better suited as part of the vector class. 

@maxteufel What do you think?

maxteufel: Yes, they should be separated.

busybody42:
Uhh... I think there was already another issue for the same thing...

Issue #793, opened 2018-03-25
Weird blocks physics


Raoul1808:
**SuperTux version: supertux2 v0.5.1-404-g1079dddc6**
**System information: Windows 10** *(details about your kernel, operating system, or distribution)*

I'm creating a "puzzle" level, where first the player has to activate a switch wich moves a platform holding some stone blocks. Once the switch activated, the stone blocks falls and goes to a pnematic platform (I don't know the english name). But when the stone blocks touches the ground, the pnematic platform goes up, and one of the blocks goes trough the platform, and is stuck un the ground, slowly descending to the void.

I join a link to the video showing the issue.
[It's here](http://www.mediafire.com/file/r1nfw1521i255cw/supertux2%20v0.5.1-404-g1079dddc6%2025_03_2018%2011_53_43.mp4)

Could you fix that?

Raoul1808

tobbi:
Our physics engine is buggy. I'm xref'ing #794 for using Box2D as a replacement.

busybody42:
We can just code the penumatic platforms to be affected by other objects and make the rising platform push ALL the stuff on it!

Alasdairbugs:
needs information, can this be reproduced in modern engine

PR #796, opened 2018-03-27
Physics: Add quadtree collision detection

christ2go:
Uses quadtree data structure for more efficient collision detection between moving objects.
(Experiemental results: Reduced number of collision tests by 1000 or more per frame)

A quadtree partitions the sector into 4 quadrants. Every node stores a list of objects in the subtree. Collisions only have to be checked against the objects in the same subtree. Whenever a node has more than 10 objects, it wil split.

### DIFF
--- a/src/math/quadtree.cpp
+++ b/src/math/quadtree.cpp
@@ -1,0 +1,133 @@
+//  SuperTux -  A Jump'n Run
+//  Copyright (C) 2018 Christian Hagemeier <christian@hagemeier.ch>
+//
+//  This program is free software: you can redistribute it and/or modify
+//  it under the terms of the GNU General Public License as published by
+//  the Free Software Foundation, either version 3 of the License, or
+//  (at your option) any later version.
+//
+//  This program is distributed in the hope that it will be useful,
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//  GNU General Public License for more details.
+//
+//  You should have received a copy of the GNU General Public License
+//  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#include "math/quadtree.hpp"
+
+QuadTree::QuadTree(const Rectf& r, int level)
+        : m_level(level),
+        m_objects(),
+        m_onlyhere(),
+        m_bounds(r),
+        next()
+{
+        for (int i = 0; i < 4; i++)
+                next[i] = NULL;
+}
+
+void QuadTree::insert(MovingObject *obj) {
+        if (next[0] != NULL) {
+                int nindex = getIndex(obj);
+                if (nindex != -1) {
+                        next[nindex]->insert(obj);
+                        return;
+                } else {
+                        m_onlyhere.push_back(obj);
+                }
+        }
+        m_objects.push_back(obj);
+        // Is our capacity fully used?
+        if (m_objects.size() > MAX_OBJECTS && m_level < MAX_LAYERS) {
+                split();
+                // Insert every object in respective quadrant.
+                for (const auto& nobj:m_objects) {
+                        int nindex = getIndex(nobj);
+                        if (nindex != -1) {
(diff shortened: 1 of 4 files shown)


busybody42:
It may cause this to happen:
10 enemies of another type are in 1 spot
*BOOM* (many die from being in a crowded spot)

christ2go:
No that won't happen. The code does not change the way collisions are handled, it only changes which objects are selected to be checked against another.

busybody42:
but what if there the game doesn't notice tux jumped on the enemy and he fell through it for a frame and then died unfairly?

christ2go:
Sorry, but I'm not able to understand how _this_ pull request will introduce such behaviour. If you experienced such problems, please open an issue about it. If you have not, could you please elaborate further how the changes in _this pull request_ will allow such situations to happen?

src/supertux/moving_object.hpp
```diff
@@ -161,6 +161,11 @@ class MovingObject : public GameObject
     bbox.set_size(w, h);
   }
 
+  Rectf get_dest()
```
tobbi: Please mark this const.

src/math/quadtree.cpp
```diff
@@ -0,0 +1,137 @@
...
+        m_onlyhere.push_back(obj);
+      }
+  }
+  m_objects.push_back(obj);
+  // Is our capacity fully used?
+  if(m_objects.size() > MAX_OBJECTS && m_level < MAX_LAYERS)
+  {
+    split();
+    // Insert every object in respective quadrant.
+    for(auto it = m_objects.begin();it != m_objects.end(); it++)
```
tobbi: Can you use the shorter
```
for(const auto& nobj : m_objects)
```
or something to that effect?

The thing is: I can never remember whether the above long way creates copies or not.

src/supertux/sector.cpp
```diff
@@ -1003,68 +1004,74 @@ Sector::handle_collisions()
...
 
-  // part3: COLGROUP_MOVING vs COLGROUP_MOVING
-  for(auto i = moving_objects.begin(); i != moving_objects.end(); ++i) {
-    auto moving_object = *i;
-
-    if((moving_object->get_group() != COLGROUP_MOVING
-        && moving_object->get_group() != COLGROUP_MOVING_STATIC)
-       || !moving_object->is_valid())
+// part3: COLGROUP_MOVING vs COLGROUP_MOVING
+for(auto i = moving_objects.begin(); i != moving_objects.end(); ++i) {
```
tobbi: Now that you decoupled the i2 iterator from the i iterator, both can use the shorter syntax :)

src/math/quadtree.cpp
```diff
@@ -0,0 +1,137 @@
...
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//  GNU General Public License for more details.
+//
+//  You should have received a copy of the GNU General Public License
+//  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#include "math/quadtree.hpp"
+
+QuadTree::QuadTree(Rectf r, int level):
```
tobbi: Kann das per Ref übergeben werden?
```
const Rectf& r
```

tobbi:
@christ2go @maxteufel Can this get merged?

maxteufel:
No. Code style still looks terrible (i.e. mixed up between old and new files, missing or too much spacing/newlines).

busybody42:
(To the March 29 comment)

...like the "bad guy turns around on slopes greater than 22.5°" bug?

christ2go:
It's (sadly) not just that, we also have some problems with moving platforms (#793) , moving solid objects (i believe there is a forum discussion on it), jittering objects and some other physics problems. 

We are currently thinking about ways to improve our physics, so if you have any idea how we could achieve better physics, they are always welcome!

christ2go:
Closing this for now. Will be replaced by something like spatial hashing in the new physics engine (and will need modifications for that).

PR #804, opened 2018-04-19
Add option to delete levelsets

### LINKED ISSUE #735: Add a button where you can delete a levelset automatically in the SuperTux Level Editor.

christ2go:
Allows levelsets to be deleted from the editor.

Implements a new menu type, the ConfirmDialog. 
ConfirmDialogs contain a query (e.g "Should this levelset be deleted") and a lambda function,
which is called when the query is confirmed. 

Deletes the levelset by recursively walking the directory and then deleting every single file.

TODO: Only delete user created levelsets (does anyone have a good idea, how this check could be done?).

Issue: Because the levelset menu can not be reloaded, the deleted levelset will still show up.  Don't know how to fix this with the current menu editor design.

Closes #735.

### DIFF
--- a/src/supertux/menu/confirm_dialog.cpp
+++ b/src/supertux/menu/confirm_dialog.cpp
@@ -1,0 +1,42 @@
+#include "supertux/menu/confirm_dialog.hpp"
+#include "gui/menu.hpp"
+#include "gui/menu_item.hpp"
+#include "gui/menu_manager.hpp"
+#include "supertux/menu/menu_storage.hpp"
+#include "util/gettext.hpp"
+
+ConfirmDialog::ConfirmDialog(const std::string& title, std::function<void(void)> callback):
+  m_title(title),
+  m_callback(callback)
+{
+  // Initialize the dialog 
+  add_label(_("Confirm dialog"));
+  add_label(title);
+  add_hl();
+  add_entry(1, _("Yes"));
+  add_entry(2, _("No"));
+}  
+
+ConfirmDialog::ConfirmDialog(const std::string& title, std::function<void(void)> callback, std::string& yes, std::string& no):
+  m_title(title),
+  m_callback(callback)
+{
+  // Initialize the dialog 
+  add_label(_("Confirm dialog"));
+  add_label(title);
+  add_hl();
+  add_entry(1, _(yes));
+  add_entry(2, _(no));
+}  
+void ConfirmDialog::menu_action(MenuItem* item)
+{
+  if(item->id == 1)
+  {
+    m_callback();
+    MenuManager::instance().pop_menu();
+  }else if(item->id == 2)
+  {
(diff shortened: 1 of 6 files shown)


maxteufel:
> TODO: Only delete user created levelsets (does anyone have a good idea, how this check could be done?).

Well, what does happen when you attempt to do that? I mean, it shouldn't be writing to the main directories anyway.

christ2go:
The game freezes

tobbi:
/home/travis/build/SuperTux/supertux/src/supertux/world.cpp: In constructor ‘World::World()’:
/home/travis/build/SuperTux/supertux/src/supertux/world.cpp:94:1: error: ‘World::m_removed’ should be initialized in the member initialization list [-Werror=effc++]
 World::World() :
 ^

src/supertux/menu/confirm_dialog.cpp
```diff
@@ -0,0 +1,42 @@
...
+  add_hl();
+  add_entry(1, _(yes));
+  add_entry(2, _(no));
+}  
+void ConfirmDialog::menu_action(MenuItem* item)
+{
+  if(item->id == 1)
+  {
+    m_callback();
+    MenuManager::instance().pop_menu();
```
tobbi: pop_menu is called in both places. You can actually simplify it to only call m_callback in case of item->id == 1 and then afterwards call pop_menu in any case.

src/supertux/menu/confirm_dialog.cpp
```diff
@@ -0,0 +1,42 @@
...
+{
+  // Initialize the dialog 
+  add_label(_("Confirm dialog"));
+  add_label(title);
+  add_hl();
+  add_entry(1, _("Yes"));
+  add_entry(2, _("No"));
+}  
+
+ConfirmDialog::ConfirmDialog(const std::string& title, std::function<void(void)> callback, std::string& yes, std::string& no):
```
tobbi:
```
const std::string& yes, const std::string& no
```

src/supertux/world.cpp
```diff
@@ -231,6 +232,36 @@ World::save(bool retry)
...
+    std::vector< std::string > files = fs.open_directory(dir);
+    for(const auto& fname : files)
+    {
+      std::string fullp = FileSystem::join(dir, fname);
+      toDelete.push(fullp);
+    }
+  }
+  // Delete in reverse order, so that directories are only 
+  // deleted, if they are empty
+  for(auto it = fileNames.rbegin(); it != fileNames.rend(); ++it)
```
tobbi: Can you use short syntax for this?
```
for(const auto& filename : fileNames)
{
}
```

weluvgoatz:
I am giving these labels because conflicting files.

busybody42:
maybe make it possible for a week after they were deleted to recover the levelsets you deleted...

PR #807, opened 2018-05-11
Adds Pandora support

rajdakin:
This pull request adds the OpenPandora support.
The OpenPandora is a little, low-speed computer. See [pyra-handheld.com/boards/pages/pandora/](https://pyra-handheld.com/boards/pages/pandora/) for more details.


maxteufel:
Thank you for the contribution!

However, we cannot merge this (as per a discussion in the developer team) for two reasons:

1. This looks like an extremely bad way of adding "support" for a platform. Changing this parameter statically, even more so with a compile-time switch is the wrong way of doing this.
2. We cannot officially support small platforms that we don't have access to, since we can't test stuff.

PR #808, opened 2018-05-11
Fixed glew on wayland.

deveee:
Glew can't open glx display when it's running on wayland session and thus returns an error. But glXGetProcAddress is fully usable on wayland, so we can just ignore the "no glx display" error.

Note that GLEW_ERROR_NO_GLX_DISPLAY needs glew >= 2.1. Older versions assume that glx display is always available and will just crash.


tobbi:
@maxteufel Do you have any further remarks before I merge this?

PR #828, opened 2018-06-01
Editor has no friend classes (Fix #786)

### LINKED ISSUE #786: Get rid of "friend" overuse in editor code
This is Bill. Bill does not use potentially unsafe usage of "friend" in modern C++ (such as we're using in SuperTux). Bill is modern. Be like Bill.

My brain hurts every time I see friend overuse. Let's use public getters and setters instead. That way we can decouple the editor subclasses as well.

This is up for grabs. I might do it once my brain is done hurting.

leper:


src/editor/editor.cpp
```diff
@@ -486,6 +491,19 @@ Editor::sort_layers() {
...
+void
+Editor::select_tilegroup(int id) {
+  tileselect.active_tilegroup.reset(new Tilegroup(tileset->tilegroups[id]));
+  tileselect.input_type = EditorInputGui::IP_TILE;
+  tileselect.reset_pos();
+  tileselect.update_mouse_icon();
+}
+
+const std::vector<Tilegroup>&
+Editor::get_tilegroups() {
```
tobbi: Mark the whole method const as well.

src/editor/editor.cpp
```diff
@@ -500,6 +518,19 @@ Editor::change_tileset() {
...
+void
+Editor::select_objectgroup(int id) {
+    tileselect.active_objectgroup = id;
+    tileselect.input_type = EditorInputGui::IP_OBJECT;
+    tileselect.reset_pos();
+    tileselect.update_mouse_icon();
+}
+
+const std::vector<ObjectGroup>&
+Editor::get_objectgroups() {
```
tobbi: Mark this method const as well.

src/editor/editor.hpp
```diff
@@ -135,12 +142,35 @@ class Editor : public Screen,
     void delete_markers();
     void sort_layers();
 
+    void select_tilegroup(int id);
+    const std::vector<Tilegroup>& get_tilegroups();
```
tobbi: s.a.

src/editor/editor.hpp
```diff
@@ -135,12 +142,35 @@ class Editor : public Screen,
...
     void change_tileset();
 
+    void select_objectgroup(int id);
+    const std::vector<ObjectGroup>& get_objectgroups();
```
tobbi: s.a.

PR #829, opened 2018-06-03
Remove a few default non-virtual dtors.

leper:
Add and remove a few related includes.


leper:
This might have some overlap with the dtor commit in #828. In case this is merged before that one, it would make sense to change the commit message of the other commit.

tobbi:
Can you rebase with latest master, please?

PR #848, opened 2018-06-27
Added a map constructor for Electrifier

### LINKED ISSUE #846: "Electrifier" should take a map of changing tiles
Currently, we add one game object per tile that needs to be changed by the electrifier. This is not ideal.

The way I imagine it, it should take an std::map data structure that should look like this:
```
{
  tile_id_from_1: tile_id_to_1,
  tile_id_from_2: tile_id_to_2
  ...
}
```
This is the file that needs to be changed:
https://github.com/SuperTux/supertux/blob/master/src/object/electrifier.cpp and the corresponding header file.

So, add a constructor that takes a map. The old uint32_t, uint32_t constructor should be changed so that it calls the other constructor. Besides, all the calls to the Electrifier constructor should get changed so that they use the new syntax.

This work would help me with integrating the storm cloud graphics from #405

InsightGit:
This pull request addresses issue number 846, where it was suggested that the Electrifier class takes a map of uint32_ts so that one Electrifier can electrify multiple tiles. This pull request implements said functionality, updates existing uses of the past constructor, and keeps the old constructor (this time calling the new map constructor). Please note that this is not fully tested(although it compiled on my Ubuntu machine) as I'm not exactly sure how to test it.

Closes #846


tobbi:
In my opinion this looks good to merge. i'd still like a second review first.

src/object/electrifier.cpp
```diff
@@ -17,20 +17,30 @@
...
+{
+  Electrifier(change_map, seconds);
 }
 
 void
 Electrifier::update(float )
 {
   if (duration.check()) {
-    Sector::current()->change_solid_tiles(change_to,change_from);
+    for(auto &tile : change_map){
```
christ2go: We don't have to use const here, @tobbi?

src/badguy/kugelblitz.cpp
```diff
@@ -142,8 +142,7 @@ Kugelblitz::active_update(float elapsed_time)
     }
 
     if (is_in_water()) {
-      Sector::current()->add_object( std::make_shared<Electrifier>(75,1421,1.5));
-      Sector::current()->add_object( std::make_shared<Electrifier>(76,1422,1.5));
+      Sector::current()->add_object( std::make_shared<Electrifier>(TileChangeMap({{75,1421},{76,1422}}),1.5));
```
christ2go: Maybe add some spacing

src/object/electrifier.hpp
```diff
@@ -19,13 +19,22 @@
...
 #include <stdint.h>
 
+namespace {
+  /**
+   * A one-to-one relation that maps one tile ID with another
+   */
+  typedef std::map<uint32_t, uint32_t> TileChangeMap;
+}
+
 //Changes all tiles with the given ID to a new one for a given amount of time, then removes itself
```
christ2go: Please update this documentation.

src/object/electrifier.cpp
christ2go: And as I said in the general comment change it to auto& tile

tobbi: Well, I think we can use ```const``` since it doesn't change any of the contents of the array.

Issue #797, opened 2018-03-29
Ability to turn Dispensers on and off with scripting?


165your4:
This would be really useful.

PR #850, opened 2018-06-27
Added ability to enable/disable Dispensers with scripting

AshishBhattarai:
[https://github.com/SuperTux/supertux/issues/797](url)


tobbi:
@AshishBhattarai The reason this doesn't work is because the name of the dispenser, although saved, does not get read from the level file when the file is loaded.

Add the following to the ```Dispenser``` constructor that takes a ```ReaderMapping&```:
```
  reader.get("name" , name, "");
```

This reminds me: We shouldn't read the name property in the sub classes of GameObject, but should rather add that code snippet to the GameObject class itself and add a constructor there that takes a ReaderMapping instance and sets the name property there. But that's something for a followup PR.

AshishBhattarai:
Thanks but it still doesn't work. I still get the warning saying "he index 'NAME' does not exist".

PR #853, opened 2018-06-29
Added support for multiple players configs in settings file

### LINKED ISSUE #852: Multiplayer: Save player controls for multiple players in settings file
Currently, only the player controls for one player are saved in the ```keymap``` property in the settings file. If we want to support multiple players, we need to save player controls for both players in the settings.

**If we do this, we still need a backwards compatible way of reading the previous keymap settings, so that those don't get lost.** 

This will need a bit of discussion, I guess. So I'm gonna mark it as such.

InsightGit:
This commit adds support for storing multiple players configs in the settings file, while attempting to maintain backwards compatibility by automatically detecting if the config file does not have a "playernum" property and if so, assigns that keymap to the first player only. This PR however, **does not** adapt KeyboardMenu or JoystickMenu to support more than one player, as it just adapts them to get the first KeyboardConfig or JoystickConfig object from vectors of shared pointers of them.

This might need a bit of discussion before it gets merged, as issue #852 said prior.

Closes #852

### DIFF
--- a/src/control/keyboard_config.cpp
+++ b/src/control/keyboard_config.cpp
@@ -19,9 +19,10 @@
 #include "util/log.hpp"
 #include "util/reader_mapping.hpp"
 
-KeyboardConfig::KeyboardConfig() :
+KeyboardConfig::KeyboardConfig(int playernumber) :
   keymap(),
-  jump_with_up_kbd(false)
+  jump_with_up_kbd(false),
+  playernum(playernumber)
 {
   // initialize default keyboard map
   keymap[SDLK_LEFT]     = Controller::LEFT;
@@ -55,6 +56,13 @@ KeyboardConfig::read(const ReaderMapping& keymap_lisp)
   if (config_is_sdl2)
   {
     keymap_lisp.get("jump-with-up", jump_with_up_kbd);
+
+    // We only want the first player to use the old config that was made for
+    // only one player.
+    if(!keymap_lisp.get("playernum", playernum) &&
+       playernum != 0){
+      return;
+    }
 
     auto iter = keymap_lisp.get_iter();
     while(iter.next())
@@ -136,6 +144,7 @@ KeyboardConfig::write(Writer& writer)
   writer.write("sdl2", true);
 
   writer.write("jump-with-up", jump_with_up_kbd);
+  writer.write("playernum", playernum);
 
   for(const auto& i : keymap)
   {
--- a/src/supertux/gameconfig.cpp
+++ b/src/supertux/gameconfig.cpp
@@ -27,7 +27,7 @@
 #include "util/log.hpp"
 #include "supertux/globals.hpp"
 
-Config::Config() :
+Config::Config(int numberofplayers) :
   profile(1),
   fullscreen_size(0, 0),
   fullscreen_refresh_rate(0),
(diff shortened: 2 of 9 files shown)


src/supertux/gameconfig.cpp
```diff
@@ -124,15 +132,31 @@ Config::load()
...
+        break;
+      }
     }
 
     ReaderMapping joystick_lisp;
-    if (config_control_lisp.get("joystick", joystick_lisp))
-    {
-      joystick_config.read(joystick_lisp);
+    
+    for(size_t i = 0; joystick_configs.size() > i; ++i) {
```
christ2go: Can you use a c++11 for loop here?

src/supertux/gameconfig.cpp
```diff
@@ -49,14 +49,22 @@ Config::Config() :
...
+  joystick_configs(),
   addons(),
   developer_mode(false),
   christmas_mode(false),
   transitions_enabled(true),
   repository_url()
 {
+  assert(numberofplayers > 0);
+
+  for(int i = 0; numberofplayers > i; ++i){
```
christ2go: For consistency with most other parts in this codebase please use `for (int i = 0; i < expr; ++i)` the other style is not used very often and imo is harder to read.

src/supertux/gameconfig.cpp
```diff
@@ -49,14 +49,22 @@ Config::Config() :
...
+    keyboard_configs.push_back(std::shared_ptr<KeyboardConfig>(
```
christ2go: Sine we use C++11 can you use std::make_shared? (https://en.cppreference.com/w/cpp/memory/shared_ptr/make_shared)

src/supertux/gameconfig.cpp
```diff
@@ -205,13 +229,19 @@ Config::save()
 
   writer.start_list("control");
   {
-    writer.start_list("keymap");
-    keyboard_config.write(writer);
-    writer.end_list("keymap");
+    assert(keyboard_configs.size() == joystick_configs.size());
```
christ2go: Can a malformed config file make this assert fail? If so, please exit more gracefully/provide a warning and initialize new default keyboards? 
(I am currently not sure if asserts are enabled in release mode @tobbi)

src/control/keyboard_config.cpp
```diff
@@ -56,6 +57,14 @@ KeyboardConfig::read(const ReaderMapping& keymap_lisp)
   {
     keymap_lisp.get("jump-with-up", jump_with_up_kbd);
 
+    if(!keymap_lisp.get("playernum", playernum)){
```
christ2go: Is there a reason for this to be two `if`s ? If not please use one condition and put the comments above.

src/supertux/gameconfig.hpp
```diff
@@ -79,8 +79,8 @@ class Config
       means autodetect. */
   std::string locale;
 
-  KeyboardConfig keyboard_config;
-  JoystickConfig joystick_config;
+  std::vector<std::shared_ptr<KeyboardConfig>> keyboard_configs;
```
christ2go: If these shared pointers are not used outside of this class, consider making them unqiue pointers.

InsightGit: I made these pointers shared not because they are necessarily used outside of this class, but because of all of this class' methods and members are public, meaning it could be used outside of this class.

christ2go: Ok, lets keep them shared

christ2go:
@InsightGit Can you reopen this against our new multiplayer branch? It would be really cool if we could get the keyboard stuff to work.

InsightGit:
Sure thing.

bou6:
Hi @InsightGit , I just tried to checkout your commit to test it but got a fatal error:
`reference is not a tree aa8adc73aaec4219ae18d268d6910c5b3629bae0` 
Any clue for this ?

PR #854, opened 2018-06-30
New physics engine (beta)

christ2go:
Alpha version of the new physics engine. 

Uses polygons for collision detection against tilemap. Inorder to mitigate the problem of "hidden edges", polygons flag adjacent vertices and ignore such normals during collision resolution. Uses SAT (seperating axes theorem) for collision detection between the polygons.

This still needs (much) better code style, seperation of collision detection from sector.cpp and a lot of documentation. For now please test!
This still needs 

- [x]  Spatial Hashing: Don't add / delete on every frame, instead reuse
- [x]  Crushing
- [x] Moving solid tilemaps
- [ ] inline Documentation
- [ ] Wiki/Latex pdf generated documentation 
- [ ]  some changes to blocks => general review of collision handling in src/object/
- [ ]  some badguys (captain snowball) no longer work like they were supposed to.
- [X] specail case code for AABB vs AABB (for speedup)
- [ ] Update license headers

Performance (would be greats)
- [X] Add AABBPolygon vs AABBPolygon collision detection (currently just calls Polygon)
- [ ] Polygon: Avoid checking parallel axes (=> speed-up)
- [ ] Look into using pooling to reduce memory allocations  
- [X] Reduce number of collision phases between moving objects (currently 3), rely on broadphase  instead

Nice (additional features) (low priority, maybe needed swimming or just out of curiosity)
- [ ] Add a sleeping mechanism
- [ ] Angular rotation
- [ ] Debug mode: Display polygons and normals
- [ ] Do continous collision detection


weluvgoatz:
Is this going to be merged any time soon? I would love to have moving tilemaps that actually work properly.

Alzter:
Can somebody please continue this? I know I'm not exactly part of the "team" but I would really appreciate the physics being improved.

HybridDog:
Adding the documentation is in the todo list and there are a lot of changes, with conflicts in 17 files.
Without the doc, continuing this may require a lot effort for everybody except christ2go.

christ2go:
Closing this pull request. The "engine" is just too resource-hungry for any practical consideration and I don't see any way to fix this.

PR #865, opened 2018-07-05
Moved Object name property parsing to GameObject class

### LINKED ISSUE #862: Object name property parsing should be done in GameObject class
Currently, we're parsing the name of a game object in the relevant sub-classes. Code-style-wise, this is inconsistent. We should add a ```GameObject``` constructor that takes a ```const ReaderMapping&``` and parse the name there. Then remove the various ```lisp.get("name", name);``` calls throughout the code.

Add constructor to this class:
https://github.com/SuperTux/supertux/blob/master/src/supertux/game_object.hpp

Remove the reader.get method from sub-classes (as shown in this example):
https://github.com/SuperTux/supertux/blob/master/src/badguy/willowisp.cpp#L55

...and initialize the GameObject constructor when calling the relevant constructors. For the above example, you can call the ```GameObject``` constructor from the ```Badguy``` class.

Please comment on this issue if something is unclear.

InsightGit:
This Pull Request moves the object name property parsing code to the GameObject class for all direct and indirect subclasses of GameObject (except for ParticleSystem, where the name parsing code was not in the constructor), as requested in issue #862.

I was unable to fully test this code (other than compiling and playing a level in the game) as I was a little unsure how to exactly test it.

Closes #862

### DIFF
--- a/src/badguy/dispenser.cpp
+++ b/src/badguy/dispenser.cpp
@@ -71,7 +71,6 @@ Dispenser::Dispenser(const ReaderMapping& reader) :
 
   type_str = get_type_string();
 
-  reader.get("name", name, "");
   reader.get("limit-dispensed-badguys", limit_dispensed_badguys, false);
   reader.get("max-concurrent-badguys", max_concurrent_badguys, 0);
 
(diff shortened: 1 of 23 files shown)


tobbi:
We also have a few ```.get``` calls in other files:
Can you search all .cpp / .hpp files for ```.get("name"``` and you should be able to find all affected files.

tobbi:
👍 Great, thank you!

PR #866, opened 2018-07-08
Moved Sector Collision code to SectorCollisionDetector

### LINKED ISSUE #729: Move collision detection from sector.cpp into new class
Sector.cpp is huge (more than 1k lines). After moving the game session recorder stuff into a new class, I figured we could also move the collision detection code into a new class (together with the can_see_player methods etc.) around 600 lines, which is almost half of the sector.cpp file.

InsightGit:
This Pull Request moves the collision code from the Sector class to a new class called SectorCollisionDetector in order to make the sector.cpp file shorter and less complex, as stated by Issue #729. This code has been tested with "Welcome To Antarctica" and all collisions seem to still work.

Closes #729

### DIFF
--- a/src/scripting/floating_image.cpp
+++ b/src/scripting/floating_image.cpp
@@ -28,8 +28,8 @@ FloatingImage::FloatingImage(const std::string& spritefile) :
 {
   using namespace worldmap;
 
-  if(Sector::current() != NULL) {
-    Sector::current()->add_object(floating_image);
+  if(::Sector::current() != NULL) {
+    ::Sector::current()->add_object(floating_image);
   } else if(WorldMap::current() != NULL) {
     WorldMap::current()->add_object(floating_image);
   } else {
--- a/src/supertux/sector.cpp
+++ b/src/supertux/sector.cpp
@@ -85,7 +85,8 @@ Sector::Sector(Level* parent) :
   player(0),
   solid_tilemaps(),
   camera(0),
-  effect(0)
+  effect(0),
+  collision_detector(moving_objects, solid_tilemaps, get_players())
 {
   PlayerStatus* player_status;
   if (Editor::is_active()) {
@@ -99,6 +100,8 @@ Sector::Sector(Level* parent) :
   add_object(std::make_shared<Player>(player_status, "Tux"));
   add_object(std::make_shared<DisplayEffect>("Effect"));
   add_object(std::make_shared<TextObject>("Text"));
+  
+  collision_detector.set_players(get_players());
 
   SoundManager::current()->preload("sounds/shoot.wav");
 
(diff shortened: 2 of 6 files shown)


src/supertux/sector.hpp
```diff
@@ -145,27 +146,20 @@ class Sector
                          MovingObject &object) const;
 
   /**
-   * Checks if the specified rectangle is free of (solid) tiles.
-   * Note that this does not include static objects, e.g. bonus blocks.
+   * Wrapper for SectorCollisionDetector's is_free_of_tiles function.
```
christ2go: I don't know if it is a good idea to completely delete this comment (@maxteufel / @tobbi)?

src/scripting/floating_image.cpp
```diff
@@ -28,8 +28,8 @@ FloatingImage::FloatingImage(const std::string& spritefile) :
 {
   using namespace worldmap;
 
-  if(Sector::current() != NULL) {
-    Sector::current()->add_object(floating_image);
+  if(::Sector::current() != NULL) {
```
christ2go: Is there a reason for this change?

src/supertux/sector.hpp
tobbi: The full comment is still here: https://github.com/SuperTux/supertux/pull/866/files#diff-3286e434910f100eed55179300cff945R45

src/supertux/sector_collision_detector.cpp
```diff
@@ -0,0 +1,645 @@
...
+//
+//  You should have received a copy of the GNU General Public License
+//  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#include "supertux/sector_collision_detector.hpp"
+
+#include "editor/editor.hpp"
+
+SectorCollisionDetector::SectorCollisionDetector(
+                                    std::vector<MovingObject*>& moving_objects,
```
tobbi: The indentation in this method is kinda...messed up.

src/supertux/sector.cpp
```diff
@@ -587,614 +590,48 @@ Sector::on_window_resize()
...
-    if (!moving_object->is_valid()) continue;
-    if ((moving_object->get_group() == COLGROUP_MOVING)
-        || (moving_object->get_group() == COLGROUP_MOVING_STATIC)
-        || (moving_object->get_group() == COLGROUP_STATIC)) {
-      if(intersects(rect, moving_object->get_bbox())) return false;
-    }
-  }
-
-  return true;
+  collision_detector.set_players(get_players());
```
tobbi: Why do you set the players again, after doing so in the constructor already?

christ2go: You could have a std::vector< Player > players member in sector. Currently the vector with players is always created whenever ``get_players``. Then you could pass the vector by reference.

src/scripting/floating_image.cpp
InsightGit: Floating_image was getting confused with the Sector scripting class and the actual Sector class.

tobbi:
Oh, looks like there are merge conflicts with master. Could you please fix them?

PR #868, opened 2018-07-09
Reformat: Branch with reformatted code

christ2go:
This branch is just an example of how our reformatted code might look like. Feel free to tamper with the clang-format settings until the style is satisfying for everyone. 

Used clang-format version 7.0.0 (tags/google/stable/2018-04-24)


HybridDog:
I think you should generally only reformat new code or code next to new code because reformatting all code would
break lots of existing Pull Requests,
ignore special cases where general code style rules make code look bad, and
change code even if it is not allowed due to licensing.
(there are probably more reasons)

PR #870, opened 2018-07-11
WalkingBadguy: intelligent badguys turn around at harmful tiles 

### LINKED ISSUE #851: Enemies that never fall off platforms should avoid spikes and harmful things on their path
@tobbi told me to make an issue on this. Enemies like bombs, smartballs, walking leaves etc. should avoid things like spikes and vines (turns around). This should be a property of the enemy so levels still work, but it by default they should turn around.

Nobelix:
This changes prevents intelligent walking bad guys from running into spikes etc...
Fixes #851

### DIFF
--- a/src/badguy/walking_badguy.cpp
+++ b/src/badguy/walking_badguy.cpp
@@ -191,6 +191,18 @@ WalkingBadguy::collision_badguy(BadGuy& , const CollisionHit& hit)
 }
 
 void
+WalkingBadguy::collision_tile(uint32_t tile_attributes)
+{
+  if ((tile_attributes & Tile::HURTS) && max_drop_height > -1 && on_ground())
+  {
+    // prevents the inteligent bad guys from running into deadly tiles
+    turn_around();
+  } else {
+    BadGuy::collision_tile(tile_attributes);
+  }
+}
+
+void
 WalkingBadguy::turn_around()
 {
   if(frozen)
--- a/src/badguy/walking_badguy.hpp
+++ b/src/badguy/walking_badguy.hpp
@@ -18,6 +18,7 @@
 #define HEADER_SUPERTUX_BADGUY_WALKING_BADGUY_HPP
 
 #include "badguy/badguy.hpp"
+#include "supertux/tile.hpp"
 
 class Timer;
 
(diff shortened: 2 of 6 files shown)


src/badguy/walking_badguy.hpp
```diff
@@ -18,6 +18,7 @@
 #define HEADER_SUPERTUX_BADGUY_WALKING_BADGUY_HPP
 
 #include "badguy/badguy.hpp"
+#include "supertux/tile.hpp"
```
christ2go: Do you need this include? Can it go into badguy.cpp?

src/badguy/walking_badguy.cpp
```diff
@@ -190,6 +190,18 @@ WalkingBadguy::collision_badguy(BadGuy& , const CollisionHit& hit)
   return CONTINUE;
 }
 
+void
+WalkingBadguy::collision_tile(uint32_t tile_attributes)
+{
+  if(tile_attributes & Tile::HURTS && max_drop_height > -1 && on_ground())
```
christ2go: Add space after if. Maybe add parantheses around first part for better readability.

src/badguy/walking_badguy.hpp
Nobelix: when it wasn't the compiler complained

src/badguy/walking_badguy.cpp
Nobelix: After what? I tried to mimic the other functions syntax

Nobelix: Oh okay I get it

src/badguy/walking_badguy.hpp
Nobelix:
```/home/noe/Bureau/SuperTux/supertux/src/badguy/walking_badguy.cpp: In member function ‘virtual void WalkingBadguy::collision_tile(uint32_t)’:
/home/noe/Bureau/SuperTux/supertux/src/badguy/walking_badguy.cpp:196:26: error: ‘Tile’ has not been declared
   if( (tile_attributes & Tile::HURTS)  &&  max_drop_height > -1  &&  on_ground())

tobbi:
This looks good, but needs a rebase. Can you please try to rebase this against current master so this can get merged?

PR #873, opened 2018-07-13
Scripting: Exposed Spotlight to the scripting API

### LINKED ISSUE #768: Spotlight: expose to scripting API
Expose spotlights to the scripting API. That means:

- adding a name property
- starting, stopping the rotation using the scripting API
- enable or disable light emission from scripting API
- change speed and direction using scripting API
- change color using scripting API

Nobelix:
_This PR exposes the Spotlight class to the squirrel API._

### Methods are the following:

- `set_emitting(bool)` enables / disables light emission
- `is_emitting()` returns true if the spotlight is emitting light
- `set_speed(int)` sets the rotation speed of the spotlight
- `get_speed()`  returns the rotation speed of the spot light
- `set_color(Color)` sets the RGBA color of the spotlight

Closes #768

### DIFF
--- a/src/object/spotlight.cpp
+++ b/src/object/spotlight.cpp
@@ -18,10 +18,12 @@
 
 #include "sprite/sprite.hpp"
 #include "sprite/sprite_manager.hpp"
+#include "scripting/spotlight.hpp"
 #include "supertux/object_factory.hpp"
 #include "util/reader_mapping.hpp"
 
 Spotlight::Spotlight(const ReaderMapping& lisp) :
+  ExposedObject<Spotlight, scripting::Spotlight>(this),
   angle(),
   center(SpriteManager::current()->create("images/objects/spotlight/spotlight_center.sprite")),
   base(SpriteManager::current()->create("images/objects/spotlight/spotlight_base.sprite")),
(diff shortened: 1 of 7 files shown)


src/badguy/mriceblock.cpp
```diff
@@ -163,6 +163,17 @@ MrIceBlock::collision_badguy(BadGuy& badguy, const CollisionHit& hit)
   return ABORT_MOVE;
 }
 
+void
+MrIceBlock::collision_tile(uint32_t tile_attributes)
```
tobbi: These are some unrelated changes. Can you remove them, please?

Nobelix: Yes I know those are changes from my previous commit, I don't know how to use git so I messed up, @maxteufel told me to make a PR and wait...

tobbi:
Is it still not working?

Nobelix:
it should be working, successfully compiled and tested on my machine

src/object/spotlight.cpp
```diff
@@ -42,6 +45,8 @@ Spotlight::Spotlight(const ReaderMapping& lisp) :
   lisp.get("speed", speed, 50.0f);
   lisp.get("counter-clockwise", counter_clockwise, false);
 
+  emitting=true;
```
tobbi: Instead of writing ```emitting = true```, please initialize this in the initialization list, behind ```speed(50.0f),``` like in the following example:
```
emitting(true),
```

PR #875, opened 2018-07-15
Clean up some includes.

leper:
src/util/reader* needs some looking at since there are forward declared classes that are returned. This needs some investigation, which I'd prefer to do in another commit.

This uses a few more forward declarations to cut down on includes in a few places, and removes the reader_fwd.hpp file which wasn't really needed anywhere as a single forward declaration did the same in pretty much all places. Also removes one unused macro from about 4 files.

Make sure this still builds without issue on as many platforms as possible.


tobbi:
AppVeyor doesn't like this:
https://ci.appveyor.com/project/supertux/supertux-9ml4d/build/2143/messages

Neither does travis:
https://travis-ci.org/SuperTux/supertux/builds/404035072

leper:
Should be fixed now for those two, I'd recommend some more testing on other platforms if those are things you care about.

tobbi:
I never test stuff. My things just work by chance because I'm awesome.

leper:
Point taken :P

tobbi:
At least you're not Irish, otherwise you'd have said _Pint taken_.

PR #885, opened 2018-07-23
Added a "Restart from Checkpoint" button

### LINKED ISSUE #800: Add Confirm Dialog for Level Restart when a Checkpoint is reached
Sometimes you accidentally restart the level because you are about to die. Then you realize that you pressed ESC DOWN and ENTER too fast without giving the fact that you reached a checkpoint before a thought. After that you are irritated by being back at the very beginning of the level and you think that would not have happened if there had been some kind of a confirmation dialog that asks you “Do you really want to start over?” and you can choose between “Start Over”, “Restart at latest checkpoint” and “Cancel”.

In my opinion this scenario is not that far-fetched and the extra button to press does not bother me in exchange for more safety. In addition, the option to restart from checkpoint is helpful in case you get soft-locked – this means you got trapped but also aren’t able to die.

InsightGit:
This PR adds a "Restart from Checkpoint" button to the pause menu in order to address issue #800, which requested for this sort of button to be added. The "Restart from Checkpoint" button will only appear underneath the "Restart Level" button if the player has more than 25 coins and has passed a valid checkpoint. If the player meets the conditions and triggers the button, it will subtract 10%(or 25) of the player's coins and restart the level at the past checkpoint.

Image of "Restart from Checkpoint" button in menu

Closes #800


tobbi:
This looks good, I'll test it later today. 👍

tobbi:
Hello @InsightGit Seems like this slipped through my fingers, sorry about that. It looks good, can you just resolve the conflicts and I'll merge this.

src/supertux/game_session.cpp
```diff
@@ -404,6 +404,13 @@ GameSession::update(float elapsed_time)
     reset_button = false;
     reset_level();
     restart_level();
+  } else if(reset_checkpoint_button) {
+    reset_checkpoint_button = false;
+    
+    PlayerStatus *player_status = currentsector->player->get_status();
```
HybridDog: maybe `get_current_sector()->player->get_status();`

tobbi:
@HybridDog I don't suppose @InsightGit is interested in continuing this. Are you willing to take over?

HybridDog:
Yes, I'll try to add the button.

PR #887, opened 2018-07-25
Added support for multiple players configs in settings file (multiplayer branch PR)

### LINKED ISSUE #852: Multiplayer: Save player controls for multiple players in settings file

InsightGit:
This commit adds support for storing multiple players configs in the settings file, while attempting to maintain backwards compatibility by automatically detecting if the config file does not have a "playernum" property and if so, assigns that keymap to the first player only. This PR however, does not adapt KeyboardMenu or JoystickMenu to support more than one player, as it just adapts them to get the first KeyboardConfig or JoystickConfig object from vectors of shared pointers of them.

This might need a bit of discussion before it gets merged, as issue #852 said prior.

Closes #852

### DIFF
--- a/src/control/joystick_config.cpp
+++ b/src/control/joystick_config.cpp
@@ -22,13 +22,14 @@
 #include "util/log.hpp"
 #include "util/reader_mapping.hpp"
 
-JoystickConfig::JoystickConfig() :
+JoystickConfig::JoystickConfig(int playernumber) :
   dead_zone(8000),
   jump_with_up_joy(false),
   use_game_controller(true),
   joy_button_map(),
   joy_axis_map(),
-  joy_hat_map()
+  joy_hat_map(),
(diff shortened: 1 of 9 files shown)


tobbi:
Preliminary review (without having tested it): Looks good!

tobbi:
Is everyone okay with merging this?

weluvgoatz:
As long as the 1-player controls still work the same, I'm fine with it.

tobbi:
Thanks! 👍 🇭🇹 🍄

PR #888, opened 2018-07-25
[In Progress] Fix Ambient Sound

lilinitsy:
This is a slight fix for ambient sound. It enables fixed-size ambient sounds to work and play. Revamped the distance calculation to be accurate.
Current issues:
- Only one sound will play.
- The first global ambient sound encountered (without a fixed size) will always be the one that plays.
- If there's no global ambient sound, the first one listed in the .stl file will be the only one that will play, but only if within the range of the ambient sound.
- Copied from ambient_sound.cpp:
/* notes:
  Only one ambient sound per frame will be played.
  How can this be fixed? 
  - Maybe have a pool of the ambientsounds to play
    (allocated with max = #sounds per level), aind then
    play all per frame within the size?
  - Don't use more than 1 global ambient sound? 
    - But his will still cancel out the fixed size ones.
  
  So far the changes to not use what seemed to be the wrong
    distance^2 calculation is helping.
*/
If anyone has any ideas on how to accomplish / implement these, please let me know!


tobbi:
@lilinitsy Are you still willing to work on this?

tobbi:
Due to lack of response I'm closing this.

PR #892, opened 2018-08-02
Text array

g--o:
Adds the text array with 3 quick test levels
Scriptable & binded to the menu selection and text deletion (defaults: ENTER & BACKSPACE)


g--o:
dammit forgot to change name in header... I'll fix it soon enough

src/object/text_array_object.cpp
```diff
@@ -0,0 +1,218 @@
...
+void TextArrayObject::set_fade_transition(bool fade_transition)
+{
+  m_fadeTransition = fade_transition;
+}
+
+TextArrayItem* TextArrayObject::get_text_item(ta_index index)
+{
+  if (m_texts.size() <= 0)
+    return nullptr;
+  return this->m_texts.at(index).get();
```
christ2go: What happens, if index is out of bounds?

src/object/text_array_object.hpp
```diff
@@ -0,0 +1,75 @@
...
+typedef unsigned int ta_index;
+
+/** A text object intended for scripts that want to tell a story */
+class TextArrayObject : public GameObject,
+                        public ExposedObject<TextArrayObject, scripting::TextArray>
+{
+public:
+  TextArrayObject(const std::string& name_ = std::string());
+  ~TextArrayObject() = default;
+
```
christ2go: It would be great to have deoxygenated comments explaining what the methods do.

tobbi:
Hey, I'm wondering: Do these .sh files need to be there? Normally we manage all the build stuff using CMake.

ubuntu-dev-installs.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
```
christ2go: Please don’t add unrelated files.

g--o:
@tobbi It's for convenience. If we don't want it I can remove them.

tobbi:
@g--o What do they do?

g--o:
@tobbi They're just for setting up the environment for ubuntu quickly.
On your first time you run ubuntu-dev-installs.sh to install all the lib packages.
And then you run build.sh to build (creates dir, runs cmake + make, adds flags)
As said, it's not neccessary, but I find it comfy and that's how I work anyway.

g--o:
@christ2go Fixed.
About the editor integration - I'm not sure I understand, do you mean an invisible block or a some dialog for inserting one? If so I'd prefer to do this in a separate PR, along with the picnic scene. For now it's scriptable like the text object. You can check the test levels to see some examples.

src/object/text_array_object.cpp
```diff
@@ -0,0 +1,222 @@
...
+{
+  if (m_isDone)
+    return;
+
+  // make sure there's anything to update
+  if (m_texts.size() <= 0)
+    return;
+
+  // detect change request
+  InputManager* current = InputManager::current();
```
tobbi: Can you use ```auto current =```

g--o:
@tobbi should be good now

tobbi:
It looks good, I just have a few questions:
* The white background looks rather opaque on the first item of the text array. Any ideas?
* Could we have it so that you can specify a duration for each text item (maybe in addition to specifying a default duration).

g--o:
@tobbi 
- Could be because of differences in fade_in vs. visible(true), I'm not sure and I honestly just forward the API to the relevant item.
- But you do have that, it's even in one of the tests =/

tobbi:
I'll give this one final test and if all is good, I'll merge it.

PR #896, opened 2018-08-08
Update target times for Icy Island levels [ci skip]

brmbrmcar:
The target times seemed harsher or much harsher than the others. They are not designed for hard-core speedrunners. This should make this more consistent with times in the rest of the worldmap and hopefully fairer, and at least make it possible for casual users.


weluvgoatz:
Yes Please!!!!! I am not able to complete most if not all levels perfectly
because oc the darn target times!!!

tobbi:
Thank you very much!

PR #897, opened 2018-08-09
Fix bug introduced by text array

christ2go:
Some changes in text array (without regenerating wrapper) had led to wrapper be non-generatable (custom typedefs, using `size_t`.


christ2go:
The file has shown up again.

tobbi:
I am wondering: Wouldn't you need to convert that int you pass to the scripting method to a size_t for the actual GameObject class?

tobbi:
Know what: Who cares... I'll fix it in a followup.

PR #898, opened 2018-08-11
Update tux.sprite

windowsdylan:
Add extra frames for new ice tux walking sprite


tobbi:
One could say I forgot to include those two frames deliberately to give you a chance to contribute, but I didn't. I just forgot them.

PR #902, opened 2018-08-14
Update tiles.strf

weluvgoatz:
Here, I have updated tiles.strf to include @Rusty-Box 's new 1-tile high forest terrain. You can find the image at #692. I am using the newest "tiles.strf" from the source so some changes weren't made by me, only adding these forest tiles were my changes...


weluvgoatz:
maybe this is too small to make a pull request for...

PR #904, opened 2018-08-17
BadGuy Squish Particle Effects & Enhance Particle Physics

lilinitsy:
Right now, this does two things
- particles can now optionally have a "gravitational" effect, or an acceleration in the y direction to make them more interesting.
- Badguy's now emit particles when they're killed by a squish, to make it a little more interesting.

TODO:
- Maybe add a more complicated subclass of Particles, PhysicsParticles, which will allow for more complicated particle movement (think effecting x, y, at t^2 and t^3 rates)
~~- Perhaps add in a "z" value to give depth to the particles and increasee / decrease size accordingly.~~ Done without storing a "z" value, just decreased the size over time.
- Particles are usually pretty gray or white... change that?
- Add different particle effects for different badguys.

### DIFF
--- a/src/math/vector.hpp
+++ b/src/math/vector.hpp
@@ -116,6 +116,11 @@ public:
     return Vector(floorf(x), floorf(y));
   }
 
+  float magnitude()
+  {
+    return sqrt(x * x + y * y);
+  }
+
   // ... add the other operators as needed, I'm too lazy now ...
 
   float x, y; // leave this public, get/set methods just give me headaches
--- a/src/math/vector.hpp
+++ b/src/math/vector.hpp
@@ -116,11 +116,6 @@ public:
     return Vector(floorf(x), floorf(y));
   }
 
-  float magnitude()
-  {
-    return sqrt(x * x + y * y);
-  }
-
   // ... add the other operators as needed, I'm too lazy now ...
 
   float x, y; // leave this public, get/set methods just give me headaches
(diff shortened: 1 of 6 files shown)


src/math/vector.hpp
```diff
@@ -116,6 +116,11 @@ class Vector
     return Vector(floorf(x), floorf(y));
   }
 
+  float magnitude()
```
christ2go: Can you use ``norm`` instead?

lilinitsy: Didn't even notice that. Sure thing.

tobbi:
Hey @lilinitsy Could you provide a youtube video of this? I currently do not have time to compile it myself.

lilinitsy:
@tobbi here's a low resolution gif https://giphy.com/gifs/5hiCfND58cHi4RCzt1/fullscreen.
Thanks @qwertychouskie, I'd missed that one. Handled.

Grumbel:
[Video of it in action](https://www.youtube.com/watch?v=w_6p1s6TMFA)

Doesn't look much good in its current form and the fillrect based particles are rather ugly to begin with.

I wouldn't be opposed to some reaction when jumping on enemies, but it needs to be more subtle and preferably sprite/surface based.

lilinitsy:
Feel free to tweak it.

tobbi:
I'll close this mainly because of https://github.com/SuperTux/supertux/pull/904#issuecomment-434519997

PR #906, opened 2018-09-01
Tweak the background image of snowy mountain

raghukamath:
This PR tweaks the background image of snowy mountain shown in first levels.
I have made the sky a little brighter and cleanup the mountain. Below is the reference picture for comparison

Screenshot with tweaked background

after

Screenshot of earlier background
before


tobbi:
Thanks, this definitely looks much better than the old one! 👍

PR #915, opened 2018-10-08
 BUG with coins

OliverBuo:
If i change the color of the coins i can't take them in them game.
Also if i change the Alpha from coins during the game. First alpha 0.1 then Alpha 1.0 ...cant thake this coins.
I think this is a BUG.

Super tux 0.5.1 

Pleas fix it.  Thank u !!  
Regards Oliver


maxteufel:
This is not how you report a bug on GitHub. There is a separate 'Issues' tab. It is how you would submit a code change, such as a fix.

Regarding your problem: we're investigating and considering the possible options.

PR #919, opened 2018-10-20
Make freetype required, libraqm optional

kernigh:
If libraqm was not found, then configure SDL_ttf --without-raqm and
don't link to libraqm.  This fixes a CMake error for linking to a
not-found library.  SuperTux built this way can't display Arabic but
can still display English.

Make Freetype REQUIRED, so not finding it is an error.  Update
INSTALL.md to list CMake 3.1, Freetype, and optional libraqm.


kernigh:
I needed this change to build supertux on my machine without libraqm. I'm running OpenBSD 6.3 on amd64 with clang 5.0.1, cmake 3.10.2, glew 2.0.0.

tobbi:
This was definitely an oversight on our side. Thanks for catching it!

PR #920, opened 2018-10-20
Fix music on big-endian platforms

kernigh:
ogg_sound_file.cpp forgot to #include <config.h>, so the definition
of WORDS_BIGENDIAN was missing, so it played noise instead of music.
wav_sound_file.cpp seems to get <config.h> from another header, but
add #include to be sure.

### DIFF
--- a/src/audio/ogg_sound_file.cpp
+++ b/src/audio/ogg_sound_file.cpp
@@ -16,6 +16,8 @@
 
 #include "audio/ogg_sound_file.hpp"
 
+#include <config.h>
+
 #include <assert.h>
 #include <physfs.h>
 
--- a/src/audio/wav_sound_file.cpp
+++ b/src/audio/wav_sound_file.cpp
@@ -16,6 +16,8 @@
 
 #include "audio/wav_sound_file.hpp"
 
+#include <config.h>
+
 #include <string.h>
 #include <stdint.h>
 #include <assert.h>


kernigh:
I have an unusual big-endian machine running OpenBSD-current on macppc. OpenBSD's package of supertux-0.5.1p1 was playing noise instead of music. I fixed the problem in 0.5.1 by adding #include <config.h> to ogg_sound_file.cpp. There was no problem with sound effects, I guess because wav_sound_file.cpp gets <config.h> from another header, but this pull request adds another #include <config.h> there to be sure.

I can't build SuperTux git master on my big-endian machine, because its newest compiler (gcc 4.9.4) gives errors about implicit constructor calls. The compiler seems to be too old. Therefore, I don't know whether git master plus this pull request works on big-endian machines. I am able to build git master plus this pull request plus #919 (I need #919 to build) on my little-endian amd64 machine with clang 5.0.1.

If you have a macppc with OpenBSD-current and try to play SuperTux 0.5.1, the game will crash. That's a bug in SDL2. After I fixed SDL's bug, I can play the game by muting the music or with this fix. All macppc machines are old. My machine has ATI Radeon Mobility M10 (ATI RV350) with only OpenGL 2.1, and is too slow to run SuperTux 0.5.1 at full speed; but I can play the game if the window is not too big. The free C++ code in SuperTux is so portable, the game works on weird machines like mine, even if music was broken until now.

tobbi:
Thanks for your thorough investigation on and your fix for this as well! Merging!

PR #983, opened 2018-12-15
Add new santa hat design for Tux

Rusty-Box:
Add new santa hat sprites
Add new frames for Santa Tux to powerups.sprite
Add new santa hat design to images/objects/particles


Rusty-Box:
@tobbi The big christmas hat for the logo will need a position adjustment!

PR #994, opened 2018-12-20
Music loop fixes

Alzter:
Fixes the loops for forest and voc_night to sound more smooth.


tobbi:
I honestly like the ending of the forest music (where it kinda tapers out and you have the cricket sound) too much to be able to merge this with good conscience.

PR #1007, opened 2018-12-25
Improve upon RustyBox's version of yeti_boss.stl

weluvgoatz:
The version of this level created by RustyBox seems rushed and feels like a normal/basic level. Also, on small screens, the stalactites are not visible and it is very easy to get on top of the yeti.

This edition changes that by changing the setting to a yeti's cave, adding more visually pleasing features, and now the stalactites are visible from smaller screens.


Alzter:
I liked the idea Grumbel had where the Yeti jumps out of a window in a different room of the castle to fight you instead of fighting you in some random ice structure.

weluvgoatz:
"I liked the idea Grumbel had where the Yeti jumps out of a window in a different room of the castle to fight you instead of fighting you in some random ice structure."

BTW, in this version, it takes place in the yeti's cave/den (which should be a bit obvious, no?) 

I still think it's better than the previous two versions.

Rusty-Box:
@weluvgoatz I think what's the issue here is what you want from a storytelling standpoint. Storywise, the intention of the current design of the boss level is that Tux is being chased by the yeti, got cornered/caught up by him later on and now has to fight him. Since they are near to the coast and the yeti first had to travel to Nolok's castle as well there is no yeti cave near by thus taking place on a random spot between the coast and the castle.
If we would now use a yeti cave like setting it would imply that Tux went into the cave on purpose which would be a questionable decicion from him, in my opinion.

weluvgoatz:
I always thought the yeti's cave/den was next to the castle and that's where they fought (since no where else I have seen says otherwise). The Yeti's fight also looks better and feels more important design-wise when it is in a special location, rather than in a random ice location. I will improve upon this pull request, but my main idea was to make it to where the stalactites were visible because I cannot see them on my screen.

Rusty-Box:
Letting it take place in the castle takes away the chasing aspect, in my opinion. A penguin's first thought probably won't be that he can battle with a yeti. But if caught up by it and there is no other way around then a fight becomes inevitable.
I agree with the fact about an "more special looking location" to some extent. A fight in a somewhat special looking location can build up on the general feel of the fight. But who says that an empty plain location can't do the same as well. In the end this all might come down to personal preverences and what you want to tell story-wise and can never be solved completely without someone disagreeing.
I'm personally not in favor with the yeti's cave as a setting due to the things I meantioned before. But I'm open to "decorational" aspects and would attempt to make the "boring somewhere" look a bit more visually pleasing without having to turn it into a yeti's cave.

weluvgoatz:
There, the level now takes place by the coast rather than in a cave.

Alzter:
I thought it should take place in a room inside Nolok's castle, when Tux chases the Yeti into there the room locks behind him leaving him no choice but to fight the yeti

weluvgoatz:
Closing this for similar reasons I had before.

Issue #987, opened 2018-12-17
Show powerup count during level


clintmoyer:
**SuperTux version:** 0.5.1
**System information:** Arch Linux (kernel 4.19.4)

It would be nice for the player to be able to visualize their weapon count. Each subsequent powerup [increments the weapon count](https://github.com/SuperTux/supertux/blob/v0.5.1/src/object/player.cpp#L1122). Perhaps the weapon count could be shown underneath the coin count in the top-right corner.

Alzter:
Wait, what?? Having more fireflowers gives you more max bullets? I never knew that! It should be visually represented, yeah.

PR #1022, opened 2018-12-31
Show power up count

clintmoyer:
Enhancement from https://github.com/SuperTux/supertux/issues/987.

- adds a counter to show the fire or ice ammo count
- hidden when no bonus
- offset fps and player position by another 20 pixels

issue-987

powerups

### DIFF
--- a/src/supertux/player_status_hud.cpp
+++ b/src/supertux/player_status_hud.cpp
@@ -30,7 +30,9 @@ PlayerStatusHUD::PlayerStatusHUD(PlayerStatus& player_status) :
   m_player_status(player_status),
   displayed_coins(DISPLAYED_COINS_UNSET),
   displayed_coins_frame(0),
-  coin_surface(Surface::from_file("images/engine/hud/coins-0.png"))
+  coin_surface(Surface::from_file("images/engine/hud/coins-0.png")),
+  fire_surface(Surface::from_file("images/objects/bullets/fire_bullet-0.png")),
+  ice_surface(Surface::from_file("images/objects/bullets/ice_bullet.png"))
 {
 }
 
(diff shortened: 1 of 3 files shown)


tobbi:
@clintmoyer Thanks for this! Very much appreciated! I would personally prefer displaying a fire flower instead of a fire bullet in the top right corner. And not showing a number but showing 2 actual fire flowers. 

But this is my personal preference, I presume. 

@SuperTux/developers @SuperTux/level-designers What do you think?

serano01:
I think an fire flower and an actual number would be best, but that is nitpicking. Otherwise an nice addition.

qwertychouskie:
IMO it is perfect just how it is, showing a Fire Flower instead may make the user think they will only loose 1 flower when getting hit, not all of them.  The way it is done in this PR avoids that confusion.

Rusty-Box:
The way it is done here is a good idea. There may be, however, a rather tricky icon decision when it comes to the air and earth powerup, as they are being affected by this multiple collecting of powerups as well. (Since they don't have stuff like bullets.)
Honestly, I'm not in favor with the "affecting of powerups by multiple collections" in general but as long as it still persists this PR shall be considered.

tobbi:
How about just showing an air / earth flower for those cases where there are no bullets?

Alzter:
Are we going to merge this soon?

HybridDog:
[Rejected-Ideas](https://github.com/SuperTux/supertux/wiki/Rejected-Ideas) mentions Power meter. I don't know what is meant by that (broken link); I think it could refer to showing something like power up count in the UI.

Alzter:
> [Rejected-Ideas](https://github.com/SuperTux/supertux/wiki/Rejected-Ideas) mentions Power meter. I don't know what is meant by that (broken link); I think it could refer to showing something like power up count in the UI.

That's a separate thing, more like a super meter.

weluvgoatz:
Is this going to be merged? If so, then the "conflicting files" have to be fixed...

clintmoyer:
I submitted this change almost a year ago. Both the PR and [enhancement report](https://github.com/SuperTux/supertux/issues/987) seemed to garner interest. Can somebody with merge permissions let me know if this will be accepted? Once I get that confirmation I will fix merge conflicts.

clintmoyer:
Can somebody with merge permissions let me know if this will be accepted? Once I get that confirmation I will fix merge conflicts.

src/supertux/player_status_hud.cpp
```diff
@@ -85,6 +87,49 @@ PlayerStatusHUD::draw(DrawingContext& context)
...
+
+  if (m_player_status.bonus == FIRE_BONUS) {
+
+    ammo_text = std::to_string(m_player_status.max_fire_bullets);
+
+    if (fire_surface) {
+      context.color().draw_surface(fire_surface,
+                                   Vector(static_cast<float>(context.get_width()) - BORDER_X - static_cast<float>(fire_surface->get_width()) - Resources::fixed_font->get_text_width(ammo_text),
+                                          BORDER_Y + 1.0f + (Resources::fixed_font->get_text_height(coins_text) + 5) + (Resources::fixed_font->get_text_height(ammo_text) + 5) * static_cast<float>(player_id)),
+                                   LAYER_HUD);
```
HybridDog: Please insert line breaks if the text exceeds 80 characters. 
If the text goes too far to the right, it is difficult to read on github (bad example [here](http://github.com/SuperTux/supertux/blob/02631a48ccaf9a3924ea4b9f7a0f8886ec38ff98/src/supertux/player_status_hud.cpp#L75)), or when using a text editor with small width.
See also: https://01.org/linuxgraphics/gfx-docs/drm/process/coding-style.html

HybridDog:
> Can somebody with merge permissions let me know if this will be accepted?

I can merge this after you fixed conflicts and add some line breaks. Support for air and other powerups  can be added later.

plonibarploni:
The FPS display is now slightly covered by this:
image

PR #1026, opened 2019-01-01
Fix uneven level terrain in or_just_me world1 level

### LINKED ISSUE #1021: Stuck in 'or is it just me?'
**SuperTux version:**
0.6.0-3-gd75e87666

**System information:**
Linux Mint 19.1 Cinnamon
Cinnamon v4.0.8
kernel: 4.15.0-43-generic

##### Expected behavior
In level 'or is it just me?' in Cave World, if you take the over-the-level shortcut, you skip a good portion of the level

##### Actual behavior
you get stuck inside the floor near the end where you're supposed to fall down (in both small and large Tux form).
tuxbug
it looks like you can't get out by jumping or even backflip

just keep moving forward without jumping; if you jump just near the end, you fall down, skipping the glitch

clintmoyer:
Tux would fall into a terrain glitch around _6976,-30_ where they would be stuck and unable to move. This touches up the upper-level tiles before the cave opening to enable Tux to continue forward-progress.

Closes #1021 

### Before

glitch

### After

fix


tobbi:
@Rusty-Box Your area of expertise.

PR #1043, opened 2019-01-13
Restore accidentally removed displacement-textures

Rusty-Box:
- Restore accidentally removed displacement-textures
- Also change lava tiles' fps back to 8


tobbi:
Wird schon passen.

PR #1070, opened 2019-02-28
    Do not enable default keys in addition to player-defined keys

### LINKED ISSUE #592: Changing the key setup still allows for default keys to be used
If you change some of the keys, the old ones will still work. This may be an annoyance to players and could potentially cause bugs.

HybridDog:
I applied what christ2go commented at #592

When I have tested it, my keyboard configuration was reset at first but it worked after configuring it again.
I also tested if you can change important keys so that you cannot move through the menu. I could not break it, i.e. the safety mechanism in keyboard_manager.cpp probably works.


tobbi:
@christ2go Can you confirm this is how it's supposed to be?

tobbi:
I'm gonna merge this.

PR #1071, opened 2019-03-05
"Document" the coin pitches in comments as approximate notes.

divVerent:
Yes, this might be useless... but might save time for whoever wants
to... "use" them next.

https://www.youtube.com/watch?v=MT9HFV-dAPI


Alzter:
Don't coins only change pitch relative to the position of the last coin collected and only when collected in quick succession?

divVerent:
Relative to the first coin in sequence, and a pause of a certain size stops
the sequence. See my level that I linked :)

PR #1079, opened 2019-03-11
Three simple Rock changes

HybridDog:
throw to ground when ungrabbing while holding down
This allows Tux to place the Rock onto ground when he wants to do it. When I recently tried to place a rock onto another one, Tux threw it away although he moved slowly.
When Tux throws the Rock down and he has a powerup, he also begins to stomp, but that's not a problem.

slow down when hurting something 
When the Rock falls onto an enemy and kills it, it looses a bit velocity because it collided.

bounce a bit from walls
If the Rock is thrown against a wall it collides but the wall does not move, so the Rock needs to bounce back a bit.

### DIFF
--- a/src/object/player.cpp
+++ b/src/object/player.cpp
@@ -897,6 +897,8 @@ Player::handle_input()
         moving_object->set_pos(dest_.p1());
         if (m_controller->hold(Control::UP)) {
           m_grabbed_object->ungrab(*this, Direction::UP);
+        } else if (m_controller->hold(Control::DOWN)) {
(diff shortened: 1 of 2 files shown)


tobbi:
I currently don't have access to a build machine. Could you make a video of the changes?

HybridDog:
Yes, here's the video: [video.mp4.zip](https://github.com/SuperTux/supertux/files/3098159/video.mp4.zip)

HybridDog:
Here's an animated webp image of the video: animated webp

Issue #990, opened 2018-12-19
Buttjumping on trampoline 'bonus block' into impossible position


plonibarploni:
Supertux 0.6.0-rc.1 (possibly also in older versions)
In 'Above the Arctic Skies', if you buttjump the bonus block that reveals the trampoline right next to the wall, Tux is moved into the wall, and you have to jump to get out of the wall (else you take a hit).

![screenshot at 2018-12-19 00-00-43](https://user-images.githubusercontent.com/44826203/50201040-89f9c980-0326-11e9-9a85-1aacb5795580.png)

I would expect that Tux would land on top of the trampoline and start bouncing away.

plonibarploni:
Similar issue also appears in 'Bouncy Coils':
![bouncycoils](https://user-images.githubusercontent.com/44826203/50259118-bbc76a80-03d0-11e9-9fef-4d6da0ab30f3.gif)

brmbrmcar:
Certainly looks like something new with the engine. This doesn't happen in my older version when buttjumping a trampoline block.

HybridDog:
![2019-10-07-113051_1920x1080_scrot](https://user-images.githubusercontent.com/3192173/66300902-696f2000-e8f6-11e9-958a-3d6a7727daa4.jpg)

Maybe it is possible to disallow the collision calculation to move Tux or other objects into tiles;
it could fix this issue, #1048, #535, #1075, and more.

Zwatotem:
> Maybe it is possible to disallow the collision calculation to move Tux or other objects into tiles;
> it could fix this issue, #1048, #535, #1075, and more.

_And by the way kill a lot of speedrunning shortcuts. :( _ 

That's a sensible solution, although player would still get hurt by the appearing trampoline, which is still not fair to the player.

HybridDog:
> And by the way kill a lot of speedrunning shortcuts. :(

It would also break one of my levels where you have to move through the floor.

> […] which is still not fair to the player, so some additional code is needed.

The spawning trampoline could probably get a small collision box which moves the player up. Currently, while the trampoline goes up, it has no collision box (as can be inferred from the screenshot).

Wuzzy2:
Reproduced in 0.6.1 beta 1.

With an important difference: This maneuver is now deadly. It insta-kills Tux, no matter if small, big, fire, etc.

Rusty-Box:
[...]

PR #1080, opened 2019-03-11
Do not automatically hurt Tux when he cannot standup

HybridDog:
The player can press Up to hurt Tux when he is stuck.
I remember that somebody created an issue for this topic but I couldn't find it.

### DIFF
--- a/src/object/player.cpp
+++ b/src/object/player.cpp
@@ -868,7 +868,7 @@ Player::handle_input()
   if (m_controller->hold(Control::DOWN) && !m_stone) {
     do_duck();
   } else {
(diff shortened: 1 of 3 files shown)


Alzter:
I like it, although pressing up to damage Tux feels a bit cryptic to me. Can you just make it not hurt him? If the player is stuck they can still restart the level.

plonibarploni:
> I remember that somebody created an issue for this topic but I couldn't find it.

#990

HybridDog:
> Can you just make it not hurt him?

done

> #990?

That's another issue.

PR #1091, opened 2019-04-24
Move simple methods from physics.cpp to physics.hpp

HybridDog:
This makes the code shorter and allows the compiler to inline the functions, as mentioned e.g. at https://en.cppreference.com/w/cpp/language/class.

### DIFF
--- a/src/supertux/physic.cpp
+++ b/src/supertux/physic.cpp
@@ -35,18 +35,6 @@ Physic::reset()
 }
 
 void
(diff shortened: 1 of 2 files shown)


HybridDog:
I've changed the commit to use a single line for the getter and setters: https://github.com/SuperTux/supertux/pull/1091/commits/1fddbfdfdf0a2e82630c305472adff4e9b0b53d5

PR #1093, opened 2019-04-25
Add a few more add-on types

weluvgoatz:
This should add 2 things: 
1) The addon can now be type "Textures" 
2) If the addon is an invalid addon type, it should default to "Add-On"

### DIFF
--- a/src/addon/addon.cpp
+++ b/src/addon/addon.cpp
@@ -51,7 +51,7 @@ Addon::Type addon_type_from_string(const std::string& type)
   }
   else
   {
-    return Addon::ADD-ON;
+    return Addon::ADDON;
   }
 }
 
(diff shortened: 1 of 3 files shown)


weluvgoatz:
Is this going to be considered?

serano01:
Do we have even any Add-Ons which are simply textures? Or is it meant for Add-Ons like the Desert World?

weluvgoatz:
A proposed idea was to have SuperTux Texture Packs, which are addons that
change the look of the game, similar to Resource Packs in Minecraft. While
there are currently no Texture Packs in SuperTux, artists might be inspired
to make some when type "Textures" is added, such as 8-bit textures,
greyscale textures, realistic textures, etc....

weluvgoatz:
@serano01 said this won't be considered, so I close it.

PR #1099, opened 2019-05-09
Make fps measurements more accurate

HybridDog:
After exceeding 0.5 s, the excess milliseconds are no longer discarded
	(the FPS shown were sometimes higher than the real FPS, especially when they were low)
minimum and maximum FPS are shown;
	these can be more interesting than average FPS when the game stutters
std::chrono is used instead of SDL_GetTicks for microsecond precision

2019-05-09-113905_1920x1080_scrot
2019-05-09-113952_1920x1080_scrot

### DIFF
--- a/src/supertux/screen_manager.cpp
+++ b/src/supertux/screen_manager.cpp
@@ -38,6 +38,7 @@
 #include "video/drawing_context.hpp"
 
 #include <stdio.h>
+#include <chrono>
 
 /** wait at least MIN_TICKS for every frame */
 static const int MIN_TICKS = 2;
@@ -51,7 +52,6 @@ ScreenManager::ScreenManager(VideoSystem& video_system, InputManager& input_mana
   m_speed(1.0),
   m_target_framerate(60.0f),
   m_actions(),
-  m_fps(0),
   m_screen_fade(),
   m_screen_stack()
 {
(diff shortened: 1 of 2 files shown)


qwertychouskie:
I think min/current/max makes more sense than min/avg/max.  Average of what?  Since the game was launched?  Since the level started?  Avg makes since for a benchmark result, but not for live measurements.

HybridDog:
The delays between adjacent frames are measured for 0.5 seconds. min is the inverse of the maximum of these delays (1/delay_max), max is analogous to min, and avg is the mean value of the delays.
current would mean that the value corresponds to the delay between the current and previous frame; showing current fps is bad because your eyeballs cannot read a number which quickly changes every frame.

tobbi:
Is this ready to get merged?

HybridDog:
> Is this ready to get merged?

I think it's ready now.

src/supertux/screen_manager.cpp
```diff
@@ -127,17 +127,102 @@ ScreenManager::get_speed() const
...
+      last_fps_min = 0;
+      last_fps_max = 0;
+    }
+    measurements_cnt = 0;
+    acc_us = 0;
+    min_us = 1000000;
+    max_us = 0;
+  }
+
+  float get_fps() { return last_fps; }
```
tobbi: Those three methods should be const

tobbi:
Please address that one review comment and I'll merge it.

src/supertux/screen_manager.cpp
HybridDog: fixed

HybridDog:
I noticed that the FPS look disturbing in the level editor.
As far as I know, this already happened before my change.
2019-05-25-154026_1920x1080_scrot

PR #1100, opened 2019-05-09
Smooth worldmap camera panning

HybridDog:
The speed does no longer depend on the framerate.
The camera position is smoothly interpolated with a cosine function instead of linear interpolation.
The panning time is mostly direct proportional to the distance between start and end position.
If the distance is very high, this time is reduced to a maximum panning time.

### DIFF
--- a/src/worldmap/camera.cpp
+++ b/src/worldmap/camera.cpp
@@ -21,10 +21,11 @@
 #include "video/viewport.hpp"
 #include "worldmap/tux.hpp"
 #include "worldmap/worldmap.hpp"
+#include "math/util.hpp"
 
 namespace {
 
-const float CAMERA_PAN_SPEED = 5.0;
+const float CAMERA_PAN_TIME_MAX = 0.52213f;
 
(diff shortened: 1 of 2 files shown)


tobbi:
/home/travis/build/SuperTux/supertux/src/worldmap/camera.cpp: In member function ‘void worldmap::Camera::update(float)’:
/home/travis/build/SuperTux/supertux/src/worldmap/camera.cpp:57:26: error: conversion from ‘double’ to ‘float’ may change value [-Werror=float-conversion]
     f = 0.5f * cosf(M_PI * f) + 0.5f;

tobbi:
While I'm at it: Are you interested in joining the team's chat here?
https://riot.im/app/#/room/#supertux:matrix.org

HybridDog:
It uses the right pi now.

> While I'm at it: Are you interested in joining the team's chat here?
https://riot.im/app/#/room/#supertux:matrix.org

Yes, but the captcha does not work for me, so I cannot join.

PR #993, opened 2018-12-20
Tux improvements

Alzter:
[ci skip] [skip ci] Adjustments/Additions to Tux's animation set.

Changes:
• Big and Small Tux's standing has been cleaned up a little.
• Big and Small Tux's idle animations have been fixed to not change position slightly.
• Big Tux's walking animation has been redone as an 8 frame animation, matching small Tux's.
• Big and Small Tux have completely new ladder animations.
• Big Tux has new work in progress sliding and swimming animations to be used in future.
• Fire and Ice Tux now have separate sprites for their hats like Earth, Air and Santa Tux, however they don't work in game yet (please fix).
• Ice Tux's hat has received some clean-ups. Fire Tux's hat may possibly receive the same treatment down the track.
• Statue Tux entirely redone.

PR #1104, opened 2019-05-20
Make fire and ice hats visible again

HybridDog:
Fixes the regression introduced by #993.


Alzter:
The issue is that I've sorted the fire and ice hats like the earth and air ones, which work differently. Now the fire and ice hats are on a separate layer, but the game's coding still treats them as if they should be one object merged with tux. Someone needs to fix this with coding, reverting the commit doesn't really help.

PR #1105, opened 2019-05-22
Adjust iceslick textures

HybridDog:
Before:
before

After (Changed iceslick):
2019-05-23-115645_1920x1080_scrot
2019-05-23-115652_1920x1080_scrot

Edit: removed the ice bridge changes images


weluvgoatz:
I prefer the old ones, these ones are too white in my opinion.  The idea with making them brighter is nice, but it really needs to be more blue.

qwertychouskie:
What would it look like with the new iceslick but the old icebridge?

Alzter:
I think it'd be a cool detail to add some icy frosting on top of the bridge, but maybe that would get confused with slippery ice so I'm not sure.

HybridDog:
> What would it look like with the new iceslick but the old icebridge?

2019-05-23-115645_1920x1080_scrot
2019-05-23-115652_1920x1080_scrot

> I think it'd be a cool detail to add some icy frosting on top of the bridge, but maybe that would get confused with slippery ice so I'm not sure.

You could also make new, similar-looking tiles for the icy bridge because the current ones are as far as I can see just like a blue version of the common brown blocks.

tobbi:
Could we have both variants?

weluvgoatz:
I want both variants, even though I prefer the older ones.

HybridDog:
> Could we have both variants?

I personally would prefer new textures over reusing old ones with new colour and contrast.
For example, that one: https://opengameart.org/content/hand-painted-texture-iceberg

busybody42:
why don't we add in some slopes for icebridge (and their normal blocks version, in the process) tiles while we're at it?

tobbi:
@HybridDog or @SuperTux/graphics-designers Would you mind assigning new tile IDs to those new textures? Then we can get this merged :)

weluvgoatz:
The images would need to be renamed as well.

Rusty-Box:
I think the ice bridge tiles are too white. A more bluish tint is prefered since the ice bridge on the worldmap is blue as well. Having both is pointless. The ice slick i don't mind.

weluvgoatz:
The iceslick looks well enough, but I agree this icebridge is really bad.

HybridDog:
I have removed the icebridge change.

Rusty-Box:
Thanks. Everything seem good to go. Does anyone else want to say something about it, before merging?

PR #1113, opened 2019-05-24
Do not assume a male player in code comments

HybridDog:
I may not have found every related comment.

### DIFF
--- a/src/object/camera.cpp
+++ b/src/object/camera.cpp
@@ -365,9 +365,9 @@ Camera::update_scroll_normal(float dt_sec)
     m_cached_translation.y = player_pos.y - static_cast<float>(m_screen_size.height) * config_.target_y;
   }
   if (ymode == 2) {
-    // target_y is the high we target our scrolling at. This is not always the
-    // high of the player, but if he is jumping upwards we should use the
-    // position where he last touched the ground. (this probably needs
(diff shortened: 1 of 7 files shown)


weluvgoatz:
Currently, the level editor defaults a level author's name to "Mr. X". You
should change this to "SuperTux Player".

HybridDog:
> You should change this to "SuperTux Player".

Done.

weluvgoatz:
Thank you. :)

src/object/camera.cpp
```diff
@@ -366,7 +366,7 @@ Camera::update_scroll_normal(float dt_sec)
   }
   if (ymode == 2) {
     // target_y is the high we target our scrolling at. This is not always the
-    // high of the player, but if he is jumping upwards we should use the
+    // high of Tux, but if he is jumping upwards we should use the
```
tobbi: Change "he is" to "they are" maybe?

HybridDog: I've changed more things in this comment now.
I don't understand the meaning of this comment. It says that in the future there probably needs to be exceptions for trampolines although these already exist in SuperTux for a long time.
https://github.com/SuperTux/supertux/blob/7d339fe2a3001c3a8c5a71094b2b3deac7bb4d52/src/object/player.hpp#L43

Issue #1045, opened 2019-01-13
Arabic is broken in 0.6.0


Fahad-Alsaidi:
**SuperTux version: 0.6.0 appimage**
**System information:  KDE neon**

I though after closing #97 that Arabic is going to displayed correctly. I don't know what's happen after that. However, Arabic is broken.

![image](https://user-images.githubusercontent.com/1632566/51088101-1c01c000-1775-11e9-8589-2e6edb34fe12.png)

christ2go:
Is the orientation still wrong or do you mean the missing translation?

Fahad-Alsaidi:
> Is the orientation still wrong or do you mean the missing translations?

orientation is wrong. It is from LTR and not connected. The right thing here:
![index](https://user-images.githubusercontent.com/1632566/51088428-f4145b80-1778-11e9-9feb-2eef117f5062.png)

tobbi:
@Grumbel Do you have any idea why this happened? Do we need another font for Arabic after all?

Fahad-Alsaidi:
@tobbi it is not about font, it is about shaping & rendering.

Grumbel:
Might be something simple like not defining `HAVE_RAQM`?

Grumbel:
Looks like it got broke in 10c6d0df7dd12506408faf81a56d48a2f029a85c, which switched from configure to using cmake for SDL_ttf.

tobbi:
@Fahad-Alsaidi Can you please verify that https://prnt.sc/ntbjz2 is correct again?

Fahad-Alsaidi:
@tobbi Yes, it is correct. Thanks for your work.

tobbi:
Fixed with #1114

PR #1114, opened 2019-05-25
Fix compilation of SuperTux with RAQM

tobbi:
This should fix issue #1045


tobbi:
Updated screenshot: https://prnt.sc/ntbjz2

PR #1125, opened 2019-06-21
Add fps limiting back

HybridDog:
This should fix the fps limit setting. Please have a look at the code and test it.

### DIFF
--- a/src/supertux/screen_manager.cpp
+++ b/src/supertux/screen_manager.cpp
@@ -181,6 +181,18 @@ struct ScreenManager::FPS_Stats
   float get_fps_min() const { return last_fps_min; }
   float get_fps_max() const { return last_fps_max; }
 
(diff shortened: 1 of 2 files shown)


src/supertux/screen_manager.cpp
```diff
@@ -181,6 +181,18 @@ struct ScreenManager::FPS_Stats
   float get_fps_min() const { return last_fps_min; }
   float get_fps_max() const { return last_fps_max; }
 
+  // This returns the highest measured delay between two frames from the
+  // previous and current 0.5 s measuring intervals
+  float get_highest_max_ms()
```
tobbi: Can you mark this const? It doesn't appear to have side effects.

HybridDog: done

tobbi:
This appears good to go, but I'd rather have this tested regardless. :)

PR #1127, opened 2019-06-24
keyboard configuration for the 2nd player

bou6:
This is a first commit to discuss the implementation for the configuration of the 2nd player.
The changes are in the **KeyboardConfig::keyMap** we add an element in the map to identify to which player belongs the key touched, in **KeyboardManager::process_key_event** we read the this element and we update the corresponding contrller  
@tobbi Lets discuss :)

### DIFF
--- a/src/control/input_manager.cpp
+++ b/src/control/input_manager.cpp
@@ -27,12 +27,15 @@
 
 InputManager::InputManager(KeyboardConfig& keyboard_config,
                            JoystickConfig& joystick_config) :
-  controller(new Controller),
   m_use_game_controller(joystick_config.use_game_controller),
   keyboard_manager(new KeyboardManager(this, keyboard_config)),
   joystick_manager(new JoystickManager(this, joystick_config)),
   game_controller_manager(new GameControllerManager(this))
 {
+	auto controller_1 = std::make_unique<Controller>();
+	auto controller_2 = std::make_unique<Controller>();
+	controllers.push_back(std::move(controller_1));
+	controllers.push_back(std::move(controller_2));
 }
 
 InputManager::~InputManager()
@@ -40,9 +43,9 @@ InputManager::~InputManager()
 }
 
 Controller*
-InputManager::get_controller() const
+InputManager::get_controller(Player::PlayerId id) const
 {
-  return controller.get();
+  return controllers[static_cast<int>(id)].get();
 }
 
 void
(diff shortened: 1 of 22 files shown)


tobbi:
Hello @bou6. Thanks for this PR. I will try to take a look at it in the upcoming days. For now: Both the appveyor and travis builds fail. This might be due to compiler warnings (we have warnings as errors enabled). If you could fix those, that would be great :)

src/control/input_manager.cpp
```diff
@@ -27,22 +27,25 @@
...
+	controllers.push_back(std::move(controller_2));
 }
 
 InputManager::~InputManager()
 {
 }
 
 Controller*
-InputManager::get_controller() const
+InputManager::get_controller(Player::PlayerId id) const
```
tobbi: Please pass id by const ref.

src/supertux/sector.cpp
```diff
@@ -88,8 +88,10 @@ Sector::Sector(Level* parent) :
   if (!player_status) {
     log_warning << "Player status is not initialized." << std::endl;
   }
-  auto player1 = std::make_shared<Player>(player_status, "Tux");
-  auto player2 = std::make_shared<Player>(player_status, "Penny");
+
+  // create players
+  auto player1 = std::make_shared<Player>(Player::PLAYER_1,player_status, "Tux");
+  auto player2 = std::make_shared<Player>(Player::PLAYER_2,player_status, "Penny");
```
tobbi: This can't be Penny since Penny was abducted :)

src/video/glutil.hpp
```diff
@@ -43,7 +43,7 @@ using namespace gl;
 #  include <GLES/glext.h>
 #else
 #  include <GL/gl.h>
-#  include <GL/glext.h>
+//#  include <GL/glext.h>
```
tobbi: Why did you remove this include? Please revert it, I don't know what side effects this might have.

bou6:
> Hello @bou6. Thanks for this PR. I will try to take a look at it in the upcoming days. For now: Both the appveyor and travis builds fail. This might be due to compiler warnings (we have warnings as errors enabled). If you could fix those, that would be great :)

Hi @tobbi, Thanks for the feedback I Just saw the Travis trace, I got the warning 

> error: ‘InputManager::controllers’ should be initialized in the member initialization list [-Werror=effc++] 

it is actually because I didn't initialize the `std::vector< std::unique_ptr<Controller> > ` in the initialiser list but I think there is no way around because the initialiser list copies the argument and we are here dealing with `unique_ptr` so can we just deactivate `-Werror=effc++`

tobbi:
I would assume that just adding
```
controllers(),
```
to the appropriate initialization list (where controller was previously initialized) should be enough, no?

tobbi:
Closing due to lack of interest. Thanks for your contributions and time anyway :)

PR #1128, opened 2019-07-13
Added a "Restart from Checkpoint" button

HybridDog:
Original PR: #885

I've changed it so that the kill function is executed, which plays the hurt sound (otherwise the player would silently loose the coins).

### DIFF
--- a/src/supertux/player_status.hpp
+++ b/src/supertux/player_status.hpp
@@ -40,11 +40,13 @@ public:
   PlayerStatus();
   void reset();
   void add_coins(int count, bool play_sound = true);
+  void take_checkpoint_coins() { coins -= std::max(coins / 10, 25); }
 
   void write(Writer& writer);
   void read(const ReaderMapping& mapping);
 
   int get_max_coins() const;
+  bool can_reach_checkpoint() const;
   std::string get_bonus_prefix() const;/**Returns the prefix of the animations that should be displayed*/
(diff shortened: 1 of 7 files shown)


src/supertux/player_status.hpp
```diff
@@ -40,11 +40,13 @@ class PlayerStatus final
   PlayerStatus();
   void reset();
   void add_coins(int count, bool play_sound = true);
+  void take_checkpoint_coins() { coins -= std::max(coins / 10, 25); }
```
tobbi: AppVeyor complains about
```
'max': identifier not found
```

HybridDog: It cannot find std::max `error C2039: 'max': is not a member of 'std'`. Maybe it's an unreliable transitive header inclusion.
https://ci.appveyor.com/project/supertux/supertux-o8t59/builds/25955016#L5821

tobbi:
I assume that the remaining build issues require a cache purge, so nothing that's caused by this PR. Thanks 👍

Alzter:
Awesome, but change the text to just "Restart from Checkpoint" to avoid padding out the pause menu.

HybridDog:
> Awesome, but change the text to just "Restart from Checkpoint" to avoid padding out the pause menu.

I think this only works as long as the message is not translated to a long one.

PR #1129, opened 2019-07-26
Enable deleting pneumatic and bicycle platforms in editor (#1018)

### LINKED ISSUE #1018: Pneumatic and bicycle platforms cannot be removed in the editor
Erasing them makes them disappear from the editor interface, but when testing the level they still appear.

martinvanzijl:
Fixes #1018. The object is now actually removed from the level (not just from the
editor interface).

Removing any child of a pneumatic or bicycle platform will delete the
whole platform. (From the code it looks like it will not work if there
is just one child.)


tobbi:
Thanks for this! Great contribution! AppVeyor build failures are unrelated!

PR #1133, opened 2019-08-05
Small tweaks to the Icy Island

weluvgoatz:
This PR is is about improving things in the Icy Island, such as level names, making the worldmap more visually pleasing by making it bigger, and getting rid of the enemy spam and the cramped feeling in some levels. I feel like the levels have too many enemies, which would be too hard for beginning players this early in the game. The overall difficulty of the world has been evened out to seem better for gameplay. Also the cave is no longer insanely difficult.  Picnic with Penny has also been updated.  Please consider @serano01 and @Rusty-Box


weluvgoatz:
@Rusty-Box The levels have been updated, please re-review.

weluvgoatz:
Closing this, as I have other, less extreme tweaks I would rather implement instead.  And no one was ever going to test the UPDATED version of this PR anyway, so I will be back with better tweaks at some point in the future.

PR #1139, opened 2019-08-13
Fix bug with falling coins (#942)

### LINKED ISSUE #942: Coins that emit from Tux after dying post checkpoint don't appear
**SuperTux version: 0.6.0 Alpha 3**

##### Expected behavior
Dying after hitting a checkpoint makes coins fly out of Tux as he falls off the screen.

##### Actual behavior
The coins appear after dying post checkpoint but only for a couple of frames.

Die.

martinvanzijl:
Fixes #942. Falling coins will now appear after dying post-checkpoint.

The problem was that the coins were spawning below the bottom of the screen, and therefore being removed immediately in the "update" method. I changed it so they will only be removed if they are also falling downwards (i.e. Y velocity is more than zero).

### DIFF
--- a/src/object/falling_coin.cpp
+++ b/src/object/falling_coin.cpp
@@ -41,7 +41,8 @@ void
 FallingCoin::update(float dt_sec)
 {
   pos += physic.get_movement(dt_sec);
-  if (pos.y > static_cast<float>(SCREEN_HEIGHT))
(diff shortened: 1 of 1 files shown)


tobbi:
Thanks for this! Looks good to me from what I can tell.

Issue #962, opened 2018-12-02
Add music fade in and out scripts


Alzter:
**SuperTux version: 0.6.0 Alpha 3**
**System information:** *Windows 10*

Title says it all. This could be used in levels like the castle of Nolok where the music could fade out at the end instead of abruptly stopping.

Alasdairbugs:
Yeah this would also be useful for custom stages, maybe an option to choose the time period at which the music fades out over?

weluvgoatz:
I'm pretty sure this has already been implemented.

tobbi:
This requires adding a fade in / fade out method to scripting/sector.cpp / .hpp.

Then you'd have to add the appropriate methods to the MusicObject class and then have that call appropriate SoundManager classes.

I'll be the mentor for this one.

oribro:
@tobbi Hi, I'm new here and trying to solve this issue. Can you give some advice?
Thanks

tobbi:
Hello @oribro ,

SuperTux uses Squirrel for scripting (but you don't need to know it in order to help with this issue). First and foremost, all the methods that are available to scripts are in the `src/scripting` directory. There you can see the different object classes, which have scripting methods themselves, which can be called from scripts.

Each sector has a MusicObject instance for the music that is played. See here for the available methods: https://github.com/SuperTux/supertux/blob/6627ddd58e60aab0781022837904eadd091e286c/src/object/music_object.cpp

You can see here how to reference the MusicObject from a sector instance:
https://github.com/SuperTux/supertux/blob/f31276d3d6a4d4ce2d593d750190197e3bb2a4e3/src/supertux/game_session.cpp#L143

The method in question that you need to add to the MusicObject would be `stop_music(float fadetime)`, `pause_music(float fadetime)` and `resume_music(float fadetime)`.

If you need more info, we're available here: https://matrix.to/#/!BfYVBjzEVacHrvDBYw:matrix.org?via=matrix.org&via=synapse.vpzom.click and hopefully someone will get back to you.

VictorJL:
I can work on this one. Just brewing my coffee, shall be ready in no-time.

Rusty-Box:
Apparently this issue was not closed after the #1370 0 was merged so ... I'm closing manually.

PR #1141, opened 2019-08-14
Add script for stopping/fading out music

martinvanzijl:
Adds a script called "stop_music". This takes a float parameter, which is the fade time in seconds. E.g. "stop_music(2.0)" stops the current music with a fade-out time of 2 seconds.

This fixes part of #962. I haven't added a way to fade *in* music yet.

Video demo below:
[stop-music-demo.zip](https://github.com/SuperTux/supertux/files/3503615/stop-music-demo.zip)

One annoying thing - the script will accept "stop_music(2.0)" but not "stop_music(2)". Can I make it accept either a float or integer? Also, can I make it no parameter at all, to stop the music immediately? I'm not very familiar with squirrel.

### DIFF
--- a/src/scripting/functions.cpp
+++ b/src/scripting/functions.cpp
@@ -235,6 +235,11 @@ void play_music(const std::string& filename)
   SoundManager::current()->play_music(filename);
 }
 
+void stop_music(float fadetime)
+{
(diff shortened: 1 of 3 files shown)


Alzter:
Awesome! This could be applied to the cutscene in The Castle of Nolok.

PR #1144, opened 2019-08-24
New Snow Tileset

Alzter:
I added the new snow tiles into the game.


serano01:
While I like them visually, I think, like RustyBox said, that these tiles should replace the old ones, not going alongside with the old stuff.

weluvgoatz:
Please have the snow top thing overlap a little bit, you don't seem to
understand entities start floating when it doesn't overlap.

Alzter:
Well I can change it but that'd mean going over all the snow tiles with snow om them and repositioning the snow. I'll do it soon-ish.

busybody42:
blergh... The snow on the top is too big...

weluvgoatz:
These tiles were implemented properly in a recent commit, thanks Alzter. :-)

PR #1148, opened 2019-08-24
[DO NOT MERGE] Sprites and animation update for the yeti

Rusty-Box:
This PR updates the sprites of the yeti entity including an additional throwing animation and a new animation planned to be used for cutscene purposes.

Here is a little demonstration on how the new sprites will look (not including the new cutscene animation as it is basicly a larger version of thes seen animation on the worldmap after the boss level has been completed).
yeti_demo2
Note: There are still some tweaks I need to do before its finished!


weluvgoatz:
It looks okay, but I recommend adding some fur texture so it looks less... bland and flat.  SuperTux textures tend to not be so basic.  Also, the light should come from the top-left, which this new sprite conflicts with.

Rusty-Box:
@weluvgoatz 1.) Like I said, there is still so stuff to do with the sprites before their 100% done. 2.) While you're right, this might be hard to achive as these sprites are being mirrored when facing in the other direction meaning that even if I would change that it would still look wrong once mirrored. So I don't know if this is even doable in the way it works right now.

qwertychouskie:
\*Sees gif\*

\*Starts humming Boss Attack\*

I agree that he needs fur, otherwise looking great!

Alzter:
Awesome! Maybe make the run look a bit more run-like though, the leg movements look a hit off

Alzter:
@Rusty-Box has there been any new progress on this?

busybody42:
it's ok...
also, what about the milestone 2 death sprites?

Alzter:
> it's ok...
> also, what about the milestone 2 death sprites?

Why are you specifying "Milestone 2"? The Yeti has only been in Milestone 2 so I don't understand what you're talking about, unless you mean Tux, which in that case you're at the wrong issue.

busybody42:
I meant when this happened when the yeti died:
https://youtu.be/8KfV0C9qSR0?t=174

busybody42:
also, what about the "yetibusted" sprites in the files?

slackstone:
OK, I just found this. Come on people, I know I'm biased but this is killer. I LOVE the Yeti update. Thanks Rusty Box!

Guih48:
> It looks okay, but I recommend adding some fur texture so it looks less... bland and flat. SuperTux textures tend to not be so basic. Also, the light should come from the top-left, which this new sprite conflicts with.

I agree. Almost all good supertux sprites in the game is a base image with a "suprrtux-style" shading and a bit pixelous details, like snowballs' eyes (for example, the mr.ice cube's old sprites are fully made with that pixelly techniqe, ans Tux's sprites almost all the shaderes wersion). So I think, theese yeti sprites are very good, but not supertux-styled engnough.

Semphriss:
Closing with permission of opener as new designs are coming.

PR #1154, opened 2019-08-25
Further Update Music

weluvgoatz:
Since Serano updated the music, it has inspired me to add some changes.

1) The editor now only lists .music files, in order to prevent duplicates.
2) Every soundtrack in the game now has a .music file to compensate.
3) Compressed some music tracks for a smaller and faster game download size, since the download size is well over 200MB compared to the 90MB of 0.5.1.

Please review @serano01 since you appear to be the music master around here. ;)


serano01:
It seems to look good (don't know about this failing check), but it also would mean, that the ogg-files would be no longer in use by the levels directly, so I again would to be forced to check every single level by hand.

weluvgoatz:
That would not be the case since the .ogg files can still be in use, they
just can't be selected from the editor.

PR #1156, opened 2019-08-26
Fix bug with killing stunned badguys (#1153)

### LINKED ISSUE #1153: Stunned Ice cubes and snails don't die, when bonus block hit from underneath
If you stun a Mr. Iceblock on a bonus tile, or a wooden box, and then you have Tux get under it, and hit it from underneath, the ice cube doesn't die. Same with Snails.  

this also happens with wooden boxes as well. If Tux is big, and he breaks the box, the stunned badguy falls right on Tux, hurting him

martinvanzijl:
Fixes #1153. Squished ("stunned") Mr. Ice Blocks and Snails are now killed if Tux hits the block underneath them, as expected.

Video demo below:

kill-stunned-badguy-demo

### DIFF
--- a/src/object/block.cpp
+++ b/src/object/block.cpp
@@ -93,18 +93,19 @@ Block::collision(GameObject& other, const CollisionHit& )
 
   // only interact with other objects if...
   //   1) we are bouncing
-  //   2) the object is not portable (either never or not currently)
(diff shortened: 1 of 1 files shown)


tobbi:
Looks good to me! :)

Issue #1150, opened 2019-08-24
Game Sucks Up Memory and Battery


weluvgoatz:
Windows 10, SuperTux since after July 2018, yadah yadah...

I have noticed that the game takes up so much battery power off my computer.  Playing the game makes my computer make very loud whirring noises, presumably trying to cool off the computer.  This was not a problem prior to the pre-0.6 updates, but now it is??? I have no clue what could be causing this, besides the apparent new rendering system that was supposed to make the game run smoother but instead fricked up my system.

Alzter:
Same for me, game takes up a lot more power ever since 0.6 and it runs slower too.

HybridDog:
#1158

weluvgoatz:
@HybridDog Nope, still exists, even under that Pull Request.

HybridDog:
Then I think that this issue could be related to #1051

Issue #1147, opened 2019-08-24
Game is extremely laggy and jittery


Alzter:
**SuperTux version: Latest Nightly**
**System information:** *Windows 10*

The game framerate is extremely jittery for me and I can't get any smooth gameplay. The game randomly speeds up and slows down due to the Vsync compensation fix, which is supposed to make the game more smooth, but just ends up making the game feel unstable. The game takes up a ton of CPU and power usage for no good reason and is ridiculously laggy and jittery. The camera jitter bug is still there, when Vsync is set to "adaptive" it's less noticable but it still happens. The game is ridiculously slow and jittery when compared to 0.5.1 or any prior version and I can't discern why or how.

weluvgoatz:
No use complaining, I've already stated these bugs so many times, but they only seem to notice you...

Alzter:
If they noticed me they'd say or do something about this, but they won't.  At this point we might as well roll everything back to 0.5.1, at least that update worked. I've been working on a new version of SuperTux that nobody cares about in Godot, hopefully if it gets good enough people will notice and it can become Milestone 3 and we can get out of this endless development hell.

HybridDog:
Does this mean that I'm the only one who tested my Pull Requests #1055, #1123 and #1125?
These changes made the game work smoothly for me on Linux.

tobbi:
@Alzter is this still the case?

weluvgoatz:
Well, this problem doesn't exist for me anymore.  The game indeed lags a bit, but it is quite playable and fun.

tobbi:
I'm gonna close this issue for now. Should it re-occur please re-open.

Issue #1134, opened 2019-08-07
Objects fall unrealistically slowly


weluvgoatz:
I am on Windows 10.  This issue occurred sometime between the release of 0.6.0 and July 1st/2nd of 2018, because that was the latest time when I knew for sure the issue didnt exist. 
Anyway, this issue happens whenever something falls, such as Tux or a fireball or an icecrusher.  Whenever something falls, it goes down very slowly, slower than normal, and also it goes diagonally a lot slower than normal too.  Please fix, since this was a major issue in release 0.6, thank you.

tobbi:
You mean *2019 right?

weluvgoatz:
I said what I meant...

HybridDog:
#1158

Issue #1120, opened 2019-06-04
Tux's physics are extremely broken


weluvgoatz:
I am on the WIP 0.6.1, but some of these errors appeared in the 0.6.0 release.

**0.6 Bug:**

When Tux falls, he falls very slowly and the game is slowed down until he stops.

**0.6.1 Bugs:**

Tux can hardly jump, he jumps 2-3 tiles instead of 4-5 and abruptly falls. When he jumps and falls, the game lags heavily and he jitters when he falls.

Because of this I cannot complete any level because Tux cannot jump over pits.

Rusty-Box:
The game indeed is lagging at the moment. However I cannot reproduce the bug involving Tux's jump height.

weluvgoatz:
The jump height was an issue with my computer, sorry.  But jumping, falling, and turning on worldmaps lags the game significantly. Also, if I haven't set my graphics settings to extremely low, the menus open extremely slowly.  Also, I want the game to be able to save my graphics settings (like setting fps to 30)

Alzter:
Ever since 0.6.0 I've been noticing the game takes up a lot more CPU power and sometimes lags more.

HybridDog:
The Framerate setting is currently broken; supertux doesn't work right if it's not set to 60.
I think that I changed screen_manager.cpp in an old supertux version first, and later, while creating that PR, I messed up the framerate setting.
https://github.com/SuperTux/supertux/pull/1055/files
The update_gamelogic function takes a dtime as argument, so I assumed that it supports variable dtime values. In some places, e.g. worldmap camera panning, the dtime is simply ignored (#1100 should fix it).
When I press F2 and significantly increase the game speed, Tux can no longer jump high, i.e. jumping probably does not work right with a large dtime.
Before the PR (e.g. in version 6.0), dtime was 1/framerate, so at that time, the framerate option probably had an impact on jump height, too.

When I made that PR, I didn't think about triple buffering yet. I presume that a high elapsed time can be followed by a low elapsed time, then the frames generated for those dtimes are stored in the buffer and after that vsync alignes both frames with 1/60 time difference.
Here are changes to compensate for this aligning when vsync is enabled:
https://github.com/HybridDog/supertux/commit/c815527316f1d3926c7a7225b167cc964ed39382
This could fix #1117.

To fix the jump height, maybe update_gamelogic can be changed to do more than one collision calculation when dtime is high.

tobbi:
[...]

PR #1158, opened 2019-08-26
Do not use a variable time step

HybridDog:
This makes the game deterministic again. If the frames are drawn too slow, the game slows down. To catch up a bit, it is allowed to process up to three steps at once; the player cannot give different keyboard input in each step when it happens.
With these changes the game generally looks a bit jittery, see https://gameprogrammingpatterns.com/game-loop.html#stuck-in-the-middle for an explanation.
Ideally, the extrapolation could be implemented in SuperTux.

I don't know if I made mistakes again, so please test it.
I especially did not test it on windows.

### DIFF
--- a/src/supertux/menu/options_menu.cpp
+++ b/src/supertux/menu/options_menu.cpp
@@ -23,11 +23,10 @@
 #include "gui/item_toggle.hpp"
 #include "gui/menu_item.hpp"
 #include "gui/menu_manager.hpp"
-#include "supertux/game_session.hpp"
 #include "supertux/gameconfig.hpp"
+#include "supertux/game_session.hpp"
 #include "supertux/globals.hpp"
 #include "supertux/menu/menu_storage.hpp"
-#include "supertux/screen_manager.hpp"
 #include "util/gettext.hpp"
 #include "util/log.hpp"
 #include "video/renderer.hpp"
@@ -56,7 +55,6 @@ enum OptionsMenuIDs {
   MNID_MAGNIFICATION,
   MNID_ASPECTRATIO,
   MNID_VSYNC,
-  MNID_FRAMERATE,
   MNID_SOUND,
   MNID_MUSIC,
   MNID_SOUND_VOLUME,
@@ -74,7 +72,6 @@ OptionsMenu::OptionsMenu(bool complete) :
   next_window_resolution(0),
   next_resolution(0),
   next_vsync(0),
-  next_framerate(0),
   next_sound_volume(0),
   next_music_volume(0),
   magnifications(),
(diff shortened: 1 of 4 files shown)


tobbi:
This at least fixes an issue whereby VSync would crash my Mac upon exiting. I tested on Mac, though, so not Windows.

tobbi:
I spoke too soon. Vsync has the ability to crash my mac, or rather: To have it hang upon exiting full screen when VSync is enabled so that I have to restart my machine,

I don't suppose you know what could be done (because it seems like an OS-related problem) but I thought I'd relay it to you anyway.

HybridDog:
I cannot help you, but you can open a new issue for this problem.

I assume that using a fixed time step could fix many issues,
for example #1150, #1147, #1134, #1120, #1117.
The FPS setting introduced by @Grumbel does no longer work; 64 logical FPS makes it impossible to exceed 64 FPS as long as there is no extrapolation.
https://github.com/SuperTux/supertux/commit/2a5486da1b2ff132fb7102b7cfb001408bf07b4b#diff-008dba05634cef29ce6e9c40cead3e5f
I think I need to remove the fps limiting from this pull request.

tobbi:
Hey, @HybridDog Can you please rebase this and fix the warning
```
/home/travis/build/SuperTux/supertux/src/supertux/screen_manager.cpp: In member function ‘void ScreenManager::run()’:

/home/travis/build/SuperTux/supertux/src/supertux/screen_manager.cpp:494:50: error: conversion from ‘Uint32’ {aka ‘unsigned int’} to ‘float’ may change value [-Werror=conversion]
```

Just so Travis is not a special snowflake anymore and we can get Windows builds? Thanks.

HybridDog:
> Can you please rebase this and fix the warning

I didn't find the warning. Clang doesn't show it to me; make --trace shows the flags, which include the warnings as far as I know: https://pastebin.com/raw/EPV38UX1

HybridDog:
I have changed the code so that the maximum logical steps executed per frame depends on the current FPS.
If the game is laggy, it calculates up to four steps at once; in this situation the same player's keyboard input is probably used for all of the four steps.
If the game does not lag, it executes not more than two steps before rendering a frame.

HybridDog:
Please test.
https://ci.appveyor.com/project/supertux/supertux-o8t59/builds/27234434/artifacts

tobbi:
@weluvgoatz I remember you testing this. What was your evaluation again?

weluvgoatz:
It fixes everything it claimed to fix.  If no more commits are needed, I
recommend merging it (Although, it does not fix the immense lag, background
shifting and battery usage)

PR #1166, opened 2019-09-12
Reduce forest dirt and rock saturation and adjust colour and brightness

HybridDog:
@weluvgoatz wrote:
> The Forest, however, […] Graphics? Absolutely horrendous, with red dirt, […]

I have changed the forest dirt textures, please tell me what you think.
Btw: As far as I know, in a real spruce forest (in Germany), there are needles on the top, under the needles there is a fluffy dark-brown layer, and further down there is yellow or grey loam.

Before:
before
After:
after


Zwatotem:
Could you please merge it, since @HybridDog already made the sprites, so it doesn't cost anything, and can reduce our eyes "pain", for the time we wait for the new forest assets?

PR #1173, opened 2019-09-19
Use Batching for the interactive particles

HybridDog:
This should increase the fps in rainy levels a bit; please test it.


tobbi:
I tested briefly and this looks good to be merged.

PR #1180, opened 2019-10-11
Debugging: Give each collision group an individual colour when drawing hitboxes

HybridDog:
2019-10-11-151056_1920x1080_scrot


tobbi:
Nice!

PR #1182, opened 2019-10-13
Poison Ivy: Slow Falling and Rename

weluvgoatz:
In this pull request, poison ivies have been renamed to "Spring Leaf", since they aren't really poison ivies. 

Also, poison ivies now float and fall slowly, just like a leaf in real life.

### DIFF
--- a/src/badguy/poisonivy.cpp
+++ b/src/badguy/poisonivy.cpp
@@ -48,6 +48,9 @@ PoisonIvy::initialize()
 void
 PoisonIvy::active_update(float dt_sec)
 {
+  if (!on_ground()) {
+  m_sprite->set_action(m_dir == Direction::LEFT ? "falling-left" : "falling-right");} else if (on_ground()) {
+  m_sprite->set_action(m_dir == Direction::LEFT ? "left" : "right");}
   WalkingBadguy::active_update(dt_sec);
 }
 
--- a/src/badguy/poisonivy.cpp
+++ b/src/badguy/poisonivy.cpp
@@ -26,14 +26,29 @@
 
 PoisonIvy::PoisonIvy(const ReaderMapping& reader)
   : WalkingBadguy(reader, "images/creatures/poison_ivy/poison_ivy.sprite", "left", "right")
-{
+ {
   walk_speed = 80;
-}
+ }
 
 PoisonIvy::PoisonIvy(const Vector& pos, Direction d)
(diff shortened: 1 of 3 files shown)


Rusty-Box:
The idea is fine but I think it should get proper floating sprites before this gets merged because otherwise it is a bit odd looking.

src/badguy/poisonivy.cpp
```diff
@@ -26,14 +26,29 @@
 
 PoisonIvy::PoisonIvy(const ReaderMapping& reader)
   : WalkingBadguy(reader, "images/creatures/poison_ivy/poison_ivy.sprite", "left", "right")
-{
+ {
   walk_speed = 80;
-}
+ }
```
HybridDog: Please don't add spaces before the braces.

src/badguy/poisonivy.hpp
```diff
@@ -18,21 +18,25 @@
...
   virtual bool is_freezable() const override;
   virtual std::string get_class() const override { return "poisonivy"; }
-  virtual std::string get_display_name() const override { return _("Poisonous ivy"); }
+  virtual std::string get_display_name() const override { return _("Spring Leaf"); }
 
 protected:
   virtual bool collision_squished(GameObject& object) override;
 
 private:
+  Physic physic;
```
HybridDog: It looks like `physic` is unused. I think that `m_physic` is from some super class.

weluvgoatz:
@Rusty-Box I've added support for a falling animation.  You can always change it when you make the new poision ivy sprite, this sprite is just a placeholder.

weluvgoatz:
Closing this, it's not ready yet.

Rusty-Box:
Reopening, as it is still considered but has to wait for required graphics. Closing might have us forgot that this PR still gets merged in the future.

Alzter:
Idea: When Poison Ivy falls it turns into a flat leaf, and sways and tilts left to right as it glides to the ground.

busybody42:
no... This is a unique idea, but poison ivies are a better enemy... also, spring leaf could be a bouncing snowball for the forest, but otherwise.... yeah, no... but the slow falling part IS ok, but could break levels...

weluvgoatz:
Closing this seen as it has been open 8 months, and I'm working on different things rather than this, which I would like to remain out of my mind.  Since we have a design document now, I added this idea to that so when the time comes we can revisit this, but as for now, I'm closing it.

PR #1183, opened 2019-10-13
Explosion: Push by default and stronger push

weluvgoatz:
This pull request makes the push explosion effect stronger, and also applies it to bombs, haywires and goldbombs so their explosions push back players.  Please test, thanks!

### DIFF
--- a/src/object/explosion.cpp
+++ b/src/object/explosion.cpp
@@ -29,7 +29,7 @@
 Explosion::Explosion(const Vector& pos) :
   MovingSprite(pos, "images/objects/explosion/explosion.sprite", LAYER_OBJECTS+40, COLGROUP_MOVING),
   hurt(true),
-  push(false),
+  push(true),
   state(STATE_WAITING),
   lightsprite(SpriteManager::current()->create("images/objects/lightmap_light/lightmap_light-large.sprite"))
 {
@@ -76,7 +76,7 @@ Explosion::explode()
 
   if (push) {
     Vector center = m_col.m_bbox.get_middle ();
-    auto near_objects = Sector::get().get_nearby_objects (center, 10.0 * 32.0);
+    auto near_objects = Sector::get().get_nearby_objects (center, 128.0 * 128.0);
 
     for (auto& obj: near_objects) {
       Vector obj_vector = obj->get_bbox ().get_middle ();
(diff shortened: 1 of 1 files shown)


src/object/explosion.cpp
```diff
@@ -90,9 +90,7 @@ Explosion::explode()
 
       /* The force decreases with the distance squared. In the distance of one
        * tile (32 pixels) you will have a speed increase of 150 pixels/s. */
-      float force = 150.0f * 32.0f * 32.0f / (distance * distance);
-      if (force > 200.0f)
-        force = 200.0;
+      float force = 150.0f * 64.0f * 64.0f / (distance * distance);
```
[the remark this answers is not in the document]
weluvgoatz: fixed because no cap could cause unexpected behaviour

src/object/explosion.cpp
```diff
@@ -76,7 +76,7 @@ Explosion::explode()
 
   if (push) {
     Vector center = m_col.m_bbox.get_middle ();
-    auto near_objects = Sector::get().get_nearby_objects (center, 10.0 * 32.0);
+    auto near_objects = Sector::get().get_nearby_objects (center, 128.0 * 128.0);
```
HybridDog: I think the 32.0 factor is there to have the value in tiles, so 10.0 * 32.0 means 10.0 tiles.
Here the argument corresponds to 512 tiles: 128.0*128.0=512.0*32.0

HybridDog:
After this change it may be possible that Tux can push himself to places where he should not be, so in some levels he may reach the goal earlier than he should. I don't know if it is important to avoid possible compatibility problems in existing levels.

weluvgoatz:
The bombs cannot cause significant pushes to make players move that far upwards, save for shortfuses.  Shortfuses are hardly used anyway, and the levels where they are used use them sparingly so they are not used in this way (Owls' Skydive Commando, some Bonus 3/4 levels)

src/object/explosion.cpp
weluvgoatz: I might change/fix that, but bigger number = more significant explosion :P

HybridDog:
> The bombs cannot cause significant pushes to make players move that far upwards […]

I've tested it (your explosion branch); the bomb can push up Tux far into the sky: 
https://streamable.com/1f7r2

PR #1184, opened 2019-10-17
Natural swimming mechanic proposal

Zwatotem:
This is a **deep** work in progress PR; dirty code, bad performance and all that stuff, but it compiles (at least at my machine) and I wanted to know your early opinions on the mechanic itself (how it feels).
Basically, when swimming there is a hidden (not graphically represented) vector pointing at current acceleration direction. Arrow keys fluently change this direction (i.e. acceleration can be at any angle, but it's going torwards 1 of 8, that is currently pointed) and apply small force. One time boost in currently pointed direction can be achived by using jump key. Besides that there is weak gravity and dynamic friction (simple models).


weluvgoatz:
Okay, I have tested it and here are my concerns:
1) You cannot control Tux's jump, and he jumps at 100% every time even when you don't want to, which is very annoying.  Not swimming related, but it was due to this PR.
2) Tux sprite does not rotate back to normal after hitting the ground.
3) It is very hard to control Tux in the water and the sprite looks very weird in the water.

Zwatotem:
1 & 2 is (I belive) repairable with some care.
3. I'm not an artist, so I used spirites I had available, but I also planned to add some for vertical movement, possibly reuse climbing. I'm worried I can't do much with the last one, (maybe changing some parameters would improve something). I'll get back with some new solutions and we'll see.

Zwatotem:
Most of problem @weluvgoatz mentioned are gone, but there are still some:
1. Boosting in upper left direcion is impossible due to issue #1188
2. The sprites are inconsistant
3. Hitbox of tux should rotate alongside his sprite
4. Underwater collision interactions have to be coded separately from normal ones

If you feel like it's okay, I'd like to merge it now (or after some cleanup commits), but to feature/swimming instead of master. That branch is pretty far behind though, so it needs to be updated first. Then I could add the rest in next pull requests.

tobbi:
@weluvgoatz You okay with pushing this to the swimming branch?

weluvgoatz:
@tobbi Yes, it looks nice, and my first 2 problems of it have been solved.  However, Tux's sprite is still messed up (but this time, only inside the water) and I would like better control of Tux in the water (he currently flies all over the place and its very hard to control him).

weluvgoatz:
@Zwatotem Proper sprites can come later, and I don't currently think it's possible to rotate sprite hitboxes.

Zwatotem:
Yes, it isn't. In last commit I added simple flip of Tux' width and height, when he moves more horizontally. It's not much, but better than nothing.

weluvgoatz:
I think this can be merged into feature/swimming, and it can be worked on from there.  It seems swimming sprites and adjustments are needed, which can be resolved there.

PR #1185, opened 2019-10-18
Fix game speed

weluvgoatz:
There are problems with the diff, I have only changed 1 line of code (line 498).  This makes the game still use the fixed time step, but also the game runs at the proper speed at all times.  I also have only tested on Windows, so Linux and Mac users please test as well.


HybridDog:
I've fixed the diff problem.

> This makes the game still use the fixed time step […]

The `dtime` value is this time step, so if you change it instead of the number of `update_gamelogic` invocations, the physics calculation can lead to unexpected behaviour.  
As far as I can see, the m_speed value is only used during the end sequence, title screen, etc., so in this case changing `dtime` doesn't affect the gameplay. 
I've added a comment about this behaviour and removed a redundant if condition which I added some time ago when I wrote the code.

weluvgoatz:
Thank you @HybridDog. Does this still make the game speed function properly?  And if so, can it be merged?

HybridDog:
> Does this still make the game speed function properly?

I've just tested it and it works. Thanks for finding this; I have somehow overlooked that there is another time multiplier in addition to the debug speed setting.

> And if so, can it be merged?

I'll click the "Squash and Merge" button.

PR #1187, opened 2019-10-20
Add a ghosforest background image

HybridDog:
The compression artifacts are intentional because looking disturbing is the image's purpose.
bg


HybridDog:
@serano01, JPG images are decoded very fast and converting them to PNG greatly increases the file size. Why did you convert the background image to PNG?

serano01:
Because most images are already in the .png format, so that it has some consistency.

PR #1190, opened 2019-10-26
Update clouds

weluvgoatz:
There is only 1 variant to the cloud particles right now because the particle engine currently cannot support multiple clouds.  Also, I updated the cloud tile for consistency and to fix the tile error.


Rusty-Box:
@tobbi Any idea on how to make it doable to have multiple variants of clouds as particles?

tobbi:
This would requiren a slight code change, but it's doable.

PR #1191, opened 2019-10-27
Fix missing music due to path changes

### LINKED ISSUE #1172: The music is missing in many levels
Since the music has been moved into subfolders, SuperTux does not find the music files in the levels which I have downloaded.
For legacy support, the music file could be searched in the subfolders if it isn't found.

HybridDog:
If the music has not been found, a fallback path is searched for.
This should fix #1172; please test it.

### DIFF
--- a/src/audio/sound_file.cpp
+++ b/src/audio/sound_file.cpp
@@ -23,6 +23,7 @@
 #include <string.h>
 #include <physfs.h>
 #include <sstream>
+#include <unordered_map>
 
 #include "audio/ogg_sound_file.hpp"
 #include "audio/sound_error.hpp"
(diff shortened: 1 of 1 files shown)


tobbi:
Thank you so much for this! I had the same idea in mind but didn't manage to do it.

HybridDog:
I forgot that the fallback paths all have a slash delimiter, so they probably do not work on Windows.

tobbi:
No problem, can you fix it?

HybridDog:
I've just seen that the paths in the level files are slash-delimited, too, (e.g. `(file "music/antarctic/chipdisko.ogg")`), so it could already work on Windows, but I haven't tested it.

tobbi:
@Rusty-Box You use this thing that they commonly refer to as "WIndows". Can you test it?

PR #1195, opened 2019-10-29
Fix a -Wreturn-std-move-in-c++11 compiler warning

HybridDog:
Clang showed a warning about StreamSoundSource to me:
```
[…]/src/audio/sound_manager.cpp:141:14: warning: prior to the resolution of a defect report against ISO C++11, local variable 'stream_source' would have been copied despite being returned by name, due to its not matching the function return type ('unique_ptr<OpenALSoundSource, default_delete<OpenALSoundSource>>' vs 'unique_ptr<StreamSoundSource, default_delete<StreamSoundSource>>') [-Wreturn-std-move-in-c++11]
      return stream_source;
             ^~~~~~~~~~~~~
[…]/src/audio/sound_manager.cpp:141:14: note: call 'std::move' explicitly to avoid copying on older compilers
      return stream_source;
             ^~~~~~~~~~~~~
             std::move(stream_source)
1 warning generated.
```

### DIFF
--- a/src/audio/sound_manager.cpp
+++ b/src/audio/sound_manager.cpp
@@ -138,7 +138,7 @@ SoundManager::intern_create_sound_source(const std::string& filename)
       auto stream_source = std::make_unique<StreamSoundSource>();
       stream_source->set_sound_file(std::move(file));
       stream_source->set_volume(static_cast<float>(m_sound_volume) / 100.0f);
-      return stream_source;
+      return std::move(stream_source);
     }
   }
 


HybridDog:
I have an offtopic question and I'm not sure where to ask it:
Why do the header files use `define`s and not `#pragma once` to avoid that header code is compiled multiple times?

tobbi:
> I have an offtopic question and I'm not sure where to ask it:
> Why do the header files use `define`s and not `#pragma once` to avoid that header code is compiled multiple times?

Reason being "It's always been done that way". Is `#pragma once` honoured by all compilers?

HybridDog:
> Is #pragma once honoured by all compilers?

It's supported in many but not all compilers: https://en.wikipedia.org/wiki/Pragma_once#Portability
_Include guards_ and _#pragma once_ do (almost) the same, so for consistency it's probably good to stick with include guards.

plonibarploni:
I get the following warning now when compiling SuperTux with gcc:
```
[…]/src/audio/sound_manager.cpp: In member function ‘std::unique_ptr<OpenALSoundSource> SoundManager::intern_create_sound_source(const string&)’:
[…]/src/audio/sound_manager.cpp:141:23: warning: redundant move in return statement [-Wredundant-move]
  141 |       return std::move(stream_source);
      |              ~~~~~~~~~^~~~~~~~~~~~~~~
[…]/src/audio/sound_manager.cpp:141:23: note: remove ‘std::move’ call
```

HybridDog:
> I get the following warning now when compiling SuperTux with gcc

I do not get a warning with the current code. I use `clang version 7.0.0-3~ubuntu0.18.04.1 (tags/RELEASE_700/final)`.
If I remove the `std::move`, the warning mentioned in the PR description (see above) still appears.

tobbi:
I got a solution for that pickle:

Use a preprocessor directive./s

No, why do both compilers behave differently?

PR #1196, opened 2019-10-29
Earth Tux: Do not rotate the lamp during a butt jump

HybridDog:
The light cone did not change although Tux' head with the lamp has rotated. This looks wrong to me.
Now the side-view butt jump textures are used so that the lamp fits to the light cone. As a side effect, this adds a bit more variety.

Earth Tux stands:
2019-10-29-113552_1920x1080_scrot

Butt jump with the lamp facing in the wrong direction:
2019-10-29-113540_1920x1080_scrot

But jump with this PR applied:
2019-10-29-113043_1920x1080_scrot


serano01:
I don't think that we want to use the old sprite ever again, since the new is an huge improvement.

Rusty-Box:
I agree. Rather we should change the light

HybridDog:
I personally like the old texture with the red eyes more than the current ones.

> Rather we should change the light

It is difficult to rotate the light cone, but you can rotate the lamp on the helmet.
The lamp is attached on the helmet, which is attached on Tux' head, so you would need to rotate the head and hat in all buttjump textures.

Alzter:
To me it looks fine as it is now?

busybody42:
maybe make it shine downwards while tux does a ground pound! (he looks down while doing it...)

serano01:
I'm closing this, since this will not be considered.

PR #1198, opened 2019-10-31
Keep collisions enabled when Tux turns invisible

### LINKED ISSUE #1003: Tux.set_visible(false); deactivates Tux’ collision
**SuperTux version:** 0.6.0
**System information:** Arch Linux

##### Expected behavior
`Tux.set_visible(false);` makes Tux invisible
##### Actual behavior
`Tux.set_visible(false);` makes Tux invisible and deactivates his collision. Tux then falls out of the level and dies.
Create a new test level and insert script trigger with `Tux.set_visible(false);`.

I think this bug is what causes “Picnic with Penny” to come to a standstill after the camera scrolls to Penny: the rap_scene never gets activated and I can’t see Tux. I went to the game files and commented `Tux.set_visible(false);` in `intro.nut` and `intro.stl`out and the cut scene ran trough without the bug.

martinvanzijl:
Fixes #1003. Below is a demo where I put a script trigger with "Tux.set_visible(false);" to the right of the coins. Note that Tux stays on the platform and can interact with the coins after turning invisible.

invisible-demo

To do this, I removed the line which moves Tux from "COLGROUP_MOVING" to "COLGROUP_DISABLED". I don't think this will have any adverse effects. Can anyone confirm?

### DIFF
--- a/src/object/player.cpp
+++ b/src/object/player.cpp
@@ -1143,7 +1143,6 @@ void
 Player::set_visible(bool visible_)
 {
   m_visible = visible_;
-  set_group(visible_ ? COLGROUP_MOVING : COLGROUP_DISABLED);
 }
 
 bool


Rusty-Box:
@tobbi you know the code better than me ;D

tobbi:
Looks good.

PR #1199, opened 2019-11-02
Credits Menu Revamp

weluvgoatz:
Remade the credits into a level, and updates to add all the new contributors.  There is a conflict but why?


serano01:
Looks good, but maybe after the end, it should go back to the main menu automatically.

tobbi:
With @serano01 's comments and please the ability to scroll through with arrow keys, I think this should be ready to be merged.

PR #1200, opened 2019-11-02
Bomb: Push when explosion

weluvgoatz:
I made a similar PR, but had to be reverted because of an uncontrolled variable.  Here, the explosion is less extreme.

### DIFF
--- a/src/object/explosion.cpp
+++ b/src/object/explosion.cpp
@@ -29,7 +29,7 @@
 Explosion::Explosion(const Vector& pos) :
   MovingSprite(pos, "images/objects/explosion/explosion.sprite", LAYER_OBJECTS+40, COLGROUP_MOVING),
   hurt(true),
-  push(false),
+  push(true),
   state(STATE_WAITING),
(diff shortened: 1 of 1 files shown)


HybridDog:
If Tux has a star, he can use the bomb to jump very high; I think this shouldn't be possible: https://streamable.com/07qqi

src/object/explosion.cpp
```diff
@@ -90,9 +90,10 @@ Explosion::explode()
 
       /* The force decreases with the distance squared. In the distance of one
        * tile (32 pixels) you will have a speed increase of 150 pixels/s. */
-      float force = 150.0f * 32.0f * 32.0f / (distance * distance);
-      if (force > 200.0f)
-        force = 200.0;
+      float force = 150.0f * 64.0f * 64.0f / (distance * distance);
+	  // If we somehow get a force of over 500, keep it at 500 because unexpected behaviour could result otherwise.
+	  if (force > 500.0f)
```
HybridDog: I can see inconsistent indentation here (maybe tabs need to be changed to spaces).

tobbi:
Can anyone that is not me take a look at this?

PR #1201, opened 2019-11-02
Game Menu: Fix very long string

weluvgoatz:
The string in the editor about the Checkpoint Restart is too long, so I have shortened it in this pull request.


tobbi:
I'm actually not sure if it's verbose enough. I think we can omit the "Level" from there.

"Restart from checkpoint" how does that sound? Checkpoint restart might make sense to some, but I am not sure if all people get it.

PR #1202, opened 2019-11-02
Tux: Various Improvements

weluvgoatz:
I have done 2 things in this PR:
1) Add a separate fall animation of Tux, because of the other Tux pull request where Tux has a new fall animation
2) Tux's friction has been improved so it is easier to control him.

### DIFF
--- a/data/images/creatures/tux/light.sprite
+++ b/data/images/creatures/tux/light.sprite
@@ -41,6 +41,16 @@
          (mirror-action "earth-jump-right"))
 
        (action
+         (name "earth-fall-right")
+         (hitbox 516 134 32 64)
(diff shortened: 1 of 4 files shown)


serano01:
Why don't you include the hat fix and the new small Tux jump fix, as well?

weluvgoatz:
@serano01 I cannot include them in this PR because the current branch of this repository doesn't have that issue.  That is the problem of another pull request, so those fixes have to be part of THAT pull request.

weluvgoatz:
I will bring back this PR once the Tux-Revamp is done, or if I am able to, I can contribute these changes to that PR.

PR #1203, opened 2019-11-03
[WIP] Editor Revamp

weluvgoatz:
Revamped some areas of the editor and added some new keyboard combos.  Also changed some naming conventions to make getting around the editor easier.

(Possibly) To-Do:

- [x] Fix conflicts + bugs (currently none exist)


src/object/ambient_sound.hpp
```diff
@@ -59,7 +59,7 @@ class AmbientSound final : public MovingObject,
   virtual HitResponse collision(GameObject& other, const CollisionHit& hit_) override;
 
   virtual std::string get_class() const override { return "ambient-sound"; }
-  virtual std::string get_display_name() const override { return _("Ambient sound"); }
+  virtual std::string get_display_name() const override { return _("Sound"); }
```
HybridDog: https://backtracks.fm/resources/podcast-dictionary/ambient+sound
> Ambient sound in relation to audio refers to the background noise present at a given scene or a location. This can include noises such as rain, traffic, crickets, birds, et cetera.

Ambient Sound is more specific than just Sound.

src/object/level_time.hpp
```diff
@@ -52,7 +52,7 @@ class LevelTime final : public GameObject,
 
   /** @} */
   virtual std::string get_class() const override { return "leveltime"; }
-  virtual std::string get_display_name() const override { return _("Level time"); }
+  virtual std::string get_display_name() const override { return _("Timer"); }
```
HybridDog: Maybe `Time Limit` (LevelTime is the time limit and not the target time, I think).

PR #1206, opened 2019-11-09
Tilegroup and objectgroup tweaks

Rusty-Box:
While the latest changes to the tile- and objectgroup have made them somewhat more comprehencible it also made some things more unnecessarily "clustered".

My proposed changes:
- restore the **projectile** and **bosses** category as objectgroups
- move "resizable objects" (secret area, script trigger, etc.) to the **sector** objectgroup
- remove the WIP forest tiles tilegroup until they are fully finished (the tiles themselve are still ingame, only the tilegroup is gone!)
- re-add the **industrial** and **forest background** tilegroup
- add a tilegroup for unisolid & lightmap tiles
- re-add the older snow cave tile to the snow tilegroup as they are still being used in the offical story mode levels


weluvgoatz:
I reccomend two things:
1) Move the triggers back into Environment (We agreed on this over Riot btw)
2) Remove the "tiles" part of "Old/Unused Tiles" for consistency
Otherwise looks good!

PR #1208, opened 2019-11-09
Buttjump hitbox adjusted to reflect sprite

Zwatotem:
Now Tux's hitbox gets squished to the center of his body while performing pre-buttjump animation. That means he can jump higher/further and avoid collisions from above (e.g. from ice spikes).
Before:
before
After:
after


tobbi:
@Rusty-Box @serano01 @weluvgoatz @Alzter You guys approve?

weluvgoatz:
I will test later.

weluvgoatz:
The buttjump hitbox is broken and Tux digs into the ground when he buttjumps.

Zwatotem:
Okay, to be fair, I also adjusted the sprite to follow intuition, e.i. to squize to the center of his body instead of lower part. And then I addjusted the intuition :-) , so that Tux can easly get one full tile higher (jump on gifs was a criterium). So in the end Tux' hitbox shirnks about 15px from the bottom, and like 2px from the top.
Maybe I should change it to more symmetric? Let me know.

> The buttjump hitbox is broken and Tux digs into the ground when he buttjumps.

Don't forget, that on Windows sprites are not reloaded on installation. For me it's fine.

HybridDog:
If I understand it correctly, this can break a lots of existing levels because Tux can jump to places which he should not be able to reach.

Alzter:
Kinda looks like he's clipping through the floor and could break existing levels as @HybridDog said, maybe not...?

Zwatotem:
I have no time to adjust it further now, I'll take care of it later. @Alzter feel free to test for yourself.
I don't think it's gonna do that much of a difference, since if somebody is grown, he's gonna reach the same height already. Although I consider redoing it since it's primary goal was to shrink collision from above, not from below.

Zwatotem:
This has been getting no interest over last few months, closing.

PR #1217, opened 2019-11-18
Fix ground pounding bug

### LINKED ISSUE #966: "Ground-pounding" against a breakable wall breaks it
v0.6.0 beta-1
win10 64bit

##### Run against a wall of a breakable tile, eg, in the "block" tilemenu, the brown bricks, and groundpound as you hit the wall. Video:

https://youtu.be/RHeBQwGCN0E

- this also exists in v0.5.1

martinvanzijl:
Fixes #966.

The bug is easiest to reproduce by standing next to some bricks, holding down and right (or left) and pressing the jump button, as shown below:

break-test-before-fix

After the fix, the problem does not happen (see below):

break-test-after-fix

I also tested with ground-pounding as per original bug description; this also seems to work fine now.

### DIFF
--- a/src/object/block.cpp
+++ b/src/object/block.cpp
@@ -86,7 +86,11 @@ Block::collision(GameObject& other, const CollisionHit& )
 {
   auto player = dynamic_cast<Player*> (&other);
   if (player) {
-    if (player->get_bbox().get_top() > m_col.m_bbox.get_bottom() - SHIFT_DELTA) {
+    bool x_coordinates_intersect =
+        player->get_bbox().get_right()  >= m_col.m_bbox.get_left() &&
+        player->get_bbox().get_left()   <= m_col.m_bbox.get_right();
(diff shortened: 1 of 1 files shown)


weluvgoatz:
Seems good enough, but this might break some levels that use this bug to their advantage.

Rusty-Box:
Why use a bug as a level gimmick? Also, using a bug for a level doesn’t prevent this bug from being fixed thus this argument in my opinion is invalid.

weluvgoatz:
I am not arguing, I am just stating a fact... I am in favor of this pull request.

Zwatotem:
You mean any specific official/bonus levels or just "some levels"?

Rusty-Box:
@welugoatz I didn’t try to blame you I was just talking about the argument in general. :)

tobbi:
Can this be merged?

weluvgoatz:
Yes please

PR #1246, opened 2019-11-27
Trampoline leaving bonus block no longer hurts Tux

### LINKED ISSUE #990: Buttjumping on trampoline 'bonus block' into impossible position

Zwatotem:
There were some official levels which contained a weird bug with a trampoline, described in existing issue thread. This commit fixes it. (Fix #990)

### DIFF
--- a/src/object/bonus_block.cpp
+++ b/src/object/bonus_block.cpp
@@ -315,7 +315,7 @@ BonusBlock::try_open(Player* player)
 
     case Content::CUSTOM:
     {
-      Sector::get().add<SpecialRiser>(get_pos(), std::move(m_object));
+      Sector::get().add<SpecialRiser>(get_pos(), std::move(m_object), true);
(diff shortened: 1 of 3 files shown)


plonibarploni:
This doesn't completely fix the issue. Pressing down still hurts Tux, and it still places Tux in an impossible position. Now it also slides down and back up again:
Peek 2020-01-08 17-35

Zwatotem:
I wasn't able to reproduce it. Does it occur on 0.6.1 or on latest master builds?

PR #1265, opened 2019-12-02
Avoid stacked objects jittering

### LINKED ISSUE #395: Stacked Objects "Jitter"
Something has changed in how objects collide with each other.  This is evident when objects are stacked atop each other, as some of the objects are continually nudged up and down.  This is most prominent when stacking objects that Tux can carry such as rocks, trampolines, and lanterns (see attached test file).  If Tux is carrying something and drops it on top of another object directly in front of him you can reliably observe the effect.  This also occurs with scripted objects.  I am surprised this hasn't been already reported as it appears in 0.4.0 and causes myriad instances of ugly behavior.

[jitter_block.zip](https://github.com/SuperTux/supertux/files/191195/jitter_block.zip)

martinvanzijl:
Fixes #395. I tested this with a large column of stacked rocks (see videos below), and it looks like they still behave as expected - without jittering, of course.

I haven't tested using other stackable objects yet.

Before:

stack-test-before

After:

stack-test-after

If anyone tests this and finds it causes undesired behaviour, please let me know.

### DIFF
--- a/src/object/rock.cpp
+++ b/src/object/rock.cpp
@@ -116,6 +116,15 @@ Rock::collision(GameObject& other, const CollisionHit& hit)
   if (is_grabbed()) {
     return ABORT_MOVE;
   }
+
+  // Don't fall further if we are on a rock which is on the ground.
+  // This is to avoid jittering.
+  auto rock = dynamic_cast<Rock*> (&other);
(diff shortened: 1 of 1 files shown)


tobbi:
Thank you so much! This issue has been bugging us for so many years!

PR #1271, opened 2019-12-08
Use green font for perfect level stats

Wuzzy2:
This PR will make change the font color to green for displayed level stats if they are “perfect”, e.g. the level target has been reached.
“Perfect” stats: All badguys killed / all coins collected / all secrets found / target time beaten.

This affects the following:
* Level intro screen
* Level outro screen
* Level stats in worldmap

The record time will become green if it is lower than the target time. If the level does not have a target time, it turns green when the level was beaten.

Examples:

* If you got 100/100 coins, the coins text will be green. If you got fewer coins, the text will be the default color.
* If you beat the level in 30 seconds, and there is a target time of 1 min, time text is green.
* If you beat the level in 4 hours, and there is no target time, time text is still green.

---

Rationale:
With this you can see which parts of the level you did perfectly at a glance. You don't need to read anymore, as soon you see green, you know it's perfect. Basically it's a small quality-of-life improvement.

### DIFF
--- a/src/supertux/colorscheme.cpp
+++ b/src/supertux/colorscheme.cpp
@@ -30,9 +30,11 @@ Color LevelIntro::s_header_color(1.f,1.f,0.6f);
 Color LevelIntro::s_author_color(1.f,1.f,1.f);
 Color LevelIntro::s_stat_hdr_color(0.2f,0.5f,1.f);
 Color LevelIntro::s_stat_color(1.f,1.f,1.f);
+Color LevelIntro::s_stat_perfect_color(0.4f,1.f,0.4f);
 
 Color Statistics::header_color(1.f,1.f,1.f);
 Color Statistics::text_color(1.f,1.f,0.6f);
+Color Statistics::perfect_color(0.4f,1.f,0.4f);
 
 Color ColorScheme::Menu::default_color(1.f,1.f,1.f);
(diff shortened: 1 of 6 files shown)


Rusty-Box:
Which type of green did you use? In general I am in favor of this idea. But depending on the type of green used it may be hard to read.

tobbi:
@Rusty-Box I assume you checked, right? This good?

HybridDog:
Here are some screenshots. The green is bright enough imo.
tmp_screenshot

Here are two more screenshots; note that I've compressed them to jpg with 4:2:0 colour sampling, so the colours look a bit wrong (unsaturated):
fot_2020-01-18-20:10:39_820767704
fot_2020-01-18-20:10:56_124248564

Alzter:
Now that is some nice quality of life stuff. I think it'd be even cooler if you made the category text green for the worldmap and level clear screen so it'd stand out more.

Wuzzy2:
> I think it'd be even cooler if you made the category text green for the worldmap and level clear screen so it'd stand out more.

Is this a requirement for getting this accepted?

Rusty-Box:
> Is this a requirement for getting this accepted?

@Wuzzy2 I wouldn't say necessarily but it would be quite nice to have everything in one PR.

Wuzzy2:
Personally, I like it the way as-is.

Issue #1111, opened 2019-05-23
Make cloned/mirrored .sprite actions have the same properties as the source animation


Alzter:
I want cloned / mirrored actions in .sprite files to get the properties of the source action. For example, if I make an action and I give it the property `(loops 1)`, and then I make an action that mirrors or clones the previous action it will have loops set to 1 for default unless otherwise specified.

LiquidityC:
I've got a PR for this. I decided to copy the props: "fps", "loops" and "hitbox". Not sure if you wanted all that.

tobbi:
Closed as per #1275

PR #1275, opened 2019-12-11
Attempts to solve task #1111

LiquidityC:
This will copy properties if none have been specified on the "copying action". Eg. you can override but you don't need to specify if you don't want anything to change.

Get back to me if I missed something vital. Hard to know all the quirks before hand.


tobbi:
Since this is mostly a feature, and we're in a feature-freeze, I'm gonna mark this for post-0.6.1

LiquidityC:
Sounds good

PR #1294, opened 2019-12-23
Fix lag in Tower of Ghosts

### LINKED ISSUE #1284: Ghost Tower level is unplayably slow
**SuperTux version:** 0.6.1
**System information:**

GNU/Linux
OpenGL version string: 3.0 Mesa 19.2.6
OpenGL core profile version string: 4.5 (Core Profile) Mesa 19.2.6
Graphics card: Haswell-ULT Integrated Graphics Controller

##### Expected behavior
Ghost Tower level is playable.

##### Actual behavior
It's unplayable. It's very very slow (less than 5 FPS).

1. Start Ghost Tower level in Story Mode

To me it happens with both renderers (both `sdl` and `opengl`).

##### Related issue

I already reported this here in a comment:

https://github.com/SuperTux/supertux/issues/1051

But I figured this might have been overlooked, so I post this again, but this time, as a separate issue. Also, it's not clear to me whether these two issues are actually the same or not.

Ordoviz:
Reduce sector size by 100 to get rid of unnecessary lava tiles. Closes #1284


serano01:
Well, it can be that easy. Thanks for your PR.

HybridDog:
I think the level should not lag even if there are redundant lava tiles.

PR #1295, opened 2019-12-25
Use preprocessor directives to disable unused code

HybridDog:
I am not sure if there is some reason why unused C++ code was not disabled with preprocessor directives, and how the compiler flag change affects compilation with various compilers, so I opened this as draft PR.


tobbi:
@HybridDog Does draft PR mean you're gonna add to it or can it be merged?

HybridDog:
> @HybridDog Does draft PR mean you're gonna add to it or can it be merged?

I have marked it as draft because I were not sure if it works correctly at that time.
I noticed another `if ((false))` and changed it now. The travis etc. builds have compiled (it was green before I force-pushed), so it should work correctly and can be merged.

PR #1296, opened 2019-12-25
Do not generate a text surface to measure text width

HybridDog:
The changes were introduced in commit a7969f.
Now only cached surfaces' widths are used.

This makes e.g. loading the Options menu the first time significantly faster.

Note that I am not very familiar with the text rendering, so I do not know if this has bad side effects.


tobbi:
You're completely correct. I tried out your commit and it sped things up significantly for me.

Thanks! :+1:

PR #1298, opened 2019-12-25
Make powerups at infoboxes not progressive

Ordoviz:
To address [feedback from tempAnon093 on the forum](https://forum.freegamedev.net/viewtopic.php?f=67&t=12828):

> During my first play of Welcome to Antarctica (tutorial), I fell
> into a spike pit before reaching the Fire Flower info box and as a
> result, it gave me an egg. This was slightly confusing

Before:

After:


weluvgoatz:
I suggest changing the block below the powerup to an empty bonus block, to give off the effect that that's where it came from.

tobbi:
We could as well change this code-wise. And have a special flag `force-contents` or something to indicate that it'll always dispense a fire flower.

tobbi:
So the issue I'm having is the following: We also want players to understand how bonus blocks normally work, if we had the bonus block always dispense a fire flower, then that would be an inconsistency. I don't know whether this is what we want. Maybe we should have the player retain their bonus after they die (and explain that this is a one time thing)?

Ordoviz:
Having a bonus block with `force-contents` is useful not only for the three scenarios in the pull request. Why not use a different color for these types of bonus blocks similar to the blocks that contain 5 coins? And the infoblock in *Welcome to Antarctica* indicates that the bonus blocks are “Super Bonus Blocks”.

Rusty-Box:
@Ordoviz This would be doable as we still have an orange variant of the bonus block that has not a purpose yet. We could use this one.

tobbi:
Different idea: How about we spawn a special character that gifts Tux his previous powerup when he dies?

busybody42:
good idea! (except the previous powerup gift...) But we need a unique block sprite... what about a star?

Guih48:
I think fixed power-up ejecting bonus-blocks would really improve functionality, and it isn't necressary to be different than an other bonus-block (we can call bonus-blocks ro mystery/lucky blocks too so is a suprise what they have inside). _(but for "smart" bonus block, that can eject both egg and advanced powerup we need a different in-editor sprite)_ (the previous-powerup-giving-back-bonus(-block?) is a good idea too)

Semphriss:
Closing due to lack of activity; feel free to re-open if further discussion happens.

Issue #1285, opened 2019-12-20
IceBlocks behave buggy when the player presses down


Ordoviz:
**SuperTux version:** 0.6.1
**System information:** Arch Linux 5.4.5-arch1-1

##### Expected behavior
Tux throws IceBlocks in the direction he is facing.

##### Actual behavior
When pressing down, Tux throws IceBlocks to the right. If he was facing left, he will take damage. If the IceBlock slides into a wall, it will stick to it.

HybridDog:
It also happens to me.

weluvgoatz:
This has been fixed.

PR #1303, opened 2019-12-28
Added downward throw for icecube

Zwatotem:
Releasing icecube while holding `down` key puts it on ground, or throws down, if released in the air. This solves #1285. Please check and review.


weluvgoatz:
I have tested this PR, and it works as intended.  I approve. 👍

Issue #1069, opened 2019-02-25
Jump on Snail does not always work


HybridDog:
If I jump on a Snail which moves down a slope, it sometimes stays unharmed.
![mpv-shot0001](https://user-images.githubusercontent.com/3192173/53341859-f7207100-390c-11e9-94a4-f8c536af07ff.jpg)
Here's a video where it happens: [v.webm.zip](https://github.com/SuperTux/supertux/files/2900808/out3.webm.zip)

tobbi:
Fixed by #1304

PR #1304, opened 2019-12-28
Snails can always be stomped

weluvgoatz:
This simply removes the flag that the snail must be on the ground to be stompable.  This seems to address #1069, but as it could cause unexpected behavior, I recommend someone else to test it as well.

### DIFF
--- a/src/badguy/snail.cpp
+++ b/src/badguy/snail.cpp
@@ -247,10 +247,6 @@ Snail::collision_squished(GameObject& object)
     case STATE_KICKED:
     case STATE_NORMAL:
 
-      // Can't stomp in midair
-      if (!on_ground())
-        break;
-
       squishcount++;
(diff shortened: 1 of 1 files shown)


HybridDog:
STATE_NORMAL probably happens when the snake is walking, and when it does not touch the ground while walking, Tux should still be able to hurt it in my opinion, so I'm in favour of this change, but I haven't tested it.

The `on_ground()` function probably works wrong because it sometimes returns false when the snail is walking on a slope (otherwise removing that code would not fix the problem); a slope is also a ground.

tobbi:
You see, looking at the code I suspect there might be some oddities going on when / if Tux should ever stomp that snail mid-fall. Not sure if that's possible gameplay-wise, though.

Zwatotem:
It is possible, and used in speedrunning.

PR #1307, opened 2019-12-29
Repurpose the Skull Tile

weluvgoatz:
This PR brings back the skull tile, which is a block that behaves like the Donut Block from Super Mario.
It also allows for it to be used in the editor, and it has gotten a new sprite (made by me).  Those unrelated changes in the diff are due to my repository being outdated.


weluvgoatz:
I am going to merge this, since graphics can always be added later.

PR #1308, opened 2019-12-31
fix FPS offset

plonibarploni:
see https://github.com/SuperTux/supertux/pull/1022#issuecomment-569969517


HybridDog:
FYI: The FPS text looks bad in the editor: https://github.com/SuperTux/supertux/pull/1099#issuecomment-495921142

PR #1312, opened 2020-01-02
Fix detection of system physfs

reinerh:
check_symbol_exists checks if compilation succeeds with the specified
symbol and header, but also fails if linking is not successful.
To detect the system library correctly, ask it to link against
${PHYSFS_LIBRARY} (which is set by find_package).


tobbi:
This breaks our Windows CI builds: 
https://ci.appveyor.com/project/supertux/supertux-o8t59/builds/29864597
https://ci.appveyor.com/project/supertux/supertux-9ml4d/builds/29864598

I'm assuming the check is added too early?

reinerh:
Sorry, I didn't check the case when no physfs is installed and found.
I'm now doing the symbol check only if a system physfs has been found.

Issue #972, opened 2018-12-04
Make climbing easier


Alzter:
**SuperTux version: 0.6.0 Alpha 3**
**System information:** *Windows 10*

I believe that climbing physics are really hard to use at the moment, so I have some suggestions for improvements.

1. Don't make Tux fall off the ladder or climbable object when reaching the edge of the climbable area, but instead just have the edge as a boundary. I keep slipping off climbables like this and it's a problem because in order to jump at the maximum height off a climbable surface, you have to carefully navigate to the top of it but don't go too high as you'd fall off, and then jump. It's inconvenient.

2. Remove action button letting go of the ladder. I constantly hold down action all the time because I am so used to run buttons, and I feel that having it take you off the ladder or climbable object is unneccesary when you can just jump off it with jump anyway.

3. Increase jump height and momentum upon jumping off a climbable. At the moment when Tux jumps off a climbable it's hard just to get up to platforms above it because if you keep holding up you fall off with no real upward velocity and jumping doesn't get you to the platform unless you're at the very top of the ladder. I suggest as well as removing falling off climbables (see point 1 & 2), and make Tux gain much more height when jumping off a climbable. Of course Tux should only get the maximum amount of height from holding jump, just tapping it should only send him a short distance up. And finally please increase Tux's horizontal movement speed upon jumping off a ladder so he can navigate easier.

(TL;DR: Make it identical to NSMB)

HybridDog:
1.
I agree. If you hold Up, Tux climbs to the top and then suddenly jumps and quickly falls down until you release and press Up again.

2.
When there is an enemy above you, you may want to fall down and not jump into it. You could make action do nothing but make Tux fall without jumping when pressing the jump button while holding Down.

3.
I disagree. Jumping from a ladder is unrealistic and a ladder should lean against something. If you could jump high from it, in some levels it may be e.g. placed so that it does not lean against a tree branch but ends somewhere under it. In my opinion, jumping low from a ladder is a neat feature.

Alzter:
> Jumping from a ladder is unrealistic and a ladder should lean against something. If you could jump high from it, in some levels it may be e.g. placed so that it does not lean against a tree branch but ends somewhere under it. In my opinion, jumping low from a ladder is a neat feature.

[...]

Issue #684, opened 2017-01-27
Improve climbing feature


Rusty-Box:
The way of how the current climbing feature works, is pretty much useless for levels, since you can't really use it for much stuff, instead of going up and down.
Sure, that's what climbing is, however its current state makes it kind of hard to combine it with proper level design. So that's why I think, that it needs improvements.

Tux should be able to jump further, so that the climbing feature could be used for cliff jumping from robe to robe, for example and should (maybe) move a little bit faster while climbing in general.
Also, it might just happens to me, but Tux always directly switch to his climbing state.

serano01:
At least the sprites should be completely redone, I think.

Alzter:
Definitely. You don't gain any height when letting go of ladders, and the action button unnecessarily makes you let go. And the sprites are crap. I may redo them.

tobbi:
@Rusty-Box Is this fixed now that #1313 is merged?

Rusty-Box:
Yes, having a normal jump height from ladders makes them much more useable.
Closing this.

PR #1313, opened 2020-01-02
Make jumps from ladder not tiny

Zwatotem:
Let me join the party :)
This addresses #972, but in a sublte way, since I don't want to introduce any bigger changes to climbing for now.


tobbi:
xref #684

PR #1319, opened 2020-01-19
Trampolines from bonus block no longer hurt Tux

Zwatotem:
Fix to accidently reverted bug.
No need to close any issue, since everything is closed anyway.


Zwatotem:
Can you merge, cause this is supposed to be quick fix and it's on my `master`.

PR #1323, opened 2020-01-21
New Snow Castle tileset

Alzter:
I made a new "snow castle" tileset, I think it's pretty cool and could be used to spruce up levels like Nolok's Castle.

SnowCastle

I've had a concept for Nolok's Castle for a while now where it could be talking place in a decrepit castle during an icy blizzard, and this tileset makes that idea closer to a reality. What do you guys think?

image


HybridDog:
Your file is 192.8 KiB, if I recompress it (export in GIMP), it takes only 81.7 KiB without quality loss.

Alzter:
image
Here's a usage example. You can make the bricks as frequent or infrequent as you like, here I made them spaced out a bit.

serano01:
Like in your concept, you could add bricks that are continuous and gap less, otherwise great work. Can't wait to use it.

PR #1324, opened 2020-01-21
New arctic background

Alzter:
I made a new arctic mountains background, it's nice and tall and wraps horizontally, so it'd be great for large scale ice levels or just for variety.

Preview:
arcticmountainssmall


Rusty-Box:
Looks good. However, I'm unsure whether it is to sharp or not. The current one arctis2 is more blurry to convey distance.

HybridDog:
Your image is too big: 11.7 MiB.
If you compress it with a higher PNG compression ratio (e.g. export it in GIMP), it takes only 517.0 KiB and is still lossless.
A lossy jpg file takes 101.3 KiB; only if you look close, you can see artifacts:
arcticmountains

> However, I'm unsure whether it is to sharp or not.

If you want a blurry background, you can simply save it in lower resolution; in the game it is scaled with (probably gamma-incorrect) linear interpolation anyways.
I've downscaled the image with a [sharp downscaling algorithm](https://gist.github.com/HybridDog/dd95a99d411972f030fb18543280ad60).
half width and height (243.2 KiB):
smaller
quarter width and height (68.4 KiB):
more_smaller

Alzter:
Compressed it with GIMP's compression level set to 9, saving 11.2 MB of space. I originally exported it compressed but when I expanded the image I forgot to re-set the compression settings, now it should be good for merge.

Alzter:
Also made it a bit more blurry using GIMP's gaussian blur tool.

Alzter:
Very cool, thanks.

PR #1327, opened 2020-01-24
Updated Badguys (WIP)

Alzter:
8 frame walk cycles for Snowball and Iceblock as well as stronger outlines to make them more distinguishable.

Previews:
SnowballwalkWIP3 Iceblockwalk2

TO-DO: Enemy variants


Nickeros:
The animation did alzter i like it drawing

busybody42:
At least just keep the old sprites in some legacy folder, in case somebody prefers the old sprites in a custom level...

busybody42:
and maybe you could add BOTH of the new spiky walk cycles, and maybe even alternate between using the 2 animations in game?

Alzter:
> and maybe you could add BOTH of the new spiky walk cycles, and maybe even alternate between using the 2 animations in game?

There's only one new spiky walk cycle?

busybody42:
I thought it was BETTER than the 2nd one...

serano01:
@Alzter: Would it be okay for you when I merge this first batch of reworked badguys to master?

Rusty-Box:
> @Alzter: Would it be okay for you when I merge this first batch of reworked badguys to master?

@serano01 The spiky animation has too much of a turn in its movement which makes it look really weird. I would like to have this addressed before merging this.

busybody42:
@Rusty-Box, that's exactly what I was thinking!

Alzter:
Wow, it's merged? Awesome!

Enemies I think need changing some time in the future:
Jumpy: The metal texture has too much contrast and doesn't look right
Spiky: Same problem as jumpy + his walk looks off
Flying snowball: Same problem as jumpy, he looks too squashed also for lack of a better term

Other enemies to redo in future: Crystallo, Walking Leaf, etc

Rusty-Box:
@Alzter I personally think in their sleeping sprite spiky's eyes should not be visible. Maybe it is just me but the way it looks here looks really weird

serano01:
Note that the old spiky-sprites are still in use, as not everyone was satisfied with them. But for the other badguys it was about time, I thought.

Issue #1262, opened 2019-12-01
Enemies disappear if Tux is very high above level bounds


Wuzzy2:
**SuperTux version:** 0.6.1 beta 1
**System information:** GNU/Linux

##### Expected behavior
If Tux is very high, enemies do not disappear.

##### Actual behavior
If you manage to get Tux very very high above the upper screen bound (outside of level), the enemies in your screen will disappear.

##### Steps to reproduce actual behavior
Use a trampoline with Air Tux. The level height should be rather low so that Tux flies very high.

You can use the cave entrance level on Icy Island from Story Mode. Start the level with Air Tux and jump on a trampoline roughly in the middle of the level. Watch what happens to the enemies.

Wuzzy2:
I should clarify: The enemies don't actually disappear, they just become invisible while Tux is very high above the level. If Tux falls down again, the enemies become visible again.

Rusty-Box:
I can confirm this issue. I assume this has to do with the fact that Air Tux jumps so high that the enemies are not inside the loading range for objects in relation to Tux's distance.

tobbi:
Fixed by #1329

PR #1329, opened 2020-01-25
Fixed offscreen definition for badguys.

Zwatotem:
This solves #1262 badguys disappearing when Tux is very high above the
screen.


tobbi:
Thanks! Can you add `Closes #1329` or `Fixes #1329`to the commit message so it auto-closes the issue? Thanks.

Issue #1289, opened 2019-12-22
Save “Show controller” option in config file


Ordoviz:
Currently, if you activate this option in the developer mode, it gets reset when you close the game. Besides, you shouldn’t need to activate developer mode to use this feature as it’s very useful for speedrunners. Therefore, I suggest making this option settable in the options menu or – if you don’t like that – make it at least settable in the config file.

tobbi:
I'll leave this open as a good first issue. If anyone wants to work on this, please ping @tobbi and I'll get you up to speed!

shawnanastasio:
Hi @tobbi, I've submitted a PR that implements this. Comments would be appreciated!

tobbi:
Fixed! Thanks!

PR #1339, opened 2020-02-08
Save debug menu's "Show Controller" option in config

shawnanastasio:
This commit moves the show_controller variable from the
Debug class to the Config class and adds a call to g_config->save()
in DebugMenu's menu_action member function so that changes are
written to the config file.

This addresses issue #1289.


tobbi:
I think this looks good to merge. Haven't tested it, but looking at the code, all seems fine.

PR #1345, opened 2020-02-16
Fixed problem with narrow levels

### LINKED ISSUE #1270: Graphics glitch for narrow levels
**SuperTux version:** 0.6.0 beta 2

##### Expected behavior
If you play a narrow level (a level that has a low overall width), the left and right level bounds are black or rendered in some other way that indicates that its out of level bounds.

##### Actual behavior
There's a weird graphical glitch at the left and right level bounds.

Evidence: Level “Flower Bonus” from Bonus Island III:

https://satoshiupload.com/images/1TScduq6Zw.png

Play with a wide screen resolution a level with a low level width.
I recommend playing a couple of levels in Bonus Island III, which has a couple of narrow levels where this can be seen.

Affected Bonus Island III levels:

* The Toilette Zone
* Flower Bonus
* Deep, deeper …
* … the deepest
* (possibly other levels)

Zwatotem:
Levels narrower than display used to be displayed on whole screen. 
Now The part out of level limits is a black stripe.
Fixes #1270

### DIFF
--- a/src/object/vertical_stripes.cpp
+++ b/src/object/vertical_stripes.cpp
@@ -1,0 +1,69 @@
+//  SuperTux
+//  Copyright (C) 2020 Grzegorz Przybylski <zwatotem@gmail.com>
+//
+//  This program is free software: you can redistribute it and/or modify
+//  it under the terms of the GNU General Public License as published by
+//  the Free Software Foundation, either version 3 of the License, or
+//  (at your option) any later version.
+//
+//  This program is distributed in the hope that it will be useful,
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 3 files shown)


busybody42:
No!
Let's just zoom in instead!
Zoom in so that the part out of bounds isn't visible, because the camera zooms in so we CAN'T see it!

src/object/vertical_stripes.cpp
```diff
@@ -0,0 +1,69 @@
...
+{
+}
+
+VerticalStripes::~VerticalStripes()
+{
+}
+
+void
+VerticalStripes::update(float dt_sec)
+{
```
tobbi: ::update() gets run once every frame. I'm wondering whether this should be placed somewhere more appropriate.

PR #1349, opened 2020-02-21
Fix Haiku build

Crestwave:
`PACKAGE_NAME` isn't defined for some reason on Haiku, so `olduserdir` resolves to `~/.`. Yikes! I just disabled the entire thing because I don't think the old config directory was ever on Haiku, anyway. I've only tested this on Haiku, as it's a pretty minor change.


tobbi:
Thanks! I'm taking your word for it that the old dir was never used. Don't have a device running Haiku, so I can't check.

PR #1350, opened 2020-02-21
Fix the secret in Semi-Frozen

brmbrmcar:
Currently, the second secret is blocked by a coin block that is unbreakable, and as such the secret is inaccessible. This increases the height of the secret slightly so it is now accessible without leaving an obvious hole such as in previous versions.


tobbi:
@Rusty-Box approve?

PR #1354, opened 2020-03-01
Rocks can activate buttons

weluvgoatz:
This PR allows rocks to be thrown onto pushbuttons in order to activate them.  While it seems this is an approved feature, I would like testing for it.

### DIFF
--- a/src/object/pushbutton.cpp
+++ b/src/object/pushbutton.cpp
@@ -65,7 +65,6 @@ PushButton::collision(GameObject& other, const CollisionHit& hit)
   auto player = dynamic_cast<Player*>(&other);
   auto rock = dynamic_cast<Rock*>(&other);
   if (!player && !rock) return FORCE_MOVE;
-  if (player || rock) {
 	  if (player) {
   float vy = player->get_physic().get_velocity_y();
   if (vy <= 0) return FORCE_MOVE;
@@ -90,7 +89,6 @@ PushButton::collision(GameObject& other, const CollisionHit& hit)
   Sector::get().run_script(script, "PushButton");
 
   return FORCE_MOVE;
-  }
 }
 
 /* EOF */
--- a/src/object/pushbutton.cpp
+++ b/src/object/pushbutton.cpp
@@ -18,6 +18,7 @@
 
 #include "audio/sound_manager.hpp"
 #include "object/player.hpp"
+#include "object/rock.hpp"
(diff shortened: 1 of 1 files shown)


src/object/pushbutton.cpp
```diff
@@ -62,15 +63,18 @@ HitResponse
 PushButton::collision(GameObject& other, const CollisionHit& hit)
 {
   auto player = dynamic_cast<Player*>(&other);
-  if (!player) return FORCE_MOVE;
+  auto rock = dynamic_cast<Rock*>(&other);
+  if (!player && !rock) return FORCE_MOVE;
+  if (player || rock) {
```
Zwatotem: I don't think you need this `if` clause, since this is the only option left. You can replace it with else or delete at all.

src/object/pushbutton.cpp
```diff
@@ -62,15 +63,17 @@ HitResponse
 PushButton::collision(GameObject& other, const CollisionHit& hit)
 {
   auto player = dynamic_cast<Player*>(&other);
-  if (!player) return FORCE_MOVE;
+  auto rock = dynamic_cast<Rock*>(&other);
+  if (!player && !rock) return FORCE_MOVE;
```
HybridDog: Please add a line break. (`return` jumps out of the function execution, so it should always stick out.)
```suggestion
  if (!player && !rock) 
    return FORCE_MOVE;
```

src/object/pushbutton.cpp
```diff
@@ -62,15 +63,17 @@ HitResponse
...
+	  if (player) {
   float vy = player->get_physic().get_velocity_y();
-
+  if (vy <= 0) return FORCE_MOVE;
   //player->add_velocity(Vector(0, -150));
   player->get_physic().set_velocity_y(-150);
```
HybridDog: Indentation and the line break are missing.

src/object/pushbutton.cpp
```diff
@@ -62,15 +63,17 @@ HitResponse
...
+	  }
 
   if (state != OFF) return FORCE_MOVE;
   if (!hit.top) return FORCE_MOVE;
```
HybridDog:
```suggestion
  if (state != OFF || !hit.top)
    return FORCE_MOVE;
```

weluvgoatz:
Closing, due to lack of activity and lack of interest.

PR #1357, opened 2020-03-09
Allow setting the layer (z-pos) of candles, torches & spotlights

### LINKED ISSUE #1011: Allow us to edit the layer/z-pos of candles, torches and spotlights
Was making a level where I needed torches in the foreground and realised this was impossible. Feel like it'd be pretty useful

martinvanzijl:
Fixes #1011. You can set the layer of candles, torches and spotlight in the level editor now. Demo below (zipped .ogv, since the .gif was too large):

[set-torch-layer-demo.zip](https://github.com/SuperTux/supertux/files/4308191/set-torch-layer-demo.zip)

Here the torch initially has layer 0, so it is behind the foreground tiles (which have layer 100). Once I change the layer of the torch to 150, it is drawn in front of the foreground tiles. The same can be done for candles and spotlights.

I made the default value for layer in these objects 0, to keep in line with existing behaviour.

Please check also my code for the torch - certain parts were drawn on `LAYER_TILES - 1`, which I changed to `m_layer - 1`. `LAYER_TILES` is defined to 0 in video/layer.hpp, so with the default value of 0 for `m_layer`, existing behaviour should be unaffected.

### DIFF
--- a/src/object/torch.cpp
+++ b/src/object/torch.cpp
@@ -37,6 +37,7 @@ Torch::Torch(const ReaderMapping& reader) :
 
   reader.get("sprite", sprite_name);
   reader.get("burning", m_burning, true);
+  reader.get("layer", m_layer, 0);
 
   m_torch = SpriteManager::current()->create(sprite_name);
   m_col.m_bbox.set_size(static_cast<float>(m_torch->get_width()),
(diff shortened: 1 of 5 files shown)


src/object/torch.cpp
```diff
@@ -51,16 +52,16 @@ Torch::draw(DrawingContext& context)
 {
   if (m_burning)
   {
-    m_flame->draw(context.color(), get_pos(), LAYER_TILES - 1);
+    m_flame->draw(context.color(), get_pos(), m_layer - 1);
```
Zwatotem: I'm pretty sure `- 1` should be removed here, since with it we're placing torches 1 layer lower, than user means to.

tobbi:
Please initialize `m_layer` in the initialiation list for all the classes:
```
/home/travis/build/SuperTux/supertux/src/object/spotlight.cpp: In constructor ‘Spotlight::Spotlight(const ReaderMapping&)’:

/home/travis/build/SuperTux/supertux/src/object/spotlight.cpp:23:1: error: ‘Spotlight::m_layer’ should be initialized in the member initialization list [-Werror=effc++]

 Spotlight::Spotlight(const ReaderMapping& mapping) :

 ^~~~~~~~~
```

martinvanzijl:
@tobbi Thanks, have corrected that now. The warning didn't seem to show up on my compiler (gcc version 7.4.0).

tobbi:
Thank you! 👍

PR #1360, opened 2020-03-26
Enable "squished" mirror action

serano01:
Allow the bouncing snowball to mirror the squished-left animation, since at the moment, it is only able to display the squished-left animation, which produces a graphical error.


serano01:
Fatbat works as intended, I'll look why this isn't the case for the other two...

serano01:
I'll close this, since in this form the PR no longer makes sense and needs a major overhaul.

PR #1362, opened 2020-03-28
Cache players in GameObjectManager

icecream95:
`get_nearest_player` was spending a lot of time searching for Player objects, so this PR adds a list for players, similar to the solid_tilemaps list.

For me, this increases performance on "A Fork in the Road" from 40fps to 60fps.

### DIFF
--- a/src/supertux/game_object_manager.cpp
+++ b/src/supertux/game_object_manager.cpp
@@ -177,10 +177,9 @@ GameObjectManager::flush_game_objects()
 
   { // update solid_tilemaps list
     m_solid_tilemaps.clear();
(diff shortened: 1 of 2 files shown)


Grumbel:
`get_players()` shouldn't be necessary, you can already do the same thing with this:

     GameObjectManager::get_objects_by_type_index(std::type_index(typeid(Player)))

Might still need some more syntactic sugar however.

Note that this is distinct from `get_objects_by_type()` as it only works for the last class in the hierarchy, unlike dynamic_cast. But Player is `final` so it should be fine.

tobbi:
Very, very nice! Also, I didn't know this existed in our codebase!

Issue #1251, opened 2019-11-28
Old levelpacks are marked as new


Wuzzy2:
**SuperTux version:** 0.6.1 beta 1
**System information:** GNU/Linux

##### Expected behavior
When I open a custom levelpack and return, the level count is not destroyed.

##### Actual behavior
When I open a custom levelpack on which I have already completed a few levels, and then I return to the main menu, and go back to the list of levelpacks, the levelpack in question is now suddenly marked as “\*NEW\*”.

This bug does ont occur in 0.6.0.

##### Steps to reproduce actual behavior
* Download and install custom levelpack
* Play a few levels
* Return

No matter what you do, the levelpack will be marked as new after you return.

weluvgoatz:
I also experience this issue.

busybody42:
On that topic, it ALSO doesn't let me use most fourm addons, no matter WHAT i do, with some exceptions... also, they would STILL show up on the addons part of the main menu, but not in the contrib levels tab itself... and on that topic...

busybody42:
...It ALSO happened in 0.6.0 when I tried to get the old version of serano's island working... remastered doesn't show up, because it wants to get downloaded again, and instead, there's the original... but it listed the original twice for some reason... and when I download remastered serano's islands, it replaced the old one! 
BUT, when I reopen the game, it's back to "download remastered serano's islands, or just play the old one instead." AGAIN!

Narre:
This issue first appeared in d119a4db (26 Dec 2018), it is in  src/worldmap/worldmap_parser.cpp

Ordoviz:
ad80130ab26cbcde36380c217a541841f892b191 temporarily hides the NEW text.

Guih48:
I found something super!

If you unpack the archives "addons" folder and move them to levels folder the bug is no more exist (yes, if you only unpack some worlds, the bug is only mend at unpacked levels) :)

Vankata453:
The `*NEW*` levelset indicator has long since been removed. Additionally, the progress functionality will be restored in #2349.

PR #1364, opened 2020-03-30
WIP: Maybe fix physfs path

tobbi:
Okay, so obviously this does not work...Or maybe not so obviously...

I tried adding the physfs paths of add-ons to a list...
And I'm basically, then mounting them when the list of contrib levels get loaded...So far so good...

The only problem is that the loading of the completed levels from the physfs shares does not seem to be working...as all the contrib levels are marked as new...So, squirrel does not load the state table properly. Why? The hell do I know...

If someone who knows about these things could take a look, that would be much appreciated.

### DIFF
--- a/src/addon/addon_manager.cpp
+++ b/src/addon/addon_manager.cpp
@@ -140,7 +140,8 @@ AddonManager::AddonManager(const std::string& addon_directory,
   m_installed_addons(),
   m_repository_addons(),
   m_has_been_updated(false),
-  m_transfer_status()
+  m_transfer_status(),
+  m_enabled_addon_paths()
 {
   if (!PHYSFS_mkdir(m_addon_directory.c_str()))
   {
@@ -433,6 +434,7 @@ AddonManager::enable_addon(const AddonId& addon_id)
   else
   {
     log_debug << "Adding archive \"" << addon.get_install_filename() << "\" to search path" << std::endl;
+    m_enabled_addon_paths[addon_id] = addon.get_install_filename();
     //int PHYSFS_mount(addon.installed_install_filename.c_str(), "addons/", 0)
 
     // std::string mountpoint;
@@ -472,6 +474,7 @@ AddonManager::disable_addon(const AddonId& addon_id)
   }
   else
   {
+    m_enabled_addon_paths[addon_id] = nullptr;
     // log_debug << "Removing archive \"" << addon.get_install_filename() << "\" from search path" << std::endl;
     // if (PHYSFS_unmount(addon.get_install_filename().c_str()) == 0)
     // {
--- a/src/addon/addon_manager.cpp
+++ b/src/addon/addon_manager.cpp
@@ -474,7 +474,7 @@ AddonManager::disable_addon(const AddonId& addon_id)
   }
   else
   {
-    m_enabled_addon_paths[addon_id] = nullptr;
+    m_enabled_addon_paths.erase(addon_id);
     // log_debug << "Removing archive \"" << addon.get_install_filename() << "\" from search path" << std::endl;
     // if (PHYSFS_unmount(addon.get_install_filename().c_str()) == 0)
     // {
--- a/src/addon/addon_manager.cpp
+++ b/src/addon/addon_manager.cpp
@@ -435,29 +435,29 @@ AddonManager::enable_addon(const AddonId& addon_id)
     log_debug << "Adding archive \"" << addon.get_install_filename() << "\" to search path" << std::endl;
     //int PHYSFS_mount(addon.installed_install_filename.c_str(), "addons/", 0)
 
-    std::string mountpoint;
-    switch (addon.get_format()) {
-      case Addon::ORIGINAL:
-        mountpoint = "";
-        break;
-      default:
-        mountpoint = "custom/" + addon_id;
-        break;
-    }
+    // std::string mountpoint;
+    // switch (addon.get_format()) {
+    //   case Addon::ORIGINAL:
+    //     mountpoint = "";
+    //     break;
+    //   default:
+    //     mountpoint = "custom/" + addon_id;
+    //     break;
+    // }
 
-    if (PHYSFS_mount(addon.get_install_filename().c_str(), mountpoint.c_str(), 0) == 0)
-    {
-      log_warning << "Could not add " << addon.get_install_filename() << " to search path: "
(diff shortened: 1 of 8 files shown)


tobbi:
@Grumbel Sorry for pinging you, but I might need your help.

I'm trying to defer mounting add-on zip paths until they're actually needed...e.g. when the Contrib levels menu is loaded or the worlds specified in that add-on are loaded for sure. 

I thought that the approach I chose here (add paths to load to a map, then load them on demand) but for some reason this messes with how Squirrel loads the level state...And I have no idea why. Could you please take a look?

tobbi:
I've investigated some more: For some reason, Squirrel can't find the "levels" table entry, despite it being created a couple calls earlier.

plonibarploni:
> The only problem is that the loading of the completed levels from the physfs shares does not seem to be working...as all the contrib levels are marked as new...So, squirrel does not load the state table properly. Why? The hell do I know...

Is this possibly related to #1251?

> I've investigated some more: For some reason, Squirrel can't find the "levels" table entry, despite it being created a couple calls earlier.

That would probably explain why this code is never executed, causing the levels be marked as new:
https://github.com/SuperTux/supertux/blob/a0cd2f4e6512ca12b0ab813fbc827ba62e1cbad3/src/supertux/menu/contrib_menu.cpp#L134-L144

Is it possible that the state is now saved differently? Older versions can't read the level state from newer versions either (I get inconsistent results, sometimes it says "new" and sometimes "0/22").

tobbi:
> > The only problem is that the loading of the completed levels from the physfs shares does not seem to be working...as all the contrib levels are marked as new...So, squirrel does not load the state table properly. Why? The hell do I know...
> 
> Is this possibly related to #1251?

I don't think so. This is an issue that got introduced through my changes...

> > I've investigated some more: For some reason, Squirrel can't find the "levels" table entry, despite it being created a couple calls earlier.
> 
> That would probably explain why this code is never executed, causing the levels be marked as new:
> 
> https://github.com/SuperTux/supertux/blob/a0cd2f4e6512ca12b0ab813fbc827ba62e1cbad3/src/supertux/menu/contrib_menu.cpp#L134-L144
> 
> Is it possible that the state is now saved differently? Older versions can't read the level state from newer versions either (I get inconsistent results, sometimes it says "new" and sometimes "0/22").

If only I knew...

tobbi:
@martinvanzijl @Zwatotem I know this is a long shot, and I also think you might not know enough about physfs to help with this (I think I already mentioned this to you, @Zwatotem). If anyone of you can find out what's wrong with this, I'd be very grateful. This is *the* issue preventing us from getting a new release.

martinvanzijl:
@tobbi I'm not familiar with physfs but I could take a look. What are the steps to reproduce the issue?

tobbi:
Install Slackstone's add-on from here:
https://raw.githubusercontent.com/SuperTux/addons/master/repository/forest_world_adventure_pack.zip

(copy it into the add-ons directory)...see how the assets from the add-on override the assets from the game. Due to physfs overriding.

tobbi:
Some further comments: If you take a look at https://github.com/SuperTux/supertux/pull/1364/files#diff-bfcd1496905218f4e56f865be1828669R358 I'm outputting all the table entries.

It produces the following output:
```
[WARNING] supertux/src/supertux/savegame.cpp:359 /levels/revenge_in_redmond/worldmap.stwm
[WARNING] supertux/src/supertux/savegame.cpp:373 Squirrel error: failed to get 'levels/revenge_in_redmond/worldmap.stwm' table entry (the index 'levels/revenge_in_redmond/worldmap.stwm' does not exist)
```

Obviously, you might be thinking "Ah, that's easy, simply add a forward slash to the search path in savegame.cpp, line 363. I tried that, to no avail.

martinvanzijl:
Thanks. After installing the add-on, when I restart SuperTux the opening screen is different and there is no music. Also some sprites are different in the levels (such as bonus blocks). Are these the things being overridden?

tobbi:
> Thanks. After installing the add-on, when I restart SuperTux the opening screen is different and there is no music. Also some sprites are different in the levels (such as bonus blocks). Are these the things being overridden?

Exactly!

martinvanzijl:
Ah, I see. And with this patch applied, the main screen and levels are now normal again, but I cannot see any levels listed for the add-on.

tobbi:
The add-on paths are loaded when the add-ons menu is loaded. But for some reason, it won't work then...It's supposed to load the add-on zips to figure out how many levels there are. But it won't display the level count.

martinvanzijl:
Ah, I see. It should put the figures for levels completed/total in the menu here:

level-info-screenshot

But instead it just says "\*NEW\*". And this is because it cannot read the key which has this information, in `savegame.cpp` with this line:

```
vm.get_or_create_table_entry(name);
```

tobbi:
> Ah, I see. It should put the figures for levels completed/total in the menu here:
> 
> level-info-screenshot
> 
> But instead it just says "*NEW*". And this is because it cannot read the key which has this information, in `savegame.cpp` with this line:
> 
> ```
> vm.get_or_create_table_entry(name);
> ```

You got it! :)

martinvanzijl:
OK, I think I made some progress. If I comment out the lines which list the keys (358 and 359), it seems to read the key on line 361 OK. At least, it does not throw an exception. Like so:

```
    //for(const auto& key : vm.get_table_keys())
      //log_warning << "Existing key: " << key << std::endl;

    vm.get_or_create_table_entry(name);
```

It seems `SquirrelVM::get_table_keys()` does some stack manipulation, which may cause the key to be missed on line 361.

However, then it still cannot find the "levels" sub-key on line 365.

tobbi:
Okay, hold on, let me see if I can fix that.

tobbi:
@martinvanzijl Thank you, I've fixed it by restoring the old stack's top. Now it even displays the level info...Next up: Loading the physfs path when loading the levelset... (currently, levelsets are empty).

tobbi:
For loading the zip file when loading the world, I'm hitting some kind of roadblock: When a level has been edited, it gets saved outside the zip file...When I'm now querying for the level's basedir, the location outside that zip file gets returned. I'm not sure how to fix that best without constantly unmounting / mounting that extra directory.

tobbi:
@christ2go suggested a way easier method. Because of this, I'm closing it.

PR #1366, opened 2020-04-02
Improve text wrapping for languages with no space characters

### LINKED ISSUE #1028: Texts in InfoBlocks are not wrapped in some languages which have no spaces between words
**SuperTux version:** 0.6.0-58-g631d2bf3a
**System information:** Linux 4.15.0-43-generic, elementary OS 5.0 Juno (built on Ubuntu 18.04.1)

### Background knowledge

Generally, Japanese does not have any spaces between words, unlike English, French, etc. For example:

**English (has a space between each words):** Bonus blocks can contain coins, power-ups or special items to help you in your quest.
**Japanese (has no space between each words):** ボーナスブロックには、冒険を助けるコインやパワーアップアイテム、特別なアイテムが入っていることがあります。

Chinese also has no spaces between words like Japanese.

### Expected behavior

screenshot from 2019-01-05 18-42-53

Texts in InfoBlocks should be wrapped even in some languages which have no spaces between words, like in English.

### Actual behavior

screenshot from 2019-01-05 18-42-17

Texts in InfoBlocks are not wrapped in some languages which have no spaces between words.

### Steps to reproduce actual behavior

* Download latest Japanese & Chinese Traditional translations of World 1 from Transifex (because built-in translations may be old)
* Run SuperTux in Japanese or Chinese
* Play some level that has InfoBlocks (e.g. "Welcome to Antarctic" in World 1)
* Hit some InfoBlock. You should see texts in InfoBlocks are not wrapped

### Additional information

screenshot from 2019-01-05 20-34-21

I face this issue in Japanese, which is my mother tongue. However, I also face this issue when I play in Chinese Traditional (zh_TW).

screenshot from 2019-01-05 18-41-10

Also, I confirmed Japanese texts were also wrapped when I added a space between words on purpose.

Considering those 2 points, I suppose this issue may occur in any languages which have no spaces between words.

martinvanzijl:
Fixes #1028. Text from Info Boxes now hard-wraps to allowed width if there are no spaces between words. Tested with both Japanese and Traditional Chinese (I don't speak either language, but the wrapping seems OK.)

Example screenshot below using Japanese, before fix:

screenshot-before-fix

After fix:

screenshot-after-fix

Video demo also (zipped):

[wrap-demo-video.zip](https://github.com/SuperTux/supertux/files/4422141/wrap-demo-video.zip)

### DIFF
--- a/src/video/bitmap_font.cpp
+++ b/src/video/bitmap_font.cpp
@@ -328,7 +328,30 @@ BitmapFont::wrap_to_width(const std::string& s_, float width, std::string* overf
     }
   }
 
-  // FIXME: hard-wrap at width, taking care of multibyte characters
+  // hard-wrap at width, taking care of multibyte characters
+  unsigned int char_bytes = 1;
+  for (int i = 0; i < static_cast<int>(s.length()); i += char_bytes) {
+
+    // calculate the number of bytes in the character
+    char_bytes = 1;
+    auto iter = s.begin() + i + 1; // iter points to next byte
+    while ( iter != s.end() && (*iter & 128) && !(*iter & 64) ) {
+      // this is a "continuation" byte in the form 10xxxxxx
+      ++iter;
+      ++char_bytes;
+    }
+
+    // check whether text now goes over allowed width, and if so
+    // return everything up to the character and put the rest in the overflow
+    std::string s2 = s.substr(0,i+char_bytes);
+    if (get_text_width(s2) > width) {
+      if (i == 0) i += char_bytes; // edge case when even one char is too wide
+      if (overflow) *overflow = s.substr(i);
+      return s.substr(0, i);
+    }
+  }
+
+  // should in theory never reach here
   if (overflow) *overflow = "";
   return s;
 }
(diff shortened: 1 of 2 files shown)


martinvanzijl:
I only updated `TTFFont`. Should this be replicated to `BitmapFont`, too?

tobbi:
> I only updated `TTFFont`. Should this be replicated to `BitmapFont`, too?

Yes please. One can toggle TTF and bitmap fonts using some menu I don't remember the shortcut of.

martinvanzijl:
Thanks, I've done that now. The option is in the Debug Menu. I tested with Japanese and it seems to work fine. Screenshot before:

bitmap-before

After:

bitmap-after

tobbi:
Thank you very much! 👍

Zwatotem:
I'm not an expert, but as far as I know the 。sign is analog of period, and should stay at the end of lines.

martinvanzijl:
@Zwatotem That makes sense. However, the glyph for the 。sign contains some spacing on the right, so it actually goes over the allowed width of the Info Box, even though in the screenshots it seems like it should fit. So it is wrapped to a new line.

A future improvement could be to use the reduced width (without spacing on the right) for this glyph when checking if the text fits within allowed width, here in the code:

https://github.com/SuperTux/supertux/blob/7fb537f9eae6e464fc7ec5c19f355cdecfcc8eae/src/video/ttf_font.cpp#L171

PR #1373, opened 2020-04-12
Avoid scrolling main area of editor and toolbox at same time

### LINKED ISSUE #1358: Level editor: scrolling on the side bar scrolls the entire level map
**SuperTux version: 0.6.1**
**System information: Linux Mint 19.3, Linux Kernel 5.3.0-40**

##### Expected behavior

Scrolling on the side bar of the level editor should just scroll the implementable level elements and not the entire map

##### Actual behavior

Scrolling on the side bar of the level editor scrolls not only the implementable level elements but also the entire map at the same time

Scroll up/down on the side bar of the level editor

martinvanzijl:
Fixes #1358. If the mouse is over the toolbox, the mouse wheel will scroll the toolbox only, not the main area of the editor. Demo below:

scroll-fix-demo

### DIFF
--- a/src/editor/editor.cpp
+++ b/src/editor/editor.cpp
@@ -635,8 +635,9 @@ Editor::event(const SDL_Event& ev)
       }
     }
 
-    // scroll with mouse wheel
-    if (ev.type == SDL_MOUSEWHEEL) {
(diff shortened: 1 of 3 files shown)


tobbi:
The jobs failed due to exceeding time limitations (nothing serious). Thanks for this! 👍

PR #1379, opened 2020-04-16
Add option for wind objects to affect badguys

### LINKED ISSUE #982: Enemies should obey wind like Tux does
Right now, I have noticed that enemies do not work in wind like Tux does. I think this should be fixed, since Tux is a creature, too, and it doesn't make sense that they don't.  I know this is a small issue/suggestion, but you guys should maybe add it. :)

martinvanzijl:
Fixes #982. This adds an option to Wind objects: "Affects Badguys". This is false (off) by default. Checking this option will make the wind affect badguys in a similar manner to Tux.

Demo below (zipped OGV file). This shows some cannon shots and bouncing snowballs being affected by an upward wind:

[wind-badguys-full-demo.zip](https://github.com/SuperTux/supertux/files/4488824/wind-badguys-full-demo.zip)

### DIFF
--- a/src/badguy/badguy.cpp
+++ b/src/badguy/badguy.cpp
@@ -915,13 +915,13 @@ void
 BadGuy::add_wind_velocity(const Vector& velocity, const Vector& end_speed)
 {
   // only add velocity in the same direction as the wind
-  if (end_speed.x > 0  &&  m_physic.get_velocity_x() < end_speed.x)
+  if (end_speed.x > 0 && m_physic.get_velocity_x() < end_speed.x)
     m_physic.set_velocity_x(std::min(m_physic.get_velocity_x() + velocity.x, end_speed.x));
-  if (end_speed.x < 0  &&  m_physic.get_velocity_x() > end_speed.x)
+  if (end_speed.x < 0 && m_physic.get_velocity_x() > end_speed.x)
     m_physic.set_velocity_x(std::max(m_physic.get_velocity_x() + velocity.x, end_speed.x));
-  if (end_speed.y > 0  &&  m_physic.get_velocity_y() < end_speed.y)
+  if (end_speed.y > 0 && m_physic.get_velocity_y() < end_speed.y)
     m_physic.set_velocity_y(std::min(m_physic.get_velocity_y() + velocity.y, end_speed.y));
-  if (end_speed.y < 0  &&  m_physic.get_velocity_y() > end_speed.y)
+  if (end_speed.y < 0 && m_physic.get_velocity_y() > end_speed.y)
     m_physic.set_velocity_y(std::max(m_physic.get_velocity_y() + velocity.y, end_speed.y));
 }
 
--- a/src/badguy/badguy.cpp
+++ b/src/badguy/badguy.cpp
@@ -905,4 +905,24 @@ BadGuy::after_editor_set()
   }
 }
 
+bool
+BadGuy::can_be_affected_by_wind() const
+{
+  return !on_ground();
+}
+
+void
+BadGuy::add_wind_velocity(const Vector& velocity, const Vector& end_speed)
+{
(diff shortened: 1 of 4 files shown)


martinvanzijl:
One difference for badguys with respect to Tux: It seems the wind object will make Tux go at _exactly_ the speed of the wind, no faster. E.g. if the wind speed is 100 in x-direction, and Tux is going at 200 in x-direction, it will actually slow Tux down till he is going 100. Badguys will not be "slowed" like this.

busybody42:
how about make an option for wind to affect tux when he's on the ground

weluvgoatz:
How about not adding separate options for things, creating different
behavior among objects with no way to tell when it will and will not work a
certain way, which just creates confusion and makes the game unfair and
uncertain in a lot of ways.

src/object/wind.cpp
```diff
@@ -109,6 +114,13 @@ Wind::collision(GameObject& other, const CollisionHit& )
     }
   }
 
+  auto badguy = dynamic_cast<BadGuy*> (&other);
+  if (badguy  &&  this->affects_badguys) {
```
tobbi, line 118: There're too many spaces around the `&&`.

src/object/wind.cpp
```diff
@@ -64,8 +68,9 @@ Wind::get_settings()
   result.add_float(_("Speed Y"), &speed.y, "speed-y");
   result.add_float(_("Acceleration"), &acceleration, "acceleration");
   result.add_bool(_("Blowing"), &blowing, "blowing", true);
+  result.add_bool(_("Affects Badguys"), &affects_badguys, "affectsbadguys", false);
```
tobbi, line 71: Can we name the lisp tag `affects-badguys` (we separate words with a dash in lisp by default).

src/object/wind.cpp
```diff
@@ -47,6 +49,8 @@ Wind::Wind(const ReaderMapping& reader) :
 
   reader.get("acceleration", acceleration, 100.0f);
 
+  reader.get("affectsbadguys", affects_badguys, false);
```
tobbi, line 52: Can we name the lisp tag affects-badguys (we separate words with a dash in lisp by default).

src/object/wind.cpp
martinvanzijl: Fixed now.

src/object/wind.cpp
martinvanzijl: Updated as suggested.

src/object/wind.cpp
martinvanzijl: Updated as suggested.

martinvanzijl:
@tobbi Thanks for the review. I've updated the code now.

src/badguy/badguy.cpp
```diff
@@ -905,4 +905,24 @@ BadGuy::after_editor_set()
...
+bool
+BadGuy::can_be_affected_by_wind() const
+{
+  return !on_ground();
+}
+
+void
+BadGuy::add_wind_velocity(const Vector& velocity, const Vector& end_speed)
+{
+  // only add velocity in the same direction as the wind
```
tobbi, line 917: I missed the superfluous space around the `&&`here as well. If you don't mind, please fix this as well, then it can get merged.

martinvanzijl: Have fixed now.

martinvanzijl:
@tobbi Thanks again, code updated now.

I wonder if wind which affects badguys should have some visual indicator, so the player knows beforehand?

tobbi:
> @tobbi Thanks again, code updated now.
> 
> I wonder if wind which affects badguys should have some visual indicator, so the player knows beforehand?

It would be very nice but at the top of my head, I couldn't imagine which indicator might work.

PR #1393, opened 2020-05-07
Prevent infinite growling sound from Haywire

### LINKED ISSUE #1260: Infinite purple bomb growl if you run away very fast from it
**System information:** 

##### Expected behavior
If you activate a purple bomb, it growls. If you run away very fast from it (so that it is out of view before it explodes), the growl should no longer be hearable.

##### Actual behavior
If you activate a purple bomb, it growls. If you now run very fast away from it (so that it is out of view before it explodes), the growl will continue to play forever.

1. Turn on sound
2. Play a relatively flat level with a purple bomb in it
3. Jump on the purple bomb to activate it
4. Run as fast as you can to the right to get out of view before the purple bomb explodes

martinvanzijl:
Fixes #1260. The growling sound from Haywire (the purple bomb) will be stopped when it goes off-screen and is deactivated before actually exploding.

### DIFF
--- a/src/badguy/haywire.cpp
+++ b/src/badguy/haywire.cpp
@@ -141,6 +141,14 @@ Haywire::active_update(float dt_sec)
 }
 
 void
+Haywire::deactivate()
+{
+  // stop ticking/grunting sounds, in case we are deactivated before actually
+  // exploding (see https://github.com/SuperTux/supertux/issues/1260)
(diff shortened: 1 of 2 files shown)


tobbi:
If I'd seen this earlier, this would've made the RC. Now it's only gonna be in the final release of 0.6.2. Thank you!

PR #1402, opened 2020-05-19
Allow scrolling in credits screen with up/down keys

### LINKED ISSUE #1231: Can no longer use arrow keys to speed up credits screen

##### Expected behavior
When you press up and down in the credits screen, you can scroll the credits.

##### Actual behavior
Can no longer use arrow keys to scroll in or speed up the credits. **This is a regression from 0.6.0!**

1. Open the credits screen
2. Press up or down

martinvanzijl:
Fixes #1231. You can now scroll in the Credits screen using the up and down arrow keys. You can also skip ahead using the action or jump keys.

Demo below. Here I use the down, up action ("a" for my setup) and jump (space) keys:

credits-scroll-demo

### DIFF
--- a/src/object/textscroller.cpp
+++ b/src/object/textscroller.cpp
@@ -38,15 +38,16 @@ namespace {
 
 const float LEFT_BORDER = 0;
 const float DEFAULT_SPEED = 60;
+const float SCROLL_JUMP = 60;
 
 } // namespace
 
 TextScroller::TextScroller(const ReaderMapping& mapping) :
-  controller(),
+  controller(&InputManager::current()->get_controller()),
   m_filename(),
   m_lines(),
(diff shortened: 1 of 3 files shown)


martinvanzijl:
Further notes - there seem to be two scenarios in which a text-scroller object is used:

1) A "textscroller" element is added directly to the level file. This is done in the Credits screen.

2) Using the script function ("display_text_file(<filename>"). An example of this is the first level in "Tux Strikes Back". This creates a new `TextScrollerScreen` object.

Keyboard handling was mostly done in `TextScrollerScreen`, but I have moved this to `TextScroller`, so that it works in both scenarios 1 and 2.

tobbi:
Looks good!

PR #1406, opened 2020-05-22
Override only configurable keys (partially fixes #1400)

### LINKED ISSUE #1400: Inconsistent confirmation keys
SuperTux 0.6.2
Ubuntu 20.04

To confirm a selection (whether it's a menu item or a level in a worldmap), the player has to press either Enter, Space, Ctrl(or Alt, or whatever ACTION is bound to), or use the mouse (where applicable). However, this is very inconsistent. For example, in the main menu, any of the options can be used, when selecting a level in a worldmap, Enter doesn't work, in certain menus in the editor, Space doesn't work, in the 'Do you want to save' menu, both Enter and Space don't work, etc.
I've also noticed that 'P' can no longer be used instead of Escape (for Pause / skip dying animation). That can get really annoying - I still keep pressing P even though it hasn't worked for a while.

I'm not sure if this was changed intentionally or not, but at least the inconsistencies should be sorted out. Maybe 'Confirm' and 'Pause' could even be settable in Options. Or at the very least, all the keys that can be used as 'Confirm' somewhere should be able to be used as 'Confirm' anywhere.

szborows:
See #1400 

Idea here is to introduce "configurable keys" because (unless there are some hacks in config file) it doesn't make sense to re-bind keys that can't be configured through GUI menu.

I'm not very familiar with the code base. I'm not sure if proposed resolution is proper one. I think it's a little bit fragile because it will be easy to introduce inconsistencies (e.g. adding more items to menu but forgetting about configurable keys set in keyboard config class).

Therefore please advise where this PR should go. And please let me know if re-binding keys that aren't present in the kbd menu should be allowed.

### DIFF
--- a/src/control/keyboard_config.cpp
+++ b/src/control/keyboard_config.cpp
@@ -24,6 +24,7 @@
 
 KeyboardConfig::KeyboardConfig() :
   m_keymap(),
+  m_configurable_controls(),
   m_jump_with_up_kbd(false)
 {
   // initialize default keyboard map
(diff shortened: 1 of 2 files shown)


tobbi:
Thanks for that!

PR #1110, opened 2019-05-23
(WIP) Tux revamp

Alzter:
I'm revamping the Tux sprites to be more consistent, lively and appealing.

PR #1407, opened 2020-05-24
Add swimming

weluvgoatz:
Here is the revived branch of swimming, which is a planned feature.  The code and sprites were created mainly by @Zwatotem and me.  Although I have tested and fixed numerous bugs with this feature, it would be nice if anyone can find any that I missed and maybe then this could be merged?  Swimming would fit right in the game anyway.


Ordoviz:
Is it intentional that you can spam the jump button to gain hyper speed? I suggest limiting Tux’ speed.

### Issues:

General:
- [ ] Hold up while you are in water. Eventually, Tux will rotate on the water surface.
- [ ] When Tux jumps into water from above, the sprite flips out for a frame. This makes the transition not very smooth.

Big Tux:
- [ ] Big Tux teleports a little when his hitbox changes from horizontally to vertically and vice versa.
- [ ] Big Tux can buttjump when he makes a high jump out of the water. This looks really odd.

Fire/ice flower:
- [ ] The bullets get shot to the right or left not the direction Tux is facing. We also need to discuss how Tux creates these bullets and how he throws them (with the right movement it looks like Tux poops them out).

Small Tux:
- [ ] When small Tux jumps out of the water to the top left, he flips after he reaches his max height and falls down on his back. This is nice for falling straight down but not for jumping in a parabola (it does not happen with big Tux or when jumping to the top right)

Missing/bad graphics:
- [ ] Big Tux looks very janky. I cannot take him seriously.
- [ ] Create a sound effect that plays when pressing JUMP while in water and make Tux swirl his arms.

Narre:
I've been working on a new add-on for SuperTux for a while now (a.k.a since 2015), and I have a good few levels centred around swimming, so I will test this feature in actual levels (rather than test levels) as soon as I can, and I will make sure to leave as much feedback as I can.

src/object/player.cpp
```diff
@@ -784,7 +988,28 @@ Player::handle_input()
     handle_input_climbing();
     return;
   }
+if (m_swimming) {
```
christ2go, line 991: Please fix the indentation.

src/object/player.cpp
```diff
@@ -324,10 +354,43 @@ Player::update(float dt_sec)
...
-  if (fabsf(m_physic.get_velocity_x()) > MAX_WALK_XM) {
-    m_col.set_width(RUNNING_TUX_WIDTH);
-  } else {
-    m_col.set_width(TUX_WIDTH);
+
+  Rectf lookahead = get_bbox();
+  lookahead.set_right(lookahead.get_left() + 62);
+  bool pathBlocked = !Sector::get().is_free_of_statics(lookahead);
+  
+  if((m_swimming || m_water_jump) && is_big() && !pathBlocked)
```
christ2go, line 362: Maybe use pathFree instead of pathBlocked and avoid double boolean negation (-> improve readability)

src/object/player.cpp
```diff
@@ -444,6 +506,149 @@ Player::update(float dt_sec)
...
+    pointx = -1.f;
+  else if (m_controller->hold(Control::RIGHT)&&!m_controller->hold(Control::LEFT))
+    pointx = 1.f;
+  bool boost = m_controller->pressed(Control::JUMP);
+
+  swim(pointx,pointy,boost);
+}
+
+void
+Player::swim(float pointx, float pointy, bool boost)
```
christ2go, line 528: Please add some documentation what all these calculations actually do. I guess they could be structured more clearly (the mix of changing vx vy and setting them more than once is a little bit confusing.) Idea: Compute the new velocities and set them just once.

Also I guess a lot of these constants should be declared as constants somewhere.

weluvgoatz:
I've fixed some bugs with this PR.  I marked them off that checklist, as well as some suggestions that aren't necessary for this PR to be merged/things I can't replicate or reproduce or see as a problem.

Ordoviz:
Your new changes removed Tux’ backflip animation.

weluvgoatz:
Tux is planned to have a new backflip animation eventually anyway that
isn't rotated.

Grumbel:
What's the plan with the backflip? I kind of liked the old behavior and haven't really seen much reason to change it.

weluvgoatz:
@Grumbel: At some point, the backflip will be redrawn to look better, as the rotating behavior definitely has room for improvement (a hand drawn flipping animation is ideal) (would be part of the Tux graphics revamp, see #1110 for more).  However, if a stable release happens before the graphics are updated, then the old behavior will probably be temporarily re-added.

PR #1408, opened 2020-05-26
Update Penny's Look

serano01:
- this will change Penny's Look from purple to black with a small scarf
- added a small idle animation similar to Tux' winking as well


qwertychouskie:
Penny still looks quite pixelated.  I wonder if it would be easier to take Tux as a base and add/tweak things to turn it into Penny.  This would hopefully avoid the pixelation issue.

Alzter:
> Penny still looks quite pixelated. I wonder if it would be easier to take Tux as a base and add/tweak things to turn it into Penny. This would hopefully avoid the pixelation issue.

I feel like it'd be more optimal to just draw her a new design. This concept art by grumbel shows what she could potentially look like, a bit slimmer than Tux with more angular features.

image

Zwatotem:
For me this graphic looks okay. Pixelated look should be easily repairable with some form of anti-aliasing effect applied.

Alzter:
> For me this graphic looks okay. Pixelated look should be easily repairable with some form of anti-aliasing effect applied.

Even then that doesn't cover the fact that both Penny and Nolok are nearly inanimate

serano01:
I'm closing this, since it needs more work.

PR #1409, opened 2020-05-26
Add missing releases to appdata file

xduugu:
This data is used by flatpak and because it was not updated for the previous releases, it lists the distributed version 0.6.2 as 0.6.1: https://flathub.org/apps/details/org.supertuxproject.SuperTux


tobbi:
Thank you!

Issue #556, opened 2016-08-17
Checkpoint Problems


Karkus476:
You won't respawn at the checkpoint if you have <25 coins. Each time you respawn at checkpoint you lose either 10% or 25 coins, whichever is higher. I know this from the code. Players do not know this. It should be explained, and we need to discuss if it should be changed.

tobbi:
The 10% or 25 coins thing is explained in the first level since https://github.com/SuperTux/data/commit/aaad57f96eeb8b3d053f4c830fb68ac9642839d4

Karkus476:
The reason I made this issue is because we are a mario clone, not a sonic clone, and I cannot work [this](http://www.wolframalpha.com/input/?source=nav&i=log+base(9%2F10)(250%2Fx)%2B10) out in my head.

qwertychouskie:
Why not just do a number of lives, as this is what most people are used to?

Hume2:
This behaviour is good when levels of an author contain loads of coins, so Tux doesn't get loads of lives. It's exponential, so he becomes only a few lives extra.

Rusty-Box:
It has been decided again the concept of lives as well as removing the link between coins and checkpoint respawning. See #1410

Issue #1293, opened 2019-12-22
Revert to the 1-up counter for players


Rusty-Box:
This might become a controversial idea but:

Instead of having the complicated coin math, percentage mumbo jumbo, the game should revert back to the 1-up counter for players. However, incase the idea of adding shops still is being consitered for the future the coin counter should not be limited to only 100 coins.

Depending over the game's overall difficulty every 100 coins can reward the player with an extra 1-up besides the Tux doll (i.e. if the game would end up too easy it would only make things worse... but I assume that is not the case here).

weluvgoatz:
I am strongly against this idea.  A one-up counter seems like it will make
the game more complicated than it already is.  Also, the 1-up counter is
one of the things that confused me the most about 0.1.3, because it was
unclear what would happen when the 1-up counter reached zero.  In my
opinion, it will be better if the coin deduction "mumbo jumbo" is just
removed, because I still like collecting coins for the sake of having
something collectible and countable.

weluvgoatz:
Also, weren't 1-ups directly related to the lives system, which is
something that was considered a rejected feature forever?

Rusty-Box:
> Also, weren't 1-ups directly related to the lives system, which is something that was considered a rejected feature forever?
Possibly, I don't know. Most likely I wasn't around at that time so I'm unaware of that statement if that's true.
> A one-up counter seems like it will make the game more complicated than it already is.
Is it really? 2 lifes -> You die -> 1 life left etc.?

weluvgoatz:
But what about when the lives counter reaches zero?

Rusty-Box:
> In my opinion, it will be better if the coin deduction "mumbo jumbo" is just removed, because I still like collecting coins for the sake of having something collectible and countable.

So no lifes, no coins related to checkpoint restarts? A bit too easy isn't it? ATleast for my taste.

> But what about when the lives counter reaches zero?

Game Over? Like... in every other game that uses this mechanic? You will replay the level again from the very beginning. Like already handled when you have no coins or less then 25, I believe.

Rusty-Box:
Maybe get kickout back to the world map first so you need to re enter the level first.

weluvgoatz:
Well "Game Over" doesn't work well in my opinion because 1) It sounds like
you must restart the game if you run out of lives, which is misleading and
2) Having to restart from just the beginning of a level is a bit tame if
you must have a "Game Over" and "Lives" screen.

weluvgoatz:
[...]

PR #1410, opened 2020-05-30
Remove link between coins and checkpoints

weluvgoatz:
This very simple change makes it to where you won't be punished for losing coins at checkpoints, as you can always respawn at them now.  However, the coins can still be removed by dying at a checkpoint, as a sort of score system, as being a score system is now coins' only purpose.  This makes hard levels much fairer, and a vote done in the SuperTux Discord proved this option more favorable than keeping it the same.

The reason this is a Pull Request and not directly implemented is because of discussion that can still be had.  If it is agreed on by the SuperTux team then this will be merged.

### DIFF
--- a/src/supertux/player_status.cpp
+++ b/src/supertux/player_status.cpp
@@ -63,8 +63,7 @@ PlayerStatus::get_max_coins() const
 bool
 PlayerStatus::can_reach_checkpoint() const
 {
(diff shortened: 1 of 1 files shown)


Ordoviz:
This would close #556 and #1293. There is no consensus on these issues, though.

PR #1414, opened 2020-06-02
Apply ice friction only when Tux toches the ground

### LINKED ISSUE #1413: Obnoxious Ice Physics
SuperTux 0.6.2
Windows 7

So, the ice physics in Supertux is really annoying. It's pretty hard to correct your jumps and stop while running or even walking sometimes. I know that it's meant to be like that but in some cases it just gets pretty annoying. In comparison to the ice physics in Mario games, the ice physics in Supertux is really annoying. So, please try to fix the ice physics and make it a little less obnoxious. I don't think so that it'll break any levels. Maybe I'm just pretty bad at this games but, please make it a little less slippery and much more controllable. Maybe, a feature can be added to control the slipperiness of the ice. Maybe, you can just ignore this issue if you want. It's not that big of a deal. 

Thanks:)

HybridDog:
This makes the slippery ice behaviour less confusing and easier.
It should fix #1413

### DIFF
--- a/src/object/player.cpp
+++ b/src/object/player.cpp
@@ -317,15 +317,15 @@ Player::update(float dt_sec)
       m_water_jump = false;
     }
   }
-  
+
   if (on_ground() && no_water) {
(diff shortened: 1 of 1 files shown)


LizzyBach:
Yeah, it's much better. Thx:-)

Issue #1317, opened 2020-01-08
Crystal cave is too dark


Zwatotem:
**SuperTux version:** 0.6.1, possibly earlier
**System information:** GNU/Linux Parabola

##### Expected behavior
Dark part of Crystal cave level is dark, but visible to naked eye, so one can activate bonus block in the range of sight and light up the level.
##### Actual behavior
On some screens dark area is unplayably dark, or even completely black.
##### Steps to reproduce actual behavior
Play Crystal cave level, go behind the door opened by 2 switches.

Zwatotem:
Two ways of solving this issue:
1. Make a slightly glowing circle around the penguin, not brighter, than the rest of the level, but bright enough, so everything can be seen easily. This circle should of course be a lot less convenient than other lights (literally inconvenient), that players would prefer to use alternatives for them by himself.
2. Make the whole section of the level brighter.

I'm in favor of first option, but maybe there are some other suggestions.

HybridDog:
I am against the first option because somebody may want to create a level which is completely dark and the player has to play blindly.

The perceived screen brightness depends on the brightness of the surrounding area (in my experience), so dark areas are a lot brighter at night.
Furthermore, humans perceive colours differently at night: https://en.wikipedia.org/wiki/Mesopic_vision

Is this the right crystal level?
![cav](https://user-images.githubusercontent.com/3192173/72173343-9cceb880-33d7-11ea-83cd-b82f92eb0503.jpg)

Zwatotem:
Yes, this is the level. I mean the part to the right behind double door.

HybridDog:
> Yes, this is the level. I mean the part to the right behind double door.

![fot_2020-01-18-17:39:05_479007068](https://user-images.githubusercontent.com/3192173/72667226-d62ba780-3a19-11ea-9f07-5e7a9eba9a66.jpg)

This is too dark for me, too. You could add mushrooms which glow when Tux walks on them; it would have an effect similar to making Tux glow. The level designer decides where to place them, so it is still possible to keep some dark areas. In comparison to torches, the mushrooms would only glow when Tux touches them.

Alzter:
I don't like pitch black levels, difficulty should stem from there being a fair challenge, not because you can't see any of the level, and having the player unable to see isn't exactly what I would consider fun, and I end up just turning up my PC gamma level to compensate anyway. I say we make it more visible.

slackstone:
[...]

PR #1416, opened 2020-06-03
Add the RubLight object

HybridDog:
It glows when it is rubbed, e.g. when Tux or a badguy walks on it.
Skidding makes it glow bright.

I think this object could be used to fix the dark crystal mine issue #1317.
fot_2020-06-03-20:05:27_545027911

I have never added a new object to SuperTux, so please tell me if I did something wrong.

It has not much to do with [that Rublight](https://www.youtube.com/watch?v=bGROryzHLII). I've given it this name because [Triboluminescence](https://en.wikipedia.org/wiki/Triboluminescence) literally means rublight and Triboluminescence is probably more complicated and realistic than what I have implemented.


src/badguy/walking_badguy.hpp
```diff
@@ -53,7 +53,8 @@ class WalkingBadguy : public BadGuy
 
   void active_update(float dt_sec, float target_velocity);
 
-  float get_velocity_y() const;
+  float get_velocity_y() const { return m_physic.get_velocity_y(); }
```
tobbi, line 56: Please re-order these so x comes before y (I know, I'm too OCD about this).

src/object/rublight.cpp
```diff
@@ -0,0 +1,117 @@
...
+    if (get_brightness() < 0.000001f) {
+      stored_energy = 0;
+      state = STATE_DARK;
+    }
+    break;
+  }
+}
+
+float
+RubLight::get_brightness()
```
tobbi, line 97: Mark this method `const`.

weluvgoatz:
Is the color of the light here configurable?  If not, it would be nice to have.

HybridDog:
> Is the color of the light here configurable? If not, it would be nice to have.

I've made the color and fading speed configurable.
The default value of the `fading_speed` appears in three places: the constructor, in the `mapping.get` invocation and in the `get_settings` method. I've looked at the spotlight code for how to do it; maybe some of them are redundant.

tobbi:
One small thing from Travis:
```
/rublight.cpp.o -c /Users/travis/build/SuperTux/supertux/src/object/rublight.cpp

/Users/travis/build/SuperTux/supertux/src/object/rublight.cpp:34:3: error: field 'fading_speed' will be initialized after field 'light' [-Werror,-Wreorder]

  fading_speed(5.0f),

  ^

1 error generated.
```

HybridDog:
> One small thing from Travis:

it should be fixed now

PR #1417, opened 2020-06-03
Remember --spawnpoint and --sector across deaths.

divVerent:
Makes playtesting nasty levels much easier.

Note: I am unsure about this implementation; I find abusing the checkpoint system rather nasty. I wonder if it would be a better idea to rather override a lot of what game_session does, by moving name of initial spawnpoint into a member or something.

Decided to not overload Sector::activate to make clearer at callsites what is a spawnpoint name (there appears to be some confusion in one spot, comment added).

### DIFF
--- a/src/editor/editor.cpp
+++ b/src/editor/editor.cpp
@@ -359,7 +359,7 @@ Editor::set_sector(Sector* sector)
   if (!sector) return;
 
   m_sector = sector;
-  m_sector->activate("main");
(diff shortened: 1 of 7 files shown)


divVerent:
Posted also https://github.com/SuperTux/supertux/pull/1419 as an alternate approach (less intrusive).

divVerent:
Closing in favor of https://github.com/SuperTux/supertux/pull/1419.

PR #1419, opened 2020-06-04
Remember --spawnpoint and --sector across deaths.

divVerent:
Makes playtesting nasty levels much easier.

A less intrusive way, as an alternative to https://github.com/SuperTux/supertux/pull/1417


tobbi:
FTR: I do prefer this solution.

divVerent:
Probably going to implement an editor feature "test from here" similar to this change.

PR #1424, opened 2020-06-10
Add a "speed" property to path nodes.

divVerent:
To not break the current level format compatibility, this property is editor
only and the "time" property remains authoritative. However, whenever speed
is set, any editor action touching the node will update time according to speed
and the distance to the next node.

### DIFF
--- a/src/editor/node_marker.cpp
+++ b/src/editor/node_marker.cpp
@@ -39,17 +39,11 @@ NodeMarker::update_iterator()
 Vector
 NodeMarker::get_point_vector() const
 {
-  std::vector<Path::Node>::iterator next_node = m_node + 1;
-  if (next_node == m_path->m_nodes.end()) {
-    if (m_path->m_mode == WalkMode::CIRCULAR) {
-      //loop to the first node
-      return m_path->m_nodes.begin()->position - m_node->position;
-    } else {
-      return Vector(0,0);
-    }
+  std::vector<Path::Node>::const_iterator next = next_node();
+  if (next == m_path->m_nodes.end()) {
+    return Vector(0,0);
   } else {
-    //point to the next node
-    return next_node->position - m_node->position;
+    return next->position - m_node->position;
   }
 }
 
@@ -64,6 +58,7 @@ NodeMarker::move_to(const Vector& pos)
 {
   MovingObject::move_to(pos);
   m_node->position = m_col.m_bbox.get_middle();
+  update_node_times();
 }
 
 void
@@ -73,6 +68,9 @@ NodeMarker::editor_delete()
   {
     return;
   }
+  std::vector<Path::Node>::iterator prev = prev_node();
+  std::vector<Path::Node>::const_iterator next = next_node();
+  update_node_time(prev, next);
   m_path->m_nodes.erase(m_node);
   Editor::current()->update_node_iterators();
 }
@@ -82,6 +80,7 @@ NodeMarker::get_settings()
 {
   ObjectSettings result(_("Path Node"));
   result.add_float(_("Time"), &(m_node->time));
+  result.add_float(_("Speed"), &(m_node->speed));
   return result;
 }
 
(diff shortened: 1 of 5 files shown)


tobbi:
I gotta ask: Why is there a const iterator for next_node but not one for prev_node ?

divVerent:
Only reason is that I didn't need a const one for prev, but had to query
the next one to simplify the existing function to find the arrow vector as
that method is const.

OTOH I needed a non const iterator too so I can modify it.

Can undo that refactor of the arrow function if you want.

tobbi:
No, it's perfectly fine :) I was just wondering, that's all.

divVerent:
Actually, I just noticed that the mutable version of next_node is unused, so I'm gonna delete it. So previous node is only ever used mutably, next node only ever used immutably, mainly because the speed for node A to node B is stored in node A.

src/editor/node_marker.cpp
```diff
@@ -78,13 +76,55 @@ NodeMarker::get_settings()
...
 }
 
 void
 NodeMarker::editor_update()
 {
   set_pos(m_node->position - Vector(8, 8));
+  update_node_time(m_node, next_node());
+}
+
+std::vector<Path::Node>::iterator NodeMarker::prev_node() {
```
Zwatotem, line 90: For me type name `std::vector<Path::Node>::iterator` seems a bit long and hard to read. It could be my lack of experience in bigger projects. Also I'm not sure, what is our policy regarding naming types.

Zwatotem:
I think you should change everything back to mutable, (or maybe that's not the cause) because I'm getting read access violation exception upon deleting nodes.

divVerent:
> I think you should change everything back to mutable, (or maybe that's not the cause) because I'm getting read access violation exception upon deleting nodes.

I cannot reproduce; however I did find an unrelated bug in node deletion. Deleting the first node of a new (size 1) path TWICE crashes:

```
888	  for (auto i = m_edited_path->m_nodes.begin(); i != m_edited_path->m_nodes.end(); ++i) {
(gdb) print m_edited_path->m_nodes.size()
$9 = -1
```

Didn't know you can make a `std::vector` have negative size ;)

Anyway, can you explain how you caused the read access violation by deleting nodes? Like, something of the sort "make new path, add 5 nodes, then delete 3rd, then 4th, then 1st, then 1st"?

divVerent:
Sent PRs for all crashes I know (all pre-existing), including this one about node deletion I just found.

Zwatotem:
So as precisely, as possible:
(I was debugging on VS2019.) Open level editor, create any new level (or go to existing one). Add a _flying platform_ object and select path object. Add new nodes to the path. Then change your tool to eraser, and select any of existing nodes.
However I just realized this doesn't occur on Release build.

divVerent:
I can't reproduce it that very way (but am on Linux/x86_64. Does it matter for you which node you delete?

Also, I am very sure the constness can't be at fault, as one would need a const_cast to cause any runtime problems with that. Can you maybe obtain a backtrace of the failure in VS2019?

Zwatotem:
No, the choice of deleted node didn't matter.

I can't provide you a backtrace anymore, since VS is making problems with building that I have no time to fix now. However I feel this is not a big problem, since on .msi from appveyor everything works smoothly (Except first node deletion)

divVerent:
I just managed to repro a crash: when the editor is in copy mode, and you use the erase tool on a node marker, then it crashes, at least with ASan enabled.

Of course, copy mode + erase tool itself is nonsense right now, as it makes a copy then erases it immediately - nevertheless this seems worth fixing.

divVerent:
Wait no - ANY erase of a platform triggers asan. In code unrelated to my change.

divVerent:
ASan fixed by https://github.com/SuperTux/supertux/commit/bfd11ac605be0abecf430b87b826b405a33fe909 - the crash that I did find fixed by https://github.com/SuperTux/supertux/commit/b9cf4f3b41fd276b09b97ee41458bf51347e8590. Can you try again on Windows?

src/editor/node_marker.cpp
tobbi: We have a couple typedef's lying around the code, so we definitely use them.

Zwatotem:
Okay, seems woking correctly. Merging.

PR #1427, opened 2020-06-11
remove extra space from function and parameter

jgarte:
Was there extra space here that should be removed in this function?


tobbi:
Thanks! 👍

PR #1431, opened 2020-06-13
Add sideways icecrushers

weluvgoatz:
This adds an option for the icecrusher to go sideways instead of up and down, and this can be activated if the player is in the line of sight or stands on top of the icecrusher and gets in the direction he wants to be launched.  The code quality of this might be poor, which I will fix, but please bugtest and see if everything works alright.  If so, hopefully this can be merged.


Zwatotem:
I spotted three things right away:
1. Icecrushers placed right by the wall (with snapping in editor) won't ever get back to their original position (pixel precision bug): https://youtu.be/fHaZs-AmkqY
2. Not really a bug, but probably unintended behavior; icecrushers keep going back when Tux just chills on top of them: https://youtu.be/dDZwa4xy_6c
3. Unlike topdown crushers you don't die upon touching the crushing side (no video)

weluvgoatz:
I've added a commit to hopefully make this build and fix that first bug you listed.  However, the other 2 things you listed are very much intentional behavior.

src/object/icecrusher.cpp
```diff
@@ -48,8 +49,10 @@ IceCrusher::IceCrusher(const ReaderMapping& reader) :
   lefteye(),
   righteye(),
   whites(),
-  ic_size(NORMAL)
+  ic_size(NORMAL),
+  sideways(false)
```
tobbi, line 53: No need to initialize this to false. Bools are false upon initialization by default.

src/object/icecrusher.cpp
```diff
@@ -81,11 +102,33 @@ IceCrusher::set_state(IceCrusherState state_, bool force)
       physic.reset ();
       physic.enable_gravity (true);
       m_sprite->set_action("crushing");
+      break;
+	case CRUSHING_RIGHT:
+      set_group(COLGROUP_MOVING_STATIC);
+	  physic.reset ();
```
tobbi, line 108: Consistent line indentation, please.

src/object/icecrusher.cpp
```diff
@@ -179,6 +221,52 @@ IceCrusher::collision_solid(const CollisionHit& hit)
         set_state(RECOVERING);
       }
       break;
+	case CRUSHING_RIGHT:
```
tobbi, line 224: Consistent line indentation and brace style...braces in a new line. I hate Java style braces, and I hate Java.

src/object/icecrusher.cpp
```diff
@@ -202,14 +291,26 @@ IceCrusher::update(float dt_sec)
   switch (state) {
     case IDLE:
       m_col.m_movement = Vector (0, 0);
-      if (found_victim())
+      if (found_victim_down() && !sideways)
         set_state(CRUSHING);
+	  if (found_victim_right() && sideways)
```
tobbi, line 296: Indentation.

src/object/icecrusher.hpp
```diff
@@ -29,7 +29,11 @@ class IceCrusher final : public MovingSprite
   enum IceCrusherState {
     IDLE,
     CRUSHING,
-    RECOVERING
+	CRUSHING_RIGHT,
```
tobbi, line 32: Indentation (check all lines)

src/object/icecrusher.cpp
weluvgoatz: Fixed.

src/object/icecrusher.cpp
weluvgoatz: Fixed.

src/object/icecrusher.cpp
weluvgoatz: Fixed... I presume?  The other cases don't do this but I did it anyway.

src/object/icecrusher.cpp
weluvgoatz: Fixed.

src/object/icecrusher.hpp
weluvgoatz: Fixed... I think.

divVerent:
On the one hand, I really hate them with a passion. Sideways crushers make for bad and surprising gameplay, and they got quite a bad track record in a certain other game with a level editor. Their only purpose ever is for trolling - anything else is better done with a platform. What's next, suddenly attaching little parachutes to them making them instantly slow down, for even more trolling?

On the other hand: making a vertical ice crusher in a troll level took me only this much code:

```
    (scriptedobject
      (name "juzkrosh")
      (sprite "/images/creatures/icecrusher/krosh.sprite")
      (x 2176)
      (y 416)
    )
    (ispy
      (script "juzkrosh.set_velocity(-500, 0)
wait(1)
juzkrosh.set_velocity(200, 0)")
      (facing-down #t)
      (direction "down")
      (x 2048)
      (y 384)
    )
```

Given this, the cost of _not_ providing them, despite them being that simple to make anyway, is that level creators who want them anyway will do so using nonstandard means. If they must exist, it's probably better to standardize their gameplay, including sensitivity to player, crushing speed, etc, rather than every level using them having somewhat different parameters on them.

src/object/icecrusher.cpp
```diff
@@ -226,6 +333,40 @@ IceCrusher::update(float dt_sec)
...
+        set_pos(start_position);
+        m_col.m_movement = Vector (0, 0);
+        if (ic_size == LARGE)
+          cooldown_timer = PAUSE_TIME_LARGE;
+        else
+          cooldown_timer = PAUSE_TIME_NORMAL;
+        set_state(IDLE);
+      }
+      else {
+        if (ic_size == LARGE)
```
tobbi, line 365: The code in both if and else is the same.

src/object/icecrusher.cpp
```diff
@@ -82,10 +103,32 @@ IceCrusher::set_state(IceCrusherState state_, bool force)
...
       break;
+    case CRUSHING_RIGHT:
+      set_group(COLGROUP_MOVING_STATIC);
+	    physic.reset ();
+      physic.enable_gravity (false);
+      m_sprite->set_action("idle");
+	  break;
+	case CRUSHING_LEFT:
+      set_group(COLGROUP_MOVING_STATIC);
+	  physic.reset ();
```
tobbi, line 114: Please fix indentation throughout the code.

src/object/icecrusher.cpp
weluvgoatz: Fixed.

src/object/icecrusher.cpp
weluvgoatz: I fixed what I saw but if there is more indentation issues let me know.

PR #1449, opened 2020-06-24
Issue 1398 fix throwable bug with coins

### LINKED ISSUE #1398: Throwable objects interact with coins
**SuperTux version:** Introduced between 0.6.0 and 0.6.1-beta.1. I assume that https://github.com/SuperTux/supertux/commit/ee86826392ef68d7e05e2a07e917e727d82a41e8 is to blame

##### Expected behavior

Throwable objects (e.g. rocks, trampoline, lanterns) do not interact with coins, checkpoints or the rotating ice balls you typically see in castles.

##### Actual behavior

Throwable objects bounce off coins etc. This causes throwable objects to land on a different spot than you indented or fall more slowly because they fall through a trail of coins.

- Throw a rock from the side on top of a coin
- Throw a rock down on a trail a of coins

martinvanzijl:
Fixes #1398. Throwable objects like rocks will no longer collide with coins or "flame" objects, like rotating ice balls.

Demo before:

rock-coin-bug-before

Demo after:

rock-coin-bug-after

### DIFF
--- a/src/object/rock.cpp
+++ b/src/object/rock.cpp
@@ -19,6 +19,8 @@
 #include "audio/sound_manager.hpp"
 #include "object/explosion.hpp"
 #include "object/coin.hpp"
+#include "badguy/badguy.hpp"
+#include "object/player.hpp"
 #include "supertux/sector.hpp"
 #include "supertux/tile.hpp"
 #include "util/reader_mapping.hpp"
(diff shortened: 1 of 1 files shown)


martinvanzijl:
Should the idea of thrown objects collecting coins (https://github.com/SuperTux/supertux/issues/1398#issuecomment-627918153) be raised as a seperate feature request?

HybridDog:
There could be an easier fix.
As far as I can see, the cases in which the rock should not be moved are objects which have the `COLGROUP_TOUCHABLE` collision group, which is visualized in orange:
fot_2020-06-24-18:37:20_029728264
fot_2020-06-24-18:37:50_023642255
fot_2020-06-24-18:37:58_590875101

When I wrote the code I didn't think about touch collisions. The problem could be fixed with `moving_object->get_group() == COLGROUP_TOUCHABLE` (and including `collision/collision_group.hpp`).

martinvanzijl:
@HybridDog Thanks for the suggestion! This seems to work, and is a lot simpler, so I changed the code accordingly.

PR #1459, opened 2020-07-13
Fix Travis build

martinvanzijl:
Follow-up from PR#1457. Fixes the type conversion warning introduced by the PR.


martinvanzijl:
The hard part was testing this locally. I could do this with `make --just-print`, copy compile command to text editor, add the flag `-Wconversion`, then paste and execute in the shell.

Is there an easy way to test the Travis build locally?

tobbi:
I'm sorry, @Zwatotem already fixed this with faccbca448f0f59c37bba360665d042e4a1b2758. Thanks, anyway 👍

PR #1460, opened 2020-07-13
Bonus Island 1 Level Improvements

Alasdairbugs:
Minor changes to all Bonus island 1 levels including:

- Enemy changes in bonus island castle (impossible enemies made the level unable to be "perfect").
- slight changes in Area 42 with jumps that were impossible as big tux.
- Stopped enemies from defeating themselves by walking into 1-wide areas in multiple levels.
- Enemy deletions due to being placed in awkward spots.
- Removed old "end sequence" and "final goal" tiles and replaced them with the red sequence trigger scripts.
- Changed some terrain slightly in other stages to allow for less awkward jumps (This is 1 or 2 tiles per level, nothing major).
- Removed 2 spikys in Cave of Mirrors that were impossible to jump over with big tux.
- Removed some Coins in "A Maze in the Sky" that made it extremely difficult to get the "perfect" goal.
- Re-alligned some badguys in some places, due to awkward placements.
- Removed the coin-box roof in the secret of "Semi-Frozen" due to impossible coins. Also removed the awkward warp that served no meaningful purpose, causes confusion and seems lazy.
- Removed a few visual bugs across some levels to do with tiling.

There's probably a couple more things i forgot but I think all the changes are reasonable and fair, don't change the terrain of the level that much, and also doesn't change the PowerUp meta, etc.


serano01:
Your work seems to be based on a older state of these levels, since I've reworked them already (not much, but the sequence trigger should be a object everywhere instead of tile-based). Also I'm missing some other improvements I made (swimming in the secret area in Semi-Frozen should be doable for example). Apart from that, your changes are very minor and since they're missing the changes I made, a merge would be counterproductive. Correct me, if I'm mistaken.

Alasdairbugs:
The changes i made were of levels i got in 0.6.2, are your changes included in the latest nightly? if so i can rework these from that nightly and make another PR of the changes i made from your improvements.

serano01:
Yes, the newest nightly would be recommend.

PR #1461, opened 2020-07-14
Level Improvements

Alasdairbugs:
All levels can now be beaten via the "perfect" objective. Enemies stop killing themselves in most cases, and all these changes were made from seranos changes to water tilemaps preparing for the swimming update.

- Removed impossible fish enemies in "Bonus Island Castle" and removed ones that were very difficult to kill.
- Removed some coins and badguys in "A Maze in the Sky" that were very hard to kill and collect, due to the scroller putting tux off-screen.
- Removed enemies across levels that killed themselves, and/or replaced them with stay-on-platform equivalents. "Cave of Mirrors" had 2 tilemap changes where i made a 3x1 hole for 2 snowballs instead of 2x2 hole that was previously there. This 3x1 hole was of course mirrored to fit the theme of the level being a mirror image.
- Made a cloud in "Tip of the Iceberg" appear solid instead of nonsolid.
- Changed some sections in Area 42 where you can get hit by an invisible ceiling. Also changed some jumps that were impossible as Big Tux.
- Removed 2 spikys in "Cave of Mirrors" that were impossible to defeat with Big Tux and Small Tux.

I think in a couple levels I removed swater tiles that didnt work properly. In Area 42 there was a very obvious graphical error with it so i overhauled it entirely. In Train Leaves in one minute the swater tile wasnt populated so i removed it and added another one. These can be easily fixed though.


papb:
> Removed enemies across levels that killed themselves

Why? I personally don't see this as necessarily a bad thing.

And does this mean you removed all bouncing snowballs from the whole game (except ones in closed spaces)?

Alasdairbugs:
Enemies that make death noises or just walk off ledges off-screen were removed. There's literally no point in having them if you cant see them dying, or hear them dying at all. Even if you do hear them dying off-screen, its very weird, and clumsy level design. It increases clarity. This really shouldn't be a discussion lol.

No i mostly removed enemies that killed themselves before the player could even touch them by squishing them (not by fire) so there was no point in having the enemies there if they don't pose any threat, and it is also lazy design because the designer couldn't bother to try and make the enemy work with terrain.

Also removed enemies that were in 1x1 spaces and killed themselves. It just leads the player to be confused.

papb:
I'm convinced, thanks. Another question:

> Also removed enemies that were in 1x1 spaces and killed themselves

What is this? I did not understand :grimacing:

Alasdairbugs:
If you place, lets say, a bomb onto a singular block, it will turn to the left and right constantly, causing it to explode. This was a mechanism introduced to stop players from cramming enemies into tiny spaces, and this is the result of that mechanism, the bomb thinks it's clustered.

serano01:
I thought for to be swim-able, water has to be placed at a z-pos with the value 0. But nevermind, thanks for your PR, merging now.

PR #1463, opened 2020-07-14
Level Improvements

Alasdairbugs:
Removed impossible collectibles, and poorly placed enemies that bad sections impossible without damage-boosting and added target times, etc.

- Removed off-screen impossible enemies in "Mario's supposed to Deal With This..."
- Adjusted Time limits to some levels that were otherwise impossible to "perfect" given the strict time limit (See: "Cliff Jumping").
- Removed off-screen enemies that killed themselves.
- Changed some on-screen enemies so they don't kill themselves before Tux can reach them physically.
- Added Level Target Times to all stages.
- Adjusted Coins, and badguys in some parts of "Think Fast, or Die" due to some collectibles being impossible to collect due to the autoscrolling.
- Removed impossible Coin Bricks in some stages; Coin bricks as walls, or in impossible or awkward positions.
- Removed a spiky in "You have one minute" and replaced it with spikes due to it being an impossible collectible.
- Removed a jumpy in "...Where Everything Is Possible" where you had to abuse a weird mechanic which involved throwing a fireball at the moment tux reverses his velocity causing a fireball to have a different, skewed trajectory... it's a mouthful but basically the enemy was very hard to kill, and given there are no ice cubes in the stage it needed to be removed, jumping down to its platform would cause a soft-lock.


papb:
Really nice!!

> jumping down to its platform would cause a soft-lock

What's a soft lock?

serano01:
There are some spikys in "...Where Anything Is Possible" which are placed in a one-tile path.

Alasdairbugs:
I can fix those spiky's when i remake the bonus levels visual effects.

@papb a softlock is when you're basically stuck and are forced to restart.

Issue #335, opened 2016-01-18
Carried objects can be released through solid magic blocks


IndianaTux:
Version 0.4.0   Windows 7 64bit
Throw any lantern at a solid magic block stack and the lantern will go through it. Make sure Tux throws the lantern half a tile close to the magic block stack. I noticed this glitch in previous SuperTux versions.

.

Hume2:
The magic blocks have a short switch delay, so when the lantern falls quick enough, it can fall trough the magic blocks before they start to be solid. This is intended feature I think.
There are two ways you can deal with that. You can make the ambient light brighter so it will trigger much earlier or you can put darker magic blocks, because they don't need so much light to switch on.

maxteufel:
> This is intended feature I think.

I agree that this can be seen as an intended feature. As such, this issue is invalid and will not be fixed.

@SuperTux/core, @SuperTux/developers: Please reopen this if you have serious concerns about this being the way it is.

LMH0013:
I think you guys have misunderstood what the bug is.  This occurs when the magic blocks are already solid, so the time delay before the block switches is irrelevant: the block is static.  Furthermore, the intended purpose of the switch delay is to prevent the blocks from flickering rapidly if the light threshold should oscillate quickly, not to allow lanterns to pass through first (such a "feature" doesn't really make any sense anyway).  I don't even think that lanterns can be dropped through magic blocks before they turn solid anymore anyway, the time delay was reduced to prevent that.

I have confirmed that the bug exists, and that it is not isolated to lanterns.  Indeed any object Tux can pick up can be passed through a solid wall of magic blocks (including badguys).  Thickening the walls causes funky behavior too.  The object progressively passes through a depth of two blocks, anything thicker and the object migrates downward.  There is definitely a glitch here.

Karkus476:
Here's a level to test the bug:
[issue335.stl.txt](https://github.com/SuperTux/supertux/files/111040/issue335.stl.txt)
and a recording of how to test it.
[issue335_recording.txt](https://github.com/SuperTux/supertux/files/111039/issue335_recording.txt)
Please tell me if I misunderstood what this bug is...
Interestingly, this makes me notice a whole other problem. Using the editor, edit the level, replace the light block objects with blocks in the tilemap, or rocks, what do you notice...?

PR #1465, opened 2020-07-15
Fixed a bug where you could travel through magic blocks while holding…

Semphriss:
… a lantern of the same color by going fast enough.

I simply removed the delay before the block becomes solid.

When swimming, Tux can gain enough velocity to pass through magic blocks while carrying a lantern of a color that should make them turn solid. Levels that relies on the player's inability to pass a magic wall block while holding a lantern of a certain color can be easily broken by this.

### DIFF
--- a/src/object/magicblock.cpp
+++ b/src/object/magicblock.cpp
@@ -37,7 +37,7 @@ const float MIN_INTENSITY = 0.8f;
 const float ALPHA_SOLID = 0.7f;
 const float ALPHA_NONSOLID = 0.3f;
 const float MIN_SOLIDTIME = 1.0f;
-const float SWITCH_DELAY = 0.06f; /**< seconds to wait for stable conditions until switching solidity */
+const float SWITCH_DELAY = 0.0f; /**< seconds to wait for stable conditions until switching solidity */
 
 } // namespace
 


tobbi:
I think this can be merged. Build failures are unrelated.

Zwatotem:
@tobbi Please, let me test this. I'm afraid that cooldown was there for a reason.

Semphriss:
I thought the same, but I don't seem to encounter any error in normal levels. If there were bugs, they might have been unknowingly fixed since the might-be-bugfix has been added.

I can try to investigate when and why this has been added, and if there are any known bugs related to that.

Semphriss:
Related issue : https://github.com/SuperTux/supertux/issues/335 : It was mentionned that it was to prevent flickering, but there is a minimum lighting time (of 1 second) which already fixes the problem.

Still looking at the commit log for other relevant information.

Semphriss:
I found 2 relevant commits :
- 4dec9ee4e :
> Author: LMH <email removed>
> Date:   Thu Jul 17 20:35:09 2014 -1000
>
>    Reduced the value of SWITCH_DELAY for magic blocks to prevent lanterns (and Tux) from prematurely passing through them while in free-fall
- 5f6321f7860167eb2d4ce6ae3f909eda555725b3
> Author: Christoph Sommer <email removed>
> Date:   Sat Aug 12 16:27:19 2006 +0000
> 
>    Reduced switching delay from 1s to 0.1s

The "feature" was added between 7ddb11ba21077cc2d7dfc2e95cbf1b3736ddbb68 and 5f6321f7860167eb2d4ce6ae3f909eda555725b3. There are no commits between them that affects magicblock.cpp. According to Git, "Reduced switching delay from 1s to 0.1s" actually *added* said delay. (That was back in 2006).

The commit adding the delay-before-switching (which was back then of one full second) doesn't seem to exist, for some reason.

Zwatotem:
I think it was there to prevent flickering in case magic block is on the edge of light source. But this value was lowered several times without complains. Merging.

PR #1466, opened 2020-07-15
Level Improvements Again

Alasdairbugs:
Changes to Bonus Island IV. Minor edits across all levels, most of the stages don't have any issues like 1,2 &3 do :)

- Fixed the Owl skydive duplication causing like 10 skydives to all fall at once causing a very loud noise when loading them in offscreen (This was in 3 stages I think).
- Added Level Target Times to all stages.
- Fixed some enemies that killed themselves (only like a few).
- Removed some unfair spike in one level, it was placed in a way that Tux would take damage from it while no where near it, and also being really hard to see in "Deeper into the mountains"

This isn't many changes since most of the levels were done pretty well (despite personally, some questionable design features).

Also note, some of these stages don't have solidwater so if serano doesnt fix that I can when I plan to revise all the aesthetics for the bonus island stages.


serano01:
Water is already swimable in these levels.

PR #1467, opened 2020-07-15
Add Circular Pathed Platforms

weluvgoatz:
This adds circular-pathed platforms to the game.

Yes, they do follow the same behavior as flames, moving around in a circle.  You can set the radius and speed of their movement as well as a delay for when they begin moving.  You can also change the sprite.

They could possibly be introduced in the ice bridge levels in Icy Island because 1) Those levels introduce no new mechanics and 2) They could "foreshadow" the introduction of the flames by introducing a docile/helpful variant and then showing you the enemy version of it.

This was quite simple of an object to add.  Please test and review, and hopefully this will get merged once it looks all good.


serano01:
Looks good.

src/object/circleplatform.cpp
```diff
@@ -0,0 +1,84 @@
+//  Copyright (C) 2020 Daniel Ward <weluvgoatz@gmail.com>
```
Zwatotem, line 1: My editor shows, that this file has mixed indentation (tabs and spaces) please change all to spaces.

data/images/engine/editor/objects.stoi
```diff
@@ -297,6 +297,9 @@
   (object
     (class "skull_tile")
     (icon "images/objects/skull_tile/skull.png"))
+  (object
+    (class "circleplatform")
+    (icon "images/objects/platforms/icebridge1.png"))
```
Zwatotem, line 302: Please, could you make dedicated icon (potentialy by modifying bicycle-platform one)?

weluvgoatz:
I updated the pull request by updating with your suggestions and requirements @Zwatotem, but I don't understand your problems with the indentation seen as I cannot see the problems you are talking about.

PR #1469, opened 2020-07-18
Add Falling Platform

weluvgoatz:
This adds the falling platform into the game.  The platform will shake when Tux is under or on it and after a short while it will fall.  This lacks a unisolid variant seen as that would be impossible/very difficult in the current way this game is set up, and a unisolid variant would remove the challenge that this obstacle is meant to have.  This comes with sprites for branches and cave terrain.

This object behaves strange with enemies and some objects, but this seems to be a problem with how collision is handled over the whole game.

This PR needs some testing and some review, once it's reviewed and fixed up a bit it can probably get merged.  This definitely needs some work done to it.


data/images/objects/fallblock/cave-5x7.sprite
```diff
@@ -0,0 +1,7 @@
+(supertux-sprite
+  (action
+    (name "normal")
+    (hitbox 0 32 160 224)
+    (images "cave-5x8.png")
```
papb, line 5: I have no idea how this works, but I noticed that in every other `.sprite` you added, this line matches the file name, but not here...

```suggestion
    (images "cave-5x7.png")
```

Or maybe this file should be called `cave-5x8.sprite` instead?

weluvgoatz: It is fixed.

PR #1473, opened 2020-07-18
(WIP) Climbing Adjustments

Alzter:
I adjusted Tux's climbing speed to be faster, and made it where pressing action doesn't take you off climbables since that seems kind of impractical.

I wanted to make it where Tux doesn't fall off climbables when he reaches the edge of the area, but he just isn't allowed to move outside them, but I wasn't sure how to code it. If anyone else could that'd be appreciated!

### DIFF
--- a/src/object/player.cpp
+++ b/src/object/player.cpp
@@ -88,9 +88,9 @@ const float MAX_RUN_XM = 320;
 /** bonus run velocity addition (pixel/s) */
 const float BONUS_RUN_XM = 80;
 /** maximum horizontal climb velocity */
-const float MAX_CLIMB_XM = 96;
+const float MAX_CLIMB_XM = 200;
 /** maximum vertical climb velocity */
-const float MAX_CLIMB_YM = 128;
+const float MAX_CLIMB_YM = 200;
 /** maximum vertical glide velocity */
 const float MAX_GLIDE_YM = 128;
 /** instant velocity when tux starts to walk */
@@ -774,7 +774,7 @@ Player::handle_horizontal_input()
     }
     else {
       // give Tux tighter air control
-      ax *= 2.f;
+      ax *= 1.5f;
     }
   }
 
@@ -1983,11 +1983,6 @@ Player::stop_climbing(Climbable& /*climbable*/)
     m_on_ground_flag = true;
     do_jump(m_player_status.bonus == BonusType::AIR_BONUS ? -540 : -480);
   }
-  else if (m_controller->hold(Control::UP)) {
-    m_on_ground_flag = true;
-    // TODO: This won't help. Why?
-    do_jump(-300);
-  }
 }
 
 void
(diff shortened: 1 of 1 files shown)


src/object/player.cpp
```diff
@@ -774,7 +774,7 @@ Player::handle_horizontal_input()
     }
     else {
       // give Tux tighter air control
-      ax *= 2.f;
+      ax *= 1.5f;
```
Zwatotem, line 777: This seems to be unrelated to climbing.

Alzter: > This seems to be unrelated to climbing.

Oops, I was messing with Tux's velocity and forgot to reset it.

Alasdairbugs:
Action should be allowed to eject tux from a ladder... otherwise how else is he going to get off the ladder if he's climbing halfway through? move over to the side until he falls off??? No, that's way too slow.

If you're going to remove action taking tux off the ladder I suggest replacing it with down, since you're never going to be ground-pounding or ducking on a ladder, it makes sense.

Zwatotem:
My plan was to make `JUMP+DOWN` work as `ACTION`, so you still can let off without turning in any direction. Additionally `JUMP + any direction combination` (including diagonals) should send Tux jumping off the ladder in that direction.

Alzter:
> If you're going to remove action taking tux off the ladder I suggest replacing it with down, since you're never going to be ground-pounding or ducking on a ladder, it makes sense.

But isn't pressing down needed to climb down the ladder?

Ordoviz:
> But isn't pressing down needed to climb down the ladder?

It would be <kbd>Down</kbd> + <kbd>Jump</kbd>.

Alzter:
> > But isn't pressing down needed to climb down the ladder?
> 
> It would be Down + Jump.

I know, but that wasn't what they said. I don't understand why it can't just be jump

Semphriss:
I can look into keeping Tux on the ladders, if help is still needed about it.

Guih48:
Yes, I think too: the current climbable "climb-off" behaviour is wery irritating. 

At the edges of climbables, Tux must stop, and don't take off without direct order (jump (or action) buttons) but the "climb-off" isn't fully a bad behaviour, it can be good with just one improvement: 
You automatically stop and the edges, and if you want to climb-off, you just release and press again the left, the right or the down key (not the up, because it's also could make irritating situations) (of course, if there is a safe ground at the climbable's bottom, if you touch the bottom of the climbable, you automatically "climb-off" to the ground).

For take-off from the climbable I agree with @Zwatotem 's idea(if we can't agree with the current action key method):

> My plan was to make `JUMP+DOWN` work as `ACTION`, so you still can let off without turning in any direction. Additionally `JUMP + any direction combination` (including diagonals) should send Tux jumping off the ladder in that direction.

I think, this is the best, if we wouldn't use action for jumping off (and ground-pounding isn't a problem, because it needs falling time to gain power, and if you don't want to gtound-pound, you just release the down key.

weluvgoatz:
Closing because these changes seem to not be planned or worked on, and even if they were, would likely be included in a much more expansive PR anyway.

PR #1474, opened 2020-07-18
Level Aesthetic Improvements

Alasdairbugs:
Bonus Island 1 Changes. Aesthetic update.

- Reworked some of the snow background stuff, there were some hideous backgrounds using only 3 tiles where there are like an option of 10. Made it look more natural, random, and irregular.
- Added the bumpy snow tiles to create more texture in some levels, as well as the bumpy snow ceiling tiles.
- Reworked the backgrounds of 2 levels.
- Added snow onto almost everything where it made sense.
- Reworked levels that were clearly snowy but didn't have their snowy block equivalent (for example, "Lies!").
- Changed some of the physical ground in "Cave Of Mirrors" due to some awkward and impossible jumps due to re-tiling.
- Stopped some more enemies from killing themselves.
- Re-tiled all levels, featuring the new snowy castle tiles.

And also, since Hume upgraded the levels to milestone 2 and put his signature in "Too Easy", I put mine and Serano's signature in the secret of "Bad Guys Stink!" at the end. Feel free to change this as you wish, I don't mind.


serano01:
Very good work! Alzter will be pleased that his castle tiles will finally be used.

PR #1475, opened 2020-07-18
Story level improvements

Alasdairbugs:
I made a level called "Learning To Fly", which, in my opinion, is more fun than "Above the Arctic Skies" Here is my reasoning:

- The terrain is more natural (Above the Arctic Skies had some rather blocky areas, particularly where the secret was).
- The terrain now has better height variation, and distinguishable features and sections of different types of platforms and solids (unisolids and normal solids).
- The level is less cluttered with badguys, less badguy dense, which is beneficial for a level where the main Power-Up is an air flower, something that is actually quite finicky to use, you move more so the player needs more time to react to enemies, and so more spread out badguys is better.
- The Platform sections are more clear and also require player interaction to make work (Button).
- The level features better secrets that both require the use of the air flower, making the air flower a required Power-Up to get the secret. An example of this can be found in "Ice in the Hole" where the fire flower is needed to access a secret, so I think this adds more flavour and variety to the stage.
- The level has more collectables that are spread out in a more fun way, making a collect-a-thon more enjoyable.
- There are less awkward jumps and platforms in my level than the old one, and the platforms have more distinct 'boarding/docking stations'.

I sent this level to Rusty a while ago. I took his feedback and put it into this stage, which I think at this point is pretty much as good as it's going to get.

I have not removed the level from the world1 file so you can chose what to do with the level, whether to keep that, delete it, keep mine or delete mine. I look forward to hearing your response.


busybody42:
I think that you should make it a level that comes before above the arctic skies.
It fits the naming, after all...

serano01:
Good work, but we came to the conclusion that "Above The Arctic Skies" won't be replaced.

PR #1476, opened 2020-07-18
Story level aesthetics

Alasdairbugs:
Story Icy Island Improvements.

- Added new bumpy snow tiles everywhere relevant, as well as bumpy ceiling snow tiles.
- Fixed the ugly spikes and replaced them with more sophisticated spikes that are 7/8ths of a tile higher than they should, giving them a distinguished view from the ground, also making the snow cover them instead of being behind them, as well as being behind tux.
- Improvements to "Welcome to Antarctica" where I added a pit, so the player knows what they are by indicating a "!" sign telling them it is bad, also changed the coins where the sign is slightly, also added a slope earlier in the level to introduce the fact that slopes exist (they weren't anywhere in levels until "A Fork in the Road" which is kind of silly). I also made the Power-Up not progressive as of someone else's suggestion, and I personally agree with the sentiment.
- Removed the impossible coins in "Entrance to the Cave"

All of these changes took about 8 hours. There should be some more changes needed to the cutscenes, with the new tiles in mind.


serano01:
While I personaly like those changes, it is planed by RustyBox and weluvgoatz to completely rework the levels of Icy Island, which will therefore turn your changes obsolete. But thanks for your PRs!

PR #1478, opened 2020-07-21
Add sideways bumpers

weluvgoatz:
This PR adds sideways bumpers to the game.  These will bounce Tux to the side stronger than he can jump, as well as up a little bit.  They do not obey gravity, but there is some unique behavior that they will stick to the side of falling blocks and will obey their falling as well.  Tux can always use them, he doesn't have to line up with them.  Also, they do not move.  They don't have collision with other objects at the moment.  They also cannot be picked up.

Due to how different they behave to normal trampolines, they are classified as different objects, but they still use similar sprites.

Note:  Github for some reason messed with some lines in fallblock.cpp, but I don't understand what those changes were since no changes seem to show up.

Please test out this PR and hopefully it can soon get merged!!! These objects are awesome!!!


AndrePinheiroPT:
FINALLY!

weluvgoatz:
I made a video showcasing this: https://youtu.be/lZhHQ5oztD0

papb:
Awesome!!

data/images/objects/trampoline/right_trampoline.sprite
```diff
@@ -0,0 +1,19 @@
...
+    (name "swinging")
+    (fps 24)
+    (hitbox 0 2 18 27)
+    (images "right-0.png"
+            "right-1.png"
+            "right-2.png"
+            "right-3.png"
+            "right-4.png"
+            "right-5.png"
+    )
```
HybridDog, line 17: Is it possible to mirror the left images?
This would avoid the (almost) duplicate image files.

weluvgoatz: I've fixed it as of the latest few commits.

HybridDog: thanks

PR #1479, opened 2020-07-21
Discord integration (Rich Presence)

Semphriss:
Integrating the Discord Rich Presence feature into SuperTux!

Screenshot from 2020-07-21 18-49-00

- Disabled by default, you can enable it at the bottom of the Options menu ("Enable Discord RPC").
- If you have Discord open while you are playing, and you allow Discord to display which game you are playing, SuperTux will display your in-game status on your profile.
- Shows if you're in the menu, in a worldmap (including the worldmap name), in a level (including the level name and its corresponding worldmap), or if you are editing a worldmap/level (will show names as well).
- Checking/unchecking the option to enable/disable Discord RPC is effective immediately, no need to restart the game or go back to the menu. Useful if you want to work on a secet level :)

It should work at least with Windows, MacOS and Linux. Tested efficiently on Linux and quickly on Windows. Android/others should be effectless.


Semphriss:
The build errors are me not having marked some overriding functions as "override" and leaving the default destructor in my C++ classes. I'll fix that right away, if anyone sees anything else, please let me know!

Semphriss:
The build fails because there is some Discord-related code that produces warnings, which are interpreted as errors by Travis. I think I'd need to move the Discord-related code as a seperate external library, but there's quite some stuff to do and my motivation doesn't go that far for today.

Semphriss:
Moving the Discord code to the "external" folder turned out much easier than expected, however the compiler seems obstinated at throwing the whole thing out the window because it doesn't like the syntax style of the Discord code.

Is there anywhere where I can put the Discord library so that Travis won't come back and scream at the implicit casts?

Semphriss:
Oh, and now there are other commits that sneaked their way in my branch. Now my PR is infected with someone else's stuff.

I'm closing this. The branch is polluted with irrelevant stuff and I'm not motivated anymore. I'll see that later.

PR #1481, opened 2020-07-22
Fix animation issue with slow framerates

### LINKED ISSUE #1453: Sprites with 1 loop don't work in laggy zones
The loops attribute of sprites allows for an action to be played only once (like the buttjump or jumpy) and it works fine normally, but in laggy areas of the game this seems to be thrown out the window and they loop indefinitely, making things like buttjump and jumpy look super weird.  If someone can fix this that'd be great.  I'm on Windows 10 and playing 0.6.2 nightly.

martinvanzijl:
Fixes #1453. Sprites with 1 animation loop now stop as expected, when playing with low frame-rates. I.e. they don't continue forever.

Demo below, recorded using the `export LIBGL_ALWAYS_SOFTWARE=true` as in https://github.com/SuperTux/supertux/issues/1453#issuecomment-654967674.

animation-fix-demo

### DIFF
--- a/src/sprite/sprite.cpp
+++ b/src/sprite/sprite.cpp
@@ -116,12 +116,9 @@ Sprite::update()
     m_frameidx++;
   }
 
-  while (m_frameidx >= get_frames()) {
+  while (m_frameidx >= get_frames() && !animation_done()) {
     m_frameidx -= get_frames();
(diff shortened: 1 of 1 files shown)


Zwatotem:
That resolves the issue, thank you for your contribution.

PR #1486, opened 2020-07-31
Use ordinary Makefile

### LINKED ISSUE #1485: CMake is a mess
**SuperTux version:**
0.6.2
**System information:**
Linux

The CMakeLists.txt is very messy, with 1000+ lines of odd CMake syntax.
The generated Makefile is 13000+ lines long.

Perhaps use a normal Makefile?
Portability may be an issue, but people who can be bothered to build Supertux by themselves probably won't be troubled by the need to install gmake.

CMake Makefile builds took 5:32
My handwritten Makefile builds took 6:13
The difference should be negligible.
Both were run as `make -j4`

erratic-c-programmer:
Fixes #1485 with a readable Makefile


config.h
```diff
@@ -0,0 +1,24 @@
...
+
+#define ICONV_CONST
+
+#define HAVE_OPENGL
+
+#define HAVE_LIBCURL
+
+#define BUILD_DATA_DIR "/home/isaac/devel/supertux/data"
+
+#define BUILD_CONFIG_DATA_DIR "/home/isaac/devel/supertux/data"
```
maxteufel, line 22: You can't simply assume that `/home/isaac` exists.

targets.mk
```diff
@@ -0,0 +1,109 @@
...
+EXT_FINDLOCALE_DIR = $(EXTERNDIR)/findlocale
+EXT_SDL_SAVEPNG_DIR = $(EXTERNDIR)/SDL_SavePNG
+EXT_SDL_TTF_DIR = $(EXTERNDIR)/SDL_ttf
+EXT_SEXP_DIR = $(EXTERNDIR)/sexp-cpp
+EXT_SQUIRREL_DIR = $(EXTERNDIR)/squirrel
+EXT_TINYGETTEXT_DIR = $(EXTERNDIR)/tinygettext
+# =================================
+# Commands and flags
+# Compiler flags
+INCLUDE_PATH = -I/usr/include/SDL2 -I/usr/include/AL -I. -Isrc -I$(EXTERNDIR)/sexp-cpp/include -I$(EXTERNDIR)/tinygettext/build/ -I$(EXTERNDIR)/tinygettext/include/ -I$(EXTERNDIR)/squirrel/include/ -Iexternal/findlocale -Iexternal/SDL_SavePNG -Iexternal/SDL_ttf/external/freetype-2.4.12/include
```
maxteufel, line 37: You can't simply assume that SDL is in `/usr/include/SDL2`, …

targets.mk
```diff
@@ -0,0 +1,109 @@
...
+.PHONY : build_extlibs all clean
+
+all : supertux2
+
+supertux2 : $(SRCDIR)/main.o libsupertux2.a
+	$V printf "Linking final binary \033[1m$@\033[0m...\n"
+	$V $(CXX) $(CXXFLAGS) -o $@ $(LINKDIRS) \
+		$< \
+		-lsupertux2 -lsquirrel_static -lsqstdlib_static -ltinygettext -lsexp -lSDL2_ttf -lsavepng -lfindlocale \
+		-lGL -lGLEW -lphysfs -lSDL2 -lSDL2_image -lopenal -lvorbisfile -lcurl -lfreetype -lboost_filesystem -lboost_locale -lpng
```
maxteufel, line 63: You can't simply assume those are the library names to be linked with.

maxteufel:
Using a build system like CMake ensures portability and makes it possible to compile SuperTux on a variety of setups:

- Versions with debug symbols can easily be built, so can optimized release builds be done quickly
- We use CMake to find the location of the libraries we depend on, which can greatly vary on Linux distributions
- CMake makes it easy to change the GL bindings from GLEW to glbinding
- CMake makes it possible to build SuperTux on Windows using VisualStudio
- CMake (and CPack) enables automated packaging in particular for Windows nightly builds

and more stuff. The proper solution would be to structure the CMake file and clean up/remove unused parts.

PR #1490, opened 2020-08-03
Dispensers can obey gravity

weluvgoatz:
This adds a small option for dispensers to obey gravity.  With this, you can have moving dispensers by toggling gravity and putting them on a moving platform.  Enjoy.

### DIFF
--- a/src/badguy/dispenser.cpp
+++ b/src/badguy/dispenser.cpp
@@ -71,6 +71,7 @@ Dispenser::Dispenser(const ReaderMapping& reader) :
   m_swivel(false),
   m_broken(false),
   m_random(),
+  m_gravity(),
   m_type(),
   m_type_str(),
   m_limit_dispensed_badguys(),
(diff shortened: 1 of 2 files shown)


src/badguy/dispenser.cpp
```diff
@@ -224,8 +226,13 @@ Dispenser::collision(GameObject& other, const CollisionHit& hit)
 }
 
 void
-Dispenser::active_update(float )
+Dispenser::active_update(float dt_sec)
 {
+  if (m_gravity)
+  {
+    BadGuy::active_update(dt_sec);
+    m_physic.enable_gravity(true);
```
tobbi, line 234: Does `Badguy::active_update` disable gravity again?

weluvgoatz: Yes, it does.

tobbi: Yeah, I know, but is it necessary to call `enable_gravity` again after calling `active_update`?

weluvgoatz: It appears not, so it's been removed.

PR #1502, opened 2020-08-13
Update Russian translation for world2

bergentroll:
Add missing messages from cutscenes.

Fix a couple of misprints.

Update Timings a bit for tower cutscene (related to all translations).


Zwatotem:
Sorry to say that, we don't support translation development here on Github. We use [Transifex](https://www.transifex.com/) instead. If you want these to be applied, please create an account there, and apply to translate [this project] (https://www.transifex.com/arctic-games/supertux/)
I think that cutscene wait times can still be applied here, if you exclude the rest.
If you have any questions feel free to ask.

bergentroll:
I sent join request on Transifex two weeks ago, but got no response. Nevermind.

Issue #401, opened 2016-04-10
Moving platforms - behaviour bug


Narre:
Tux's behaviour is different on object platforms, moving solid tiles, and moving unisolid tiles. It appears in both 0.4.0 and current version on branch master

Object platforms work fine, but moving tilemaps cause strange jumping behaviour of Tux.

https://youtu.be/bye-VqU7wGE

HybridDog:
I think the second one looks best.

tobbi:
I'm not sure if you still have the example level for this bug, but it would be great if you do, @Narre

Narre:
Sorry for the late reply,
no, I don't have it anymore but I've quickly created another one so you can try for yourself, @tobbi 
[platform_test.zip](https://github.com/SuperTux/supertux/files/4466715/platform_test.zip)

The difference between the object and solid tiles is negligible, but the unisolid platforms should probably be looked into.

tobbi:
Fixed by #1514

PR #1507, opened 2020-08-14
Support for easing on any path object

Semphriss:
Instead of the mandatory flat linear movement, platforms and tilemaps can now move at varying speed more easily with easing!

Demo : https://youtu.be/gmWf3KeeiSs

supertux_easing


mbernard2:
Everything works fine with the "platform" object, but not with moving tilemaps.

There are a few problems when Tux in standing on a downwards-accelerating platform (as a tilemap). Sometimes, Tux tends to move "into" the platform, and eventually either jumps or dies. **However**, this is a known issue regarding moving platforms, and is related to issue #401 on which I've been working lately. In other words: it doesn't have to be fixed in this PR.

There's just one comment I would make: I suggest you remove the "elastic" mode, because Tux either goes right through it or gets killed by it (when it's slower), so it's pretty useless:

 &nbsp;&nbsp;  

Apart from that, I made recordings of the other bugs I've seen, but again, the fundamental problem here is the collision mechanism. I'm about to write a fix for that and submit a PR soon. I don't know if it should be merged in the meantime (as it doesn't affect platform objects), or if we should wait for #401 to be fixed. @tobbi @weluvgoatz,  what do you guys thinks ?

But anyway, here's what went wrong:

- Tux getting "into" the platform by a few pixels, than jumping by itself:
  

- Tux getting killed by a platform that moves down too fast:
  

- Same as before, but this time the platform is moving up:
  

- Same as the first, but Tux is on a slowly accelerating platform:

Semphriss:
Ah, that's problematic.

I'd still like to retain easing as it can be used correctly if, for example, it is used on non-solid tilemaps or on solid tilemaps on which Tux can't be standing when it moves (like a tile-based door that only moves during a scripted animation).

But I get your point, it can easily be misused and it's not necessarily obvious to an unexperienced or unaware level designer, and it can be a very frustrating bug to encounter as a player.

I can make a few temporary changes while the bug is being fixed :
- Show a warning dialog to the user when changing the easing mode of a tilemap to anything else than "No easing"
- Only show the most common easing modes (those that are known for not causing problems) and add an option to show all of them; when the player would activate it, it'd display a warning instructing the designer to use those easings with care.
- Temporarily forbid easing on solid tilemaps (and display a warning when setting an easing for a solid tilemap)

Would any of them be useful?

Also, if you want some help to fix that bug, I'm all in :)

tobbi:
You forgot one option:

4) Changing our collision detection code so it works with the changes. Have fun with that ;)

weluvgoatz:
I would prefer if all options were available for usage, on all path objects.

Semphriss:
> You forgot one option:
> 
>     1. Changing our collision detection code so it works with the changes. Have fun with that ;)

@mbernard2 said he was working on it; Imma wait for his reply, so we don't do the same fix twice :)

maxteufel:
I think the included files are "small" enough to be put in an appropriate subdirectory of `src/` (perhaps they should go into `src/math/`) rather than `external/`, similar to how we have `src/addon/md5.[ch]pp`.

Semphriss:
That's true, plus the files have been heavily modified, and I never managed to find how to include them without typing "#include "../external..."" so it would be good if I moved it to src/math. I'll ask around a bit, and if people approve it I'll do that!

src/math/easing.hpp
```diff
@@ -0,0 +1,102 @@
...
+double ElasticEaseInOut(double p);
+double BackEaseIn(double p);
+double BackEaseOut(double p);
+double BackEaseInOut(double p);
+double BounceEaseIn(double p);
+double BounceEaseOut(double p);
+double BounceEaseInOut(double p);
+
+typedef double(*easing)(double);
+easing getEasingByName(EasingMode ease_type);
```
tobbi, line 95: change the signature to getEasingByName(`const EasingMode& easy_type)`

src/math/easing.hpp
```diff
@@ -0,0 +1,102 @@
...
+const char* getEasingName(EasingMode ease_type);
```
tobbi, line 96: Here accordingly.

src/math/easing.hpp
```diff
@@ -0,0 +1,102 @@
...
+
+}
+
+EasingMode EasingMode_from_string(std::string ease_name);
```
tobbi, line 100: ...and here as well.

Semphriss: I corrected the signatures in easing.cpp and easing.hpp, GCC told me nothing else had to be changed (it didn't print any warning). I'll wait for the current build to finish, so that if there's any other error clang finds, I'll correct it in the same commit to save some time.

PR #1508, opened 2020-08-15
Added fading gradients through Squirrel

Semphriss:
Added four functions through Squirrel to fade colors over time : 
```c++
// In addition to set_colo1 and set_color2, set_colors allows you to set both colors in one function
set_colors(float red, float green, float blue, float time);

fade_color1(float red, float green, float blue, float time);
fade_color2(float red, float green, float blue, float time);
fade_colors(float red1, float green1, float blue1, float red2, float green2, float blue2, float time);
```
supertux_fading_gradients

### DIFF
--- a/src/object/gradient.cpp
+++ b/src/object/gradient.cpp
@@ -45,7 +45,13 @@ Gradient::Gradient(const ReaderMapping& reader) :
   m_gradient_bottom(),
   m_gradient_direction(),
   m_blend(),
-  m_target(DrawingTarget::COLORMAP)
+  m_target(DrawingTarget::COLORMAP),
+  m_start_gradient_top(),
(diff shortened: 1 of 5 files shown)


Semphriss:
**Note** : I'd like someone to review my coding style, if possible; in gradient.hpp I declared object properties as private and starting with "m_"; is that the correct/recommended way of proceeding? How is it best to do?

tobbi:
I thought we already had that? I'm a bit confused.

Semphriss:
> I thought we already had that? I'm a bit confused.
@tobbi 

There was an option to change the color instantly (`set_color`), but not to fade progressively over time. This PR adds this small missing feature :)

Issue #1492, opened 2020-08-03
Hebrew letters didn't appear [status: awaiting merge]


omeritzics:
**SuperTux version: v0.6.2**
**PC information: A 64-bit PC running Windows 10 and Linux.**

##### Expected behavior
The Hebrew and English/Latin characters both are displayed well.
##### Actual behavior
The Hebrew characters are not displayed in language auto-detect, And the English/Latin characters are not displayed when choosing Hebrew (non auto-detect).

*I found a better Hebrew supporting font than Yehuda and Noto Sans, and it supports English/Latin characters. Link: http://hebrew-font-shuneet.com/font/index.html.

**Ideal steps to reproduce actual behavior**
1. Merge the Hebrew font (Shuneet or Yehuda CLM) into SuperTux-Medium.ttf.
2. Modify resources.cpp.

**Alternative steps to reproduce actual behavior**
*There is an pending pull request with theese steps. Link: #1520.
1. Use Shuneet font instead of Yehuda CLM.
2. Modify resources.cpp if needed.

omeritzics:
I submitted a pull request to fix this issue, please merge it.
Link: https://github.com/SuperTux/supertux/pull/1495

Zwatotem:
@omeritzics Does https://github.com/SuperTux/supertux/commit/583adad770ae2fb86ee7d28bc6e2f1f8f03bced3 solve the issue listed?
Note, that some texts in Hebrew are missing due to incomplete translation. We rely on community translating the game to different languages.

omeritzics:
I don't think this solve the issue. I recommend to merge Yehuda CLM into SuperTux-Medium.ttf (as in PR https://github.com/SuperTux/supertux/pull/1495) and remove the return to the Hebrew font in the resources file.

omeritzics:
Hello, FontForge crashes almost every time I open SuperTux-Medium.ttf, and I found bugs while trying to merge Hebrew into SuperTux-Medium. Is there another way to merge it?

omeritzics:
Hello, pull request #1520 may solve the issue, Please check it out.

Issue #1383, opened 2020-04-20
Clarify license of fonts


vpzomtrrfrt:
data/AUTHORS contains information about the licensing of various data files, but I don't see anything about the font files in the repository. What are the licenses of those?

omeritzics:
Hello, The license of Noto Sans font is SIL Open Font License, and the license of Yehuda CLM is GNU GPL 2.

tobbi:
Closed by #1520

tobbi:
Re-opened as per personal request of @Zwatotem Sorry, there was a miscommunication or rather no communication.

omeritzics:
Hello, the license file of Noto Sans is already in the fonts folder and under SIL 1.1 license. Shuneet font is under SIL 1.1 license too. The license of Noto Sans is in the downloaded compressed folder, and there is a link to the license of Shuneet in the website of the font.

huntekye:
What's left for this issue? The only font files I see left without licenses are Roboto-Regular.ttf and SuperTux-Medium.ttf . I suppose that SuperTux-Regular doesn't need a license file, and Roboto's license is here: [https://fonts.google.com/specimen/Roboto/license]( https://fonts.google.com/specimen/Roboto/license). (It's another SIL v1.1)
If it's just the Roboto license, I'll make a PR to add it; let me know.

Alasdairbugs:
@huntekye if you could make a PR for that it'd be appreciated

PR #1520, opened 2020-08-27
Fix issue #1492 with the alternative steps + Fix issue #1383

omeritzics:
Hello, this pull request solves issue #1492 with the alternative steps and solves issue #1383.


src/supertux/resources.cpp
```diff
@@ -92,8 +92,8 @@ Resources::get_font_for_locale(const std::string& locale)
   if(locale == "cmn" || locale == "ja" || locale == "zh_CN" || locale == "zh_TW")
     return "fonts/NotoSansCJKjp-Medium.otf";
   if(locale == "he")
-    return "fonts/Yehuda CLM Bold.ttf";
-  return "fonts/SuperTux-Medium.ttf";
+    return "fonts/shuneet3-medium.otf";
+  return "fonts/Supertux-Medium.ttf";
```
Zwatotem, line 96: This line has to stay here. It does not mean "set Supertux-Medium for Hebrew", it means  "set Supertux-Medium for all other languages"

Zwatotem (approved):

src/supertux/resources.cpp
tobbi, line 96: Please keep the casing of that string exactly like it was before: "fonts/SuperTux-Medium.ttf";

Otherwise, this is good to be merged.

Zwatotem: Thanks, didn't notice that.

Zwatotem:
Won't set #1383 as fixed yet; it requires more care (licenses for other fonts as well).
@omeritzics Thanks for your contribution.

omeritzics:
Hello, I took a look in the license files again, there is no issue with the font licenses.

Issue #1523, opened 2020-08-30
Problems with the Hebrew font


omeritzics:
#1681  fixes this issue, please check it out.

omeritzics:
Pull request https://github.com/SuperTux/supertux/pull/1524 may solve this issue, please check it out.

Zwatotem:
Does https://github.com/SuperTux/supertux/commit/e6094875d346157d5169aeb7864f96b12fb851b4 solve this issue? 
Please test, by downloading appropriate version from https://download.supertux.org/ (auto-locale-font branch). I can't test myself, since none of my devices are in Hebrew.

omeritzics:
Hello, this doesn't solve the issue. Replacing `    return "fonts/shuneet3-medium.otf";
  return "fonts/SuperTux-Medium.ttf"; with `    return "fonts/shuneet3-medium.otf";
 } 
else
{
 return "fonts/SuperTux-Medium.ttf";`
may solve it.

omeritzics:
Hello, https://github.com/SuperTux/supertux/commit/1ef369a0a6cf815b03231efa09d3c10943f48679 solved this issue, but there is a problem with Shuneet in SuperTux.

omeritzics:
> Hello, I tried the last nightly build and the problem is solved, but there is a problem with Shuneet in SuperTux somehow.

This issue is back.

omeritzics:
Hello, pull request solves this issue and also adding the Devanagari letters to SuperTux Medium. Link: #1534

omeritzics:
Hello, PR https://github.com/SuperTux/supertux/pull/1534 is waiting for review, and it fixes a missing-glyph problem in Devanagari too.

omeritzics:
PR #1681 is waiting for review and provides a better solution for Hebrew and Devanagari by replacing the fonts.

PR #1524, opened 2020-08-30
Fix issue #1523 with option 1

omeritzics:


tobbi:
If we return before the switch, SuperTux-Medium.ttf will be returned every time. Can you check what auto-detect returns on your system, e.g. by adding a ` log_debug << "some string here " << some_variable << std::endl;`?

omeritzics:
Hello, how to check it?

Zwatotem:
> If we return before the switch, SuperTux-Medium.ttf will be returned every time. Can you check what auto-detect returns on your system, e.g. by adding a ` log_debug << "some string here " << some_variable << std::endl;`?

In this case _some_variable_ should be `std::locale().name()`. This string is supplied by system, so I'm concerned it can be different system to system, even for the same locale, so for exaple it can be "he" on some systems, and "he_IL" on others.

So the thing, that should be passed to this function, should be `std::locale().name()` in case auto locale is set.  
I recommend changing all the `==` comparisons in this font seletion code with `string::find()` for broader selection of possible values.

omeritzics:
Hello, this pull request moved to [here](https://github.com/SuperTux/supertux/pull/1525).

PR #1528, opened 2020-09-05
Layout widget in editor is now scrollable

Semphriss:
Having dozens of tilemaps and other sector elements is no longer a pain to manage - now you can scroll the toolbar and access those sneaky little icons!

----------

> Urgh! I can't access all my tilemaps! They're going off-screen!

Screenshot from 2020-09-05 01-19-34

> Ah, much better.

Screenshot from 2020-09-05 01-19-47

### DIFF
--- a/src/editor/layer_icon.cpp
+++ b/src/editor/layer_icon.cpp
@@ -65,9 +65,7 @@ LayerIcon::draw(DrawingContext& context, const Vector& pos, int pixels_shown)
   ObjectIcon::draw(context, pos, pixels_shown);
   int l = get_zpos();
   if (l != std::numeric_limits<int>::min()) {
-    //context.color().draw_text(Resources::small_font, std::to_string(l),
-    //                            pos + Vector(16,16),
-    //                            ALIGN_CENTER, LAYER_GUI, ColorScheme::Menu::default_color);
+    // Don't draw the text if the icon is not 100% visible
     if (TileMap* tilemap = dynamic_cast<TileMap*>(m_layer)) {
       if (tilemap->m_editor_active) {
         context.color().draw_surface(m_selection, pos, LAYER_GUI - 1);
--- a/src/editor/layer_icon.cpp
+++ b/src/editor/layer_icon.cpp
@@ -57,6 +57,25 @@ LayerIcon::draw(DrawingContext& context, const Vector& pos)
   }
 }
 
+void
+LayerIcon::draw(DrawingContext& context, const Vector& pos, int pixels_shown)
+{
+  if (!is_valid()) return;
+
+  ObjectIcon::draw(context, pos, pixels_shown);
+  int l = get_zpos();
+  if (l != std::numeric_limits<int>::min()) {
+    //context.color().draw_text(Resources::small_font, std::to_string(l),
+    //                            pos + Vector(16,16),
+    //                            ALIGN_CENTER, LAYER_GUI, ColorScheme::Menu::default_color);
+    if (TileMap* tilemap = dynamic_cast<TileMap*>(m_layer)) {
+      if (tilemap->m_editor_active) {
+        context.color().draw_surface(m_selection, pos, LAYER_GUI - 1);
+      }
+    }
+  }
+}
+
 int
 LayerIcon::get_zpos() const
 {
--- a/src/editor/layers_widget.cpp
+++ b/src/editor/layers_widget.cpp
@@ -137,7 +137,7 @@ EditorLayersWidget::update(float dt_sec)
   {
     m_scroll -= 2;
   }
-  else if (m_scroll_speed > 0 && m_scroll < static_cast<int>(m_layer_icons.size()) * 35)
+  else if (m_scroll_speed > 0 && m_scroll < (static_cast<int>(m_layer_icons.size()) - 1) * 35)
   {
     m_scroll += 2;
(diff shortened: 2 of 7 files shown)


serano01:
Awesome, especially for the forest worldmap, since it has so many tilesets.

src/editor/layer_icon.cpp
```diff
@@ -57,6 +57,25 @@ LayerIcon::draw(DrawingContext& context, const Vector& pos)
...
+LayerIcon::draw(DrawingContext& context, const Vector& pos, int pixels_shown)
+{
+  if (!is_valid()) return;
+
+  ObjectIcon::draw(context, pos, pixels_shown);
+  int l = get_zpos();
+  if (l != std::numeric_limits<int>::min()) {
+    //context.color().draw_text(Resources::small_font, std::to_string(l),
+    //                            pos + Vector(16,16),
+    //                            ALIGN_CENTER, LAYER_GUI, ColorScheme::Menu::default_color);
```
Zwatotem, line 70: Looks like there's some leftover testing code here

src/editor/object_icon.cpp
```diff
@@ -68,4 +68,16 @@ ObjectIcon::draw(DrawingContext& context, const Vector& pos)
...
 
+void
+ObjectIcon::draw(DrawingContext& context, const Vector& pos, int pixels_shown)
+{
+  auto cropped_surface = m_surface->region(Rect(32-pixels_shown,
+                                                0,
+                                                32,
+                                                32));
+  context.color().draw_surface_scaled(cropped_surface,
+                                      Rectf(pos + Vector(32-pixels_shown,0) + m_offset, pos + Vector(32,32) - m_offset),
```
Zwatotem, line 79: Operator spacing is inconsistent here and in some other places. Please adjust it to be always spaced (except commas of course).

Zwatotem (commented):
Looks good, but I would block it from further scrolling if there's only one icon left on the screen. Currently you can flush them all to the left.
Btw, I expected mouse scrolling, by the title, maybe I expected too much 😆.

tobbi:
Approval from me once the review comments were addressed.

tobbi:
Oh, and check the travis warnings and fix them.

Semphriss:
I'm gonna fix all of this, I'd just have one question :

> Btw, I expected mouse scrolling, by the title, maybe I expected too much.

I'm not sure to understand... It scrolls by hovering the first or last 32 pixels of the widget bar, isn't that the way it should work? I wanted to make it work the same way as the tilesets widget. Do you want me to make it another way? What's the method you originally understood from the title?

src/editor/layer_icon.cpp
Semphriss: Dang, I though I'd catch 'em all with a `git diff`... Gonna review the whole thing :laughing:

Zwatotem:
Sorry, my comment didn't make much sense. What I had in mind was mouse wheel scrolling, just like the tile selection panel does, but horizontally.

Semphriss:
Oh, okay. Should I also make it scrollable with mouse wheel?

Zwatotem:
If it isn't much of a deal for you then sure.

Semphriss:
I'll look into this; I'm pushing my first fix batch in a few moments, I'll use that first batch to see if there are any other Travis errors.

I'll add the mouse wheel scroll (if doable in less than an hour :laughing:) in the second next commit, along with any other Travis errors if applicable.

Semphriss:
Alright, I couldn't resist and I did everything in a single commit.

I do have a problem, though; scrolling with the mouse wheel also triggers the editor behind the widget. It seems like my `return true;` in the mouse wheel function isn't enough to prevent propagation.

Semphriss:
Nevermind, it's a code 18. It works fine now, and I fixed all the requested changes; I'll commit now.

src/editor/layers_widget.cpp
```diff
@@ -108,7 +110,11 @@ EditorLayersWidget::draw(DrawingContext& context)
   int pos = 0;
   for (const auto& layer_icon : m_layer_icons) {
     if (layer_icon->is_valid()) {
-      layer_icon->draw(context, get_layer_coords(pos));
+      if (pos*35 >= m_scroll) {
+        layer_icon->draw(context, get_layer_coords(pos));
+      } else if ((pos+1)*35 >= m_scroll) {
+        layer_icon->draw(context, get_layer_coords(pos), 35 - (m_scroll - pos*35));
+      }
```
Zwatotem, line 117: Operators still glued to operands here...

src/editor/object_icon.cpp
```diff
@@ -65,7 +65,26 @@ void
...
+                                        pos + m_offset,
+                                        pos + Vector(32, 32) - m_offset
+                                      ),
+                                      LAYER_GUI - 9);
+}
+
+void
+ObjectIcon::draw(DrawingContext& context, const Vector& pos, int pixels_shown)
+{
+  auto cropped_surface = m_surface->region(Rect(32-pixels_shown,
```
Zwatotem, line 78: ...and here

Zwatotem (commented):
There is an unintended behaviour in that new feature, the bar keeps scrolling, even when I take the mouse.
See video:
https://youtu.be/BV5S4DYmnbU

PR #1535, opened 2020-09-10
(Fixes #1510) Seperated BadGuy classes between custom-scripted and non-scripted

### LINKED ISSUE #1510: Will-O-Wisp not exposed to scripting anymore
**SuperTux version:** since d5a5b62ebfd0f8250cac2432c7f59dc44dbdb4bd (found with `git bisect`)

##### Expected behavior

Will-O-Wisps are scriptable.

##### Actual behavior

    Error running script: Squirrel error: Couldn't start script (the index 'willy' does not exist)
    [SCRIPTING] AN ERROR HAS OCCURRED [the index 'willy' does not exist]

Place a Will'O'Wisp and call it _willy_. Open the [console][1] and type `sector.willy`.

[1]: https://github.com/SuperTux/supertux/wiki/Console

Semphriss:
Fixes #1510.

I have to do the same for another PR I'll make, so I thought I'd train myself with a basic fix :)

Now you can name your Will-o-wisp "willy" and your mr. ice cube "icy" and both will be accessible from the console, as expected.


Zwatotem:
I'm against this solution, it looks 'hacky' and introduces inconsistencies in code. It shouldn't be the case to not be able to expose to scripting classes that are in inheritance relations. Apart from that this unables derivative classes to inherite scripts from base classes. I would accept a temporary solution, however this would require dedicated revert before so I can hardly agree on this.

Semphriss:
> it looks 'hacky' and introduces inconsistencies in code

What do you mean by that? I can try to look into this and adapt my code.

Right now the problem is: If a given object inherits from `ExposedObject<>` more than once, the code breaks. An alternative solution would be to find the code that handles ExposedObject and somehow tell it which one it should use for every object, though I'm not sure if that's possible.

Zwatotem:
Maybe changing inheritance by `ExposedObject<>` to composition could allow to have multiple scripting interfaces without incompatibilities?

Semphriss:
Is that possible in C++? I remember reading an SO question saying it's burdensome (if possible at all) to handle ambiguities when inheriting multiple times from the same class with differnet templates...

I'd need your help (or someone else's) on that one, not sure I can do it myself. There seems to be a lot to take in consideration (like the wrapper; will we have to change the way it works and gets generated?)

Zwatotem:
> Is that possible in C++? I remember reading an SO question saying it's burdensome (if possible at all) to handle ambiguities when inheriting multiple times from the same class with differnet templates...

That's the reason, why I consider conposition here instead of inheritence. So for example have:
```c++
class Willowisp
{
  ExposedObject<Willowisp, scripting::Willowisp> scr1;
  ExposedObject<BadGuy, scripting::BadGuy> scr2;
  ...
}
```
instead of
```c++
class Willowisp : public ExposedObject<BadGuy, scripting::BadGuy>,
                  public ExposedObject<Willowisp, scripting::Willowisp>
  
{
  ...
}
```
> I'd need your help (or someone else's) on that one, not sure I can do it myself. There seems to be a lot to take in consideration (like the wrapper; will we have to change the way it works and gets generated?)

Yes, I would probably require an overhaul of current scripting system, including wrapper, so I'm defenately not an expert here.

Semphriss:
This seems like quite a large undertaking. At the moment, I'm focusing the the overhaul of the particle system (which, I confess, uses the same "hacky" trick to expose rain, clouds and the rest to scripting).

I'd recommend assigning someone else to that job so that you don't have to wait for me; I'm ready to adapt my code to any solution the team comes up with.

I'm leaving this PR open for the moment, at least until we discussed with the rest of the team about what we should do with the scripting system.

tobbi:
In any case, we will have to fix the referenced bug "WillOWisp not exposed anymore".

PR #1541, opened 2020-10-08
Make badguys' pattern not random-based (WIP)

### LINKED ISSUE #959: Make Flying Snowball's flight pattern not random
**SuperTux version: 0.6.0 Alpha 3**

I think Flying Snowball should have a set path of vertical movement that he follows rather than randomly jetting up or down. Would help speedruns and means you don't have to wait as long for him to lower in tricky parkour sections.

Semphriss:
(Edit:) Resolves #959

A badguy that's fun to beat should have a well-defined, easy-to-understand pattern so that new players that aren't well used to the game dont get surprised by a sudden, unexpected change of movement from a badguy.

It also helps speedrunners by not throwing a whole run simply because mister pilot decided it'd hover at the worst height possible and stay there on the best run of the day.

**Currently:**
Pilots accelerate randomly. Each time you load a level, the pilot's pattern will be different.

**At the moment, this PR makes it so that:**
Pilot have a deterministic, but hard-to-predict pattern. That is, playing the same level and going at the same pace should result in similar encounters with the badguys. This solves the speedrunners' problems, but does not yet solve the predictability problem.

**To-do:**
- [x] Make the pilot/flying snowball follow a simple, clearly-defined movement. It could be a simple sine wave movement, or something a bit more complex like this : https://www.desmos.com/calculator/aw4dpgqntf (if link dies, the formula is `sin(x)^3 + sin(3 * (x - pi/3))/3`)
- [x] Investigate and remove randomness from the skullyhops, I've been told they have random movements too.

----------

I opened this draft PR so that people could share their thoughts about what should be done and how. Above are my opinions, but I'd like to hear about other team members, level designers, speedrunners and players too.

### DIFF
--- a/src/badguy/flyingsnowball.cpp
+++ b/src/badguy/flyingsnowball.cpp
@@ -32,7 +32,7 @@ FlyingSnowBall::FlyingSnowBall(const ReaderMapping& reader) :
   total_time_elapsed(),
   puff_timer()
 {
-  m_physic.enable_gravity(true);
+  m_physic.enable_gravity(false);
 }
 
 void
@@ -51,6 +51,7 @@ bool
 FlyingSnowBall::collision_squished(GameObject& object)
 {
   m_sprite->set_action(m_dir == Direction::LEFT ? "squished-left" : "squished-right");
+  m_physic.enable_gravity(true);
   m_physic.set_acceleration_y(0);
   m_physic.set_velocity_y(0);
   kill_squished(object);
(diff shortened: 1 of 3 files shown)


weluvgoatz:
Very nice indeed!
~~The name "pilot" is cringe, blech :P~~

Semphriss:
**To-do complete!**

The pull request is ready for review!

predictable_flying_snowballs

src/badguy/flyingsnowball.cpp
```diff
@@ -69,30 +69,18 @@ FlyingSnowBall::collision_solid(const CollisionHit& hit)
...
-
-    m_physic.set_velocity_y(m_physic.get_velocity_y() * 0.99f);
-
-  } else if (get_pos().y < m_start_position.y - 2*32) {
-
-    // Flying too high - decreased propeller speed
-    m_physic.set_acceleration_y(-grav*0.8f);
+  // Put that function in a graphing calculator :
+  // sin(x)^3 + sin(3(x - pi/3))/3
+  float targetHgt = std::pow(std::sin(total_time_elapsed), 3.f) +
```
Mathnerd314, line 76: sin can overflow. I think you should have something like:
```C++
if (total_time_elapsed > TAU) {
  total_time_elapsed = total_time_elapsed % TAU
}
```

src/badguy/flyingsnowball.cpp
```diff
@@ -69,30 +69,18 @@ FlyingSnowBall::collision_solid(const CollisionHit& hit)
...
+                    std::sin(3.f *
+                             ((total_time_elapsed - 3.14159f) / 3.f)
```
Mathnerd314, line 78: 3.14159f should be `PI` (from src/math/util.hpp).

serano01:
Very well, I've tested it and will now merge it. Thanks!

PR #1543, opened 2020-10-10
Corner-based autotile system

Semphriss:
For smoother experience with certain tilesets.

The snow background tileset is the first to be implemented with this system, as the tiles are designed to imply solidity on the corners rather than on the tile's center, like traditional snow would.

This results in a much smoother and intuitive experience, as well as more precise tile configurations.

*I also took the freedom to add two tiles so that the tileset fit that system better.*

**This does not replace the old system!** It's an addition that makes sense with how certain tilesets are made. If I have some time, the worldmap tileset should also be converted to this system.

----------

Fun fact: the classic system can require up to 256 tiles with multimasking excluding optional non-solid tiles; the corner-based system requires only 16 tiles (including the single empty tile) and doesn't need to consider solidity or multi-masking :^)


Zwatotem:
I tested it, and everything works, except trying to punch a hole in the snow background using an empty tile results in a game crash.

Semphriss:
I'm going to investigate on that, I'll try to make it so that tile 0 acts like an eraser, like a special case. I think it messes up with both systems when autotiling with a non-autotileable system :D

Semphriss:
It should now be fixed, and the erasing system is now implemented directly so the erasing experience should be smooth.

Now, autotiling shouldn't give invalid configuration no matter what you do:
- Erasing topping snow won't work, you have to autoerase the tile below it. No more gaps :)
- Erasing corner-based tilesets erases in corners
- Erasing any kind of tile (regardless of whether the erased tile was autotilable, corner-autotilable or not autotilable at all) should autotile all surrounding tiles properly

**Only possible problem:** If corner-autotilable tiles are drawn on corners with regular tiles, it will adapt smoothly to the surface; however, erasing the regular tiles will *not* add the now-missing corner autotiles.

tobbi:
👍 approved!

PR #1547, opened 2020-10-15
Improve teleporter dot FPS

MrIceblock:
I mostly think that the teleporter dot FPS should be increased to 10, but if you think it should be increased to something else or keep the current, please close this pull request. I don't have much info as this is just a FPS increase.


serano01:
Why do you think this change is needed?

MrIceblock:
> Why do you think this change is needed?

I will close this, as it's unwanted.

PR #1548, opened 2020-10-15
SuperTux launcher

### LINKED ISSUE #248: Update check for SuperTux
Currently, it is not possible to check whether a new version of SuperTux was released on the website. In order to get people up-to-date, adding an update check would make sense.

Semphriss:
Closes #248

The SuperTux launcher will allow users to install and manage multiple versions of SuperTux simultaneously. Nostalgics, speedrunners, enthousiasts, you will be able to have versions 0.1.3, 0.5.1, 0.6.2 and the latest nightly at the same time on the same computer!

unknown

The launcher will also expose some features accessible only from the command line, such as immediately opening a level for plaing, recording gameplay, playing a recorded gameplay, and more!

----------

*As of writing these lines, the launcher isn't yet in a working state. There is still much to be done before you can enjoy those features - but fear not, I'm working hard to release this as soon as possible!*
<sub>*I'm also very glad if you'd like to help me!*</sub>

- [X] The CMakeLists.txt adaptation, the creation of a new executable, and the folder for the launcher
- [x] Setup a window with decoration
- [x] Launch a game and report whether or not it executed properly
- [ ] List multiple executable paths with option to add/remove some
- [ ] Save executable paths
- [ ] Handle launching options (developer mode, start a level immediately, record gameplay, and other features currently CLI-exclusive)
- [ ] Give each version its own user data folder
- [ ] Crash reporter, with exit code and full output
- [ ] Manage installing new versions
- [ ] Check if new versions are available from time to time
    - [ ] Make a server to host API-friendly binaries and info


papb:
Cool, recently I tried to play a level I used to play ~15 years ago (in 0.1.3) and it didn't work well in 0.6.2... It had some hidden blocks that no longer appeared once being "jumped from below" (is there a word for that?)

LizzyBach:
It'll be a pretty neat feature

ScriptForAll:
This seems a pretty good idea. This surely will make it easier for speedrunners and non-windows users that wish to play older versions.

However, I have some questions regarding it:
- Will profiles be "attached" to each version? (e.g. 0.3.3c has its own profile folder) 
- How would addons be dealed with? Would they even be integrated? If so, would they match the version at hand?
- Only the version listed above would be supported or would other versions be also integrated?

Another thing to point out is that 0.1.3 crashes sometimes due to its obsolete version of SDL. Updating the library, at least on Windows, to its newer version 1 instance should fix this problem.

Semphriss:
>     Will profiles be "attached" to each version? (e.g. 0.3.3c has its own profile folder)

Each and every version will have its own user directory. That means versions will not share user profiles, configurations, add-ons, custom levels, etc. That also means a new install done through the launcher will give a completely fresh install, with no add-ons or progress at all.

The reason behind this is because some features might not be compatible across multiple SuperTux versions, therefore the user should take explicit steps to transfer potentially incompatible data.

Don't worry though - you will still be able to transfer your progress from version to version easily with the launcher! I aim to add options to the launcher to transfer parts or all of the progress from a version to another.

>     How would addons be dealed with? Would they even be integrated? If so, would they match the version at hand?

As stated above :) Indeed certain add-ons might work only with a certain range of versions; therefore, the user will have to either re-download the add-ons for each version, or they will have to transfer data using the methods described.

>     Only the version listed above would be supported or would other versions be also integrated?

Every single version of SuperTux you can possibly imagine will be supported :) 0.6.3, 0.5.1, 0.3.0, 0.1.3, 0.0.4, every other release/development build, every nightly, even custom builds will be compatible.

The way it's currently planned, there will be a server to offer ready-made builds for all releases (0.x.x) and the few latest nightlies (as available [here](https://download.supertux.org)), and the launcher would use this to fetch official versions. It will also be possible to install/compile a custom build manually, then add its install path to the launcher, which will then treat it as a valid SuperTux version, so if someone gives you their own build, you can still integrate it to the launcher :)

Semphriss:
This PR has been abandoned due to the ever-changing design choices and due to the higher priority of other PRs; if I ever get back to this, I will probably restart from scratch.

PR #1554, opened 2020-10-18
Editor autosave

### LINKED ISSUE #1551: [SUGGESTION] Auto-Save for the level editor (Should be off by default to prevent accidental changes to the level we're editing)
**SuperTux version:** SVN via .appimage
**System information:** Ubuntu 20.04 LTS Focal Fossa

Hello, I would like an auto-save system for the level editor that's off by default (In case we don't want accidental changes to the level we're editing), since the game crashed while I'm doing unsaved changes to my level, and now I lost my work. The auto-save system for the level editor may be useful in-case I'm lazy to hit CTRL+S, or even a crash, so can you add this feature?

Semphriss:
Closes #1551 

The new autosave feature will automaticaly save the current level to a backup file (**not** the level file, but a different one) each time the level is tested or at most 5 minutes after the level was last saved/auto-saved (amount of time customizable through the editor pause menu).

If the editor crashes, the recovery file will survive and will be detected by the editor on next launch. Level editors will have the option to open the recovery file or not.
- If they choose to open the recovery file, they must save the file manually to apply the recovery data to the main save file. Not saving before quitting the editor will discard the recovery data permanently.
- If they choose to *not* open the recovery file, the regular file will be opened instead and progress will be where it was at the last *manual* save (Ctrl+S, or through the menu). Although from this point onwards, the recovery file is already assumed to be gone, it will not actually be deleted/overwritten unless the user takes one of the actions listed below.

**The recovery file is deleted when:**
1. _The user saves the level manually._
The recovery file is deleted until a new autosave request is initiated by the game. The reason is that manually saving makes the regular file more recent than the recovery file; the auto-save would then be useless, as it is older than the actual save.

2. _The user quits the editor._
Quitting the editor is considered a normal operation. The user is assumed to be aware of the state of the save file upon quitting the editor (they will be told if their level has unsaved changes and asked to confirm their choice). The recovery file is then needless (no need to offer a user to recover unsaved changes if they explicitly chose to discard them).

### DIFF
--- a/src/supertux/menu/editor_level_select_menu.cpp
+++ b/src/supertux/menu/editor_level_select_menu.cpp
@@ -155,7 +155,7 @@ EditorLevelSelectMenu::menu_action(MenuItem& item)
 
     if (PHYSFS_exists((file_name_full + "~").c_str())) {
       auto dialog = std::make_unique<Dialog>(/* passive = */ false, /* auto_clear_dialogs = */ false);
-      dialog->set_text(_("An auto-save file has been found. Would you like to open the auto-save file?\n"));
+      dialog->set_text(_("An auto-save recovery file was found. Would you like to restore the recovery\nfile and resume where you were before the editor crashed?"));
       dialog->clear_buttons();
       dialog->add_default_button(_("Yes"), [this, file_name] {
         open_level(file_name + "~");
--- a/src/supertux/menu/editor_level_select_menu.cpp
+++ b/src/supertux/menu/editor_level_select_menu.cpp
@@ -135,15 +135,43 @@ EditorLevelSelectMenu::create_item(bool worldmap)
 }
 
 void
+EditorLevelSelectMenu::open_level(const std::string& filename)
+{
+  auto editor = Editor::current();
+  editor->set_level(filename);
+  MenuManager::instance().clear_menu_stack();
+}
+
+void
 EditorLevelSelectMenu::menu_action(MenuItem& item)
 {
   auto editor = Editor::current();
   World* world = editor->get_world();
   if (item.get_id() >= 0)
   {
-    editor->set_level(m_levelset->get_level_filename(item.get_id()));
 
-    MenuManager::instance().clear_menu_stack();
+    std::string file_name = m_levelset->get_level_filename(item.get_id());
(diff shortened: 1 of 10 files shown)


src/supertux/menu/editor_level_select_menu.cpp
```diff
@@ -134,16 +134,44 @@ EditorLevelSelectMenu::create_item(bool worldmap)
...
   {
-    editor->set_level(m_levelset->get_level_filename(item.get_id()));
 
-    MenuManager::instance().clear_menu_stack();
+    std::string file_name = m_levelset->get_level_filename(item.get_id());
+    std::string file_name_full = FileSystem::join(editor->get_level_directory(), file_name);
+
+    if (PHYSFS_exists((file_name_full + "~").c_str())) {
+      auto dialog = std::make_unique<Dialog>(/* passive = */ false, /* auto_clear_dialogs = */ false);
+      dialog->set_text(_("An auto-save file has been found. Would you like to open the auto-save file?\n"));
```
tobbi, line 158: Not sure about the string here. I'd suggest something more user-friendly, such as "Would you like to restore your the level you were last working on?" or something? Or other suggestions?

Semphriss: I'm open to changing the line if needed, but I'd like to keep the part that says the message is shown because a recovery file was found. There might be bugs I haven't noticed that would leave the recovery file there (say, the file deletion fails for some reason), and seeing the warning without knowing why it pops in the first place could cause some confusion.

Semphriss: A few suggestions:
- "It looks like the editor crashed last time you opened this level. Would you like to continue where you left off?"
- "An auto-save recovery file was found. Would you like to restore the recovery file and resume where you were before the editor crashed?"
- "It looks like you have some unsaved progress on that level. This can happen if the editor crashes. Would you like to resume where you left off?"

weluvgoatz: I'd suggest making one that is as short but effective as possible.  Players do not like to read one large string that can be easily summarized and have the same effect.

PR #1560, opened 2020-10-21
Support for zooming from scripting

Semphriss:
## Zooming now available to level designers!

zooming

If you have a portion of you level that is bigger or smaller, you can now use zooming to increase the effect on the player :)

Accessible from `Camera.set_scale(scale)`, `Camera.scale(scale, time)` and `Camera.ease_scale(scale, time, easing)` where:
- **scale** is a decimal number which is the zoom multiplier. 1 is default zoom, The gif uses 0.75. A smaller number = a larger portion of the level visible to the screen.
- **time** is a decimal number that represent the time it will take to reach the scale.
- **easing** is a string with the name of the easing you want to use, following the format "Ease`[Mode][Direction]`" where "Mode" is the easing mode (Sine, Exponential, Back, Bounce, etc.) and Direction is In, Out, or InOut. For a ful llist of possibilites, see: https://easings.net/ (note that the name are not exactly the name, the order is different). Examples: "EaseSineInOut", "EaseBounceOut", "EaseBackIn"

### DIFF
--- a/src/object/camera.hpp
+++ b/src/object/camera.hpp
@@ -79,7 +79,7 @@ public:
   void reset(const Vector& tuxpos);
 
   /** return camera position */
-  const Vector& get_translation() const;
+  const Vector get_translation() const;
   void set_translation(const Vector& translation) { m_translation = translation; }
 
   /** shake camera in a direction 1 time */
@@ -97,12 +97,22 @@ public:
   Vector get_center() const;
 
   void set_mode(Mode mode_) { m_mode = mode_; }
+
+  /** get the exact scale at this exact moment */
+  float get_current_scale() { return m_scale; }
+
+  /** get the scale towards which the camera is moving */
+  float get_target_scale() { return m_scale_target; }
+
+  /** smoothly slide the scale of the camera towards a new value */
+  void ease_scale(float scale, float time, easing ease);
   /** @} */
 
 private:
   void update_scroll_normal(float dt_sec);
   void update_scroll_autoscroll(float dt_sec);
   void update_scroll_to(float dt_sec);
+  void update_scale(float dt_sec);
   void keep_in_bounds(Vector& vector);
   void shake();
 
@@ -134,6 +144,13 @@ private:
   float m_scrollspeed;
 
   std::unique_ptr<CameraConfig> m_config;
+
+  float m_scale,
+        m_scale_origin,
+        m_scale_target,
+        m_scale_time_total,
+        m_scale_time_remaining;
+  easing m_scale_easing;
 
 private:
   Camera(const Camera&) = delete;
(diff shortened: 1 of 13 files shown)


Rusty-Box:
Is this linked to the magnification setting? And if so, can you also return the zoom back to the original setting?

Semphriss:
This works in a similar way but is completely independant from the magnification setting. The scale is tied to the camera object, and is applied only to the sector the camera is linked to, similarly to the camera translation.

Rusty-Box:
So, a 200% zoom when magnification is 100 is different from a 200% zoom when magnification is 160?

Semphriss:
Yes, indeed. A magnification of 200 with a zoom of 200% would result in an overall zoom of 400%.

Rusty-Box:
Okay, but doesn't this provide a problem for players with different different settings. For example, if a cutscene zoomes in (based on a magnification of 100) important things may get lost if the same cutscene is viewed by players having a magnification of 200?

Semphriss:
This problem isn't introduced by this feature, though. Players on smaller/bigger resolutions have the same problem. The difference between what players on 640x480 see and those on 1920x1080 see will be the same at 100% zoom or 500% zoom.

This is a level design thing, and since the feature is for level designers, it will be their job to make sure what they want to show is clear and visible on small resolutions. This is true without this feature, and it will be the same with it :)

src/object/camera.hpp
```diff
@@ -97,12 +97,22 @@ class Camera final : public GameObject,
   Vector get_center() const;
 
   void set_mode(Mode mode_) { m_mode = mode_; }
+
+  /** get the exact scale at this exact moment */
+  float get_current_scale() { return m_scale; }
```
tobbi, line 102: This can be marked `const`

src/object/camera.hpp
```diff
@@ -97,12 +97,22 @@ class Camera final : public GameObject,
...
+
+  /** get the scale towards which the camera is moving */
+  float get_target_scale() { return m_scale_target; }
```
tobbi, line 105: This too.

src/video/canvas.cpp
```diff
@@ -127,7 +127,9 @@ Canvas::draw_surface(const SurfacePtr& surface,
   request->blend = blend;
 
   request->srcrects.emplace_back(Rectf(surface->get_region()));
-  request->dstrects.emplace_back(Rectf(apply_translate(position), Size(surface->get_width(), surface->get_height())));
+  request->dstrects.emplace_back(Rectf(apply_translate(position)*scale(),
```
tobbi, line 130: Add spaces around the `*`

Semphriss: I've rarely seen so many consecutive closing parentheses.

PR #1563, opened 2020-10-22
Coyote time

Semphriss:
Coyote time: If you run and jump from a ledge but press jump just too late, instead of falling of the ledge without jumping, it'll give you a very small amount of extra "gace" time where you can jump despite no longer being on the platform.

This is useful for jumping long ledges where you often have only one chance (as in, if you miss, you fall down and die immediately).


weluvgoatz:
I'm going to merge this since it works very well as intended and there doesn't seem to be any issues caused by this PR.  Thank you!

PR #1568, opened 2020-10-27
Update Danish translation

AJenbo:


serano01:
You should know that translations are working through transifex. So if you change a file on github, it will overwritten with the next translation for a release.

PR #1572, opened 2020-11-08
Align selection

Semphriss:
select_snap

Selected tiles in editor will now snap to the grid.


Semphriss:
I just realised #1571 is in there...

PR #1585, opened 2020-11-15
Remove extraneous font image causing path conflict

eric135:
This file does not appear to be used in code and causes a warning while cloning on Windows and likely other case-insensitive file systems.


tobbi:
Thanks! 👍

PR #1591, opened 2020-11-29
Add new signs

DwarflinDeveloping:


serano01:
I'm not sure if those are needed, because we already have billboards for that purpose.

Rusty-Box:
The buttjump sign may be worth adding but I’m not so sure about the others.

DwarflinDeveloping:
I have closed this pull request and will start a new one. In the new one, I will remove unnecessary files like the big signs.

ScriptForAll:
I can see the empty tall sign being used as an alternative to the green info boxes ~(i don't know if this is its official name if even they have one)~. 

For instance, this sign could be employed to "in-plot" text (e.g. a warning from Nolok) and info boxes could be used for meta-plot messages (e.g. game mechanics tutorials).

Ordoviz:
I like the separation of in-plot and meta-messages but all text needs to be translatable. Text in different languages takes a different amount of space, so we would need to adjust the size of the signs accordingly.

ScriptForAll:
Each sign size don't have to correspond with the amount of text of the correspondent message. The same happens with the info box, therefore, the same principle should apply to signs.

If one wants to distinguish even further these two, a text balloon could be used instead of a text box (as it's being currently used). This text balloon's dimensions would be dependent on the amount of text, and thus, simplifying the issue mentioned by @Ordoviz.

Issue #1592, opened 2020-11-29
..Or Is It Just Me : Two small improvements


Gnu-Bricoleur:
**SuperTux version: 0.6.0**
**System information:** *Linux 4.19.0-12-amd64 #1 SMP Debian 4.19.152-1 (2020-10-18) x86_64 GNU/Linux
*

It is not a bug strictly speaking, just two small things that could be improved in the level "or is it just me" from the main story mode

##### Expected behavior
Tux should not be able to stick to the cliff like that
Tux should not be able to leave the level by going up

##### Actual behavior
![Capture d’écran du 2020-11-29 14-27-10](https://user-images.githubusercontent.com/15331611/100549879-0f59b880-3276-11eb-953c-992ce8f6f904.png)
![Capture d’écran du 2020-11-29 15-03-38](https://user-images.githubusercontent.com/15331611/100549882-1254a900-3276-11eb-8f33-b31a5285a033.png)
##### Additional debugging information

*If available, please attach a backtrace of a crash (you can get one using GDB or another debugger), or a core dump. If you're on GNU/Linux and your distro is using systemd, it's highly likely that a core dump is available using `coredumpctl(1)`.
If the issue only occurs in a custom level, please attach that too.*

##### Guidelines for reporting issues

- [ ] I have read <https://github.com/SuperTux/supertux/blob/master/CONTRIBUTING.md#bug-reports>.
- [ ] I verified this isn't an issue that's already been reported.
- [ ] This report only includes details about **one** (1) issue, feature request or topic.
- [ ] If I make a mistake while submitting this report I will use the edit feature instead of closing this issue and opening a new one.
- [ ] This issue is **not** about wrong translations (use Transifex for those), or anything unsupported (e.g. third-party add-ons).
- [ ] Once you're done, please remove the section 'Guidelines for reporting issues', the 'Additional debugging information' if it is empty, and any text in *italic font*.

Gnu-Bricoleur:
See Pull request #1593

Zwatotem:
This level is designed to let Tux go on the top. There is no mistake here. The hangover glitch is known, and applies to all instances of triangle hitbox underneath a square one.

PR #1593, opened 2020-11-29
Small edits on the Or Just Me level 

Gnu-Bricoleur:
Small edits on the Or Just Me level to prevent Tux sticking on the edge of a cliff or escaping by the top of the level see Issue #1592 on SuperTux/supertux


Gnu-Bricoleur:
Closing the top of the level
Capture d’écran du 2020-11-29 15-03-38
Capture d’écran du 2020-11-29 20-13-23

Smoothing the cliff to prevent Tux sticking
Capture d’écran du 2020-11-29 14-27-10
Capture d’écran du 2020-11-29 20-13-16

Zwatotem:
As I said in #1592 the alternative route on top of the level is intentional, so there's no need to repair anything. Also, the glitch can still be performed in other places:
SuperTux v0 6 2-532-g3a19fb2e4 2020-11-29 20-37-00_Moment
so I see no sense in deleting individual points like that, this problem requires a universal solution in game code.
Thanks for your encouragement.

PR #1598, opened 2020-12-04
Prevent player from falling off the screen while swimming

Semphriss:
It is no longer possible to die by falling off the screen while in a swimming state.

This will save level designers the burden of manually placing blocks at the bottom of their water bodies to prevent fortuitous deaths.

(This will indeed require an even more urgent revision of levels using water without swimming in mind)

### DIFF
--- a/src/object/player.cpp
+++ b/src/object/player.cpp
@@ -1849,7 +1849,12 @@ Player::check_bounds()
   }
 
   /* fallen out of the level? */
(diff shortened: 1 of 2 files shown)


serano01:
Or, instead of the revision, a invincible badguy could be created, or some other method of an instant death. Otherwise some levels could get unrecognizable for the original author to begin with...

Semphriss:
My first thoughs would've been "make the water in these levels non-solid", but there would need to be a clear visual distinction between which water is swimmable and which isn't.

PR #1602, opened 2020-12-09
[WIP] Move to C++17

strikersh:


strikersh:
C++ 20 introduce a few new things :
boost::algorithm::ends_with can be replaced by std::string::ends_with
boost::format by std::format
and <chrono> adds what was missing to replace boost::gregorian

PR #1611, opened 2020-12-20
Editor Scroll Speed small tweaks

Raoul1808:
I wanted to push a small tweak I made to the editor when working on editor samples.
Basically you can now hold the Control and Shift key to scoll respectively slower (0.5x) and faster (3x).

The first commit has been tested on Windows and Ubuntu, but the second one only on Windows.
I consider this tweak done, it could be improved to allow creators to enter other scroll values (with a limit) or speed modifiers.

### DIFF
--- a/src/editor/editor.cpp
+++ b/src/editor/editor.cpp
@@ -670,16 +670,20 @@ Editor::event(const SDL_Event& ev)
 		redo();
 		}
 
-  if (ev.type == SDL_KEYDOWN &&
-    ev.key.keysym.mod & KMOD_SHIFT) {
-    m_scroll_speed = 96.0f;
-  }
-  else if (ev.type == SDL_KEYDOWN &&
-    ev.key.keysym.mod & KMOD_CTRL) {
-    m_scroll_speed = 16.0f;
-  }
-  else {
-    m_scroll_speed = 32.0f;
+  if (ev.type == SDL_KEYDOWN)
+  {
+    if (ev.key.keysym.mod & KMOD_SHIFT)
+    {
+      m_scroll_speed = 96.0f;
+    }
+    else if (ev.key.keysym.mod & KMOD_CTRL)
+    {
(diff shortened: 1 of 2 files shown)


src/editor/editor.cpp
```diff
@@ -669,6 +670,20 @@ Editor::event(const SDL_Event& ev)
 		redo();
 		}
 
+  if (ev.type == SDL_KEYDOWN &&
+    ev.key.keysym.mod & KMOD_SHIFT) {
+    m_scroll_speed = 96.0f;
+  }
+  else if (ev.type == SDL_KEYDOWN &&
```
Ordoviz, line 677: Using a nested if statement, you can avoid checking `ev.type == SDL_KEYDOWN` twice.

Semphriss: While you're changing this part of the code  -  please format your `if` statemets as:
```cpp
if (condition)
{
  ...
}
else if (condition)
{
  ...
}
else
{
  ...
}
```

.. with all brackets of their own lines. Saves some code cleaning time :^)

Raoul1808: I didn't think of nested if statements, I'm working on it rn.
@Semphriss shouldn't I remove the brackets since the speed changes are one line only?

Semphriss: Usually, brackets are omitted when the operation is trivial and the body of the `if` statement is practically guaranteed to never need additions (like `if (must_be_false) return;`).

With if-else statements, it's even more important since it's more likely to cause trouble with maintenance later (imagine someone adding a line when you didn't use brackets). For the moment, I'd advise keeping the brackets even though it's only one line, but I'm open to discussion.

tobbi:
Is this ready to be merged?

tobbi:
According to @Semphriss this is ready to be merged. Thanks for this, @Raoul1808

PR #1612, opened 2020-12-22
Added ladders to forest worldmap

Semphriss:
Screenshot from 2020-12-22 16-30-00

Also made some minor decorational fixes.


ScriptForAll:
I don't want to sound very picky but the ladders don't stand out very much from the background. 

I'd suggest darkning the transparent background a bit so that it's more clear what's background and what's "foreground".

PR #1613, opened 2020-12-22
Create github actions matrix build

maths22:
Now that travis is dead, move the build over to github actions


Semphriss:
Much, much cleaner than my own file, I have to say!

Question though - can the Release trigger be limited to tag pushing easily?

maths22:
@Semphriss I need to figure out how to make it not build on both push and pull_request for the same commit, because that's useless.  Regarding the release vs debug thing, on travis we used to do all 6 builds for every commit which is why I pulled that forward here.  Currently the ones we upload to https://download.supertux.org/ are only ever "Release" builds, not "Debug" builds, but if we wanted to change that, at least for the pre-merge build that would sound reasonable to me

PR #1618, opened 2020-12-24
Compress the data folder

### LINKED ISSUE #1224: Optimize/decrease PNGs sizes with OptiPNG
ST uses PNG images a lot and there are tools for optimizing PNGs w/o losing any information. For example - http://optipng.sourceforge.net/ . I think that would be a good practice to start using such tools before the every release. It would provide smaller packages and installations.

Here is a result of using OptiPNG on `SuperTux\data\images` folder with `-o2` option (fast but efficient):
- Overall size before: 51.4 MiB
- Overall size after: 42.3 MiB
- Overall saving: ~9 MiB

Comparison of subfolder sizes in Double Commander:
image

Most noticeable comparison:
image

It might be easier to try it on Linux, but here is how to use it on Windows:
- Download http://prdownloads.sourceforge.net/optipng/optipng-0.7.7-win32.zip?download
- Extract `optipng.exe` into `C:\Windows`
- Apply the following reg-file:
```
Windows Registry Editor Version 5.00
[HKEY_LOCAL_MACHINE\SOFTWARE\Classes\Folder\shell\OptiPNG]
@="Run OptiPNG on Folder"
[HKEY_LOCAL_MACHINE\SOFTWARE\Classes\Folder\shell\OptiPNG\command]
@="cmd.exe /c \"TITLE Running OptiPNG on %1 && FOR /r \"%1\" %%f IN (*.png) DO optipng.exe -o2 -strip all \"%%f\" \""
```
- Restart PC
- Run OptiPNG on folder by right mouse click menu and wait
- To remove the previous registry change apply the following:
```
Windows Registry Editor Version 5.00
[-HKEY_LOCAL_MACHINE\SOFTWARE\Classes\Folder\shell\OptiPNG]
[-HKEY_LOCAL_MACHINE\SOFTWARE\Classes\Folder\shell\OptiPNG\command]
```

Semphriss:
Closes #1224 
(I suppose)

1. Optimized the image files and the audio files, saving ~50Mb
2. Package the data folder as a .zip archive to save even more space (the levels' source take 55Mb; compressed, they take **1.5Mb, more than 97% efficiency**)

This build is expected to fail a few times, as I can't really install over and over again on my system (no easy uninstaller on Linux, unfortunately). Don't merge too quickly!


Semphriss:
Closing this, if I ever want to redo this I'll open a new PR because a bunch of files are stale and should be re-optimized. Also, I never managed to make those levels efficiently compressed.

HybridDog:
I assume most of the size of a level is caused by the tilemaps.

The tilemaps are similar to an image with indexed colour, so image compression algorithms could be used to achieve a good compression ratio. As far as I know, PNG does not use a special algorithm to compress the 2D array with index entries since the Paeth and other predictors don't work well for indexed images. JPEG XL and its predecessor FLIF (and FUIF) however compress indexed images better than PNG in common cases.

There's also the lz4 compression (https://github.com/lz4/lz4) which is very fast but produces approximately twice the output size than zip.

PR #1619, opened 2020-12-26
In-game progress stats

Semphriss:
unknown

- Displays all stats in the pause menu, so player can check them quickly.
- Appears during playtime and notifies the player with a sound effect whenever a stats reach 100% completion

https://user-images.githubusercontent.com/66701383/103158154-bf0d4200-47b2-11eb-858f-21e01faf68ea.mp4

TODO:
- [x] Check for better icons
- [x] Check for better sound effects
- [x] Add easing?


ScriptForAll:
Personally, I would put those stats at the top left corner instead at the bottom right one.
But apart from that detail this looks very nice! This is a great implementation to the game and a good enhancement to the gameplay (like I've said before)!

Thanks for the work you've been putting on!

serano01:
I like it as it is. Most of the HUD is already on the right corner, if something suddenly pops out on the other side, it may go unnoticed.

PR #1621, opened 2020-12-27
Added Color (with alpha) to background objects

Semphriss:
Now you can tint your backgrounds and make them transparent :)


tobbi:
Very nice!!! 👍

PR #1622, opened 2020-12-27
Fixed bug where ScriptTrigger areas wouldn't save their name

Semphriss:
Note: Adding the key name to the ObjectSettings on line 75 of src/trigger/scripttrigger.cpp (old version) wasn't enough, I had to refactor the code a bit (made it cleaner too)

### DIFF
--- a/src/supertux/moving_object.cpp
+++ b/src/supertux/moving_object.cpp
@@ -18,6 +18,7 @@
 
 #include "editor/resize_marker.hpp"
 #include "supertux/sector.hpp"
+#include "util/reader_mapping.hpp"
 #include "util/writer.hpp"
 
 MovingObject::MovingObject() :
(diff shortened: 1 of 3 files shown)


src/supertux/moving_object.cpp
```diff
@@ -29,6 +30,16 @@ MovingObject::MovingObject(const ReaderMapping& reader) :
...
 {
+  float height, width;
+
+  if (reader.get("width", width))
+    m_col.m_bbox.set_width(width);
+
+  if (reader.get("height", height))
+    m_col.m_bbox.set_height(height);
+
+  reader.get("x", m_col.m_bbox.get_left());
```
tobbi, line 41: I'm confused about this. the get_left() and get_top() methods are marked const. However, they're passed by reference to the reader.get() method which might attempt to save a value, which should fail. Can you please test this?

Semphriss: @tobbi There is another place in the code that uses the exact same trick, the author called it a "temporary hack" (which I (or anyone) should fix someday). Two overloads for that function, of which one is a non-const reference.
image

tobbi: Very interesting. OKay, I'm gonna merge this.

PR #1625, opened 2020-12-29
Fixed tilemap tiles converted to objects

### LINKED ISSUE #1378: BonusBlocks are placed in the wrong position when transformed to objects
SuperTux version: current git version (commit f6ea1f85a)
System information: Ubuntu 19.10 64bit

When a level is started, the game transforms all bonusblocks placed in tilemaps to objects, but:

Expected behaviour: bonusblocks are placed in the same position as the tile-bonusblock

Actual behaviour: bonusblocks are placed in the wrong position

The position is not entirely incorrect - but it is relative to the entire sector while it should be relative to the position of the tilemap the tile-bonusblocks were in.

Steps to reproduce actual behaviour: Create a new level, in that level, create a tilemap and put it anywhere but x=0, y=0. Put bonusblocks(or coins, or anything that transforms into an object) in the top-left corner. Now, if you play the level, the bonusblocks won't be in the top-left of the tilemap, but in the top-left of the entire sector. (top-left is just for illustration, it works the same anywhere)

editor
game

I'm not sure when this bug first appeared but I know for sure this worked fine in 0.5.1

Semphriss:
Fixes #1378

### DIFF
--- a/src/supertux/game_object_manager.cpp
+++ b/src/supertux/game_object_manager.cpp
@@ -69,6 +69,30 @@ GameObjectManager::process_resolve_requests()
   m_name_resolve_requests.clear();
 }
 
+void
+GameObjectManager::try_process_resolve_requests()
+{
+  assert(m_gameobjects_new.empty());
+  std::vector<GameObjectManager::NameResolveRequest> new_list;
+
+  for (const auto& request : m_name_resolve_requests)
+  {
+    GameObject* object = get_object_by_name<GameObject>(request.name);
+    if (!object)
+    {
+      // Unlike process_resolve_requests(), we just keep that one in mind
+      new_list.push_back(request);
+    }
+    else
+    {
+      request.callback(object->get_uid());
+    }
+  }
+
+  m_name_resolve_requests.clear();
+  m_name_resolve_requests.assign(new_list.begin(), new_list.end());
+}
+
 const std::vector<std::unique_ptr<GameObject> >&
 GameObjectManager::get_objects() const
 {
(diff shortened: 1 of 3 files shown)


src/supertux/game_object_manager.hpp
```diff
@@ -184,6 +185,9 @@ class GameObjectManager
 protected:
   void process_resolve_requests();
 
+  /** Saem as process_resolve_requests(), but those it can't find will be kept in the buffer */
```
tobbi, line 188: typo.

src/supertux/game_object_manager.cpp
```diff
@@ -69,6 +69,30 @@ GameObjectManager::process_resolve_requests()
...
 
+void
+GameObjectManager::try_process_resolve_requests()
+{
+  assert(m_gameobjects_new.empty());
+  std::vector<GameObjectManager::NameResolveRequest> new_list;
+
+  for (const auto& request : m_name_resolve_requests)
+  {
+    GameObject* object = get_object_by_name<GameObject>(request.name);
```
tobbi, line 80: Can you maybe change this to 
```c++
auto object = ...
```

Semphriss: `auto*` ?

Our guidelines recomment receiving pointers as `auto*`

tobbi: Interesting. https://stackoverflow.com/questions/12773257/does-auto-type-assignments-of-a-pointer-in-c11-require

I believe our guidelines should be changed, but I will have to discuss that with the rest of the programmers.

PR #1630, opened 2020-12-31
WIP: Update eo.po

jorgesumle:


Semphriss:
For translations, please use Transifex.

jorgesumle:
Can I upload it or send somewhere the PO file when ready? I kinda work better with an offline editor, because I don't usually have an Internet connection. Also, I never used Transifex, as it's proprietary.

jorgesumle:
I found that you can upload it, but you need to give me access to the Esperanto translation team, so that I can submit my translation there.

PR #1631, opened 2021-01-03
Wall Jumping

weluvgoatz:
This PR allows Tux to walljump upon collision with certain, new special tiles.  This is to allow old levels to remain as they are, while also having the new feature exist.  Tux can walljump both with and without sliding down the wall, which is also now possible.  

Demo: https://youtu.be/wFMRfa60GLI (However, I recommend playing the actual build if you want the complete experience)

Comments? Suggestions? Demands? Tell me, and I'll try to fix them in order for this to get merged.  Be sure to playtest it!

Note: I'm considering this incomplete/not final, as feedback is still needed.


src/object/player.cpp
```diff
@@ -360,6 +367,58 @@ Player::update(float dt_sec)
...
   // extend/shrink tux collision rectangle so that we fall through/walk over 1
   // tile holes
+  
+  //wallclinging and walljumping
+  
+  Rectf wallclingleft = get_bbox();
+  wallclingleft.set_left(wallclingleft.get_left() - 8);
+  bool trygrabwallleft = !Sector::get().is_free_of_statics(wallclingleft);
+  m_on_left_wall = ((trygrabwallleft) ? true : false);
+
```
Semphriss, line 377: The code is literally perfect in my eyes except for this line.

This line.

_screeches loudly_

--------------------------

This line (and another one 5-ish lines below) could be simplified to `m_on_left_wall = trygrabwallleft;`
Considering this would make both variables identical, do you need `trygrabwallleft`? Can you replace it by `m_on_left_wall`?

src/object/player.cpp
```diff
@@ -360,6 +367,58 @@ Player::update(float dt_sec)
...
+  Rectf wallclingright = get_bbox();
+  wallclingright.set_right(wallclingright.get_right() + 8);
+  bool trygrabwallright = !Sector::get().is_free_of_statics(wallclingright);
+  m_on_right_wall = ((trygrabwallright) ? true : false);
+  
+  if (m_on_wall)
```
tobbi, line 383: There is a lot of code duplication from in lines 383-400. You can do better.

src/object/player.cpp
```diff
@@ -360,6 +367,58 @@ Player::update(float dt_sec)
...
+  {
+    if (((trygrabwallleft) || (trygrabwallright)) && !on_ground() && !m_does_buttjump && !m_swimming)
+    {
+      m_on_wall = true;
+    }
+    else
+    {
+      m_on_wall = false;
+    }
```
Semphriss, line 392: `m_on_wall = ((trygrabwallleft) || (trygrabwallright)) && !on_ground() && !m_does_buttjump && !m_swimming`

src/object/player.cpp
```diff
@@ -360,6 +367,58 @@ Player::update(float dt_sec)
...
+  }
+  else
+  {
+    if (((trygrabwallleft) || (trygrabwallright)) && !on_ground() && !m_does_buttjump && !m_swimming)
+    {
+      m_on_wall = true;
+    }
```
Semphriss, line 399: When entering this point of the code, `m_on_wall` is guaranteed to be `false`, so that code does the same job as the code in the part above it. Replace the whole `if (m_on_wall) { ... } else { ... }` with the code I gave you in the other review

serano01:
What's about these changes in snowball.cpp? There aren't supposed to be part of this, are they?

weluvgoatz:
The changes in snowball shouldn't be appearing in the latest commit?  Anyways, no, they were part of a test change that I reverted.
Also, yes I will update the code quality, as I can improve it.

Rusty-Box (changes requested):
Could you please remove the IDs you used from the "left over ID" list so we know which have already been used.

Rusty-Box (approved):
Tile-ID issue has been solved!

Semphriss:
I find it surprisingly comfortable at use. Good work! :+1:

Semphriss:
A few minor gameplay details:
- If you go away from a walljumpable wall while sliding it down, Tux will face the wrong way. It's minor though, since I didn't manage to hold an incorrect direction for longer than 0.25 seconds :laughing: 
- When using the Earth Flower, the light coming from Tux's helmet should be displayed the same way as when he is skidding. Since the hat is slightly rotated todards the camers, a different light texture is used, and since the walljump sliding helmet sprite is rotated too, they should use this light sprite instead of the default directional light.

Zwatotem:
Here are some tricks you can do as of https://github.com/SuperTux/supertux/pull/1631/commits/2c8893cb6aba618f1914794fb3f82661bc9347b2:
climb
There's also this:
yahoo

The first one is just pressing left, jump, right in the correct timing.
Second one is backflipping onto a wall and spamming jump. You're guaranteed to not fall off.

Zwatotem:
This is, how Tux looks like when sticking to the wall directly after water jump:
image
Also this doesn't end the waterjump phase, so after jumping off the wall waterjumping animation is back. I don't know, if it should cancel, or if it can stay, that's something to be decided, but it doesn't break anything.

Zwatotem:
This is, how Tux looks like when sticking to the wall directly after water jump:
image
Also this doesn't end the waterjump phase, so after jumping off the wall waterjumping animation is back. I don't know, if it should cancel, or if it can stay.

PR #1632, opened 2021-01-03
Let's climb that mountain: crystal repair

Zwatotem:
* repair crystal tiles
* add paralax foreground


serano01:
Oh, thanks. I knew somewhere it still uses the old tileset arrangement. I'll merge this soon.

PR #1633, opened 2021-01-09
Added complete scripting support for custom particles

Semphriss:
Exposed API to allow scripting to:
- Spawn particles (Any amount, instantly or with respect to the "delay-between-each-spawn" attribute, etc.)
- Clear all particles on screen
- Get and set almost any property, and fade and ease any property for which fading and easing are relevant

### DIFF
--- a/src/object/custom_particle_system.cpp
+++ b/src/object/custom_particle_system.cpp
@@ -541,7 +541,7 @@ CustomParticleSystem::update(float dt_sec)
   //  update the already existing particles regardless, if any
 
   // Handle easings
-  for (auto req : script_easings)
+  for (auto& req : script_easings)
   {
     req.time_remain -= dt_sec;
     if (req.time_remain <= 0)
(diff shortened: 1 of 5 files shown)


Semphriss:
NOTE: If you merge, please merge using **squash**. Only the message of the first commit is relevant.

src/object/custom_particle_system.cpp
```diff
@@ -538,6 +540,23 @@ CustomParticleSystem::update(float dt_sec)
   // "enabled" being false only means new particles shouldn't spawn;
   //  update the already existing particles regardless, if any
 
+  // Handle easings
+  for (auto req : script_easings)
```
tobbi, line 544: Can `req` be passed by reference?

Semphriss: It can, will fix in a few moments :+1:

Semphriss: Done :+1:

PR #1634, opened 2021-01-10
Editor region/rectangle tool will no longer paint tiles until the mouse button is released

Semphriss:
https://user-images.githubusercontent.com/66701383/104114459-a184c000-52fc-11eb-8f44-49b8d27b58b2.mp4


Ordoviz:
Are you sure this is not a feature?

PR #1637, opened 2021-01-12
Added usage of skippable cutscenes in cutscene levels

Semphriss:
Cutscene levels will now be iinstantly skipped on escape press.

I intend to re-add in-level small cutscenes and make them skippable with escape key in the same way.


weluvgoatz:
The Picnic with Penny one does not skip the cutscene.

PR #1639, opened 2021-01-12
Code cleanup: Simplified cutscene text rendering

Semphriss:
My original code implied the creation of a single game object held in the sector, which would exist with the sector and be rendered with it. It would move with the camera every frame, and check for the level status. All of this, only to display "Press escape to skip" at the top left corner of the screen. I also had left a bunch of broken, partially disabled code, inside the GameSession file from when I was doing tests.

My new code uses the damn `draw()` function like a normal programmer would do.


tobbi:
Eh, yeah, that seems reasonable.

PR #1641, opened 2021-01-14
Added fade helpers to help clean code in game objects

Semphriss:
The FadeHelper will make it much easier to write fading code inside GameObjects and any `update()`-able class.

**Note:** This helper has yet to be used inside the code. A refactor would be needed. I can do this, but if you think this PR is a good idea, I would really appreciate your help :)

### DIFF
--- a/src/util/fade_helper.cpp
+++ b/src/util/fade_helper.cpp
@@ -50,19 +50,18 @@ FadeHelper::update(float dt_sec)
       *m_value = m_target;
     return m_target;
   }
-  else
-  {
-    // FLOAT/DOUBLE CONVERSION
-    // If at some point in development, floats are changed to doubles, or the
-    // ease funciton return type change from double to float, remove the casts
-    // here (it takes a lot of space).               ~ Semphris
-    float progress = static_cast<float>(m_ease(static_cast<double>(
-                        m_start + (m_target - m_start) * (m_time / m_total_time)
-                        )));
-    if (m_value)
-      *m_value = progress;
-    return progress;
-  }
+
+  // FLOAT/DOUBLE CONVERSION
+  // If at some point in development, floats are changed to doubles, or the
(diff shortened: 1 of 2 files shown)


src/util/fade_helper.cpp
```diff
@@ -0,0 +1,74 @@
...
+                        m_start + (m_target - m_start) * (m_time / m_total_time)
+                        )));
+    if (m_value)
+      *m_value = progress;
+    return progress;
+  }
+}
+
+bool
+FadeHelper::completed()
```
tobbi, line 69: const.

src/util/fade_helper.hpp
```diff
@@ -0,0 +1,58 @@
...
+  /**
+   * Increases the internal timer of the FadeHelper.
+   * 
+   * @param dt_sec The increase in time since last call to update().
+   * @returns The new value for the internal value.
+   */
+  float update(float dt_sec);
+
+  /** @returns true if the FadeHelper has completed fading. */
+  bool completed();
```
tobbi, line 43: And here too.

src/util/fade_helper.cpp
```diff
@@ -0,0 +1,74 @@
...
+  m_time += dt_sec;
+
+  if (completed())
+  {
+    m_time = m_total_time;
+    if (m_value)
+      *m_value = m_target;
+    return m_target;
+  }
+  else
```
tobbi, line 53: Since you return'ed from the last if, the else statement is redundant. You can reduce the level of nesting here.

Semphriss: I'm not sure to understand... The value computed for each `return` statement is different between the if and the else.

Do you suggest I put a `float return_value;` that I set in both the if and the else, and then have only 1 return statement after the whole if/else?

Semphriss: Fixed :+1:

PR #1642, opened 2021-01-14
Replace value copying with std::move and const references

HybridDog:
I've applied many changes suggested by clang-tidy.

### DIFF
--- a/src/editor/button_widget.hpp
+++ b/src/editor/button_widget.hpp
@@ -31,7 +31,7 @@ private:
 public:
   ButtonWidget(SpritePtr sprite, const Vector& pos, std::function<void()> m_sig_click = {});
   ButtonWidget(const std::string& path, const Vector& pos, std::function<void()> callback = {}) :
-    ButtonWidget(SpriteManager::current()->create(path), pos, callback)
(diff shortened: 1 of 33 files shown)


Semphriss:
What does std::move do?

I read up documentation online and apparently std::move destroys the original object (hence why you had to remove the `const`'s), and considering some of them are passed by reference, it might have unintended side effects... (those who aren't passed by reference are copied too anyways, so little is being saved by using std::move at another point in the code)

tobbi:
https://stackoverflow.com/questions/51705967/advantages-of-pass-by-value-and-stdmove-over-pass-by-reference I have to say I'm not really knowledgeable about this but I trust clang-tidy and HybridDog to do the right thing.

HybridDog:
I didn't know that std::move can be used for constructor parameters before I created this Pull Request with the help of clang tidy output. I assume std::move is advantageous because [an rvalue argument is moved instead of copied](https://stackoverflow.com/a/51706522) and the function signatures are shorter and indicate that an argument will be copied.

> considering some of them are passed by reference, it might have unintended side effects...

In my changes I used either std::move or a reference function parameter but not both at once unless I did a mistake.

PR #1644, opened 2021-01-15
little problem with a french translation and replacement of an old reference on the readme

hogren:
I detected one problem with a french translation. And, searching a documentation for the level editor, I found an old reference. I look into the git log to see that the documentation is on the wiki.


Semphriss:
The wiki is already accessible in the repository menu bar:
Untitled

For translations, please use [Transifex](https://www.transifex.com/arctic-games/supertux/).

PR #1648, opened 2021-01-18
Fix bag badguy animation

Rusty-Box:
Bag's bouncing and landing animation now works and looks as intended.


serano01:
Late to the party, this works for the bouncing snowballs and the Mr. Jumpy badguys, right?

Rusty-Box:
> Late to the party, this works for the bouncing snowballs and the Mr. Jumpy badguys, right?

Only for Jumpy. The bag badguy was originally what became Jumpy. All instances of bouncing snowball in this add-on have been reverted back to bouncy snowball for consistency to the old Revenge on Redmond.

serano01:
Okay. And thanks for your PR, I wasn't aware that it was (still) broken.

PR #1653, opened 2021-01-22
Added support for getting/setting the player's position

### LINKED ISSUE #1121: Scripting: Teleport to sector with the same position
I think it'd be cool to have a script that lets you teleport to different sectors of a level while keeping the same position.

Semphriss:
Closes #1121 

This allows to very precisely set Tux's position in a given sector.


tobbi:
Merci beaucoup! 👍

PR #1659, opened 2021-01-29
Refactored interface textbox

Semphriss:
- Improved selection: pressing left or right while having text selected will move the caret to the beginning or the end of the text selection, respectively (and clear the selection)
- Refactored the handling of the backspace key to make it similar to the handling of the delete key. This allows the usage of the regular update() function and makes controls compatible with Widgets again, in addition to make the code generally cleaner.


Semphriss:
See #1662 instead.

PR #1660, opened 2021-01-29
Fixed raw pointers in custom interface objects

Semphriss:
Some places in the custom interface controls used raw pointers. I replaced those with smart pointers, or no pointers at all where relevant.


Semphriss:
See #1662 instead.

PR #1662, opened 2021-01-29
UI improvements

Semphriss:
Includes #1659 and #1660.

List of changes:
- Added containers for UI components.
- Some places in the custom interface controls used raw pointers. I replaced those with smart pointers, or no pointers at all where relevant.
- Improved text selection: pressing left or right while having text selected will move the caret to the beginning or the end of the text selection, respectively (and clear the selection)
- Refactored the handling of the backspace key to make it similar to the handling of the delete key. This allows the usage of the regular update() function and makes controls compatible with Widgets again, in addition to make the code generally cleaner.


tobbi:
What is the status of this one? Can it get merged?

Semphriss:
Yes, in fact it's part of the editor revamp that I never finished. If I tagged it as ready for review, it should be ready for merge.

Issue #1635, opened 2021-01-10
Tux with Ice power can throw bullets through 1-tile wide walls


Semphriss:
Reproducible on any OS, on the lastest version @ master.

https://user-images.githubusercontent.com/66701383/104114612-36d48400-52fe-11eb-98c6-736807c7af95.mp4

HybridDog:
The problem does not happen if I change the x offset in the spawn position from 0 to 8 when facing left and from 32 to 24 when facing right in this line:
https://github.com/SuperTux/supertux/blob/c54014ac60f6f14fc96cadcc12f8b688268e89a2/src/object/player.cpp#L1115

tobbi:
I cannot reproduce this issue anymore with latest master. Closing.

PR #1665, opened 2021-02-07
Cheat Menu: Set a high number of the bonus count e.g. the number of fire bullets

HybridDog:
This can be helpful to test issues such as #1635.

### DIFF
--- a/src/supertux/menu/cheat_menu.cpp
+++ b/src/supertux/menu/cheat_menu.cpp
@@ -30,10 +30,10 @@ CheatMenu::CheatMenu()
   add_label(_("Cheats"));
   add_hl();
   add_entry(MNID_GROW, _("Bonus: Grow"));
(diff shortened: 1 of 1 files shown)


Semphriss:
Is there any interest in having cheats that give only x1 bonus anyways?

I thought instead of having an entry for each bonus type and another separate entry for giving a bunch of bonuses, we could make it so that the already existing entries directly give 64 bonuses.

(Also, for my personal comprehension - any particular reason for 61 specifically?)

HybridDog:
I thought for debugging purposes it could make sense to add only x1 bonus, e.g. to test the earth flower expiration time. However, actual powerup tests can also be done in an example level with the powerup, so I now think directly giving a 64x makes more sense.
I set the number to 61 because I can, i.e. there's no particular reason.

Semphriss (approved):

PR #1668, opened 2021-02-11
Expanding of the retro snow tileset

Rusty-Box:
- Several more tiles, plus redrawn snow pile background element
- Additional info block in retro style
- More fluid retro coin animation
- Adding more old powerup sprites along side coffee and other retro elements


serano01:
Will the reworked levels follow with a separate PR?

Rusty-Box:
@serano01 Yes, they will. I wanted to keep both in seperate PRs.

PR #1669, opened 2021-02-11
The Revenge In Redmond Revamp

Rusty-Box:
This revamp attempts to make the retro addon feel even more like the very old versions of SuperTux (e.g. v0.0.4 and v0.0.6) but with some better visuals and presentation.

All badguys aside the three original badguys Bsod, Laptop and Bag have been removed from the levels as well as some more references or easter eggs to past elements of Revenge In Redmond have been added. Addtionally, the cave level now uses the old M1 tileset rather than the current one.


serano01:
The second dot on the worldmap leads to a not available level named "you_know_rule" instead of "who_is_dawn". Also I think that the laptop should have a stunned animation. In the orginal RiR, he was fold in.

Rusty-Box:
> 
> 
> The second dot on the worldmap leads to a not available level named "you_know_rule" instead of "who_is_dawn". Also I think that the laptop should have a stunned animation. In the orginal RiR, he was fold in.

A small oversight on my part. I changed the level name pretty late. I'll fix go that. Regarding the laptop I agree. Its .sprite file wasn't designed around the iceblock badguy so the animation doesn't work properly. I'd rather do that in our master branch though. Not in this PR.

serano01:
Aside from that, I'm fine with this changes.

serano01:
I just noticed after merging that some decals are missing, for example the one at the end of the first level, or the chopper in "Get to da Choppa!".

Rusty-Box:
I get the first one since I forgot to remove it. It was  just for testing. I don’t know why the helicopter is missing. Unless I messed up the image path it should be there.

I’ll go check that asap.

Rusty-Box:
Okay, the level "Get to da Choppa!" seems to work fine. I assume the decal was missing since in this PR build the decal wasn't added yet while it is present in our main build. I have now also removed the unneeded decal from the first level so everything should be golden now.

PR #1670, opened 2021-02-12
[QoL] Climbable small tweak

Raoul1808:
This PR adds a limit for when Tux reaches the top of a climbable area, where he will stop before jumping out of the area.

This was originally made for ladders to make it easier to jump off the top, but this applies to every climbable zone.
Possible improvement: make this an option to all climbable zones to restrict *some* boundaries

### DIFF
--- a/src/object/player.hpp
+++ b/src/object/player.hpp
@@ -26,6 +26,7 @@
 #include "supertux/player_status.hpp"
 #include "supertux/sequence.hpp"
 #include "supertux/timer.hpp"
+#include "trigger/climbable.hpp"
 #include "video/surface_ptr.hpp"
 
 class BadGuy;
--- a/src/object/player.hpp
+++ b/src/object/player.hpp
@@ -26,7 +26,6 @@
 #include "supertux/player_status.hpp"
 #include "supertux/sequence.hpp"
 #include "supertux/timer.hpp"
-#include "trigger/climbable.hpp"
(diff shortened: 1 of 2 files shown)


serano01:
Is this related to the changes which were made here? https://github.com/SuperTux/supertux/pull/1473

src/object/player.hpp
```diff
@@ -26,6 +26,7 @@
 #include "supertux/player_status.hpp"
 #include "supertux/sequence.hpp"
 #include "supertux/timer.hpp"
+#include "trigger/climbable.hpp"
 #include "video/surface_ptr.hpp"
```
Semphriss, line 30: No need to include a file in the header if it's only used in the body; include in the body instead. If we include too much in the headers, we risk creating a dependency loop

Semphriss (changes requested):

src/object/player.hpp
Raoul1808: I thought it would be in the header since there is a `Climbable&` definition for `m_climbing`. I'm fixing that right now

Semphriss: If it wasn't there then there is no need to add it. Most of the time, when you need to add a certain class, you can do just `Class NameOfTheClass;` in the header, since you don't actually use the class in the header, and you just include in the body.

Zwatotem:
This is working as intended, and the code seems fine. I'm merging.

PR #1673, opened 2021-02-13
Added Windows workflow

Semphriss:
Builds and uploads a debug/release installer for Windows.

TODO: Upload installer files to the nightlies, @maths22 perhaps you can help me there?

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -26,7 +26,7 @@ on:
   pull_request: {}
 # TODO the glbinding build
 jobs:
-  build-supertux:
+  build-unix:
 
     strategy:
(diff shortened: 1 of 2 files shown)


maths22 (commented):
Generally looks good to me.  If you more or less copy lines 131 to 148 that should work for posting to the download.supertux.org.  You will probably need to add an explicit `shell: bash` to the `Post uploaded file` part, but it should more or less otherwise work.  You may need to make a tweak to that script to find msi file names

Semphriss:
I didn't know Windows supported the Bash shell :D

Alrighty, I'll look into that whenever possible. Thanks for the review!

PR #1679, opened 2021-02-24
 Bumper: push the player only if it moves against the bumper 

HybridDog:
I changed the bumper so that it only pushes the player if he/she moves against it:
https://user-images.githubusercontent.com/3192173/109016663-ab992d00-76b6-11eb-956b-065835691d66.mp4

This fixes a problem where the player is sometimes bumped many times by one bumper and somehow looses its new velocity. It also changes the general behaviour of bumpers.

### DIFF
--- a/src/object/bumper.cpp
+++ b/src/object/bumper.cpp
@@ -33,11 +33,11 @@ Bumper::Bumper(const ReaderMapping& reader) :
   physic(),
   left()
 {
-	reader.get("left", left);
+  reader.get("left", left);
   m_sprite->set_action(left ? "left-normal" : "right-normal");
-	physic.enable_gravity(false);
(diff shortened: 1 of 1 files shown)


weluvgoatz:
The other new PR solves a lot of issues with the bumper, and I don't really like the changes introduced by this PR since it makes the bumper feel less impactful than before.  This PR doesn't seem to solve any of the actual problems with bumpers and only changes things for the sake of changing...

weluvgoatz:
The indent improvements are good though.

HybridDog:
I mainly opened this PR because I noticed that sometimes when Tux is bumped many times at once, the velocity is somehow set to zero. There's a new PR for bumper changes, so I can close this.

PR #1680, opened 2021-02-25
Changed handling of Tux movement

Semphriss:
- Removed speed ceiling that gets applied whenever the player presses a certain direction. Now, limits are applies regardless of what the player presses, and the speed is brought back within limits smoothly rather than instantly.
- Added a boost mechanism, to help Tux gain extra velocity above speed limits for a short amount of time (useful for bumpers).
- Fixed the one-wall walljump-climbing bug.

### DIFF
--- a/src/object/bumper.cpp
+++ b/src/object/bumper.cpp
@@ -67,7 +67,8 @@ Bumper::collision(GameObject& other, const CollisionHit& hit)
   if (player)
   {
 	  float BOUNCE_DIR = left ? -BOUNCE_X : BOUNCE_X;
-	  player->get_physic().set_velocity(BOUNCE_DIR, BOUNCE_Y);
+	  player->get_physic().set_velocity(0.f, BOUNCE_Y);
+    player->sideways_push(BOUNCE_DIR);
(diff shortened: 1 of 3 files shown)


Semphriss:
@Zwatotem Any comments? I can't get very far with just an emoji :^)

Zwatotem:
I just liked the climbing bug. But that's okay if you fixed it. Can't give feedback on the rest right now.

Semphriss:
I removed the climbing bug because it opened the door to poor level design, both by level designers that will expect players to use it and those who don't.

We also discussed about Tux's moveset the other day, and we agreed that double-wall walljumping would be the ceiling in terms of mobility for Tux. (Ngl, I would prefer not have a fat penguin have the moveset of a ninja :eyes:)

weluvgoatz:
This feature is good, I approve of the merge but I'll wait for others to chime in. :)

weluvgoatz:
I remember this PR working quite well, are there any objections to it?

Semphriss:
Merge conflicts are solved, so whenever you want :^)

weluvgoatz:
Gonna merge this now, thanks.

PR #1687, opened 2021-03-06
Bugfixes in the particle editor

Semphriss:
- [x] Fixed bug mixing slashes and backslashes in Particle Editor filenames
- [ ] ~~Investigate on the particles not spawning under specific circumstances~~


Semphriss:
Since I couldn't reproduce the second bug in a long time, I'll let this PR be merged and open another if the bug is discovered again.

PR #1689, opened 2021-03-07
Added scripting to rotating spotlight object + some changes

### LINKED ISSUE #768: Spotlight: expose to scripting API

Semphriss:
Fixes #768 

Added setters, faders, and easers for angle, rotation speed, and light color.

Also changed the direction variable to consider not rotating at all.

Also changed GameObjects to handle FadeHelpers globally. This will require a refactor of all GameObject-inheriting classes to call GameObject::update(float) where necessary.

Consideration: Although the "direction" property seems needless (as clockwise/counter-clockwise rotation can be achieved using positive or negative values), I chose to leave it there so that scripters can easily change the direction of a spotlight - or stop it completely - without having to know/remember its original speed in some way if they want to reset it at a later point.

### DIFF
--- a/src/object/gradient.hpp
+++ b/src/object/gradient.hpp
@@ -17,8 +17,8 @@
 #ifndef HEADER_SUPERTUX_OBJECT_GRADIENT_HPP
 #define HEADER_SUPERTUX_OBJECT_GRADIENT_HPP
 
-#include "squirrel/exposed_object.hpp"
 #include "scripting/gradient.hpp"
+#include "squirrel/exposed_object.hpp"
 #include "supertux/game_object.hpp"
 #include "video/drawing_context.hpp"
 
--- a/src/object/spotlight.cpp
+++ b/src/object/spotlight.cpp
@@ -20,7 +20,38 @@
 #include "sprite/sprite_manager.hpp"
 #include "util/reader_mapping.hpp"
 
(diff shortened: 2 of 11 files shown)


Semphriss:
Fixes #768

weluvgoatz:
This works great! I do have a small suggestion though; the "time" and "angle" parameters seem to be reversed in the scripting functions sometimes (see lines 59-62 of the src/scripting/spotlight.cpp, but this issue shows up throughout the file).

I don't know if this is intentional, as normal players won't see the difference if it's documented one way or another, but if it ends up being an oversight it should be fixed.  Either way, I will merge it once that is fixed or turns out to be intentional.  Thanks for this PR!

Semphriss:
image

If you are refering to this, it's normal; it's because the fade helper expects time to be first, but in scripting, I wanted to make it consistent (value is always the first argument, time is the second, and easing the third) to make it easier for scripters.

I double-checked, and arguments are ordered properly everywhere :D

weluvgoatz:
Thanks for clarification, and since it all works as intended, I'll merge this.

PR #1690, opened 2021-03-07
Worldmap tux swimming animation no longer hardcoded

Semphriss:
Worldmap animation handling improvements. Now any sprite can support up/down/left/right animations, it they don't exist it fallbacks to the generic directionless animation, if it doesn't exist either it fallbacks to not changing the animation.


weluvgoatz:
Looks good, will merge now.

PR #1699, opened 2021-03-14
Fix explosion sprite FPS.

GreenMouseFan:
A small pull request fixing explosion sprite framerate. Now it should explode for approx 1 second.

### DIFF
--- a/data/images/objects/explosion/explosion.sprite
+++ b/data/images/objects/explosion/explosion.sprite
@@ -1,7 +1,7 @@
 (supertux-sprite
   (action
     (name "default")
-    (fps 15.0)
+    (fps 8.0)
     (hitbox 36 36 48 48)
     (images 
       "explosion-0.png"


weluvgoatz:
What? There is literally nothing wrong with it, don't know why this was made...

GreenMouseFan:
There was a change to FPS of the animation but they do it to the wrong sprite and explosions lasted for less than a second as a result. I think this fix will allow more time for explosions and making the game easier.

Semphriss:
At 8 FPS, the animation is too jerky and irregular. It cannot fit with its graphical surroundings.

I haven't seen any other complaints about the bomb so far, but if we are to adjust it, we should reduce the cooldown time rather than the FPS. I have an easier time understanding that the bomb's explosion shouldn't hurt after 0.75 seconds than that poor graphical performance is worth the shot.

GreenMouseFan:
Maybe 12 fps, cuz there are now 12 frames of animation?

tobbi:
@SuperTux/graphics-designers I need your input on this one.

weluvgoatz:
I would prefer closing it.

tobbi:
Can I get the two sprites side-by-side with the different fps'es?

PR #1700, opened 2021-03-14
Temporary fix to visually show blocked paths

### LINKED ISSUE #1694: "A fork in the road"
**SuperTux version:** 0.6.2-616-gdd472a4a1
**System information:** NixOS-unstable updated within the past few weeks

##### Expected behavior

Completing "A Fork in the Road" with one of the paths, then completing the other levels around the loop in a counter-clockwise direction, allows returning to "A Fork in the Road" from the other direction.

##### Actual behavior

Tux gets stuck on the road and doesn't return to "A Fork in the Road".

image

worldmapfinish() in the console, complete the lower path in "A Fork in the Road", try to circle around.

##### Proposed solution

The easiest way to fix this is to hide the whole road, so that Tux doesn't exit from the map. But then we need special unexplored-road tiles to show that a path exists.

Semphriss:
Fixes #1694

image

I agree it's not the most beautiful addition, but at least it'll give visual cues as for which paths are open.


Rusty-Box:
Perhaps we could simply turn the blocked path slightly translucent instead?

PR #1704, opened 2021-03-18
Swimming Improvements

weluvgoatz:
This Pull Request fixes many if not most of the problems introduced by the inclusion of swimming into SuperTux.  Please review the PR so that it can be merged when ready!  Note that the graphics were not changed, as this will happen when the Tux graphics revamp is done, and many examples of improvable code were not changed in order to remain consistent until an overall code quality improvement is done.


mbernard2:
I noticed it's easier to jump out of the water (especially in *Shocking* which was incredibly difficult before this PR) but it's still too difficult sometimes to reach higher platforms. Take a look :

st-shocking

If you don't have enough speed when trying to reach the platform, it's pretty much too late...

Perhaps we could give Tux a boost no matter its current velocity every time the player presses Jump while swimming ?

Or keep it as it is, but then adjust the levels to make sure they're not too hard ?

What do you guys think ?

Zwatotem:
When you use air flower and jump out of the water pressing the spacebar all the time Tux will start to glide, which looks exremally silly with the way his sprite is actually rotating.

weluvgoatz:
Since this has gotten approval from RustyBox and Semphris (the two currently most active developers by far), I will merge it.  Finally!

PR #1705, opened 2021-03-18
Major improvements to UBports build

Semphriss:
The game actually runs on mobile devices now (those supporting SDL at least), and on all architectures.

There is also a beginning of mobile interface (the escape button is buggy, though). The game is more or less playable for people who are used to mobile controls.

TODO:
- [x] Actually make a decent-looking interface for mobile.
- [x] Disable the mobile interface inside the level editor (will be done when the new editor will arrive).
- [x] Fix the escape button calling pressed() repeatedly even when holding the button constantly.
- [x] Generally speaking, fix the code quality:
  - [x] The `UBUNTU_TOUCH` compiler option used in the C++ code at some places should actually be a more general variable, like `MOBILE` or similar, for places where iOS and Android should have the same features.
  - [x] There are a few TODO's, FIXME's and poor code practices (hardcoding) that might need to be fixed.


tobbi:
🍗

PR #1708, opened 2021-03-27
Added color support for Will'O'Wisps

Semphriss:
Will'O'Wisps can now be given any color, and empty lanterns will be filled with the given color.

### DIFF
--- a/src/badguy/willowisp.hpp
+++ b/src/badguy/willowisp.hpp
@@ -71,7 +71,7 @@ public:
   /** make WillOWisp vanish */
   void vanish();
 
-  Color get_color() { return m_color; }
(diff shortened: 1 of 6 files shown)


src/badguy/willowisp.hpp
```diff
@@ -70,6 +71,8 @@ class WillOWisp final :
   /** make WillOWisp vanish */
   void vanish();
 
+  Color get_color() { return m_color; }
```
tobbi, line 74: const

Semphriss: consted :+1:

tobbi:
Thanks! Disc'o'wisps!!!

PR #1713, opened 2021-03-28
Skull tiles now respawn 10 seconds after crumbling

### LINKED ISSUE #1712: Unstable tiles should be respawning
If you mess up a section that relies on unstable tiles to cross a bottomless pit (for example: a bomb explodes, destroying them or just bad timing on the players end) and the unstable tiles are gone you basically have to restart the level.

Instead unstable tiles should respawn after a certain amount of time or after being offscreen.

Semphriss:
Fixes #1712

### DIFF
--- a/src/object/skull_tile.cpp
+++ b/src/object/skull_tile.cpp
@@ -24,7 +24,7 @@
 
 static const float CRACKTIME = 0.3f;
 static const float FALLTIME = 0.8f;
-static const float RESPAWNTIME = 10.f;
+static const float RESPAWNTIME = 5.f;
 static const float FADETIME = 0.5f;
 
 SkullTile::SkullTile(const ReaderMapping& mapping) :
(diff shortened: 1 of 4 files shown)


serano01:
That changes the skull tiles, but what about the unstable tiles? They are more used after all. And if I'm not mistaken, skull tiles are unaffected by explosions.

Semphriss:
Odd, I thought both were the same thing. Was there a reason to design both seperately at the moment of creation?

serano01:
Skull tiles only fall when the player has been on it for a sufficient time, unstable tiles on the other hand will always fall regardless on the time.

Semphriss:
@serano01 Done; you may test changes to see if everything is good this way. Unstable tiles now respawn too, and both skulls and unstable tiles respawn after 5 seconds.

PR #1714, opened 2021-03-28
Added new lightmap tiles

Semphriss:
New ones fade from black/white to transparent. It also includes corner tiles.


weluvgoatz:
This works as intended, merging now.

PR #1722, opened 2021-04-13
Editor: Put tiles without holes when holding the left mouse button

HybridDog:
When the user puts many single tiles while holding the left mouse button and moves the cursor quickly, tiles are now also put at positions between two successive cursor positions.
The line interpolation is disabled when there's a relatively long time between two successive movements to avoid undesired behaviour when the frame rate is very low or the user scrolls around without releasing the left mouse button.

Current master branch:
https://user-images.githubusercontent.com/3192173/114600623-d81ffd00-9c94-11eb-84ad-2034053496f1.mp4

Proposed changes (without the time check):
https://user-images.githubusercontent.com/3192173/114600651-deae7480-9c94-11eb-9219-404adb06f6a9.mp4

To do:
* Test it thoroughly
* Code style related changes if needed

### DIFF
--- a/src/editor/overlay_widget.cpp
+++ b/src/editor/overlay_widget.cpp
@@ -57,10 +57,11 @@ bool EditorOverlayWidget::alt_pressed = false;
 EditorOverlayWidget::EditorOverlayWidget(Editor& editor) :
   m_editor(editor),
   m_hovered_tile(0, 0),
-  m_hovered_corner(0, 0),
+  m_hovered_tile_prev(0, 0),
   m_sector_pos(0, 0),
   m_mouse_pos(0, 0),
   m_previous_mouse_pos(0, 0),
+  m_time_prev_put_tile(std::chrono::steady_clock::now()),
   m_dragging(false),
   m_dragging_right(false),
   m_scrolling(false),
@@ -243,8 +244,9 @@ EditorOverlayWidget::input_autotile_corner(const Vector& corner, uint32_t tile, 
 }
 
 void
-EditorOverlayWidget::put_tile()
+EditorOverlayWidget::put_tile(const Vector& target_tile)
 {
+  Vector hovered_corner = target_tile + Vector(0.5f, 0.5f);
   auto tiles = m_editor.get_tiles();
   Vector add_tile(0.0f, 0.0f);
   for (add_tile.x = static_cast<float>(tiles->m_width) - 1.0f; add_tile.x >= 0.0f; add_tile.x--) {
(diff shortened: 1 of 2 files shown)


Semphriss:
Quick bug: Corner autotiling no longer works as expected. Reproducible by attemping to autotile these tiles before vs after PR: 

image

`m_hovered_corner` is not simply the currently hovered corner + 0.5 in each dimension, it's the corner between 4 tiles that's the closest to the mouse position. The easiest way would be to keep `m_hovered_corner` and use it for autotiling, and use the other variables for your other purposes.

HybridDog:
Thank you for the testing.
```c++
  m_hovered_tile = sp_to_tp(m_sector_pos);

  float half_tile = 16.f;
  m_hovered_corner = sp_to_tp(m_sector_pos + Vector(half_tile, half_tile));
[…]

Vector
EditorOverlayWidget::sp_to_tp(const Vector& sp, int tile_size) const
[…]
  Vector sp_ = sp - tilemap->get_offset();
  return sp_ / static_cast<float>(tile_size);
}

```
If I understand the code correctly, `sp_to_tp(m_sector_pos + Vector(half_tile, half_tile))` is the same as `m_hovered_corner + Vector(0.5f, 0.5f)` because `tile_size` is 32.

I changed it so that it now calculates a line on a finer grid. As far as I can see, corner autotiling works now.

Semphriss:
> If I understand the code correctly, `sp_to_tp(m_sector_pos + Vector(half_tile, half_tile))` is the same as `m_hovered_corner + Vector(0.5f, 0.5f)` because `tile_size` is 32.

It isn't; `sp_to_tp` converts a **s**creen **p**oint to a **t**ile **p**osition. This ensures that the hovered tile is always a valid tile.

`Vector hovered_corner = target_tile + Vector(0.5f, 0.5f);`

`target_tile` is already a valid tile; therefore, adding half a tile can only give a position between tiles, which is not what the autotile function expects.

I really recommend you keep `m_hovered_corner` for readability, and also because I'm currently revamping the autotile system and I expect to use that variable for other purposes.

Semphriss:
So I tested a bit...

Dragging around sometimes leaves holes in the strikes:

https://user-images.githubusercontent.com/66701383/115041178-3967ec00-9ea0-11eb-9fbb-c569fb88b503.mp4


... and I can confirm corner autotiles still don't work:

https://user-images.githubusercontent.com/66701383/115041227-44bb1780-9ea0-11eb-9cd9-3c85f5de7d06.mp4


I think the universe doesn't like that solution either :smile:
I think the best would be to keep variables for autotiling and those for this PR seperated. To come back on my comment above, it'd be easier to keep `m_hovered_tile`; if you remove it, you'd have to redo the same calculations with `m_sector_pos` and `sp_to_tp` anyways, which require passing through a member variable IIRC.

HybridDog:
> It [is not the same]; sp_to_tp converts a screen point to a tile position. This ensures that the hovered tile is always a valid tile.

The only difference I can see is the existence check for the tilemap.

Here is a comparison between `m_hovered_corner + Vector(0.5f, 0.5f)` and `sp_to_tp(m_sector_pos + Vector(half_tile, half_tile))`:
```c++
// sp_to_tp(m_sector_pos + Vector(half_tile, half_tile))
Vector sp_ = m_sector_pos + Vector(half_tile, half_tile) - tilemap->get_offset();
return sp_ / static_cast<float>(tile_size);
	// = (m_sector_pos + Vector(16.0, 16.0) - tilemap->get_offset()) / 32.0
	// = (m_sector_pos - tilemap->get_offset()) / 32.0 + Vector(0.5, 0.5)

// m_hovered_tile = sp_to_tp(m_sector_pos);
Vector sp_ = m_sector_pos - tilemap->get_offset();
return sp_ / static_cast<float>(tile_size);
	// = (m_sector_pos - tilemap->get_offset()) / 32.0

// m_hovered_corner + Vector(0.5f, 0.5f)
(m_sector_pos - tilemap->get_offset()) / 32.0 + Vector(0.5f, 0.5f)
```


> I really recommend you keep m_hovered_corner for readability, and also because I'm currently revamping the autotile system and I expect to use that variable for other purposes.
[…] I think the best would be to keep variables for autotiling and those for this PR seperated.

The Pull Request is still a Draft, so I can change most of the code when needed once it works.
I think I need to use both the target_tile and hovered_corner for this Pull Request so that autotiling works.


> Dragging around sometimes leaves holes in the strikes

I cannot reproduce this. If the std::chrono time between two movements is large (> 70 ms), e.g. due to low FPS, strikes are omitted. Otherwise holes shouldn't appear.


> and I can confirm corner autotiles still don't work

I cannot reproduce the problem; it works for me:
https://user-images.githubusercontent.com/3192173/123839359-52affe80-d90d-11eb-965f-7c2549748811.mp4


Sorry for the late response.

Semphriss:
Well, my turn to apologize for the slightly greater delay :smile: 

@HybridDog How should we proceed with this PR?

HybridDog:
I don't know how to proceed. For now I've rebased it and it still works for me.

Semphriss:
Can't reproduce either bug anymore, assuming it's either fixed somehow, or that it was a punctual bug. I'll open an issue if that ever happens again.

Should be good now.

Semphriss:
@HybridDog With the bugs out of the way, is there anything else to do before merging? I'll leave it to you to mark this PR as ready for review when you're ready.

HybridDog:
In my opinion, the code changes are ready. Is the removal of the member variable `m_hovered_corner` still a problem for you?

Semphriss:
I have abandoned trying to enforce proper code quality, given the lack of standards to follow and the lack of expertise to maintain them. I'll leave it to the generic principle: if it works, it works.

PR #1728, opened 2021-04-19
Move SuperTux to C++20

Grumbel:
This is restoring #1602 back into a compilable state. Let's see what breaks.


Grumbel:
Needs a bit of cleanup, but everything outside of UBPorts compiles with C++20 set. UBPorts is a bit problematic, as it's stuck on an ancient Ubuntu 16.04, which is EOL in a week. No idea how far away the UBPorts 20.04 update is or if there are any PPAs that could give us a newer gcc/clang on that system.

Semphriss:
Consideration: Ubuntu 18.04 can no longer compile SuperTux out of the box. Neither CMake nor the default GCC, as given by the repositories, support C++20; for the compiler, Canonical provides later versions of Clang; for CMake, I couldn't find later versions that support C++20.

I have no idea how other distros are affected, as I only run Ubuntu 18.04 and 20.04.

Grumbel:
I am not too worried about Ubuntu 18.04, as newer compiler are easily available from PPA and we have appimage/flatpak/guix packages on top of that.

But UBPorts will be difficult, as there doesn't seem to be an easy way to get a new compiler onto that. We might need to compile one ourselves.

src/supertux/main.cpp
```diff
@@ -265,20 +264,20 @@ std::string olduserdir = FileSystem::join(physfs_userdir, PACKAGE_NAME);
...
+  std::filesystem::path userpath(userdir);
 
-  boost::filesystem::directory_iterator end_itr;
+  std::filesystem::directory_iterator end_itr;
 
   bool success = true;
 
   // cycle through the directory
-  for (boost::filesystem::directory_iterator itr(olduserpath); itr != end_itr; ++itr) {
+  for (std::filesystem::directory_iterator itr(olduserpath); itr != end_itr; ++itr) {
```
jwakely, line 275: `filesystem::directory_iterator` can be used directly with range-based `for`:

```suggestion
  for (auto file : std::filesystem::directory_iterator(olduserpath)) {
```

src/supertux/main.cpp
```diff
@@ -265,20 +264,20 @@ std::string olduserdir = FileSystem::join(physfs_userdir, PACKAGE_NAME);
...
   try
   {
-    boost::filesystem::rename(itr->path().string().c_str(), userpath / itr->path().filename());
+    std::filesystem::rename(itr->path().string().c_str(), userpath / itr->path().filename());
```
jwakely, line 278: Why create a temporary string by calling `string()` and then get a `const char*` instead of just `itr->path.c_str()`?

But why get a `const char*` from it at all, when that just has to be converted to a new `filesystem::path` (which allocates more memory and parses the path again)?

```suggestion
    std::filesystem::rename(itr->path(), userpath / itr->path().filename());
```

src/supertux/game_session.cpp
```diff
@@ -16,6 +16,8 @@
 
 #include "supertux/game_session.hpp"
 
+#include <cfloat>
+
```
jwakely, line 20: This change is needed independent of any move to C++20, see #1804

Semphriss:
Related:

I've investigated supporting the Steam Linux Runtime to fix #2112 and #2083, everything went well until the compilation hit the missing Boost dependencies. It appears SLR doesn't include Boost, and [the documentation instructs not to add any extra dependency](https://gitlab.steamos.cloud/steamrt/scout/sdk/-/blob/steamrt/scout/README.md#apt-packages).

Since this PR seems to remove all dependencies to Boost, I believe I'll have to wait until this gets merged before resuming my work on the Steam version.

mrkubax10:
This is probably replaced by #2473

PR #1730, opened 2021-04-21
Replace boost::format with fmt

Grumbel:
This goes together with #1728. The fmt library would be a temporary stand-in for `std::format()`, as that isn't yet widely enough available. Both share more or less the same API and format strings, while `boost::format` is quite a bit different.

This change is necessary as `boost::format` in boost-1.72 has a conflict with g++-10, which affects quite a few distributions, see https://github.com/SuperTux/supertux/pull/1728/commits/5f484f49e7ef00170d63be0bb06f0f7958fafabd.

This also breaks UBPorts as libfmt-dev only shows up in 18.04. This shouldn't be too hard to fix, but UBPorts will break  due to C++20 anyway.

This patch would change a few strings that are used in translation.


Grumbel:
As this doesn't trivially compile on UBPorts I moved it over into #1728, no point in doing this in isolation without also moving to C++17/20.

Issue #1725, opened 2021-04-16
Idea: More lenient spike hitboxes


weluvgoatz:
As of the latest commit to the swimming revamp branch, the game can now detect if a certain tile type is within a certain range of a certain object.  Once that is merged, we should add in a feature where the hitbox that spikes are able to touch for killable objects is smaller than the static hitbox, making it more lenient (badguys and Tux will have to actually be in the spike somewhat instead of just lightly breezing by it, possibly not even seeming to touch it).
This would also fix that one bug where Tux standing on a rock gets hurt by spikes even though the rock is covering the spike.

I had this idea already but I'm logging it here so it does not get lost.  Any feedback to this idea?

Semphriss:
I support!

Many other games I know have a certain margin of contact and it turns out to play quite nicely. I'd be interested to see what it plays like in SuperTux; I know for instance that sometimes Tux standing exactly 1 tile above spikes (rock overlapping with spikes, tiles next to the spikes, etc.) and it happened that Tux took damage when unexpected. It'd be good to see that gone.

Issue #1175, opened 2019-09-29
Dangerous tiles hurt Tux even when they are blocked off


HybridDog:
**SuperTux version:** v0.5.1-1937-g4d1c419ea
**System information:** lubuntu

##### Expected behavior

![2019-09-29-113032_1920x1080_scrot](https://user-images.githubusercontent.com/3192173/65830124-cf332a80-e2ac-11e9-80b4-49eb8c2f9af2.png)
Tux does not get hurt when he jumps onto the rock

##### Actual behavior
The spikes kill Tux although they are blocked by a rock object, which is as high as one tile

##### Steps to reproduce actual behavior
See the screenshot, jump onto the rock

tobbi:
I feel like this is a collision rect problem where the collisiion rect of the spikes is maybe 1 or 2px larger than that of the block.

HybridDog:
If I understand the collision system correctly, it firstly moves objects to target positions, which depend on the objects' velocities. Then it tests if the object at the target position intersects with a tile, and after that it tests collision with other objects.
https://github.com/SuperTux/supertux/blob/3827b0555c294e69f100525f7360b2b652362f4d/src/collision/collision_system.cpp#L494
If it works this way, Tux would also get hurt when his velocity is high enough and the rock is bigger than the spikes.

tobbi:
That doesn't seem to be a good idea. If that's indeed the case, we'd have to change it.

HybridDog:
Here's a similar problem:
When I press left while falling right of a wall and then release left before the end of the wall, Tux nonetheless gets hurt by the spike:
https://user-images.githubusercontent.com/3192173/105752916-2f40ec00-5f48-11eb-99d6-3bc4f7ea88b0.mp4

PR #1741, opened 2021-05-02
More lenient spike hitboxes

weluvgoatz:
This is a little PR that adds a bit of leniency to when Tux and badguys hit spikes and are actually hurt by them.  This makes the game a little more comfortable to play, especially in cases when Tux is right up against spikes (See issues #1725 and #1175, which would be fixed by this PR).

Only thing is, there has got to be a better method to defining these "hitboxes"... but if that isn't an issue then don't consider it one.


Rusty-Box:
Seems good. Tried reproducing ##1741 with it and it seems to have fixed that.

PR #1750, opened 2021-05-12
Fix trampoline bug

mbernard2:
Tux can now jump on trampolines stacked over other objects.

Cause of the bug : no proper collision was detected between the player and the trampoline in some circumstances.


mbernard2:
Hold on, info blocks just stopped working because of this.

Semphriss:
@mbernard2 Is there any update about this PR? The bug is still present and frequently reported, so it would be cool to get it merged whenever possible.

mbernard2:
Sorry for the late response. Unfortunately I don't think I will have time to work on SuperTux anytime soon.

weluvgoatz:
Closing due to lack of inactivity.  It can be returned to, if the author eventually wants.

PR #1753, opened 2021-05-19
Fix glm check failing on Nix and EmuELEC

### LINKED ISSUE #1752: After the change to GLM it now fails to compile
**SuperTux version:** 221e10b77e1ce2591fcfee91199b1e1e909507e4
**System information:** EmuELEC aarch64

##### Expected behavior
Finish compiling
##### Actual behavior
Errors out in 
```
CMake Error at CMakeLists.txt:893 (add_executable):
  Target "supertux2" links to target "glm::glm" but the target was not found.
  Perhaps a find_package() call is missing for an IMPORTED target, or an
  ALIAS target is missing?
  ```
  Even though glm is reported as found `-- Found glm` and it is installed

```
cmake -DBoost_NO_BOOST_CMAKE=ON -DENABLE_OPENGL=ON -DENABLE_OPENGLES2=ON -DBUILD_DOCUMENTATION=OFF -DENABLE_DISCORD=Off -DCMAKE_BUILD_TYPE=Release -DIS_SUPERTUX_RELEASE=ON
```

compiles and works fine on: 9bc3014b7598c41a71945d8819dafe342a83ad79

Grumbel:
Fixes #1752

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -176,7 +176,16 @@ link_directories(${Boost_LIBRARY_DIRS})
 
 find_package(glm QUIET)
 if(glm_FOUND)
-  message(STATUS "Found glm")
+  if(TARGET glm::glm)
+    message(STATUS "Found glm")
+  else()
+    # fallback for old glm version in EmuELEC and Nix
+    get_target_property(GLM_INCLUDE_DIR glm INTERFACE_INCLUDE_DIRECTORIES)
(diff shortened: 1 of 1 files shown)


weluvgoatz:
I get the error that the ALIAS target "glm::glm" cannot be created because target "glm" is imported but not globally visible.

Grumbel:
Older versions of cmake don't allow aliases on imported targets. We can either up the requirement to cmake 3.11 (or 3.18?) or do the aliasing manually.

PR #1755, opened 2021-05-20
Fix for hat sprite desync

### LINKED ISSUE #1719: Hat sprite desyncs when collecting powerup
This issue has been discussed, but I will file it here to keep track of it.
In the current nightly build version of SuperTux (any OS) Tux's hat desyncs when he collects a powerup.  The hat sprite starts at its first frame but the body starts at the frame that it is already at, causing the hat to jitter and desync until Tux takes a new action.  This is especially noticeable when swimming, but happens regardless of his swimming status.  I hope this can be fixed!!!

weluvgoatz:
This pull request fixes the bug when swimming where the powerup hats would desync from the body when collected, producing an odd jittering animation.  I made it to where the hat sprite is forced into the same frame and frame progress as the body, but only when they share the same amount of frames in order to avoid bugs with other actions.  Also, this has been applied only to the player.  It would close #1719.

### DIFF
--- a/src/object/player.cpp
+++ b/src/object/player.cpp
@@ -1683,8 +1683,20 @@ Player::draw(DrawingContext& context)
   if (m_player_status.has_hat_sprite())
   {
     m_powersprite->set_action(m_sprite->get_action());
+    if (m_powersprite->get_frames() == m_sprite->get_frames())
+    {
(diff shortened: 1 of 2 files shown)


tobbi:
Is this good to go?

weluvgoatz:
Yes, it has been for quite a while.

PR #1756, opened 2021-05-21
Walljump Fixes and Improvements

weluvgoatz:
This quick PR makes horizontal and vertical movement while wall jumping more configurable, as well as fixing a bug where Tux would appear to cling to the air sometimes.

### DIFF
--- a/src/object/player.cpp
+++ b/src/object/player.cpp
@@ -942,7 +942,7 @@ Player::do_backflip() {
 
 void
 Player::do_jump(float yspeed) {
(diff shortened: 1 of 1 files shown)


tobbi:
clang does not like this, can you fix?

weluvgoatz:
I'm not sure if I can fix this since it seems clang has an issue with the Discord integration, which I did not touch during this PR.  I might be wrong, though.

Rusty-Box:
Looks good! Not sure what's the staus on the "clang" thing but outside of that I'd say it is ready for merging

Issue #924, opened 2018-10-27
Ideas for making Haywire more challenging


Rusty-Box:
I personally always saw the enemy Haywire as a more difficult and "stronger" version of Mr. Bomb. However, in my opinion Haywire does not seem more challenging than his blue counterpart. Sure, he can run after being hit but even then once the player for example jumps over a small gap, Haywire cannot follow which is pretty boring as due to complex level design this and many occasions where Haywire cannot follow will happen a lot making Haywire not really diffrent from Mr. Bomb. So here are three ideas that would make Haywire a little bit more challenging:

1. Let Haywire be able to jump over a short distance or from higher platforms (similar to captain snowball) when running after Tux so he can follow the player awhile longer. (Would require additional sprite work)
2. Maybe increase their speed a little bit so it is a bit harder to out run them.
3. Make Haywire avoid enemies when chasing Tux so Haywire does not get turend around when bumping into another enemy on their path. An idea for that could be that Haywire jumps over them or is simply not affected by their hitboxes (If possible).

brmbrmcar:
Maybe a speed increase could work, but the other two look like they'd make some things impossible and make level design harder. Older levels may break especially because of the first idea. I'd support the idea that there would be a customisable speed. But I think in cases where they're meant to be hard, they work quite well as they are, and when they're meant to be easier on longer platforms or just not hellish levels they are. It works quite well in my opinion.

Rusty-Box:
It wouldn’t  make level design harder as they would be similar to captain snowball then, movement wise. And using them isn’t hard too, isn’t it. Besides it's the level creator’s responsibility to make their levels work with the latest version of the game, not necessarily only ours.

brmbrmcar:
It isn't just incompatibility, it's the removal of a potentially vital feature. It's not the job of the level creator to predict the future of how the game might change and completely ruin a level.

weluvgoatz:
Can this be closed?

PR #1757, opened 2021-05-24
Haywire Updates

weluvgoatz:
This PR makes the haywire more effective in his pursuit of Tux by turning around at a faster speed and chasing him slightly faster as well.  Also, I have fixed the bug where the recently stunned haywire would start drifting before he appeared to be running.

IMPORTANT EDIT: I decided to try and include the requested changes of #924 in this, as well!

### DIFF
--- a/src/badguy/haywire.cpp
+++ b/src/badguy/haywire.cpp
@@ -183,7 +183,7 @@ Haywire::active_update(float dt_sec)
     else if (p && time_stunned == 0.0f)
     {
       /* Player is on the right or left*/
-        target_velocity = (p->get_pos().x > get_pos().x) ? walk_speed : (-1.f) * walk_speed;
+      target_velocity = (p->get_pos().x > get_pos().x) ? walk_speed : (-1.f) * walk_speed;
     }
 
     WalkingBadguy::active_update(dt_sec, target_velocity, 3.f);
@@ -303,7 +303,9 @@ HitResponse Haywire::collision_badguy(BadGuy& badguy, const CollisionHit& hit)
     return FORCE_MOVE;
   }
   else
+  {
     WalkingBadguy::collision_badguy(badguy, hit);
+  }
   return ABORT_MOVE;
 }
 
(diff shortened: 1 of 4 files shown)


Rusty-Box:
I think this is pretty good. Haywire is much more mobile. A personal nitpick would be his jump height/distance.
I was thinking more like how far captain snowball jumps but the way it is now I'd be able to settle on.

src/badguy/haywire.cpp
```diff
@@ -139,19 +176,20 @@ Haywire::active_update(float dt_sec)
...
-        target_velocity = (-1.f) * walk_speed;
+    if (stomped_timer.get_timeleft() >= 0.05f)
+    {
+      target_velocity = 0.f;
+    }
+    else if (p && time_stunned == 0.0f)
+    {
+      /* Player is on the right or left*/
+        target_velocity = (p->get_pos().x > get_pos().x) ? walk_speed : (-1.f) * walk_speed;
     }
```
Semphriss, line 187: Two too many spaces :p

src/badguy/haywire.cpp
```diff
@@ -255,4 +295,16 @@ void Haywire::play_looping_sounds()
...
 
+HitResponse Haywire::collision_badguy(BadGuy& badguy, const CollisionHit& hit)
+{
+  if (is_exploding)
+  {
+    badguy.kill_fall();
+    return FORCE_MOVE;
+  }
+  else
+    WalkingBadguy::collision_badguy(badguy, hit);
```
Semphriss, line 306: It would be better with brackets here, to avoid any confusion.

Semphriss (commented):
It's actually a surprisingly good pathfinding logic! I've tried finding flaws in it (e. g. if Tux gets in a position where he'll die no matter what, like haywire jumping if Tux jumps but not if he doesn't, etc.) but I found nothing remarkably annoying about it.

Only detail (very minor one, it doesn't really require a fix), on downwards ledges that require a jump, the haywire will simply fall instead of jumping, like:

image

It takes the solid path in red, but it could take the dotted path in green. It doesn't matter much, though, it's just in case it's something that slipped through and easy to fix. If it's intended or too long to implement, leave it out, it's already working super well :D

Oh, also - two small code quality changes.

Semphriss (approved):

PR #1758, opened 2021-06-02
Fix segfault when quitting a level while grabbing

### LINKED ISSUE #1739: Quitting a level while not carrying a lantern properly causes a segfault
On latest @ master:

1. Open any level with any carriable object.
2. Grab the carriable object, walk to a wall, then release the action key; as to let the object go, but it won't actually go because of the wall.
3. Press escape, restart the level.
4. Repeat steps 2 and 3.
5. The game will crash with a segmentation fault.

When a sector is reset, its destructor is called, except the first time for some reason. Upon calling the sector's destructor, each object will be destroyed. Carriable objects will have an on-destroy listener bound to the player; that is, when they will be destroyed, they'll instruct the player to ungrab said object. However, at this point in the code, the player object will already be destroyed, which the code does not check, therefore accessing dead memory through a dangling pointer and causing a segfault.

seanyeh:
Fixes #1739

Reproduction steps:
Jump on Mr ice block, grab, and abort level while grabbing. Segfaults

As explained in the issue, the player is destroyed before the grab listener. I believe it segfaults here:
https://github.com/SuperTux/supertux/blob/1df160fa556f6f1875b869562bb579c3bab8163b/src/supertux/game_object.cpp#L54-L56
when the grab listener (associated with the now-destroyed player) does something.

By ungrabbing in the player destructor, we also remove this listener and avoid this issue. Let me know if this seems correct, thanks

### DIFF
--- a/src/object/player.cpp
+++ b/src/object/player.cpp
@@ -234,6 +234,7 @@ Player::Player(PlayerStatus& player_status, const std::string& name_) :
 
 Player::~Player()
 {
+  ungrab_object();
   if (m_climbing) stop_climbing(*m_climbing);
 }
 


Semphriss:
I'll check this ASAP - the thing I'm afraid about is if the lamp gets destroyed first, then the player might cause a segfault now when ungrabbing the now-destroyed lamp.

tobbi:
@Semphriss Did this work? From a quick glance, it looks like it should.

Grumbel:
To avoid these issues in the future I would recommend not using pointers directly, but holding `GameObject` by their UID and using `GameObjectManager::get_object_by_uid()` to retrieve the pointer when needed. Makes it much easier to deal with disappearing objects.

Semphriss:
On quick first try, it seems to have fixed the problem :+1:

Semphriss (approved):

Issue #806, opened 2018-05-10
Make AmbientSound actually work


tobbi:
Our AmbientSound object doesn't actually work (also check out the test level pertaining to Ambient Sound). One should actually look into how it's supposed to work and why it doesn't.

tobbi:
Okay, the test level is broken for some reason. AmbientSound as it is works, but only when you set distance factor / the other distance setting to 0. I don't know what the settings mean, though.

Rusty-Box:
I personally think that the AmbientSound object is made unnecessarily complicated. It should only have a volume setting and a distance setting.

Additionally I would recommend, adding a sector setting where you can choose an ambient sound that plays throughout the entire sector. (e.g. rain)

maxteufel:
So, plan for an update as discussed with @Rusty-Box is:

- The size of the AmbientSound object determines the area where the volume is at 100%
- A distance setting specifies, where compared to the center of the AmbientSound object, the volume drops to 0%
- We interpolate outside the specified area
- Additionally, an option to specify the base volume of the sound (in case the provided sound is too loud)

Rusty-Box:
Considering the AmbientSound object is still rather messy regarding its editor settings: we should remove the "distance bias" setting, because honestly no one seems to know what it even does, and have the "distance factor" increase the area where the sound plays at 100% volume from the main object.

Rusty-Box:
This was adressed by #2688 - Closing this.

PR #1763, opened 2021-06-12
Fix ambient sound (Issue #806)

mrkubax10:
This pull request adds features from maxteufel's comment: [comment](https://github.com/SuperTux/supertux/issues/806#issuecomment-394170973) to AmbientSound

### DIFF
--- a/src/object/ambient_sound.cpp
+++ b/src/object/ambient_sound.cpp
@@ -142,7 +142,7 @@ AmbientSound::update(float dt_sec)
       if (sound_source==nullptr)
         start_playing();
       else if (m_effect_distance == -1)
-        sound_source->set_gain(0.9f);
+        sound_source->set_gain(targetvolume);
       else
         sound_source->set_gain((1.0f - (dist / m_effect_distance)) * targetvolume);   
     }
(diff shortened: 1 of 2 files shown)


Rusty-Box:
I like this change to the ambient sound. However, there then should also be a way to play an ambient sound such as rain for an entire sector without its volume changing. I personally think this should be a sector option, like how we handle editing the music.

mrkubax10:
Idea with ambient sound for sector is nice, I will try to add this

tobbi:
Can this be merged?

mrkubax10:
Problem is that it will make existing ambient sounds not working.

tobbi:
What needs to be done in order to fix them?

mrkubax10:
Ambient sound now uses property effect-distance instead of multiple properties to determine volume at certain distance, so every ambient sound needs to have this property set.

tobbi:
There's actually not that many levels that use ambient-sound. Not in the data directory of the main game anyway. Not sure about add-ons.

Semphriss:
Will this be overridden by #2172 ?

Vankata453:
We should probably revive this PR, ambient sound does seem to need fixing.

mrkubax10:
> We should probably revive this PR, ambient sound does seem to need fixing.

Yes but probably as a different PR

PR #1764, opened 2021-06-15
Better "transportation" sprites for worldmaps

Rusty-Box:
With the help of @Alzter, this PR adds improved boat sprites as well as a new transportation methode aside the boat, granito.

There are still a few things that I need help fixing as I have no idea how to solve/what causes them:

- [x] Boat doesn't show proper empty sprite when not in use, however granito does
- [x] Changing power-up via cheat menu doesn't change sprite immediately, only upon changing directions


weluvgoatz:
The first point of issue in this PR can probably be fixed with some code adjustments, as the game might not recognize some of the sprite actions you specified in boat.sprite as actions that the game needs to make happen.

Guih48:
I think that also would be good, if small and big tux could have different sprite (ok, where is the sprite???)

Rusty-Box:
@Guih48 What do you mean “where is the sprite”? There are separate sprites for small and bug Tux in both boat and granito mode.

Guih48:
> @Guih48 What do you mean “where is the sprite”? There are separate sprites for small and big Tux in both boat and granito mode

Yes, but they aren't look different

Rusty-Box:
> > @Guih48 What do you mean “where is the sprite”? There are separate sprites for small and big Tux in both boat and granito mode
> 
> Yes, but they aren't look different (and if we would make they different we need a new sprite)

Uh… yes they are different… one is taller than the other

Guih48:
> > > @Guih48 What do you mean “where is the sprite”? There are separate sprites for small and big Tux in both boat and granito mode
> > 
> > 
> > Yes, but they aren't look different (and if we would make they different we need a new sprite)
> 
> Uh… yes they are different… one is taller than the other

Oh sorry, i don't noticed the difference :(
(Because we need to make sprites more different?)

Semphriss:
@Rusty-Box Can this be rebased on latest master?

Rusty-Box:
Okay, I went back to this and currently it seems the only bug still present is that the sprite can't quite update when moving up and/or down.

**An example:**
As Big Tux I go onto a boat/granito and move on an up or down path. Now I wanna change my powerup state with the cheat menu to Fire Tux, while I'm on said path. However, instead of showing Fire Tux now in the boat/granito, I still see Big Tux in it until my boat/granito either begins to move to the right or the left. Only then it seems to update the sprite it is supposed to show.

PR #1766, opened 2021-06-16
Rename ubports mk dir to ubuntu-touch

jonnius:
UBports is the name of the foundation which should not be used for referring to the OS, which is Ubuntu Touch.


Semphriss:
If it's going to be renamed, it should be renamed to "clickable" instead, given that the configuration is specific to clickable.

jonnius:
I find it a bit suprising to name it after the tool rather than the target, but I changed it accordingly.

Semphriss:
My usual logic to name workflow it according to the most specific thing it is designed for. For example, the workflow for Linux/macOS isn't specific to a single compiler (it isn't even a single OS - FreeBSD would be there too if GitHub offered it directly). That's why it's named "Unix".

For Ubuntu Touch, the workflow is designed to work specifically with Clickable; if there is any other framework/toolchain/etc., it would need its own workflow. (I'm not aware of any framework for UT other than Clickable - are there more?)

Also - did you push?

jonnius:
I pushed to the wrong branch before 🤦🏽 

Clickable is a tool to facilitate cross-compiling and packaging click packages for Ubuntu Touch. The packaging itself is done by the tool `click`. Ubuntu Touch is not like a traditional distro. It rather defines the framework apps can expect. Clickable provides docker images based on that framework coming from the Ubuntu Touch repositories provided by UBports.

In theory, Clickable could be used to build packages for other systems. But as this is not the case, Clickable is more specific than Ubuntu Touch.

Semphriss (approved):

PR #1769, opened 2021-06-25
Fix low trampoline jumps

### LINKED ISSUE #1656: Low trampoline jumps
**SuperTux version:** supertux2 v0.6.2-614-gd86dae0e9
**System information:** Lubuntu 18.04

##### Expected behavior

When holding the jump button, Tux should always jump high after landing on a trampoline.

##### Actual behavior

When I hold the jump button shortly before Tux lands on the trampoline, he jumps only a low height. 

[Here](https://user-images.githubusercontent.com/3192173/105749990-74fbb580-5f44-11eb-9634-7e6e5f1562d4.mp4) is a video of the problem. To hear the sound you may need to download the video.

* Jump from a solid tile so that Tux will eventually land on a trampoline
* Release the jump button mid-air
* Shortly before landing on the trampoline, press and hold the jump button

As far as I know, there's a feature which lets Tux jump off solid ground if the player presses and holds the jump button shortly before landing. Perhaps the implementation of this feature conflicts with the high trampoline jumps.

If I remember correctly, there may be another old issue about this problem which is closed now.

kmatebotond:
Fixes #1656

### DIFF
--- a/src/object/player.cpp
+++ b/src/object/player.cpp
@@ -619,7 +619,7 @@ Player::swim(float pointx, float pointy, bool boost)
     if(is_ang_defined)
     {
       delta = pointed_angle - m_swimming_angle;
-  
+
       if(std::abs(delta) > math::PI)
         delta += delta > 0 ? -math::TAU : math::TAU;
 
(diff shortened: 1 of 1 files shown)


HybridDog:
Thank you. I tested it and low trampoline jumps didn't happen when I held jump shortly before landing on the trampoline.

During testing I noticed that "jumping" from a trampoline is still possible if I hold jump shortly after leaving the trampoline (instead of shortly before), but this is less problematic in my opinion because it doesn't reduce my intended jump height. Here's a video about holding jump shortly afterwards:
 
https://user-images.githubusercontent.com/3192173/135086555-1a5f315a-ae2a-404e-b1c9-e9087c03f7a4.mp4

In my opinion Tux should never slow down in upwards direction after jumping, so I think this Pull Request can be merged.

Issue #1724, opened 2021-04-14
Suggestion: Groundpounding on trampolines should do a high jump


Semphriss:
### The issue

Trampolines require holding the jump key to perform high bounces. With the air flower, this clashes with the gliding ability, which often triggers right above the trampoline, delaying the excepted bounce by a short but annoying amount of time, which becomes an even greater problem when moving horizontally at the same time and Tux ends up missing the platform.

### The proposed solution

Groundpounding on trampolines should also result in a high jump, similar to holding space. This seems intuitive enough (more force downwards on a trampoline will give more momentum upwards) and easy to execute. This would be in addition to holding space, as I see no need to remove that way of working.

I'm leaving this issue open as a request for comments, feedback, concerns, propositions and other thoughts from both the team and the public.

Alzter:
I like this idea as buttjumping on trampolines currently feels underwhelming

Guih48:
This is a good idea, but it's makes many levels too easy, so i would suggest a new trampoline (or just bump block?) type that if you ground-pound it launches you high.

mrkubax10:
Added in #1772

PR #1772, opened 2021-06-29
Jump higher on trampoline when groundpounding

mrkubax10:
This pull request adds feature from #1724

### DIFF
--- a/src/object/trampoline.cpp
+++ b/src/object/trampoline.cpp
@@ -91,9 +91,23 @@ Trampoline::collision(GameObject& other, const CollisionHit& hit)
       //player is falling down on trampoline
       if (hit.top && vy >= 0) {
         if (!(player->get_status().bonus == AIR_BONUS))
-          vy = (player->get_controller().hold(Control::JUMP) || player->get_controller().hold(Control::DOWN)) ? VY_MIN : VY_INITIAL;
+        {
+          if (player->get_controller().hold(Control::JUMP))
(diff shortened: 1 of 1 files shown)


src/object/trampoline.cpp
```diff
@@ -85,13 +85,15 @@ Trampoline::collision(GameObject& other, const CollisionHit& hit)
     auto player = dynamic_cast<Player*> (&other);
     //Trampoline works for player
     if (player) {
+      if (player->m_does_buttjump)
+        player->m_does_buttjump = false;
```
Semphriss, line 89: TODO: Check the pros and cons of this

mrkubax10: This is workaround for this: [video](https://cdn.discordapp.com/attachments/396022213100437504/859345418441850910/2021-06-29_10-09-15.mp4). `m_does_buttjump` is set back to true later anyway.

Semphriss: Ah, okay. In this case I retract my comment.

Guih48:
This is a good idea, but it's makes many levels too easy, so i would suggest a new trampoline (or just bump block?) type that if you ground-pound it launches you high.

mrkubax10:
Actually jump height is slightly smaller than normal jump. Try doing this in latest version or nightly and check what is happening.

Semphriss:
Seems to work as intended, merging.

PR #1775, opened 2021-06-30
added bonuscave forest remix

RandomPenguin1778:
this is a remix of the bonuscave music . this is made for forest-themed bonus cave levels


Semphriss:
It would be better to share this on social media; it'll be added only when something in-game will use that song.

RandomPenguin1778:
should i share this on discord ?

RandomPenguin1778:
does the game have bonus-cave levels in the forest world ?

weluvgoatz:
I agree with Semphris, closing this (but that does not mean that this will go unconsidered.  Just share it somewhere else first and get approval there)

PR #1776, opened 2021-07-01
Fix emscripten build

mrkubax10:


Semphriss:
I'm on it - that's not the issue. The issue is that some flags that once were allowed both to the compiler and the linker are now linker-only.

I'm already on my way to fix this as I told the team - I'll close this PR now.

PR #1779, opened 2021-07-06
Fix ability to put portable object into other

mrkubax10:


Semphriss:
What behavior does this change in the game?

mrkubax10:
Fixes this:
https://user-images.githubusercontent.com/63350514/124615349-76c39080-de75-11eb-81b5-a8c250020ed3.mp4

Ordoviz:
I support the change. While I found this way of stacking rocks very convenient, it is not very intuitive and it can lead to weird interactions if you throw a rock from a very small distance into a rock stack. Level designers should stop making levels requiring rock stacks.

mrkubax10:
This change doesn't only apply to rocks.

weluvgoatz:
> Fixes this:
> https://user-images.githubusercontent.com/63350514/124615349-76c39080-de75-11eb-81b5-a8c250020ed3.mp4

Why is this an issue? I thought this way of stacking rocks was nice...

Semphriss:
How is this a bug/broken behavior? This is the default way of handling colliding collideables...

PR #1783, opened 2021-07-07
Crystallo Improvements and Additions

weluvgoatz:
This PR makes significant changes to the Crystallos so that they feel more unique as Tux traverses the Crystal Mine:
- The sprite has been remade to match the high-fps animations and higher-quality sprite work of the other Icy Island enemies.
- The crystallo now turns around by slowing and turning instead of abruptly turning.
- Two new crystallo variants: Roof Crystallo and Sleeping Crystallo.  Roof Crystallo walks along the roof and drops down when it sees Tux, shattering into 4 shards that the crystallo splits into after hitting the ground, flying in the air, and sticking to the ground and wall where they landed for a few seconds before despawning.  Sleeping Crystallo sleeps peacefully in a sleeping area above the ground but when it becomes aware of Tux it jumps out of its sleeping place and walks around normally.
- A decal has been added for Sleeping Crystallo's sleeping area.

Todo:

- [x] Proper sleeping crystallo sleeping area.
- [x] Sound design for Sleeping Crystallos waking, Roof Crystallos shattering and Shards hitting their target.


Rusty-Box:
This is really good. Something I'd suggest is to have the default action of sleeping crystalo in the editor be its sleeping mode.
Also, I'm not quite sure on the easing, it looks a bit off though I'm not sure what to do instead? Maybe change the type of easing, idk...

Semphriss:
Really cool!

I'd have two suggestions, :
- Shattering sound when the crystallo breaks?
- Smaller particles/dust when the crystallo breaks in small parts, so it doesn't look like half the crystallo disappeared :D

PR #1789, opened 2021-07-12
Add delete level option to editor

mrkubax10:


Guih48:
Yes i think level deleting option is a good idea, and it's also true to add-ons, because if we can create levels, and download add-ons throught the game we also need to can remove they trought the game, and not searching for game's specific folders and delete throught the file manager.

weluvgoatz:
A few things:
- This seems to detect whether to delete based on the level name and not the filename, I suggest deleting based on the filename just in case the player has 2 levels of the same name and cannot delete the 1st if they are currently in the 2nd.
- Add a HL below the "Delete Level" header, for consistency.

mrkubax10:
I will do that

PR #1801, opened 2021-08-02
Remove sky from snow backgrounds

Rusty-Box:
The affected snow backgrounds are: arctis2.png and arcticmountain.png

I still kept the not transparent version of arctis2 to be used as the background for the level editor's start screen without the black void visible. It is located separately from the other backgrounds in `images/engine/menu/` to not cause confusion.


serano01:
I'm not against it, but what's the occasion? Plus, after merging, someone needs to add a gradient for the levels which are using them, otherwise they will have a black void visible.

Rusty-Box:
@serano01 This allows these backgrounds to be layered and combined together. This is also beneficial for having them set at different times of day or varing sky colors in general.

Also, I personally see them having much more of a future then parallax tiles (atleast the way we use them right now).

PR #1807, opened 2021-08-17
Minor Level Fixes

serano01:
- Removed unnecessarily tiles in "Crystal Beauty"
- Added Gradient in "After the Glaciers"
- Added Gradient in "Pink Snow"


Semphriss (approved):
I'll skip reviewing bonus islands, as I'm not familiar enough with them.

PR #1809, opened 2021-08-19
Add extended level done sound

mrkubax10:


Semphriss:
Unfitting, IMO. Has there been any discussion from the artists about it?

Semphriss:
Given the lack of discussion, and the lack of request by the artistic community, I will be closing this.

If it gets revived someday, there should be some discussion prior to addition.

PR #1810, opened 2021-08-29
Update CMakeLists.txt to enable GLM Experimental

christianhaitian:
Build fails on Debian Buster without these entries for either Release or Debug builds.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -71,8 +71,10 @@ endif(NOT EMSCRIPTEN)
 ## Add definitions
 if(CMAKE_BUILD_TYPE MATCHES "Release|RelWithDebInfo")
   add_definitions(-DRELEASE)
+  add_definitions(-D GLM_ENABLE_EXPERIMENTAL)
 elseif(CMAKE_BUILD_TYPE MATCHES Debug)
   add_definitions(-DDEBUG)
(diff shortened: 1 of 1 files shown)


tobbi:
```
In file included from /home/runner/work/supertux/supertux/src/audio/sound_manager.hpp:28:0,
                 from /home/runner/work/supertux/supertux/src/audio/openal_sound_source.cpp:19:
/home/runner/work/supertux/supertux/src/math/vector.hpp:20:0: error: "GLM_ENABLE_EXPERIMENTAL" redefined [-Werror]
 #define GLM_ENABLE_EXPERIMENTAL
```

Probably needs an `#ifndef`. Can you fix that as well?

Grumbel:
Probably best to remove the `#define GLM_ENABLE_EXPERIMENTAL` from all the source and switch to just compiler flags.

`-D GLM_ENABLE_EXPERIMENTAL` shouldn't have a space.

christianhaitian:
It compiled fine on my end with that inclusion as shown in the commit but I don't know what the implications are for non debian/ubuntu systems.  If there's a more efficient solution for this that allows building on debian/ubuntu, I am in full support.

Grumbel:
Fixed with b320c27513858be65d38ef7e87437c230b9b3d8f

PR #1527, opened 2020-09-01
Allow turning to stone midair

Ordoviz:
The Earthflower is underpowered and I never used the stone ability. This PR enables you to press <kbd>Down</kbd>+<kbd>ACTION</kbd> midair to turn Tux into a heavy invincible stone that can squish badguys and destroy brick blocks.

The sprites/animations have some issues. Let us come to a consensus whether this feature should be added before fixing these issues. I encourage you to play around with this PR locally.
  
video

PR #1820, opened 2021-09-19
New rock bonus

Semphriss:
Closes #1527 

https://user-images.githubusercontent.com/66701383/133940232-c64b7428-462f-4e04-b3e0-4c544c9f5cca.mp4

The feature is not polished, but I open the PR so people can start giving feedback and playing around with it.


tobbi:
I like this but how do we make this not overkill?

Semphriss:
A good first step would be to refactor the player.cpp file to fix the huge technical debt we have.

In all seriousness, to have a better management (better physics on slopes and all) would require a refactor of the player code and the physics engine. It will become increasingly difficult to maintain most player-related features until then.

Busybody64:
You should have a cooldown between using this powerup.
Also, you should still be able to turn into a statue if you're not holding left or right keys.

Guih48:
I think, this is very good, but a bit too overpowered, so i think, it would be good if this ability could only available on slopes, or until full-speed running, and at other cases tux would have a tanooki-statue-like ability.

Busybody64:
> I think, this is very good, but a bit too overpowered, so i think, it would be good if this ability could only available on slopes, or until full-speed running, and at other cases tux would have a tanooki-statue-like ability.

I agree with this statement.

Guih48:
> I think, this is very good, but a bit too overpowered, so i think, it would be good if this ability could only available on slopes, or until full-speed running, and at other cases tux would have a tanooki-statue-like ability.

Like this: [https://github.com/SuperTux/supertux/pull/1527#issue-690358030](https://github.com/SuperTux/supertux/pull/1527#issue-690358030)

pelya:
> I suggest to add friction, so if you are not holding any directional key, the rock should begin to slow down.

Stone Tux should slow down even when holding a directional key, that would be more fair.
So Tux can turn to stone while running, but will stop after some short distance or after bumping into enemy, and there's no control over Tux once he starts rolling. Tux will need some minimal speed to flatten the enemy, if the speed is not enough Tux just stops rolling. And he won't be able to flatten tall enemies at all, like Stumpy or Snowman.

Semphriss:
> So Tux can turn to stone while running, but will stop after some short distance or after bumping into enemy

The problem with this is that Tux can get accidentally stuck in narrow areas without possibility to leave. I added a small movement possible to let Tux creep out of such places, but not in a convenient/satisfying enough way to make it usable through regular gameplay.

pelya:
That's also a long-standing problem with physics - it doesn't look good
when Tux stands up in  a narrow area and takes damage, and these little
crouching jumps are silly. Make big Tux roll or slide on his belly instead,
when he cannot stand up.

weluvgoatz:
Closing this because I am working on a significantly more expansive and modern version of this PR.

PR #1825, opened 2021-09-23
Fixed poor feeling when playing with bumpers

### LINKED ISSUE #1824: Bumper momentum inconsistency
SuperTux-v0.6.2-829-g63a326685
Windows 10

Unsure what the expected behaviour is for the momentum, but I feel like it is inconsistent. sometimes tux is very heavy, and other times, he turns around too fast. The momentum applied is such that when tux dies, even his hitbox is nudged. I feel like there is speed being added to tux's original speed instead of multiplying his current speed, which would make more sense... the harder you hit the bumper, the more it bumps.

https://youtu.be/v89G2mRryWs showcases this.

I think this needs more discussion. Bumpers, currently, are very volatile though however.

Semphriss:
Fixes #1824 

Might need extensive, thorough active testing since this relate to ease of use and user satisfaction.

### DIFF
--- a/src/object/player.cpp
+++ b/src/object/player.cpp
@@ -2206,6 +2206,7 @@ void
 Player::sideways_push(float delta)
 {
   m_boost = delta;
+  m_physic.set_velocity_x(delta > 0 ? MAX_RUN_XM : -MAX_RUN_XM);
 }
 
 void


Semphriss:
(Closed given lack of support)

PR #1827, opened 2021-09-24
Attempt to fix missing SDL.h with VCPKG builds

Semphriss:


Grumbel:
> Do you not need to pass the new "TINYGETTEXT_SDL2_INCLUDE_DIRS" variable to the subproject?

Yes.

> Also, don't have the previous changes be reverted?

Semphriss first one was still necessary, as there was unnecessary code that added an extra 'SDL2' to the path and that would still make the build fail, though I only realized that after having dropped it already.

> Why did this show up so suddenly? Is a recent change in cmake responsible for the breakage?

Some change in vcpkg or SDL I would assume (25382e39f7bf93e1d501d9342c6845e7bb0d2574 maybe, update to SDL 2.0.16), though really the fault is in SuperTux.

PR #1832, opened 2021-10-04
Improvements to Ice Powerup

weluvgoatz:
Frozen enemies can now be picked up and carried around like rocks, and the ice powerup in general has a few improvements.  There are also some minor fixes/adjustments made in this PR.  Given that it changes so much, it might be prone to a few issues, so I will iron those out as they are discovered as well.


Guih48:
I think, it's a good idea to make some frozen enemies can slide like mr. Ice blocks, but for the difference they can't stoppable by jump, and also, thely can't turn back at walls, because they break.
What do you think about theese mechanics?

Issue #1833, opened 2021-10-05
Music sometimes not playing after death


LouriNoel:
v0.6.2-842-gb1c41b6df
Ubuntu 20.04.3 (64-bit) (virtual machine, but also happens occasionally on my dual-boot)

##### Expected behavior

When Tux dies, the music fades out. When respawning (at the level's beginning or at the checkpoint), the music fades in from where it stopped.

##### Actual behavior

In some cases, the music does not fade in after respawning: we don't hear anything. Taking a checkpoint does not affect this behaviour. Pressing the escape key and resuming the game, or changing the music volume in the options, restarts the music normally.

##### Steps to reproduce actual behavior

This happens more frequently when the game lags (I tried running the game in a virtual machine), but it can also happen under normal conditions. I will explain why. Just take a level like Nolok's castle and keep dying at the beginning on the first badguy, in this level you don't even have to move.

##### Additional debugging information

I already investigated the issue, and I will do a PR to fix it.

Normally when Tux dies, the music fades away because we decrease the gain in StreamSoundSource::update. The music fades in after respawning because GameSession calls MusicObject::resume_music, which ultimately increases the gain in StreamSoundSource::update.
But sometimes resume_music is not called soon enough, so the music is paused at the end of the fade time in StreamSoundSource::update. Then the gain increases again after respawning, but the music is not playing because it is paused.

What I suggest (and this is what I do in the PR that will soon follow this post) is to resume m_music_source on top of calling set_fading in SoundManager::resume_music (which is called by GameSession).
This way, this additional call will play the music if it is paused, and otherwise do nothing (ie. return immediately if the music is not paused).

PR #1834, opened 2021-10-05
fixing music not playing after death sometimes (#1833)

LouriNoel:
Resuming m_music_source on top of calling set_fading will play
the music if it has been paused, and will do nothing if the music
is not paused yet.


Semphriss (approved):
I remember running across that bug a few times. Props for finding how to fix it, and big thanks for the PR!

PR #1835, opened 2021-10-05
Editor load filesystem

### LINKED ISSUE #1819: CLI option --edit-level takes path relative to data directory, inconsistent with normal play invocation
Playing a SuperTux level located in the current working directory is as easy as `supertux2 ./level.stl`. However, attempting to edit the same file does not work by simply appending `--edit-level`, since this option has a different file lookup mechanism.

Both situations should use the same lookup mechanism, that is, find the file using absolute path or relative path from current working directory.

tobbi:
Load levels and worldmaps from the filesystem instead of using a PHYSFS local reference.

Fixes #1819

Please test this!

### DIFF
--- a/src/supertux/level.cpp
+++ b/src/supertux/level.cpp
@@ -71,6 +71,11 @@ Level::save(const std::string& filepath, bool retry)
   try {
     { // make sure the level directory exists
       std::string dirname = FileSystem::dirname(filepath);
+      if (dirname == "./")
+      {
+        dirname = "";
+      }
+
       if (!PHYSFS_exists(dirname.c_str()))
       {
         if (!PHYSFS_mkdir(dirname.c_str()))
@@ -104,6 +109,11 @@ Level::save(const std::string& filepath, bool retry)
       log_warning << "Failed to save the level, retrying..." << std::endl;
       { // create the level directory again
         std::string dirname = FileSystem::dirname(filepath);
+        if (dirname == "./")
+        {
+          dirname = "";
+        }
+        
         if (!PHYSFS_mkdir(dirname.c_str()))
         {
           std::ostringstream msg;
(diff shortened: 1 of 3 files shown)


Semphriss:
> @Semphriss Can you please specify in what order we should check paths?
> 
> My suggestion would be: a) Absolute paths in the user's filesystem b) Relative paths in the user's filesystem c) PhysFS mounts
> 
> You concur?

I'd propose *only* absolute/relative from CWD; testing multiple paths might be confusing for users when paths clash. If we really need to have both options available, it would be wise to let the user specify their preference with a CLI flag, to avoid making certain level files unplayable under certain circumstances.

src/supertux/command_line_arguments.cpp
```diff
@@ -115,6 +116,7 @@ CommandLineArguments::print_help(const char* arg0) const
     << _("Game Options:") << "\n"
     << _("  --edit-level                 Open given level in editor") << "\n"
     << _("  --resave                     Loads given level and saves it") << "\n"
+    << _("  --from-datadir               Indicates that the LEVELFILE path is relative to the data dir. Only used if --edit-level is specified") << "\n"
```
Semphriss, line 119: IMO just `--from-datadir` isn't clear enough, it leaves confusion as to whether only the following path is from the datadir or if all of them are. To fix this, I'd suggest `--paths-from-datadir` (It also makes it clear in an instant that this is about paths).

I also recommend changing "LEVELFILE path" to just "paths" <sub>(plural)</sub> since all paths on the CLI should be affected by this.

mrkubax10:
Better late than never :^)

tobbi:
What paths should actually be affected by this?

Hypernoot:
Apparently, this functionality has been already added, possibly by the editor improvement. Closing as unnecessary.

PR #1842, opened 2021-10-07
Fix not working unisolid tiles

### LINKED ISSUE #1839: Unisolid hurting tiles don't hurt Tux
Current ST version from master (as of commit 718fe34bf8)
LMDE 4 64-bit (kernel 4.19.0-17-amd64)

Unisolid hurting tiles (the spikes Hume added a few years back, for example) are supposed to only hurt Tux when he comes from a certain direction or directions, but after the change that made regular hurting tiles more forgiving, the unisolid ones stopped hurting Tux entirely.

I noticed this issue recently while replaying some Narre's Multiworld levels where I use these extensively (I know I haven't touched that for a while as I'm quite short on time right now, but I'm pretty sure this issue first appeared with the more lenient spike hitboxes).

image

mrkubax10:
Fixes #1839

### DIFF
--- a/src/badguy/badguy.cpp
+++ b/src/badguy/badguy.cpp
@@ -318,7 +318,7 @@ BadGuy::collision_tile(uint32_t tile_attributes)
   if (tile_attributes & Tile::HURTS && is_hurtable())
   {
     Rectf hurtbox = get_bbox().grown(-6.f);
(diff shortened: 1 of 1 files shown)


tobbi:
@Semphriss Can you check this again?

PR #1850, opened 2021-10-14
Fix relative paths with .. in file selection

mrkubax10:
Fixes problem when user goes one directory up and selects file when using basedir


Zwatotem:
I know you asked about if it breaks something already, but I wasn't and still am not sure, whether it can affect some built-in levels or add-ons (especially the latter, since they tend to use a lot of customization). I would allow relative links, provided they are inside supertux home directory or inside basedir.

PR #1865, opened 2021-11-03
Added world selection menu

### LINKED ISSUE #1036: Add world select
Modern Mario platformers, or just platformers in general have a world select menu to choose which world to warp to on the fly. Currently in SuperTux you have to go through all the levels to move worlds. Can we get a world select menu for convenience?

Semphriss:
https://user-images.githubusercontent.com/66701383/140189067-fc63d7ac-c425-47f5-9624-de8b1da62946.mp4

The interface and the controls will be changed, but the choice of what is best cannot be determined by me alone.

I will also make it possible to lock certain worlds.

Closes #1036


Busybody64:
This reminds me of the unused world selection screen in NSMB DS!
In all seriousness, though, this would be a GREAT addition. Would help limit backtracking.
Maybe make it so you cannot enter any new worldmaps unless you actually accessed it normally at least once? (Like via teleporter to the other worldmap?)

Also, what happens if you try to enter any of the worlds after the forest world via this methood?

tobbi:
There are compiler warnings in the clickable build. Please fix them.

Semphriss:
@tobbi I forgot to mark this PR as draft - there are still many features and design decisions missing, please don't merge it too soon :D

Semphriss:
https://user-images.githubusercontent.com/66701383/140635262-3fcb2cfc-428b-4945-b4ef-439c527b9e2d.mp4

Busybody64:
Epic! This should've been in the game sooner.

tobbi:
I've convered this to a draft now.

Semphriss:
I removed an unnecessary variable, so *now* it should be ready :smile:

src/math/util.hpp
```diff
@@ -17,6 +17,8 @@
 #ifndef HEADER_SUPERTUX_MATH_UTIL_HPP
 #define HEADER_SUPERTUX_MATH_UTIL_HPP
 
+#include <math.h>
```
tobbi, line 20: Is this include needed?

Semphriss: It appears so, despite that I didn't change anything else...

tobbi:
You're right, it is.

Semphriss:
I just found out why it worked before without the include but now it needs the include: because `positive_fmodf` is marked inline and this PR uses that include for the first time, so it's the first time it needs to actually use fmodf.

tobbi:
#1865

tobbi:
#1036

weluvgoatz:
I already said this elsewhere, but I'll leave it here so it doesn't get forgotten:  The worldmap state isn't reset when re-entering that world, so, for example, if you were in the ice cave, world-select to the forest, and world-select back to the Icy Island, you start at the beginning of the Icy Island but the cave mode is still active, so you cannot see anything properly.

Semphriss:
FIXME: Given #1993, per Daniel's bug, it should also reset the ambient light. Will do after #1993 is merged.

PR #1872, opened 2021-11-07
Fix Windows nightly uploads

tobbi:


tobbi:
Oh look, this fixes the nightlies!

PR #1874, opened 2021-11-07
Textscroller Updates

weluvgoatz:
This PR adds the Text Scroller to the object menu, so level designers can add and remove them.  Also, the following options were added to it:
- whether or not it is controllable
- a script for when it is finished (no script just ends the level)
- anchor for the position of the text (and an offset to move the text to the left or right to a certain degree)
- anchor of the text itself (so left-aligned, right-aligned, or center-aligned [default])

This should allow for more text scroller configurations than before, such as if the designer wants to have one anchored to each side of the screen, ones that are timed (so controlling it doesn't work), and ones where they want something else to happen when the text ends, etc...


serano01:
Does this effect the old Textscroller levels in some way? Can they be reworked with the new object tool? Just curios, the PR itself looks good to me.

weluvgoatz:
I checked, and the old TextScrollers seem to be fine.  Not sure if they can be reworked with this new tool, though, because the TextScrollerScreens seem to use a different format than the level object...

PR #1878, opened 2021-11-10
Text + Menu box improvements

weluvgoatz:
This PR brings a few improvements to text and menu boxes:
- Infoblock transition mode, colors and roundness can be edited
- There is a developer mode option to change the color scheme of the menus
- Scriptable text boxes can now have their box and text colors, as well as roundness and transition mode changed (you can now also add them in yourself with the level editor instead of having to use an unseen text object)


src/scripting/text.cpp
```diff
@@ -16,10 +16,19 @@
 
 #include "scripting/text.hpp"
 
+#include <stdexcept>
```
tobbi, line 19: What do you need that include for?

weluvgoatz: It seems to be leftover from another thing I was working on.  I'll fix it.

PR #1881, opened 2021-11-11
Added option to save level under a different name

Semphriss:
https://user-images.githubusercontent.com/66701383/141243384-9274485e-4137-45c4-b68e-113bb652c17c.mp4

This PR adds two options to the level editor menu:
- Save Level as
- Save Copy

Both save the current level under a different filename; the difference between both is that "Save Level as" will make the editor track the new given filename, and subsequent saves will be written to that new file, whereas "Save Copy" will write to the given file only once and the editor will keep tracking the regular file.

### DIFF
--- a/src/editor/editor.cpp
+++ b/src/editor/editor.cpp
@@ -91,6 +91,8 @@ Editor::Editor() :
   m_reactivate_request(false),
   m_deactivate_request(false),
   m_save_request(false),
+  m_save_request_filename(""),
+  m_save_request_switch(false),
   m_test_request(false),
   m_particle_editor_request(false),
   m_test_pos(),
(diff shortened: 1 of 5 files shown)


Zwatotem:
Does this feature have file extension inference? Should we force one? Should it be OS-dependent?

Semphriss:
It does not enforce any filename restriction. I originally wanted to copy the mechanism to create a level, but I realized it does not take a filename, it automatically generates one. I'm open to any recommendation about validation.

However, it does handle path separator conversion properly (forward vs backwards slashes).

Zwatotem:
Most windows programs automatically add a file extension after the name, if you don't input it yourself. Users can be used to this. There is also an option to change the file type in a drop down to "All file types" position, and then it saves it literally, like here.

Issue #1149, opened 2019-08-24
Backgrounds Shift Out of Place


weluvgoatz:
I am using Windows 10, SuperTux version is at least less than a year old?  I'm not sure how long this issue has existed since my computer won't let me install ANY version between 0.4.0 and 0.6.0 and the older nightly builds of 0.6.

Anyway, this bug has the backgrounds of the game slowly move out of place, which is really annoying and bad when it comes to top or bottom aligned backgrounds.

serano01:
I don't know why this happens, but I was able to fix it per editor.

"(background
      (scroll-offset-x 0.5)
      (scroll-offset-y 0.5)
      **(x 9120)
      (y 192)**
      (alignment "bottom")
      (scroll-speed-x 0.5)
      (scroll-speed-y 0.5)
      (speed 0.5)
      (image-top "/images/background/bluemountain2.png")
      (image "/images/background/bluemountain2.png")
      (image-bottom "/images/background/bluemountain2.png")"

The bold lines had to be removed, then it worked. Maybe Tobbi knows more.

Rusty-Box:
@tobbi Would you look into it, please?

Rusty-Box:
It seems that every new added background gets a x an y cordinated which messes up the backgrounds alignment. I don't know what causes this in the code but this should defenitly be fixed! Currenty it can only be solved manully after removing the x and y cordinats from the .stl file.

Rusty-Box:
In case you need stuff to do @tobbi, I recommend this one ;)

weluvgoatz:
This seems to just be an issue of converting an Android version level to the official version.  In official levels, this issue doesn't exist.

Rusty-Box:
It does. Like I mentioned when adding a background layer the game gives them a x and y coordinate. This causes the background too shift out of place. Most notably when choosing an alignment.

Rusty-Box:
Unless you talk about the background slowly moving. Then you might want to check whether or not you have given the background scrolling speed

Semphriss:
The problem was found: creating any object with the editor by clicking on the screen will automatically assign an X and an Y corresponding to the point clicked on-screen.

Since,in the game, "spawning an object at a given position" means to create a mock save file containing the object and an X and Y property, it will assign values based on whether or not the object saves something to their X and Y properties in save files - which the background does.

[...]

PR #1886, opened 2021-11-13
Fix background shifting

weluvgoatz:
Solution for #1149 (or at least what that issue turned into), the background position is now added to the scroll offset so that the editor can edit it and set it back to (0, 0) if they want.


tobbi:
How many levels use the x and y properties of the background object?

PR #1890, opened 2021-11-14
Added add-on creator

### LINKED ISSUE #717: Export to add-on option for the editor

Semphriss:
It is now possible to pack a world or a levelset as an add-on zip archive automatically, form inside the game.

Features:
- Create an add-on automatically from an option in the editor menu; no user input needed.
- Add-on generated as a full, ready-to-ship zip archive in the `addons` folder.
- Automatically finds the necessary data to fill the .nfo file.
- **Versioning system**: Packaging the same add-on multiple times increases the version number automatically.

Flaws and shortcomings:
- Author and license for the add-on automatically set to be the same as the level from which the option is called, which might not be consistent/compatible with other licenses if many assets.
- Does not seek for files outside the world's folder; images and other assets will not be packaged if they are not located within `levels/<world-id>/`.

Closes #717


tobbi:
#717

PR #1895, opened 2021-11-17
Add generic list selection menu and use it in an editor menu

### LINKED ISSUE #1894: Replace arrow selectors with too many choices with a better selection menu
They are good for situations where there are few options to choose from but some have way to many options
Here's an example from the dispenser object's settings:
This image 
The widget's options include every single badguy in the game and is annoying to use.

Rami-Slicer:
Add a generic list selection menu and use it.

* Add menu_list.cpp and its header
* Replace a somewhat annoying string selector used to configure the badguys emitted by a dispenser who's options were _every single_ badguy.

Closes #1894

### DIFF
--- a/src/gui/menu_list.cpp
+++ b/src/gui/menu_list.cpp
@@ -2,14 +2,11 @@
 #include "gui/menu_list.hpp"
 #include "gui/menu_manager.hpp"
 
-ListMenu::ListMenu(std::vector<std::string>* items_, int* selected_) : 
-  items(items_), 
-  selected(selected_)
+ListMenu::ListMenu(const std::vector<std::string>& items, int* selected) : 
+  m_selected(selected)
 {
-  int i = 0;
-  for(auto& item : *items) {
-    add_entry(i, item);
-    i++;
+  for(uint i = 0; i < items.size(); i++) {
+    add_entry(i, items[i]);
   }
   add_hl();
   add_back("OK");
@@ -17,7 +14,9 @@ ListMenu::ListMenu(std::vector<std::string>* items_, int* selected_) :
 
 void
 ListMenu::menu_action(MenuItem& item) {
-  *selected = item.get_id();
+  if(m_selected) {
+    *m_selected = item.get_id();
+  }
   MenuManager::instance().pop_menu();
 }
 
--- a/src/gui/menu_list.cpp
+++ b/src/gui/menu_list.cpp
@@ -1,0 +1,24 @@
+#include "gui/menu_item.hpp"
+#include "gui/menu_list.hpp"
+#include "gui/menu_manager.hpp"
+
+ListMenu::ListMenu(std::vector<std::string>* items_, int* selected_) : 
+  items(items_), 
+  selected(selected_)
+{
+  int i = 0;
+  for(auto& item : *items) {
+    add_entry(i, item);
+    i++;
+  }
+  add_hl();
+  add_back("OK");
+}
+
+void
+ListMenu::menu_action(MenuItem& item) {
+  *selected = item.get_id();
+  MenuManager::instance().pop_menu();
+}
+
+/* EOF */
--- a/src/gui/menu_list.cpp
+++ b/src/gui/menu_list.cpp
@@ -21,8 +21,8 @@
 ListMenu::ListMenu(const std::vector<std::string>& items, int* selected) : 
   m_selected(selected)
 {
-  for(uint i = 0; i < items.size(); i++) {
-    add_entry(i, items[i]);
+  for(size_t i = 0; i < items.size(); i++) {
+    add_entry(static_cast<int>(i), items[i]);
   }
   add_hl();
   add_back("OK");
--- a/src/gui/menu_list.cpp
+++ b/src/gui/menu_list.cpp
@@ -1,3 +1,19 @@
+//  SuperTux -- List menu
+//  Copyright (C) 2021 Rami <rami.slicer@gmail.com>
+//
+//  This program is free software: you can redistribute it and/or modify
+//  it under the terms of the GNU General Public License as published by
+//  the Free Software Foundation, either version 3 of the License, or
+//  (at your option) any later version.
+//
+//  This program is distributed in the hope that it will be useful,
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//  GNU General Public License for more details.
+//
+//  You should have received a copy of the GNU General Public License
+//  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
 #include "gui/menu_item.hpp"
 #include "gui/menu_list.hpp"
 #include "gui/menu_manager.hpp"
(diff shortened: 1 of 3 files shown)


src/gui/menu_list.cpp
```diff
@@ -0,0 +1,24 @@
+#include "gui/menu_item.hpp"
+#include "gui/menu_list.hpp"
+#include "gui/menu_manager.hpp"
+
+ListMenu::ListMenu(std::vector<std::string>* items_, int* selected_) :
```
Semphriss, line 5: Why are the arguments taken as pointer? Can they be null?

src/gui/menu_list.cpp
```diff
@@ -0,0 +1,24 @@
+#include "gui/menu_item.hpp"
+#include "gui/menu_list.hpp"
+#include "gui/menu_manager.hpp"
+
+ListMenu::ListMenu(std::vector<std::string>* items_, int* selected_) : 
+  items(items_), 
+  selected(selected_)
```
Semphriss, line 7: Why is the list of items saved in the object? The list might not exist anymore if the list was created dynamically, like:
```cpp
...
{
  std::vector<std::string> items = {};
  int selected;
  MenuManager::instance().push_menu(std::make_unique<ListMenu<(&items, &selected));
}
// Both `items` and `selected` are freed, the menu now holds dangling pointers
...
```

src/gui/menu_list.cpp
```diff
@@ -0,0 +1,24 @@
...
+    i++;
+  }
+  add_hl();
+  add_back("OK");
+}
+
+void
+ListMenu::menu_action(MenuItem& item) {
+  *selected = item.get_id();
+  MenuManager::instance().pop_menu();
```
Semphriss, line 21: If `selected` is null, the behavior here is undefined (Segmentation fault on most systems). Even if erronous, the behavior should be well-defined to avoid needless crashes.

src/gui/menu_list.cpp
```diff
@@ -0,0 +1,24 @@
...
+
+ListMenu::ListMenu(std::vector<std::string>* items_, int* selected_) : 
+  items(items_), 
+  selected(selected_)
+{
+  int i = 0;
+  for(auto& item : *items) {
+    add_entry(i, item);
+    i++;
+  }
```
Semphriss, line 13: Less important, but might be good to consider: what are the tradeoffs of this vs. a regular `for (int i = 0; i < items->size(); i++)` loop?

Semphriss (changes requested):
Heya, thanks for the PR!

I gave it a quick look and I have a few questions on choices regarding code design. I'm not sure what's intended and what's a mistake, so if there's a reason you made certain things this way, let me know!

src/gui/menu_list.cpp
Rami-Slicer: I don't think the pointer is necessary. I'll fix that.

src/gui/menu_list.cpp
Rami-Slicer: I didn't think of that at all, yeah I'll remove that part since it's not needed at all.

src/gui/menu_list.cpp
Rami-Slicer: I've added a check for the pointer to make sure it's not null.

src/gui/menu_list.cpp
Rami-Slicer: Not much. I'll replace this too.

Busybody64:
I think you should also try to add a dropdown menu for the owl, too!
We don't want the game to crash because you made a typo.

src/gui/menu_list.cpp
```diff
@@ -0,0 +1,24 @@
+#include "gui/menu_item.hpp"
+#include "gui/menu_list.hpp"
+#include "gui/menu_manager.hpp"
+
+ListMenu::ListMenu(std::vector<std::string> items_, int* selected_) :
```
Semphriss, line 5: Please pass arguments as references to avoid an unneeded copy call, and as `const` if you don't modify the object for the caller.

src/gui/menu_list.cpp
```diff
@@ -0,0 +1,24 @@
...
+  add_hl();
+  add_back("OK");
+}
+
+void
+ListMenu::menu_action(MenuItem& item) {
+  // prevent segmentation fault should selected be null
+  if(selected) {
+    *selected = item.get_id();
+  }
```
Semphriss, line 20: No need to document pointer checks, they're pretty basic and already easy to understand just by looking at the code :)

Semphriss (changes requested):
I'm not sure if you finished all the edits you planned to do, but in case you did, I have two extra comments:

src/gui/menu_list.cpp
```diff
@@ -0,0 +1,23 @@
+#include "gui/menu_item.hpp"
+#include "gui/menu_list.hpp"
+#include "gui/menu_manager.hpp"
+
+ListMenu::ListMenu(const std::vector<std::string>& items_, int* selected_) : 
+  selected(selected_)
+{
+  for(int i = 0; i < items_.size(); i++) {
+    add_entry(i, items_[i]);
```
Semphriss, line 9: I missed that one - member variables should always be prefixed with `m_` per the [coding style](https://github.com/SuperTux/supertux/blob/master/CODINGSTYLE.md), so after that there's no need for trailing underscores for the argument names in the constructor.

(`items` also no longer has a member variable with the same name :D)

mrkubax10:
Files menu_list.cpp and menu_list.hpp are missing license info, check other files.

Rami-Slicer:
Woops I forgot about that

src/gui/menu_list.hpp
```diff
@@ -0,0 +1,40 @@
...
+#ifndef HEADER_SUPERTUX_GUI_MENU_LIST_HPP
+#define HEADER_SUPERTUX_GUI_MENU_LIST_HPP
+
+#include "gui/menu.hpp"
+
+class ListMenu final : public Menu 
+{
+public:
+  ListMenu(const std::vector<std::string>& items, int* selected_);
+
```
Semphriss, line 26: Probably my last comment - is that trailing underscore there for something?

Rami-Slicer: Yes I forgot to update that during PR whackamolle

Semphriss:
Please address the build issues. (Most are due to code quality problems, like a missing `override`)

Semphriss:
Pro tip: before pushing, you can enable code quality checks by passing `-DWARNINGS=ON` and `-DWERROR=ON` and rebuild SuperTux with both GCC and Clang, and it'll show where the problems are without needing to wait for the CI.

src/gui/menu_list.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+#include "gui/menu_item.hpp"
+#include "gui/menu_list.hpp"
+#include "gui/menu_manager.hpp"
+
+ListMenu::ListMenu(const std::vector<std::string>& items, int* selected) : 
+  m_selected(selected)
+{
+  for(uint i = 0; i < items.size(); i++) {
+    add_entry(i, items[i]);
```
Semphriss, line 25: MSVC doesn't recognize `uint`, please use `size_t` instead

Rami-Slicer:
Ah I was wondering how to get the errors to appear, thanks

Semphriss:
That last build error is not part of the code, it's a known issue with the environment.

Semphriss (approved):

mrkubax10 (approved):

Issue #1420, opened 2020-06-04
Display Cut-scene text without coding


LizzyBach:
Supertux 0.6.2
Windows 7 32-bit

Currently, you can only display cut-scene text by typing a 4-5 line code for each sentence. Yes, I know that you can display text with the message box but, they can only be activated by hitting them. So, I guess a new script trigger for displaying cut-scene text with fade in/out, code to process before and after the text is displayed and selecting font, maybe?, I don't know. I think so that it'll be a nice idea. What do you think about that?

Alzter:
We should really have a designated text box for cutscenes.

PR #1915, opened 2021-12-01
Add text area object

mrkubax10:
Adds feature from #1420


src/editor/object_option.hpp
```diff
@@ -443,6 +443,21 @@ class ButtonOption : public ObjectOption
   ButtonOption& operator=(const ButtonOption&) = delete;
 };
 
+class StringArrayOption : public ObjectOption
+{
+public:
+  StringArrayOption(const std::string& text, const std::string& key, std::vector<std::string>* items);
+
```
Semphriss, line 450: Is there a reason to pass a pointer there?

The problem with passing a pointer is that the caller will bear the responsibility of managing its lifetime, which prevents simple calls like `ctor({ "foo", "bar" }, ...)` or similar. Also, is it expected to pass a null pointer there?

I believe it would be preferable to use `const&` and to copy the vector into the member variable, if possible.

src/editor/object_option.hpp
```diff
@@ -443,6 +443,21 @@ class ButtonOption : public ObjectOption
...
+  virtual void save(Writer& write) const override;
+  virtual std::string to_string() const override { return {}; }
```
Semphriss, line 452: Why an empty string there?

It's used at least for displaying on tooltips, it would be preferable to return something if possible.

src/editor/object_option.hpp
```diff
@@ -443,6 +443,21 @@ class ButtonOption : public ObjectOption
...
+  virtual void add_to_menu(Menu& menu) const override;
+private:
+  std::vector<std::string>* m_items;
```
Semphriss, line 455: (Same as first comment, marking this so it isn't forgotten)

src/editor/object_settings.cpp
```diff
@@ -292,6 +292,12 @@ ObjectSettings::add_sexp(const std::string& text, const std::string& key, sexp::
   add_option(std::make_unique<SExpObjectOption>(text, key, value, flags));
 }
 
+void
+ObjectSettings::add_string_array(const std::string& text, const std::string& key, std::vector<std::string>* items)
```
Semphriss, line 296: (Same as first comment, marking this so it isn't forgotten)

src/gui/item_string_array.cpp
```diff
@@ -0,0 +1,36 @@
...
+//
+//  You should have received a copy of the GNU General Public License
+//  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#include "gui/item_string_array.hpp"
+
+#include "gui/menu_manager.hpp"
+#include "gui/menu_string_array.hpp"
+
+ItemStringArray::ItemStringArray(const std::string& text, std::vector<std::string>* items, int id) :
```
Semphriss, line 22: (Same as first comment, marking this so it isn't forgotten)

src/gui/item_string_array.hpp
```diff
@@ -0,0 +1,37 @@
...
+#ifndef HEADER_SUPERTUX_GUI_ITEM_STRING_ARRAY_HPP
+#define HEADER_SUPERTUX_GUI_ITEM_STRING_ARRAY_HPP
+
+#include "gui/menu_item.hpp"
+#include <vector>
+
+class ItemStringArray final : public MenuItem
+{
+public:
+  ItemStringArray(const std::string& text, std::vector<std::string>* items, int id = -1);
```
Semphriss, line 26: (Same as first comment, marking this so it isn't forgotten)

src/gui/item_string_array.hpp
```diff
@@ -0,0 +1,37 @@
...
+
+  virtual void process_action(const MenuAction& action) override;
+private:
+  std::vector<std::string>* m_items;
```
Semphriss, line 30: (Same as first comment, marking this so it isn't forgotten)

src/gui/menu.cpp
```diff
@@ -340,6 +341,15 @@ Menu::add_badguy_select(const std::string& text, std::vector<std::string>* badgu
   return *item_ptr;
 }
 
+ItemStringArray&
+Menu::add_string_array(const std::string& text, std::vector<std::string>* items, int id)
```
Semphriss, line 345: (Same as first comment, marking this so it isn't forgotten)

src/gui/menu.hpp
```diff
@@ -96,6 +97,7 @@ class Menu
     bool is_linear = false);
   ItemColorChannelOKLab& add_color_channel_oklab(Color* color, int channel);
   ItemPaths& add_path_settings(const std::string& text, PathObject& target, const std::string& path_ref);
+  ItemStringArray& add_string_array(const std::string& text, std::vector<std::string>* items, int id = -1);
```
Semphriss, line 100: (Same as first comment, marking this so it isn't forgotten)

src/gui/menu_string_array.cpp
```diff
@@ -0,0 +1,61 @@
...
+//
+//  You should have received a copy of the GNU General Public License
+//  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#include "gui/menu_string_array.hpp"
+
+#include "gui/menu_item.hpp"
+#include "util/gettext.hpp"
+
+StringArrayMenu::StringArrayMenu(std::vector<std::string>* items) :
```
Semphriss, line 22: (Same as first comment, marking this so it isn't forgotten)

src/gui/menu_string_array.hpp
```diff
@@ -0,0 +1,41 @@
...
+#define HEADER_SUPERTUX_GUI_MENU_STRING_ARRAY_HPP
+
+#include "gui/menu.hpp"
+
+#include <vector>
+
+class StringArrayMenu final : public Menu
+{
+public:
+  StringArrayMenu(std::vector<std::string>* items);
```
Semphriss, line 27: (Same as first comment, marking this so it isn't forgotten)

src/gui/menu_string_array.hpp
```diff
@@ -0,0 +1,41 @@
...
+
+  virtual void menu_action(MenuItem& item) override;
+private:
+  std::vector<std::string>* m_array_items;
```
Semphriss, line 31: (Same as first comment, marking this so it isn't forgotten)

src/object/text_area.hpp
```diff
@@ -0,0 +1,49 @@
...
+//  You should have received a copy of the GNU General Public License
+//  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#ifndef HEADER_SUPERTUX_OBJECT_TEXT_AREA_HPP
+#define HEADER_SUPERTUX_OBJECT_TEXT_AREA_HPP
+
+#include "supertux/moving_object.hpp"
+#include "supertux/timer.hpp"
+
+class TextArea final : public MovingObject
```
Semphriss, line 23: Why doesn't the class inherit the Trigger class? (It's designed specifically for that kind of object)

Note: I won't otherwise review the code for this class, as it may require a rewrite.

Semphriss (changes requested):
I reviewed the code; there are a few things I noted that I believe might have been mistakes, so I've listed them all there. If anything was intended to be the way it is, let me know!

For the sake of best production, I've been more verbose than brief, just in case any detail could be extra helpful; most of this is just redundant (there's a question I had in the code that occurred 8 to 10 times, I marked each place).

src/trigger/text_area.hpp
```diff
@@ -0,0 +1,49 @@
...
+  virtual ObjectSettings get_settings() override;
+  virtual std::string get_class() const override { return "text-area"; }
+  virtual std::string get_display_name() const override { return _("Text Area"); }
+  virtual bool has_variable_size() const override { return true; }
+private:
+  bool m_started, m_inside, m_once, m_finished;
+  std::vector<std::string> m_items;
+  float m_delay, m_fade_delay;
+  unsigned int m_text_id;
+  Timer m_update_timer, m_fade_timer;
```
Semphriss, line 42: The reason why I suggested using TriggerBase as base is because it has an `event` function which helps to avoid having to store all these variables :smile: 

The class should be refactored to match the structure of its parent and sibling classes.

mrkubax10: `m_started` and `m_inside` are used to make sure that player won't restart text sequence by leaving and entering text area during sequence. When `m_once` is set to true text sequence will play only once so when user leaves area after sequence finishes it won't play again. `m_finished` is true after text sequence. I'm not sure how to replace these variables by using `TriggerBase`

Semphriss: Please take a look at `TriggerBase::event()`

Semphriss:
TODO: Change the way string input/editing is handled for StringArrays

src/gui/menu_string_array.hpp
```diff
@@ -0,0 +1,42 @@
...
+  StringArrayMenu(std::vector<std::string>& items);
+
+  virtual void menu_action(MenuItem& item) override;
+private:
+  std::vector<std::string>& m_array_items;
+  std::string m_text;
+  int m_selected_item;
+
+  void reload();
+private:
```
Semphriss, line 36: Small code quality comment: private member functions should go below public member functions, under their own `private` tag; those visibility tags should be preceded by an empty line, except the one at the top of the class.

Semphriss (changes requested):

Semphriss (approved):

src/trigger/text_area.cpp
```diff
@@ -25,6 +25,11 @@
 #include "video/drawing_context.hpp"
 #include "video/layer.hpp"
 
+namespace
+{
+std::string anchor_points[] = { "middle", "topleft", "top", "topright", "left", "right", "bottomleft", "bottom", "bottomright" };
+}
+
```
Semphriss, line 32: ~~Is there a reason to put it in `namespace {}`? Also, it can be made `static` if it's used only in this cpp file, and it can be marked as `const` if it isn't edited at runtime~~

I believe you won't need this anyways, given the possible changes below.

src/trigger/text_area.cpp
```diff
@@ -149,6 +159,8 @@ TextArea::get_settings()
   settings.add_bool(_("Once"), &m_once, "once");
   settings.add_float(_("Text change time"), &m_delay, "delay");
   settings.add_float(_("Fade time"), &m_fade_delay, "fade-delay");
+  settings.add_string_select(_("Anchor"), &m_anchor,
+    { _("Middle"), _("Top-left"), _("Top"), _("Top-right"), _("Left"), _("Right"), _("Bottom-left"), _("Bottom"), _("Bottom-right") }, 0, "anchor-point");
```
Semphriss, line 163: `add_enum` would be better adapted for this, since it refers to a logically ordered enum; you may find an example in `src/objects/custom_particle_system.cpp:429`. It helps avoid having to declare `anchor_points`, lets you use the `AnchorPoint` enum as a type for `m_anchor`, and removes the need to call `string_to_anchor_point`.

I'll mark the other areas for clarity :smile:

src/trigger/text_area.hpp
```diff
@@ -54,6 +54,7 @@ class TextArea final : public TriggerBase
   size_t m_current_text;
   Status m_status;
   Timer m_timer;
+  int m_anchor;
```
Semphriss, line 58: In reference to above, this variable can be declared with type `AnchorPoint`, which is more logical to use and is more verbose; it'll help future maintainers to easily understand the variable.

src/trigger/text_area.cpp
```diff
@@ -88,6 +96,7 @@ TextArea::event(Player& player, EventType type)
         m_current_text = 0;
         m_status = Status::FADING_IN;
         m_timer.start(m_fade_delay);
+        text_object.set_anchor_point(string_to_anchor_point(anchor_points[m_anchor]));
         text_object.set_text(m_items[m_current_text]);
```
Semphriss, line 100: With the two edits above, this line becomes simply `text_object.set_anchor_point(m_anchor);`

src/trigger/text_area.cpp
```diff
@@ -130,6 +139,7 @@ TextArea::update(float dt_sec)
         {
           m_status = Status::FADING_IN;
           m_timer.start(m_fade_delay);
+          text_object.set_anchor_point(string_to_anchor_point(anchor_points[m_anchor]));
```
Semphriss, line 142: Same as above

src/trigger/text_area.cpp
```diff
@@ -45,6 +51,7 @@ TextArea::TextArea(const ReaderMapping& mapping) :
   mapping.get("delay", m_delay);
   mapping.get("once", m_once);
   mapping.get("fade-delay", m_fade_delay);
+  mapping.get("anchor-point", m_anchor, 0);
```
Semphriss, line 55: With this method, the value will appear in the file as a number instead of a string, which isn't very intuitive and memorable for those editing the files manually.

With the New Method:tm:, it will be saved as strings (:tada:), so this line will need to be changed to reading a string and then assigning that string passed to `string_to_anchor_point` to `m_anchor` manually.

src/trigger/text_area.cpp
```diff
@@ -33,7 +38,8 @@ TextArea::TextArea(const ReaderMapping& mapping) :
   m_fade_delay(1.0f),
   m_current_text(0),
   m_status(Status::NOT_STARTED),
-  m_timer()
+  m_timer(),
+  m_anchor(0)
 {
```
Semphriss, line 43: It should now default to the old default option, `ANCHOR_MIDDLE`.

Semphriss (changes requested):
Since I'm requesting a design change, I "enabled verbose mode" (e. g., wrote rather more than less) to leave a trace of my thoughts. I think I marked every single line you edited :smile: 

Anyways, sorry for the comment spam, there really is only one change to do and the rest should be natural. I only commented everywhere to provide explanations for anyone who might be interested.

src/trigger/text_area.cpp
mrkubax10: I was trying to do like this but AnchorPoint uses bitmasks so it is not ordered enum

src/trigger/text_area.cpp
mrkubax10: I was trying to do like this but AnchorPoint uses bitmasks so it is not ordered enum

Semphriss: It should be fixed though; the current method has many flaws.

I can take care of this in another PR when I'll have some time.

mrkubax10: I will try to come up with different method then

Semphriss:
If there is still interest, bitmask support for anchors can be dropped and replaced with linear enums. If that would cause too many complex changes, then forget about add_enum and try to implement the other changes if possible (e. g. save string keys instead of numbers).

Semphriss:
@mrkubax10 Are the changes I pushed good?

mrkubax10:
Fine for me

Semphriss (approved):

PR #1921, opened 2021-12-08
New industrial tiles

cnub66:
Redid industrial tileset to be less crunchy-looking and hopefully more fitting. Also adds unisolid tiles.
Demo


Semphriss:
Moving to discussions; closing

PR #1922, opened 2021-12-09
Refactor level flipping

mrkubax10:
Fixes issues with level flipping.

### DIFF
--- a/src/object/platform.cpp
+++ b/src/object/platform.cpp
@@ -147,17 +147,6 @@ Platform::editor_update()
 }
 
 void
-Platform::editor_delete()
-{
-  auto path_obj = get_path_gameobject();
-  if(path_obj != nullptr)
-  {
-    path_obj->editor_delete();
-  }
-  GameObject::editor_delete();
-}
-
-void
 Platform::on_flip(float height)
 {
   if (Path* path = get_path()) {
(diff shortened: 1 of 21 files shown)


Semphriss:
Self-requested a review because I've noticed a few ways to improve the code, I'll check that in-depth and I''ll take the time to write my comments properly whenever possible (might take a few days, feel free to remind me if I seem to have forgotten).

src/supertux/flip_level_transformer.hpp
```diff
@@ -33,20 +33,11 @@ class Decal;
 /** Vertically or horizontally flip a level */
 class FlipLevelTransformer final : public LevelTransformer
 {
+public:
+  static Flip transform_flip(Flip flip);
```
Semphriss, line 37: Is there any upside to use this rather than taking the flip as (non-const) reference and modifying it directly that way?

(Suggesting that option as it would avoid an extra assignment in the code from callers)

src/object/platform.cpp
```diff
@@ -156,6 +157,14 @@ Platform::editor_delete()
   GameObject::editor_delete();
 }
 
+void
+Platform::on_flip(float height)
+{
+  if (Path* const path = get_path()) {
```
Semphriss, line 163: You have used `Path* const` at a few places (I won't mark them all as I usually do), which is a const pointer to a non-const object; is there a particular reason to do it this way?

tobbi:
Please rebase!

mrkubax10:
I will do that as soon as possible

PR #1927, opened 2021-12-12
An option for enabling and disabling swimming in the editor

bruhmoent:
Some levels might not be designed for swimming, or simply swimming breaks the creator's idea of how a level is supposed to be played. I think making an option for disabling/enabling it might be a good idea. (default: ON)
obraz_2021-12-12_225245
(ex.)


mrkubax10:
I think that you used wrong branch when opening pull request

bruhmoent:
oh i didnt notice, sorry

PR #1928, opened 2021-12-12
Show selected enemy in dispenser menu

Narre:
After the recent change to the enemy select menu, the game no longer showed what enemy is currently selected (the only way to know was to either select it again or add a new one and delete it if it wasn't the right one). This pull request makes the menu show "Select enemy (currently selected)" instead of just "Select enemy".

refresh_menu() is changed to refresh() because refresh() is defined in the generic Menu class, whereas refresh_menu() was only in menu_badguy_select, so using it would mean eiher restricting menu_list to badguy select menu only or having two functions with identical functionality defined in Menu.

It also changes OK to Cancel in the list menu as OK makes no sense if no changes are made (and it also makes it translatable).

### DIFF
--- a/src/gui/menu_badguy_select.cpp
+++ b/src/gui/menu_badguy_select.cpp
@@ -21,6 +21,8 @@
 #include "gui/menu_manager.hpp"
 #include "gui/menu_list.hpp"
 
+#include "boost/format.hpp"
+
 std::vector<std::string> BadguySelectMenu::all_badguys;
 
 BadguySelectMenu::BadguySelectMenu(std::vector<std::string>* badguys_) :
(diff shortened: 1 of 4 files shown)


Narre:
image

mrkubax10:
Could you use C format string: `_("Select enemy (%s)")`

Narre:
How? gettext doesn't have any string formatting, and even if it had, wouldn't doing this make it untranslatable?

Narre:
OK, I've done it. But it definitely does break the translations, so if this is merged, it will have to be updated on Transifex.

mrkubax10:
> I still can't see how this won't break translations but I will try later today.
RTL languages

Narre:
Yeah, I know it's better to make it possible to put the %s anywhere in the string, I'm just referring to the fact that it will have to be updated on Transifex. Anyway I've done it now.

mrkubax10:
Could you address build issues? It fails due to `void refresh();` not being marked as `override`.

PR #1931, opened 2021-12-13
Added reference counter for PathObjects

Semphriss:
Automatically manages a path's lifetime based on whether or not it is referenced by other objects.

Also fixes a bug where node handles in the editor would persist after a path is deleted, resulting in undefined behavior. It does not, however, remove nodes upon deselection; it only removes nodes when the path itself is deleted.


tobbi:
Looking good!

PR #1955, opened 2021-12-15
Update worldmap_cheat_menu.cpp bonuses.

bruhmoent:
This updates the worldmap's cheat menu bonuses to be the same as in-level cheat menu.


bruhmoent:
Note: Mistaken awarded bonuses in the worldmap's cheat menu.

PR #1956, opened 2021-12-15
Update worldmap_cheat_menu.cpp bonuses.

bruhmoent:
Changed from awarded one from x64. Weird it gives only 1, as in the level you can get x64 instantly.


mrkubax10:
This needs to wait for 0.6.3 release as it changes strings

Semphriss:
Note: I'd like to request the further delay of this PR because I already fixed this in another, upcoming PR, which I expect to make available shortly after the release of 0.6.3.

Semphriss:
Overridden by #2043, which was the magical PR in question. Closing with approval of opener.

PR #1959, opened 2021-12-15
Make ping-pong the default path mode if the path mode is invalid

Narre:
As discussed on Discord, oneshot is not the best option for the default path mode, especially for levels where unordered was used before it was removed (because it makes the platform stop after a single "cycle"). This PR changes the default path mode to pingpong (which was decided to be the best option in the same Discord discussion), so the affected platforms don't stop moving after reaching the last node.

### DIFF
--- a/src/object/path.cpp
+++ b/src/object/path.cpp
@@ -36,8 +36,8 @@ string_to_walk_mode(const std::string& mode_string)
   else if (mode_string == "circular")
     return WalkMode::CIRCULAR;
   else {
-    log_warning << "Unknown path mode '" << mode_string << "'found. Using oneshot instead." << std::endl;
-    return WalkMode::ONE_SHOT;
+    log_warning << "Unknown path mode '" << mode_string << "'found. Using pingpong instead." << std::endl;
+    return WalkMode::PING_PONG;
   }
 }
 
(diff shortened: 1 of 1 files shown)


Semphriss (changes requested):
Design decision; there should be further discussion in here about the pros and cons of each option.

Semphriss:
My thoughts:

I have seen many discussions about with option would be the most *convenient*. I believe such a factor is irrelevant to consider for the following reason:

Given that this happens only in case of invalid/missing key, it would be irrelevant to consider regular use cases. Instead, we should ask: in which circumstances is this part of the code relevant?

Some possible cases:
- Someone edits the level fine manually and does an error.
  - In this case, the user should fix the problem in all circumstances.
- Older levels which use an incompatible format.
  - This is a tougher issue, as it's difficult to guess what the original intent was; but I think we can agree that the level should be fixed in the shortest delays.

I am not aware of any other relevant use cases; please add comments if I forgot about any.

Given those two circumstances, there is no interest in picking any option above the other. It would therefore be preferable to keep the behavior consistent with other cases where keys are missing, and putting the first option as default (in this case, oneshot). Putting another option (in this case, the second) as the fallback option would be confusing for users who aren't aware of the discussion we're having, although this alone is a minor issue.

**However:** It is important to make sure the user realizes there is an error that happened. The user may of may not have developer mode enabled and notice the error, and even if they do, they may or may not notice the warning immediately. Therefore, it is important that the default path chosen reflects clearly that an error happened.

Given this requirement, putting a *convenient* option would be counter-productive. It is much more important to put an error the user can recognize as "error-ful". The fact that oneshot is the first option (and that most invalid states fallback to the first option), the fact that it is difficult to work with and therefore easily noticeable, and most of all the fact that it is not frequent on modern level designing and therefore will stand out upon happnening all push me towards voting to keep oneshot as the default fallback option.

Narre:
There doesn't seem to be much interest in discussing this. I'm closing it since it will be obsoleted when a way to make n-state lifts possible is implemented in the future.

PR #1960, opened 2021-12-15
Fix tiles disappearing when changing size and offset

### LINKED ISSUE #1945: Shrinking a tilemap while using a negative offset looses tiles
Discovered while investigating a fix for #1943

https://user-images.githubusercontent.com/66701383/146117642-0c0c2b8e-2f6b-42fe-960f-f552acb9527f.mp4

The created empty lines at the bottom are not empty lines added by offset; if you check the horizontal line of snow tiles and count the number of tiles between it and the bottom of the tilemap, it will become apparent that the 5 empty lines at the bottom should have kept the base of the block.

mrkubax10:
Fixes #1945

### DIFF
--- a/src/object/tilemap.cpp
+++ b/src/object/tilemap.cpp
@@ -485,6 +485,22 @@ void
 TileMap::resize(int new_width, int new_height, int fill_id,
                 int xoffset, int yoffset)
 {
+  //Apply offset
+  if (xoffset || yoffset) {
+    for (int y = 0; y < m_height; y++) {
+      int Y = (yoffset < 0) ? y : (m_height - y - 1);
+      for (int x = 0; x < m_width; x++) {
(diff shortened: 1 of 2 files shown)


Semphriss (changes requested):
New issue: When resizing a tilemap to be *larger* with an offset to displace tiles in the area that doesn't exist yet, the tiles are lost.

https://user-images.githubusercontent.com/66701383/146259572-1740eb99-d1f3-4f38-a76a-25eedbb135d3.mp4

I've already tried fixing it this way but then I thought about it, and I realized it would introduce another bug :^) It seems like this is the easy solution everybody will think about :laughing:

Semphriss (approved):
Works well :+1: Thanks for the fix!

PR #1964, opened 2021-12-17
Attempt to fix changing sprites of certain objects in editor

### LINKED ISSUE #1961: Object sprite changes don't update unless editor is reloaded
v0.6.3-beta.1.2-gdb0689f97
win 10

changing an object to a different sprite wont change it unless you reload the level editor, by exiting and re-entering it, or going to another stage and coming back, this gets quite annoying and discourages use of decals.

Semphriss:
Fixes #1961

Or rather, should.

Please test this PR to ensure that all objects are indeed fixed.

@HybridDog The rublight's settings builder pulls from grandparent MovingObject instead of parent MovingSprite; was it intended to be that way? It seems like an accident, but I don't want to risk introducing unintended side effects.
image

### Important notes
- **Spotlight** currently cannot have its sprite changed because it is composed of several layers of different sprites; it would need to be refactored before supporting sprite changing. I did not do it in this PR because I wanted to stick to simple bugfixes strictly, to have it merged in time for 0.6.3 if possible (and to avoid introducing new bugs, as much as possible).

### DIFF
--- a/src/badguy/badguy.cpp
+++ b/src/badguy/badguy.cpp
@@ -865,6 +865,8 @@ BadGuy::get_settings()
 void
 BadGuy::after_editor_set()
 {
+  MovingSprite::after_editor_set();
+
   if (m_dir == Direction::AUTO)
   {
     if (m_sprite->has_action("editor-left")) {
--- a/src/badguy/dispenser.cpp
+++ b/src/badguy/dispenser.cpp
@@ -464,6 +464,8 @@ Dispenser::get_settings()
 void
 Dispenser::after_editor_set()
 {
+  BadGuy::after_editor_set();
+
   set_correct_action();
 }
 
--- a/src/badguy/flame.cpp
+++ b/src/badguy/flame.cpp
@@ -17,18 +17,18 @@
 #include "badguy/flame.hpp"
 
 #include "audio/sound_manager.hpp"
-#include "audio/sound_source.hpp"
 #include "editor/editor.hpp"
 #include "math/util.hpp"
 #include "object/sprite_particle.hpp"
 #include "sprite/sprite.hpp"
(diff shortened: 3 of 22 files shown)


Narre:
I had a quick look (but I didn't test everything thoroughly). This is what noticed:

- The editor now renders Ice Flames as regular flames by default, even though the game doesn't.
- Same for leafshot - the editor renders them as kamikazesnowballs, the game doesn't.
- Decal still doesn't update its hitbox after a sprite change

image
I don't know exactly what you meant by this, but flames, iceflames and ghostflames are not just retextures, they have different behaviours to fire/ice flowers.

- So the sprite of ghost flames should still be settable. The user might have their own sprite for an addon and may not want their flames to be killable with the iceflower(flame) or fireflower(iceflame).

Semphriss:
@Narre I've updated the PR to include those changes. Big thanks for the catch about flames, I wasn't aware of that!

Could you (or anyone familiar with C++) review my modifications with the #include's in the flame files? Since I removed some stuff to make it more flexible, I saw I could simplify the whole thing, but I'm not sure if my edits are optimal, so I'd like to be reviewed on this by someone, if possible.

Narre:
I tested it and it all seems to be working correctly :)

As for the review, I'm only just beginning with C++, I've only ever worked in C nonplusplus before, so I don't dare to comment on any C++ code. You'll have to ask someone else about that.

Narre:
Just one more thing - please remove the ability to change the sprite of worldmap-spawnpoint. Spawnpoints can't be rendered and setting their sprite just crashes the game.

HybridDog:
> The rublight's settings builder pulls from grandparent MovingObject instead of parent MovingSprite; was it intended to be that way?

If I remember correctly, it wasn't intentional.

PR #1970, opened 2021-12-18
Move icecrusher code to the badguy folder

Rami-Slicer:
Icecrushers are classified as badguys just like angrystones but were in the "objects" folder.


src/object/brick.cpp
```diff
@@ -20,7 +20,7 @@
 #include "badguy/badguy.hpp"
 #include "object/bouncy_coin.hpp"
 #include "object/explosion.hpp"
-#include "object/icecrusher.hpp"
+#include "badguy/icecrusher.hpp"
 #include "object/player.hpp"
```
Semphriss, line 24: Please move the #include's to remain in alphabetical order

src/supertux/game_object_factory.cpp
```diff
@@ -91,7 +91,7 @@
 #include "object/ghost_particle_system.hpp"
 #include "object/gradient.hpp"
 #include "object/hurting_platform.hpp"
-#include "object/icecrusher.hpp"
+#include "badguy/icecrusher.hpp"
```
Semphriss, line 94: Same here

Semphriss (changes requested):

Semphriss (approved):

PR #1971, opened 2021-12-18
Fix path node flipping

Narre:
This PR fixes broken path node flipping, which seems to have been introduced by separating paths from objects, and then made worse by adding bezier curves. Since paths can now be used by multiple objects, offsetting nodes according to the object's height is no longer possible - this means that flipped objects that are on a path will be offset by their height after a levelflip. There's not much that can be done about this in flip-handling and will have to be tackled in a more generic way (by not limiting nodes to represent the top-left corner of a platform's position).

This should fix flipping of platforms, node-following tilemaps and path-coins. If I forgot about something, please let me know.

Please test! I have relatively little experience with C++ and (especially) the SuperTux code. This is one of my first attempts at anything non-trivial in SuperTux.

**before:**

https://user-images.githubusercontent.com/14074789/146653425-d2b180ec-78f6-4492-b798-ec1139c9026c.mp4

**after:**

https://user-images.githubusercontent.com/14074789/146653438-7a97be25-104c-4e87-94e1-d92d8fb6d7bd.mp4

### DIFF
--- a/src/badguy/ghoul.hpp
+++ b/src/badguy/ghoul.hpp
@@ -44,8 +44,6 @@ public:
 
   void move_to(const Vector& pos) override;
 
-  Path* get_path_if_exists() override { return get_path(); }
-  
 protected:
   bool collision_squished(GameObject& object) override;
   
--- a/src/badguy/ghoul.hpp
+++ b/src/badguy/ghoul.hpp
@@ -43,6 +43,8 @@ public:
   void stop_moving();
 
   void move_to(const Vector& pos) override;
+
+  Path* get_path_if_exists() override { return get_path(); }
   
 protected:
   bool collision_squished(GameObject& object) override;
--- a/src/badguy/willowisp.hpp
+++ b/src/badguy/willowisp.hpp
@@ -68,8 +68,6 @@ public:
     ExposedObject<WillOWisp, scripting::WillOWisp>::unexpose(vm, table_idx);
   }
 
-  Path* get_path_if_exists() override { return get_path(); }
-
   /** make WillOWisp vanish */
   void vanish();
 
--- a/src/badguy/willowisp.hpp
+++ b/src/badguy/willowisp.hpp
@@ -68,6 +68,8 @@ public:
     ExposedObject<WillOWisp, scripting::WillOWisp>::unexpose(vm, table_idx);
   }
 
+  Path* get_path_if_exists() override { return get_path(); }
+
   /** make WillOWisp vanish */
   void vanish();
 
--- a/src/object/camera.hpp
+++ b/src/object/camera.hpp
@@ -106,8 +106,6 @@ public:
 
   /** smoothly slide the scale of the camera towards a new value */
   void ease_scale(float scale, float time, easing ease);
-
-  Path* get_path_if_exists() override { return get_path(); }
   /** @} */
 
 private:
--- a/src/object/camera.hpp
+++ b/src/object/camera.hpp
@@ -106,6 +106,8 @@ public:
 
   /** smoothly slide the scale of the camera towards a new value */
   void ease_scale(float scale, float time, easing ease);
+
+  Path* get_path_if_exists() override { return get_path(); }
   /** @} */
 
 private:
--- a/src/object/coin.hpp
+++ b/src/object/coin.hpp
@@ -48,8 +48,6 @@ public:
 
   virtual void move_to(const Vector& pos) override;
 
-  Path* get_path_if_exists() override { return get_path(); }
-
   void collect();
 
 private:
--- a/src/object/coin.hpp
+++ b/src/object/coin.hpp
@@ -48,6 +48,10 @@ public:
 
   virtual void move_to(const Vector& pos) override;
 
+  Path* get_path_if_exists() override {
+    return get_path();
+  }
+
   void collect();
 
 private:
--- a/src/object/coin.hpp
+++ b/src/object/coin.hpp
@@ -48,9 +48,7 @@ public:
 
   virtual void move_to(const Vector& pos) override;
 
-  Path* get_path_if_exists() override {
-    return get_path();
-  }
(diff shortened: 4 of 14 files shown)


src/badguy/ghoul.hpp
```diff
@@ -43,6 +43,8 @@ class Ghoul final : public BadGuy,
   void stop_moving();
 
   void move_to(const Vector& pos) override;
+
+  Path* get_path_if_exists() override { return get_path(); }
```
Semphriss, line 48: `get_path()` is already public; you may use it directly instead.

Generally, it's not a good idea to add "if_exists" to the names of functions that return a pointer; if a pointer is returned, it is already assumed that the pointer might be null.

src/badguy/willowisp.hpp
```diff
@@ -68,6 +68,8 @@ class WillOWisp final :
     ExposedObject<WillOWisp, scripting::WillOWisp>::unexpose(vm, table_idx);
   }
 
+  Path* get_path_if_exists() override { return get_path(); }
+
```
Semphriss, line 72: Same her

src/object/camera.hpp
```diff
@@ -106,6 +106,8 @@ class Camera final : public GameObject,
 
   /** smoothly slide the scale of the camera towards a new value */
   void ease_scale(float scale, float time, easing ease);
+
+  Path* get_path_if_exists() override { return get_path(); }
```
Semphriss, line 110: Same (Also, you placed this function inside the functions for scripting, between the Doxygen braces)

src/object/coin.hpp
```diff
@@ -48,6 +48,8 @@ friend class HeavyCoin;
 
   virtual void move_to(const Vector& pos) override;
 
+  Path* get_path_if_exists() override { return get_path(); }
+
```
Semphriss, line 52: Same

src/object/path.cpp
```diff
@@ -231,4 +231,14 @@ Path::is_valid() const
   return !m_nodes.empty();
 }
 
+void
+Path::on_flip(float height)
+{
+  for (auto& nod : m_nodes) {
+    nod.position.y = height - nod.position.y;
```
Semphriss, line 238: Was there a reason why the variable isn't named `node`?

src/object/platform.hpp
```diff
@@ -65,6 +63,8 @@ class Platform : public MovingSprite,
   /** Updates the platform to the given action  */
   void set_action(const std::string& action, int repeat);
 
+  Path* get_path_if_exists() override { return get_path(); }
+
```
Semphriss, line 67: Same, and same with scripting :stuck_out_tongue:

src/object/tilemap.hpp
```diff
@@ -203,7 +203,9 @@ class TileMap final :
   void set_tileset(const TileSet* new_tileset);
 
   const std::vector<uint32_t>& get_tiles() const { return m_tiles; }
-  
+
+  Path* get_path_if_exists() override { return get_path(); }
+
```
Semphriss, line 208: Same here

src/supertux/flip_level_transformer.cpp
```diff
@@ -30,9 +31,19 @@ void
 FlipLevelTransformer::transform_sector(Sector& sector)
 {
   float height = sector.get_height();
+  std::set<Path*> paths;
 
   for (auto& object : sector.get_objects()) {
     object->on_flip(height);
+    if (Path* path = object->get_path_if_exists()) {
```
Semphriss, line 38: You can then use just `get_path()` here.

Semphriss, line 38: All objects that have a path are `PathObject`'s; you can use a `dynamic_cast<PathObject*>(object)` here to check if the object is a PathObject and use it afterwards.

(I think there's one or two objects that have paths but don't inherit from PathObject, but don't take them in consideration, these objects are the ones that need to be fixed)

src/supertux/game_object.hpp
```diff
@@ -163,6 +164,8 @@ class GameObject
       together (e.g. platform on a path) */
   virtual void editor_update() {}
 
+  virtual Path* get_path_if_exists() { return nullptr; }
+
```
Semphriss, line 168: And same here :D

Semphriss (changes requested):
Thanks for the PR!

I left a few comments at some places with suggestions on how to better design stuff.
Something I forgot to specify in my comments; you added a virtual function, but when you overrode it, you forgot to mark it as virtual again :stuck_out_tongue: But it isn't important, as the function is actually not needed in this case.

src/object/path.cpp
Narre: I named it `nod` because that is the name used in `Path::move_by()`. It is weird, I will change it to `node`.

src/supertux/flip_level_transformer.cpp
Narre: Thank you! Those are the C++ things I will have to learn. I knew there must be a nicer solution.

src/supertux/flip_level_transformer.cpp
```diff
@@ -30,9 +31,20 @@ void
...
+    if (po && po->get_path()) {
+      if (paths.find(po->get_path()) == paths.end()) {
+        paths.insert(po->get_path());
+      }
+    }
+  }
+
+  for (Path* path : paths) {
+    path->on_flip(height);
   }
```
Semphriss, line 49: I'm sorry for only mentioning this now, I just thought about it:

Would it not be easier to put this code in the `PathGameObject` class? That class holds directly the interesting paths, and putting the relevant code there removes the needs for:
- `dynamic_cast`'ing and treating PathObjects specially;
- check for duplicate paths;
- adding code in every class that makes use of a path.

Semphriss (approved):

weluvgoatz (approved):

PR #1980, opened 2021-12-19
Fix input in editor when dialog is opened

mrkubax10:
Fixes ability to scroll level and werid animation error when escape is pressed when dialog is opened

### DIFF
--- a/src/editor/editor.cpp
+++ b/src/editor/editor.cpp
@@ -406,25 +406,28 @@ Editor::update_keyboard(const Controller& controller)
     return;
   }
 
-  if (controller.pressed(Control::ESCAPE)) {
-    esc_press();
-    return;
-  }
+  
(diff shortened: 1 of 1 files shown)


Semphriss:
I struggle to reproduce the issue; do you have steps?

I agree the way of doing it that you propose should be favored instead of using `m_enabled`, but I can't find where it fails.

mrkubax10:
https://user-images.githubusercontent.com/63350514/146728283-b8968843-4e06-418e-a888-5271bcaabedf.mp4

Semphriss:
Self-requesting a review because I believe there could be a better way to handle this code-wise, I'll investigate and come back with a solution if I find one.

Semphriss:
Couldn't think of one. I'll come back later if I find one.

Semphriss (approved):

Issue #1965, opened 2021-12-17
`g_config->locale` is empty (or broken) when calling `get_font_for_locale()`


gnh1201:
**SuperTux version:** 5e50a35e678f2b441f47ed5684a37889b44460ee
**System information:** Lubuntu 20.04 LTS

##### Expected behavior

Issue #1885 should have been solved.

##### Actual behavior

The problem of Issue #1885 has not been solved.

The variable `g_config->locale` did not contain a valid value as a string. (`get_font_for_locale()` in src/supertux/resources.cpp)

gnh1201:
I submitted the pull request to fix this problem. Please check.

PR #1990, opened 2021-12-20
Fix the locale problem #1965

gnh1201:
Fix the locale problem #1965


Semphriss (changes requested):
This code does not belong in this function; loading resources requires additional configuration which is already managed in src/supertux/main.cpp. The issue should not be hidden behind patch code, it should be fixed where it is already located.

I already have this in my to-do list. If you could provide reproduction steps valid for any platform, it would be helpful.

gnh1201:
#1992

PR #1992, opened 2021-12-20
Use g_dictionary_manager for locale

### LINKED ISSUE #1965: `g_config->locale` is empty (or broken) when calling `get_font_for_locale()`

Semphriss:
g_config->locale is set only if the user manually expressed a preference; automatic detection wouldn't pick the right font in that case.

Closes #1990
Fixes #1965

### DIFF
--- a/src/supertux/resources.cpp
+++ b/src/supertux/resources.cpp
@@ -66,7 +66,7 @@ Resources::load()
   {
     console_font.reset(new TTFFont("fonts/SuperTux-Medium.ttf", 12, 1.25f, 0, 1));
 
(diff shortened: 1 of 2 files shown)


gnh1201:
It works!

PR #1999, opened 2021-12-21
Adjust mushroom decals

serano01:
Made the images significant smaller and better to use.


Semphriss:
Do the ghost forest levels need to be adjusted? <sub>(If yes, does it matter?)</sub>

serano01:
I don't think so. But a follow-up PR is coming anyway (the user bar from discord is working on it).

Narre:
Well, you removed not only the transparent bits to the right and bottom, but also left and top, which means the existing decals will be off by a few pixels, and bar's PR will probably be done before this is merged, so I'd say adjustments will be needed.

weluvgoatz:
I played around with it, and it doesn't seem like these changes affect the placement of the decals in the levels.  If they somehow do, I'll fix up the levels.

PR #2003, opened 2021-12-22
Attempt to add FreeBSD packages

Semphriss:
This adds automatic FreeBSD packaging in the CI.


Semphriss:
FreeBSD will most likely not be updated to a version with a functioning FreeBSD CPack generator in the foreseeable future; closing.

PR #2004, opened 2021-12-22
Replace boost::format with fmt::format

Grumbel:
This is in preparation for C++20. fmt shares mostly the same API as std::format(), which isn't in a usable state yet.

Using the system provided fmt ended up being too brittle, as 6, 7 and 8 all have incompatibilities with each other when it comes to having runtime generated format strings (e.g. as returned from gettext()), thus external/fmt/ submodule was added.

Since the syntax of format strings changes, merging this has to wait until after 0.6.3.

fmt should be usable with C++14.

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -29,7 +29,7 @@ jobs:
     env:
       # Revision of https://github.com/pelya/commandergenius.git - SDL2 repo is a submodule
       # We check out a specific stable revision, so the build won't break if I modify my SDL repo
-      SDL_ANDROID_REV: 532acc9192
+      SDL_ANDROID_REV: d3d938c27d66fe568637a7fe0dc0112cf6e76217
       # It takes 35 minutes per one architecture to build SuperTux, of which 20 minutes are for building Boost and ICU libraries
       # Enable additional architectures when we have these libraries inside the cache, or create a separate Android script for each architecture
       ARCH_LIST: armeabi-v7a arm64-v8a x86 x86_64
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -67,6 +67,10 @@ jobs:
         run: |
           sed -i "s/MultiABI=.*/MultiABI='${ARCH_LIST}'/g" project/jni/application/supertux/AndroidAppSettings.cfg
 
+          # FIXME: Tweaks to get fmt compiled, should probably be upstreamed
+          sed -i "s#\(AppSubdirsBuild='[^']*\)'#\1 supertux/external/fmt/src supertux/external/fmt/include'#" project/jni/application/supertux/AndroidAppSettings.cfg
+          mv -v ../external/fmt/src/format.cc ../external/fmt/src/format.cpp
+
       - name: Copy precompiled libraries from cache
         uses: actions/cache@v2
         id: cache
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -67,10 +67,6 @@ jobs:
         run: |
           sed -i "s/MultiABI=.*/MultiABI='${ARCH_LIST}'/g" project/jni/application/supertux/AndroidAppSettings.cfg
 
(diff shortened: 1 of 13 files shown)


CMakeLists.txt
```diff
@@ -204,6 +204,13 @@ endif()
 
 find_package(PNG REQUIRED)
 
+option(USE_SYSTEM_FMT "Use preinstalled fmt, must be 8.0.0 or newer" OFF)
+if(USE_SYSTEM_FMT)
+  find_package(fmt 8.0.0 REQUIRED)
+else()
+  add_subdirectory(external/fmt EXCLUDE_FROM_ALL)
+endif()
```
Semphriss, line 212: Would it be possible (and if yes, would it be a good idea) to always `find_package(fmt 8.0.0)` without making it required, and then include the subdirectory based on whether or not the version was found?

Semphriss:
While I've been assisting with translations, I've noticed the current system does not allow certain changes that might be necessary in some languages, for example:
- Reordering keywords: Let say in English we have `format(_("Date: {} {}, {}"), month, day, year)`, but certain languages would prefer to use a different order, like DD/MM/YYYY instead of MM/DD/YYYY.
- Grammatical differences: In certain languages, "A worldmap" and "Download the worldmap" need "worldmap" to be spelled differently.

Since we're looking for replacements for the string translation formatting, should we take the opportunity to try and find a new system that would help fix those issues?

Grumbel:
Argument ordering is already supported. In boost you can do:

    boost::format("Date %3% %2% %1%") % month % day % year

 and in fmt you can do:

    fmt::format("Date: {3} {2}, {1}", month, day, year)

So that's up to the translator, the capabilities are already there and just require changing the format string in the translation.

Named arguments are supported as well:

    fmt::format("Date: {year} {month}, {day}", fmt::arg("month", month), fmt::arg("day", day), fmt::arg("year",year))

Though they might cause more problems than they solve if people accidentally start translating the keywords, so I am not sure if it's worth using them.

As for resolving ambiguities, gettext (and I assume tinygettext) supports an additional context argument that can be used to resolve ambiguities when identical strings need different translations depending on where they are used:

https://www.gnu.org/software/gettext/manual/html_node/Contexts.html

CMakeLists.txt
Grumbel: Should be an easy enough change if desired, though I can of like sticking with one library version than hoping that whatever gets shipped with a random distribution actually works.

Semphriss:
Ah, interesting, I wasn't aware of that.

Should we inform the translators in some way?

The keyword proposition sounds interesting, although indeed we would have to make sure translators don't accidentally translate them.

Perhaps a keyword-looking format like `{FMT_MONTH_STR}` and `{FMT_DAY_INT}` would give enough of a hint to translators? I saw Transifex supports checking for presence of `%s`'s, links and parentheses, maybe it (and/or other services) would detect that words in brackets are keywords and should stay the same?

Semphriss:
Android builds are failing, it doesn't detect fmt. I'll try to see if it can be added.

.github/workflows/android.yml
```diff
@@ -67,6 +67,10 @@ jobs:
         run: |
           sed -i "s/MultiABI=.*/MultiABI='${ARCH_LIST}'/g" project/jni/application/supertux/AndroidAppSettings.cfg
 
+          # FIXME: Tweaks to get fmt compiled, should probably be upstreamed
+          sed -i "s#\(AppSubdirsBuild='[^']*\)'#\1 supertux/external/fmt/src supertux/external/fmt/include'#" project/jni/application/supertux/AndroidAppSettings.cfg
```
Semphriss, line 71: @pelya Can this be sent upstream?

pelya:
Yes, please create a merge request. I cannot edit these files myself
because of the war.

Semphriss:
Opened as https://github.com/pelya/commandergenius/pull/136

PR #2012, opened 2021-12-23
Use separate sector for cutscenes, instead of file

Vankata453:
Fixes an issue, where cutscenes will set the best level statistics of a level to only the best level statistics of the cutscene (all equal to 0, most of the time). Cutscenes are now used in a level's file, just in a separate sector. The cutscene level files are also deleted. Levels changed are: The Castle of Nolok, Tower of Ghosts.

This PR also removes some `wait()` scripts from the trigger, which starts the cutscene in Tower of Ghosts and fixes some minor issues in the Tower of Ghosts cutscene:

* Fixes a typo.
* Uses initialization script, instead of a trigger.
* Adds `start_cutscene()` and `end_cutscene()` to the script of the cutscene.
* Makes all texts (excluding the first one) stay on the screen for a second or two more.


Semphriss (approved):
Didn't test much, but codewise this looks good.

Rusty-Box:
Wouldn’t that mean that the time for the cutscene is also counted towardsthe total play time? I don’t think that’s how this should be handled. Should we rather fix the initial bug?

Vankata453:
Also added a commit that changes a text in Tower of Ghosts's cutscene:

Old:

> It's from Penny, telling him that Nolok took her to an Tropical World. He seemed very upset about it.

New:

> It's from Penny, telling him that Nolok took her to a Tropical World and seemed very upset about it.

It fixes a typo + makes it a little more clearer that Nolok is upset, in my opinion.

Semphriss:
The conjunction of both parts with "and" sounds off to me. They have a different purpose in the text (one is what Nolok wrote, and the other is Tux's impression of the letter).

With that version, it seems like Penny told him in her letter that Nolok was very upset; if that is the case, I'd suggest:
> It's from Penny, telling him that Nolok took her to a Tropical World, and that he seemed very upset about it.

...making it clearer that it is neither Penny nor her letter itself who's upset about it, but rather Nolok.

Another possibility would be to keep the original version but replace the last "he" with something that refers to Nolok, like "The creature" or similar.

Vankata453:
The first suggestion seems very good, thanks!

PR #2014, opened 2021-12-23
Don't count play time when in cutscene

mrkubax10:


src/supertux/game_session.cpp
```diff
@@ -409,8 +409,11 @@ GameSession::update(float dt_sec, const Controller& controller)
     assert(m_currentsector != nullptr);
     // Update the world
     if (!m_end_sequence) {
-      m_play_time += dt_sec; //TODO: make sure we don't count cutscene time
-      m_level->m_stats.finish(m_play_time);
+      if (!m_level->m_is_in_cutscene)
+      {
+        m_play_time += dt_sec;
+        m_level->m_stats.finish(m_play_time);
```
Semphriss, line 415: Not on you, but generally speaking (and if anyone knows); why is it the "finish" function that's called here? It updates the time but it also sets the time as "Final", which doesn't seem to be the intended effect...

mrkubax10: Yeah I also was wondering why

PR #2023, opened 2021-12-25
Fix Putrid Pools (and other Halloween levels)

Narre:
When the Halloween Special was added, the final level exploited a lighting bug to make the lava appear green. Since this bug was fixed a long time ago, the lava is now orange, so I changed its tint to resemble the originally intended colour (the lighting around it still isn't green, but I think that can only be fixed by making lava shine the colour of the tilemap it's in).

I also changed the z-pos of the foreground darkness in all Halloween levels from 500 to 200 since at 500, the coin and powerup count was completely obscured.

**0.3.5:**
putrid_035

**0.6.3:**
putrid_063

**this PR:**
putrid_pr


Semphriss:
Hmm, the green lava with orange lighting looks quite odd... Is there a particular reason for lava there to be green?

data/levels/halloween2014/pools.stl
```diff
@@ -990,34 +643,13 @@
...
       (x 9632)
       (y 864)
     )
-    (candle
-      (flicker #f)
-      (sprite "images/objects/candle/no_candle.sprite")
-      (color 0.01960784 0.6078432 0.01960784)
-      (x 9504)
-      (y 992)
-    )
```
Semphriss: Why have so many candles been removed? Was it intended?

data/levels/halloween2014/pools.stl
```diff
@@ -1807,18 +1243,18 @@
...
... 3081 3081 3081 3081 3078 3077 3077 3081 3082 3081 3081 3077 3077 3077 3078 3077 3078 3078 3077 3077 3077 3078 3081 3078 3077 3081 3077 3077 3082 3077 3082 3081 3077 3077 3077 3082 3082 3078 3081 3077 3078 3082 3082 3081 3081 3081 3077 3081 3081 3082 3078 3078 3082 3081 3077 3078 3077 3081 3081 3077 3078 3082 3081 3081 3078 3081 3082 3082 3082 3077 3082 3077 3077 3082 3077 3081 3081 3082 3078 3077 3078 3081 3077 3078 3082 3081 3081 3081 3077 3081 3082 3077 3077 3081 3082 3077 3082 3077 3077 3081 3081 3092 3073 3075 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 3080 3081 3081 3082 3082 3081 3077 3102 3103 3081 3082 3077 3082 3081 3077 3077 3077 3077 3077 3077 3078 3077 3082 3081 3082 3094 3095 3078 3078 3077 3082 3077 3081 3081 3082 3081 3078 3077 3081 3081 3081 3078 3077 3096 3097 3081 3078 3077 3078 3081 3082 3078 3094 3095 3082 3082 3077 3077 3078 3082 3082 3077 3081 3096 3097 3078 3081 3078 3082 3102 3103 3081 3082 3077 3082 3081 3082 3078 3078 3082 3078 3077 3082 3078 3077 3081 3081 3077 3082 3078 3082 3082 3082 3082 3081 3081 3078 3077 3077 3081 3077 3078 3077 3077 3077 3077 3081 3081 3081 3078 3077 3077 3078 3082 3077 3081 3094 3095 3078 3082 3082 3081 3081 3081 3082 3081 3078 3082
```
Semphriss: What are the tile changes?

Semphriss (changes requested):

data/levels/halloween2014/pools.stl
Narre: Yes, the candles were there to make the lava green; the bug this level exploited meant that the lava wouldn't emit any light and these candles (placed in the lava) are what made it green and also glow green, but since that bug has been fixed and the glow of the lava is stronger than the glow of the candles, the candles didn't do anything.

data/levels/halloween2014/pools.stl
Narre: I removed the lava from the untinted "Interactive" tilemap and put them to the new green-tinted tilemap.

Narre:
> Hmm, the green lava with orange lighting looks quite odd... Is there a particular reason for lava there to be green?

As far as I understand, it's not supposed to be lava. It's supposed to be some kind of putrid substance that hurts Tux, hence the name "Putrid Pools". Problem is, there's currently no way to make lava glow green, which I think can be considered a bug since the lava should probably glow the colour of the tinted tilemap, which, in this case, would be (0, 0.6, 0).

Semphriss:
Just to let the info here: Before this gets merged, I'd like to try and see if I can do a complementary fix to allow tilemap lightmap to be tinted as well.

Semphriss:
Given the excessive amount of time that passed without me finding a solution, I retract any interest in reviewing this and will leave it to the team to decide whether or not to accept this fix.

Grumbel:
I suggest adding new slime tiles for this. The lava lighting is hardcoded (sector.cpp:689)  and there is no easy may to override it. A slime graphic can be found here: https://github.com/Pingus/pingus/blob/master/data/images/liquids/slime.png

Alternative solution would be to introduce a liquid game object that allows more configuration than raw tiles.

PR #2025, opened 2021-12-25
Tiles from other tilesets can now be imported from the strf files

### LINKED ISSUE #325: Allow custom tiles for levels
In reference to 057aa7dbf6257d5b4bd697b4080b25ef352cc1d9

Removed support for multiple Tilesets inside a level file, i.e. the (tilesets ...) tag

Neither the official nor addon levels make use of this feature, it
isn't support by any of the editors either. Addons that use custom
tilesets are just defining a new tileset, instead of mixing them in
the level file, so the ability to use custom tilesets inside a addon
is still available.

If there is any demand for this feature it should be brought back as
function inside the .strf files, not the .stl files, i.e. something
like:

```
(supertux-tiles
   (import-tileset
      (file "tilesetname.strf")
      (start 0)
      (end   2153)
      (offset 0))

   ;; custom tiles go here

   ...))
```

The .stl file would then refer to a single tileset file, instead of multiple as once. Thus allowing sharing of tilesets between level files without the need to duplicate the imports in every .stl

Narre:
Fixes #325

Implements tile importing in the way Grumbel suggested in the issue. This only imports tiles, since tilegroups would be full of non-imported tiles (therefore holes) and the user may want to define their own tilegroups anyway. Similarly, autotiles may depend on non-imported tiles and are therefore not imported.

The most obvious use case for this feature is that if a player wants to add custom tiles into their add-on while keeping the vanilla tiles.strf tiles, they previously had to either redefine tiles.strf (which broke a lot of stuff when the add-on was used), or to create a new strf file and copy all the contents of tiles.strf manually, which could break if tiles.strf got updated (for example, an image was changed and the existing ids were remapped to the new image). Now the player could just create a new strf, import tiles they need, add their own and mix & match them all into their own tilegroups.

Note: a full tileset (with no limit and no offset) can be imported using (start 0)(end 0)(offset 0) - this will also import autotiles and tilegroups, however, if new tiles are then added into the imported tileset, the user may redefine their ids by having used them before

Please test! I haven't had time to throw every possible problem at it (like offsets making the tiles go below 1, and other edge cases). I will test this myself later today, but the more people test it, the better!\
_edit:_ I have now tested (and fixed) a few things (see the commits)

example: if I want to make an add-on, and I need to add two tiles, I can now just make a strf that looks like this:

```
(supertux-tiles
  (import-tileset
    (file "images/tiles.strf")
    (start 0)   ;; I want the tiles from the beginning
    (end 4326)  ;; This is the last tile in the file at this time. I will use the higher ids myself
    (offset 0)  ;; I want the tiles to retain their original ids without an offset
  )
  (tiles
    (image "images/tiles/my_tiles.png")
    (ids 4327 4328)
    (attributes ...)
    (data ...)
  )
  (import-tileset
    (file "images/tiles.strf")
    (start 4327) ;; New tiles were added since made this tileset, I want to import them (the new ones only)
    (end 4400)   ;; This is the current last id
    (offset 2)   ;; Since I have already used two ids for my tiles, I need to offset these tiles by 2
  )
)
```

I would need to define tilegroups and autotiles manually, but it's still less tedious than defining everything manually and I would be able to mix my new tiles into them.
If you have any better ideas for tilegroup and/or autotile importing, please tell me. Thank you.

### DIFF
--- a/src/supertux/tile_set_parser.cpp
+++ b/src/supertux/tile_set_parser.cpp
@@ -39,7 +39,7 @@ TileSetParser::TileSetParser(TileSet& tileset, const std::string& filename) :
 }
 
 void
-TileSetParser::parse(uint32_t start, uint32_t end, int32_t offset)
+TileSetParser::parse(uint32_t start, uint32_t end, int32_t offset, bool imported)
 {
   if (offset && static_cast<int32_t>(start) + offset < 1) {
     start = -offset + 1;
@@ -70,8 +70,8 @@ TileSetParser::parse(uint32_t start, uint32_t end, int32_t offset)
     else if (iter.get_key() == "tilegroup")
     {
       /* tilegroups are only interesting for the editor */
-      /* ignore tilegroups for imported tilesets unless there's no limit and no offset*/
-      if (start || end || offset) continue;
+      /* ignore tilegroups for imported tilesets */
+      if (imported) continue;
       ReaderMapping reader = iter.as_mapping();
       Tilegroup tilegroup;
       reader.get("name", tilegroup.name);
@@ -85,8 +85,8 @@ TileSetParser::parse(uint32_t start, uint32_t end, int32_t offset)
     }
     else if (iter.get_key() == "autotileset")
     {
-      /* ignore autotiles for imported tilesets unless there's no limit and no offset */
-      if (start || end || offset) continue;
+      /* ignore autotiles for imported tilesets */
+      if (imported) continue;
       ReaderMapping reader = iter.as_mapping();
       std::string autotile_filename;
       if (!reader.get("source", autotile_filename))
@@ -111,7 +111,7 @@ TileSetParser::parse(uint32_t start, uint32_t end, int32_t offset)
       reader.get("end", import_end);
       reader.get("offset", import_offset);
       TileSetParser import_parser(m_tileset, import_filename);
-      import_parser.parse(import_start, import_end, import_offset);
+      import_parser.parse(import_start, import_end, import_offset, true);
     }
     else
     {
@@ -119,7 +119,7 @@ TileSetParser::parse(uint32_t start, uint32_t end, int32_t offset)
     }
   }
   /* only create the unassigned tilegroup from the parent strf */
-  if (g_config->developer_mode && !end)
+  if (g_config->developer_mode && !imported)
   {
     m_tileset.add_unassigned_tilegroup();
   }
--- a/src/supertux/tile_set_parser.cpp
+++ b/src/supertux/tile_set_parser.cpp
@@ -43,7 +43,11 @@ TileSetParser::parse(uint32_t start, uint32_t end, int32_t offset)
 {
   if (offset && (int)start + offset < 1) {
     start = -offset + 1;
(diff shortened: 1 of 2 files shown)


mrkubax10:
Please address build issues. It fails due to variables using the same name as method parameters

tobbi:
I kinda wish we could auto-import tilesets that can be found in the PhyFS path, but that's just my opinion.

Narre:
Wouldn't any kind of automatic importing inevitably result in id conflicts?

tobbi:
Okay, yes. I was imagining something like add-on specific tilesets. And if there's ID conflicts, we either take the first or the last or something.

src/supertux/tile_set_parser.cpp
```diff
@@ -70,10 +81,12 @@ TileSetParser::parse()
     else if (iter.get_key() == "tiles")
     {
       ReaderMapping tiles_mapping = iter.as_mapping();
-      parse_tiles(tiles_mapping);
+      parse_tiles(tiles_mapping, start, end, offset);
     }
     else if (iter.get_key() == "autotileset")
     {
+      /* ignore autotiles for imported tilesets unless there's no limit and no offset */
+      if (start || end || offset) continue;
```
Semphriss, line 89: It would be possible to load autotilesets by giving the loader the offset and handling it inside the parser.

That being said, how are imported tilesets handled in the editor? Is it a manual-only feature?

Narre: The offset isn't the problem, the problem is that you can only import a part of the tileset (for example, in a hypothetical situation, you may only want to import tiles 100-500 from a tileset, but its autotiles try to use tiles 501, 502 and 424242 with tile 500. What now?) If you import a reasonable range from tiles.strf without an offset, you can easily use the autotiles by just putting `(autotileset (source "autotiles.satc"))` in your strf.

As for the editor, there's currently no way to define your own tilesets from within the game, right? So imported tilesets are a feature you can use when defining your custom strf. `(import-tileset)` is just a more convenient way of using existing tiles (than manually copying them from said tileset and possibly remapping their ids), and it also guarantees that if something like a path to an image changes in the imported tileset, yours won't need to be changed to accommodate for that changed file.

Semphriss: It isn't possible to *write* a custom .strf file from withing the editor, but it is very possible to *assign* a .strf file in the editor, hence the question: how will the editor handle levels that use multiple tileset files?

Narre: The editor doesn't even know the tileset uses imported tiles, it works as it should. Quick demonstration:

https://user-images.githubusercontent.com/14074789/147425995-cc76f80f-587f-40bd-893c-34dc1fb3123d.mp4

The `test_tiles.strf` tileset imports tiles 0-500 from `tiles.strf`, assigns them ids 1000-1500, and defines 4 new tiles with ids 1-4. Then there are 3 tilegroups defined, one with some imported tiles, one with some new tiles, and one with a mix of both. Everything else is correctly put into the unassigned "Others" tilegroup, with the correct ids.

Narre:
I just realised that nesting imported tilesets doesn't work as it should, fixing that now.

Semphriss (approved):

PR #2031, opened 2021-12-27
Add delete world option to editor

mrkubax10:


src/physfs/util.cpp
```diff
@@ -60,6 +60,21 @@ bool remove(const std::string& filename)
...
+  for (const char* const* file = files; *file != nullptr; file++)
+  {
+    std::string path = FileSystem::join(filename, *file);
+    if (is_directory(path))
+      remove_with_content(path);
+    PHYSFS_delete(path.c_str());
+  }
+  PHYSFS_freeList(files);
+  PHYSFS_delete(filename.c_str());
+}
```
Semphriss, line 76: Small detail, but it would be preferable to keep `free` functions at the very end, so that if anyone else comes to maintain the function, they can immediately spot the manually managed lifetime.

src/supertux/menu/editor_levelset_select_menu.hpp
```diff
@@ -21,6 +21,7 @@
 
 class EditorLevelsetSelectMenu final : public Menu
 {
+  friend class EditorDeleteLevelsetMenu;
```
Semphriss, line 24: Please don't use `friend`; add the function needed as appropriate.

`EditorLevelsetSelectMenu` is a menu class; it would be preferable to find an appropriate utility class to handle deleting a world.

Semphriss (changes requested):

src/physfs/util.cpp
```diff
@@ -60,6 +60,21 @@ bool remove(const std::string& filename)
   return PHYSFS_delete(filename.c_str()) == 0;
 }
 
+void
+remove_with_content(std::string& filename)
```
tobbi, line 64: I think you can mark the filename reference `const`

tobbi: Or does that lead to conflicts with line 75?

Semphriss: In any case, the caller does not need to have the string edited, so it should be passed by `const&` and copied inside the function if necessary.

src/supertux/menu/editor_delete_levelset_menu.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+void
+EditorDeleteLevelsetMenu::menu_action(MenuItem& item)
+{
+  int id = item.get_id();
+  if (id >= 0)
+  {
+    std::vector<std::string>& contrib_worlds = m_editor_levelset_select_menu->get_contrib_worlds();
+    if (Editor::is_active() && Editor::current()->get_world() && Editor::current()->get_world()->get_basedir() == contrib_worlds[id])
+      Dialog::show_message(_("You cannot delete world that you are editing"));
```
Semphriss, line 72: "...delete *the* world..."

Semphriss:
The code looks all good, but imo it would need more spacing - would it be fine if I opened a PR to your branch?

mrkubax10:
What do you mean by more spacing?

Semphriss:
Add some whitespace to separate the different parts of the code, to give some breathing air

src/supertux/menu/editor_levelset_select_menu.hpp
```diff
@@ -32,6 +32,8 @@ class EditorLevelsetSelectMenu final : public Menu
   void initialize();
   void reload_menu();
 
+  std::vector<std::string>& get_contrib_worlds() { return m_contrib_worlds; }
```
tobbi, line 35: I believe this can be marked `const`

PR #2039, opened 2021-12-28
Fix ghost tree not appearing properly

bruhmoent:
https://user-images.githubusercontent.com/69918580/147585225-58fa1e2c-7be3-4b4d-921d-5e5ae6dba5f3.mp4

Before ^

https://user-images.githubusercontent.com/69918580/147584927-ee0be863-9490-4ef9-9295-8ae5d16a30c4.mp4

After ^


tobbi:
Can you explain what the bug was? I am not sure I've ever experienced it.

bruhmoent:
> Can you explain what the bug was? I am not sure I've ever experienced it.

There was a mistake, when someone put a normal tree in a different layer, that appeared after passing goto ghostforest. I removed the tree and placed it in the correct layer.

tobbi:
Ah, alright. Thanks 👍

PR #2040, opened 2021-12-28
Show progress in contrib level menus

### LINKED ISSUE #1251: Old levelpacks are marked as new

Narre:
Fixes #1251

I fixed the feature that was broken for so long it got removed entirely. And I made it better.

progress_official
progress_user


bruhmoent:
I love it!

Narre:
Feedback please! Should I remove the `worlds[i]->get_title() == wm_title` in 
image
?? I'm not sure if it's a good idea to separate the progress indicator of the add-on and the worldmap just because their titles don't match.

mrkubax10:
Also maybe don't show progress for levelsets

Narre:
Why though? Is there a particular reason why the progress of levelsets shouldn't be shown? There is no rule saying that all non-test levels must be in a worldmap, it's just that most people choose to create a worldmap for their levels, but there's nothing stopping them from creating a levelset of playable levels, and then the progress makes sense there.

Narre:
Also I decided to remove that check that would separate world and worldmaps if their titles didn't match since I don't think it makes much sense. I originally had that in place before I implemented the total world level count so I couldn't check if the numbers were the same but now I feel it makes things less intuitive and unnecessary.

data/levels/christmas2016/worldmap.stwm
```diff
@@ -1,6 +1,6 @@
 (supertux-level
   (version 3)
-  (name (_ "Christmas 2016"))
+  (name (_ "Christmas Special"))
   (author "Carsten Wirtz")
```
Semphriss, line 4: Were the name changes intended to be part of this PR? (They weren't mentioned in the PR desc)

There's also `Community Island 2020` -> `Community Island` and `Haloween 2014` -> `Halloween Special`.

src/supertux/menu/sorted_contrib_menu.cpp
```diff
@@ -17,15 +18,17 @@
 #include "supertux/menu/sorted_contrib_menu.hpp"
 
 #include <sstream>
-#include "util/gettext.hpp"
+#include "boost/format.hpp"
+#include "gui/menu_manager.hpp"
+#include "gui/menu_item.hpp"
```
Semphriss, line 23: Please separate system includes, dependency includes and SuperTux includes. You can find the details in [the coding style document](https://github.com/SuperTux/supertux/blob/master/CODINGSTYLE.md#includes).

src/supertux/menu/sorted_contrib_menu.cpp
```diff
@@ -38,10 +41,78 @@ SortedContribMenu::SortedContribMenu(std::vector<std::unique_ptr<World>>& worlds
...
+          title_str = str(boost::format(_("[%s] (%u/%u; %u%%)")) % worlds[i]->get_title() %
+                          solved_count % level_count % (100 * solved_count / level_count));
+        }
+      }
       else
-        title_str = worlds[i]->get_title();
+      {
+        uint32_t island_level_count = 0, island_solved_count = 0,
+                 world_level_count = 0,  world_solved_count = 0;
+        auto wm_filename = savegame->get_player_status().last_worldmap;
```
Semphriss, line 71: Please add some empty lines in here, it feels cramped :smile:

src/supertux/menu/sorted_contrib_menu.cpp
```diff
@@ -38,10 +41,78 @@ SortedContribMenu::SortedContribMenu(std::vector<std::unique_ptr<World>>& worlds
...
+          }
+          else
+          {
+            /* This is translatable since RTL languages may prefer to put the progress
+               info to the left of the title */
+            title_str = str(boost::format(_("%s (%u/%u; %u%%) - %s (%u/%u; %u%%)")) % worlds[i]->get_title() % 
+                            world_solved_count % world_level_count % (100 * world_solved_count / world_level_count) %
+                            wm_title % island_solved_count % island_level_count %
+                            (island_level_count ? (100 * island_solved_count / island_level_count) : 100));
+          }
```
Semphriss, line 113: Is there a reason to duplicate the count for a world and for a worldmap? Isn't the world's stats enough?

src/supertux/savegame.cpp
```diff
@@ -323,6 +323,13 @@ Savegame::get_worldmaps()
 WorldmapState
 Savegame::get_worldmap_state(const std::string& name)
 {
+  std::string path;
+  if (name[0] == '/') {
+    path = name;
+  }
+  else {
+    path = "/" + name;
+  }
```
Semphriss, line 332: This check seems needless - it should be on the caller to ensure that there is a leading slash if a leading slash is necessary. (Also, it would be preferable to use the FileUtils where possible.)

src/supertux/player_status.hpp
```diff
@@ -61,6 +61,7 @@ class PlayerStatus final
 
   std::string worldmap_sprite; /**< the sprite of Tux that should be used in worldmap */
   std::string last_worldmap; /**< the last played worldmap */
+  std::string last_worldmap_title; /**< the title of the last played worldmap */
```
Semphriss, line 65: Marking this in case the worldmap data gets removed, this variable will be unnecessary.

Semphriss (changes requested):

data/levels/christmas2016/worldmap.stwm
Narre: I renamed them because I originally had that check in place that would show both names from `info` and the name of the worldmap if the titles didn't match even if the level counts did. I removed that check now (that's what I wanted feedback about, but ultimately decided was pointless and removed it), so the titles don't need to be changed anymore. I still think it's better this way (the worldmap titles are consistent with the title in the `info` file), but if you think I should change them back, I will. These don't show up anywhere in the game anyway.

src/supertux/menu/sorted_contrib_menu.cpp
Narre: I originally thought it may be enough the other way round :D

I think having both counts is good for bigger worlds with more than one worldmap, like 
image
For worlds which only have one worldmap (ie the level count in the worldmap and the world are identical), only one will show.

data/levels/christmas2016/worldmap.stwm
Semphriss: As a matter of principle, it would be preferable to discuss add-on renamings with the community, just to ensure that the new name makes sense and that everybody is fine with it.

src/supertux/menu/sorted_contrib_menu.cpp
```diff
@@ -38,10 +42,89 @@ SortedContribMenu::SortedContribMenu(std::vector<std::unique_ptr<World>>& worlds
...
+        if (!island_level_count && !world_solved_count)
+          title_str = str(boost::format(_("%s *NEW*")) % worlds[i]->get_title());
+        else
+        {
+          const auto wm_title = savegame->get_player_status().last_worldmap_title;
+
+          if (island_level_count == world_level_count)
+          {
+            // This is translatable since RTL languages may prefer to put the progress
+            // info to the left of the title
```
Semphriss, line 113: Same here

Narre:
Those titles aren't the names of the Add-Ons (those are in the `info` file). These are the names of the worldmaps (the stwm "levels"), which don't show up anywhere in the game. I just made them consistent with the Add-On names in `info`.

src/supertux/menu/sorted_contrib_menu.cpp
```diff
@@ -17,15 +18,18 @@
 #include "supertux/menu/sorted_contrib_menu.hpp"
 
 #include <sstream>
-#include "util/gettext.hpp"
+
+#include "boost/format.hpp"
+#include "gui/menu_manager.hpp"
```
Semphriss, line 23: One small detail: There should be an empty line between headers from dependencies (SDL, Physfs, Boost, etc.) and headers from the source directory

Semphriss (changes requested):

Semphriss (approved):

src/supertux/menu/sorted_contrib_menu.cpp
```diff
@@ -38,10 +43,83 @@ SortedContribMenu::SortedContribMenu(std::vector<std::unique_ptr<World>>& worlds
...
+      {
+        uint32_t level_count = 0, solved_count = 0;
+        const auto& state = savegame->get_levelset_state(worlds[i]->get_basedir());
+
+        for (const auto& level_state : state.level_states)
+        {
+          if (level_state.filename.empty() || level_state.filename.back() == '~') continue;
+          if (level_state.solved) ++solved_count;
+          ++level_count;
+        }
```
Zwatotem, line 58: I feel like this counting logic doesn't belong here, and could be better reused, if it was moved to a dedicated `LevelsetState` method.

src/supertux/menu/sorted_contrib_menu.cpp
```diff
@@ -38,10 +43,83 @@ SortedContribMenu::SortedContribMenu(std::vector<std::unique_ptr<World>>& worlds
...
+
+        for (const auto& world : world_list)
+        {
+          const auto& world_state = savegame->get_worldmap_state(world);
+          for (const auto& level_state : world_state.level_states)
+          {
+            if (level_state.filename.empty()) continue;
+            if (level_state.solved) ++world_solved_count;
+          }
+        }
```
Zwatotem, line 101: Same as before, but for `WorldmapState` as well. The code is being repeated at this point.

Zwatotem (changes requested):
Please apply the suggestions that I highlighted. (Or inform me why they shouldn't be)

src/supertux/menu/sorted_contrib_menu.cpp
```diff
@@ -1,6 +1,6 @@
 //  SuperTux
 //  Copyright (C) 2021 mrkubax10 <mrkubax10@onet.pl>
-//                2021 Jiri Palecek <narre@protonmail.com>
+//                2022 Jiri Palecek <narre@protonmail.com>
```
Semphriss, line 3: You can put `2021-2022` for proper copyright

src/supertux/menu/sorted_contrib_menu.cpp
```diff
@@ -61,54 +61,40 @@ SortedContribMenu::SortedContribMenu(std::vector<std::unique_ptr<World>>& worlds
...
 
-        const auto& state = savegame->get_worldmap_state(wm_filename);
-        for (const auto& level_state : state.level_states)
-        {
-          if (level_state.filename.empty()) continue;
-          if (level_state.solved) ++island_solved_count;
-          ++island_level_count;
-        }
-
+        auto world_level_count = std::make_pair(0u, 0u);
```
Semphriss, line 71: Declaring `std::pair<uint32_t, uint32_t>` helps avoid an unnecessary construction.

src/supertux/menu/sorted_contrib_menu.cpp
```diff
@@ -61,54 +61,40 @@ SortedContribMenu::SortedContribMenu(std::vector<std::unique_ptr<World>>& worlds
...
-          const auto wm_title = savegame->get_player_status().last_worldmap_title;
+          const std::string wm_title = savegame->get_player_status().last_worldmap_title;
 
-          if (island_level_count == world_level_count)
+          if (island_level_count.second == world_level_count.second)
           {
             title_str = str(boost::format(_("%s (%u/%u; %u%%)")) % worlds[i]->get_title() %
-                            island_solved_count % island_level_count % (100 * island_solved_count / island_level_count));
+                            island_level_count.first % island_level_count.second %
+                            (100 * island_level_count.first / island_level_count.second));
```
Semphriss, line 89: I'm not sure how safe this is (in terms of readability). Normally it would be preferable to use `static_cast<float>` for both, and then `floor()` if necessary. That would make the line quite long though, so using a temporary variable to hold the completion percentage would be preferable.

src/supertux/menu/sorted_contrib_menu.cpp
```diff
@@ -61,54 +61,40 @@ SortedContribMenu::SortedContribMenu(std::vector<std::unique_ptr<World>>& worlds
...
           else
           {
             title_str = str(boost::format(_("%s (%u/%u; %u%%) - %s (%u/%u; %u%%)")) % worlds[i]->get_title() % 
-                            world_solved_count % world_level_count % (100 * world_solved_count / world_level_count) %
-                            wm_title % island_solved_count % island_level_count %
-                            (island_level_count ? (100 * island_solved_count / island_level_count) : 100));
+                            world_level_count.first % world_level_count.second %
+                            (100 * world_level_count.first / world_level_count.second) %
+                            wm_title % island_level_count.first % island_level_count.second %
+                            (island_level_count.second ? (100 * island_level_count.first / island_level_count.second) : 100));
```
Semphriss, line 97: Same here

mrkubax10:
What is left to do for this pull request besides resolving merge conflict?

Narre:
I think I'm (for the most part) done with this PR. However, there is one minor annoyance I'm not sure how to fix and a few design choices that may be worth considering, but can be addressed in another PR.

- **The minor annoyance** is that the title of the current worldmap is saved in the savegame file in the language SuperTux is set to at the time of saving (loading the particular worldmap), and if the language is changed afterwards, the title will only update after reloading the worldmap. This can be fixed by saving the last worldmap title without translation and translating it directly in `sorted_contrib_menu.cpp`, however, the `m_name` variable loaded in `worldmap.cpp` loads the already translated string and I couldn't figure out where it's pulling it from (or rather, how it's pulling it). _Maybe using an actual IDE instead of a plain text editor would help..._

- **The design choices** include the most obvious one - whether and how (not) to count cutscenes and similar towards the level count. This could be achieved with a bool in the `.stwm` map, but then the total level count (between all worldmaps) would not see it. This would probably be better addressed separately in another PR since I'm not sure how I would do it and I'm sure there are people better at tackling this than me.

- **The other design choice** I thought about is where to show the progress. Currently, in this PR, the progress is shown the way it used to be before this feature broke (next to the world title), but since this line is kinda long now (because for worlds that consist of more than one worldmap, it shows overall progress as well as the progress in the current worldmap). I thought it may be better to show the progress in another manner, for example, at the bottom of the screen alongside the description, but this would require some discussion and when decided, can be done in a single simple commit.

Narre:
Also, before anyone asks, I removed the Community Island to resolve a merge conflict since it was already removed on `master`

mrkubax10:
> The other design choice I thought about is where to show the progress. Currently, in this PR, the progress is shown the way it used to be before this feature broke (next to the world title), but since this line is kinda long now (because for worlds that consist of more than one worldmap, it shows overall progress as well as the progress in the current worldmap). I thought it may be better to show the progress in another manner, for example, at the bottom of the screen alongside the description, but this would require some discussion and when decided, can be done in a single simple commit.

Since descriptions aren't usually long I think that moving progress information there is a good idea.

mrkubax10:
> 

The minor annoyance is that the title of the current worldmap is saved in the savegame file in the language SuperTux is set to at the time of saving (loading the particular worldmap), and if the language is changed afterwards, the title will only update after reloading the worldmap. This can be fixed by saving the last worldmap title without translation and translating it directly in sorted_contrib_menu.cpp, however, the m_name variable loaded in worldmap.cpp loads the already translated string and I couldn't figure out where it's pulling it from (or rather, how it's pulling it). Maybe using an actual IDE instead of a plain text editor would help...

This is because the lisp files can contain translations inside of them. Look at this `name` entry from `data/world1/worldmap.stwm`:
`(name (_ "Icy Island"))` (`_` means translated string). If I understand this correctly worldmap name is automatically translated in `get` method of `ReaderMapping`: https://github.com/SuperTux/supertux/blob/master/src/util/reader_mapping.cpp#L131, which is called there: https://github.com/SuperTux/supertux/blob/master/src/worldmap/worldmap_parser.cpp#L69

Narre:
I'm closing this since it's going to be replaced by #2349.

Issue #1552, opened 2020-10-17
Improve general code quality


Semphriss:
Right now:
- [ ] The code contains several `FIXME`s and `TODO`s which are undealt with
- [ ] The clang compiler generates extra warnings which I believe should be corrected
- [ ] The [guidelines](https://github.com/SuperTux/supertux/blob/master/CODINGSTYLE.md) are not always respected <sub>Not gonna lie, I'm one of the most guilty people about that</sub>
- [ ] The documentation is probably incomplete

Although I don't think this is a priority, I think the quality of our code should be improved and better match the above criteria.

I'm opening this issue to get some feedback on its relevace. Would it be useful to eventually spend some time on this?

erratic-c-programmer:
Related: can we use tabs instead of spaces?
Then people who like two spaces can configure their editor; those who hate it (like me) can set tab width to 4/8

erratic-c-programmer:
Also I see a lot of assert(false)'s, wouldn't that just either i) make the game crash unconditionally or ii) do nothing at all if NDEBUG is set?

HybridDog:
In my opinion one-line methods, for example [LayerIcon::is_tilemap()](https://github.com/SuperTux/supertux/blob/master/src/editor/layer_icon.cpp#L99), should be defined in the header files.

HybridDog:
I've passed `-DCMAKE_EXPORT_COMPILE_COMMANDS=ON` to cmake and executed `python3 /usr/bin/run-clang-tidy-10.py -checks='-*,boost-*,bugprone-*,performance-*'` in the build folder to get clang tidy warnings: 
[clang_tidy_output.txt](https://github.com/SuperTux/supertux/files/5757855/clang_tidy_output.txt)

It detects, for example, that the first `break` in this switch-case in yeti.cpp is redundant:
https://github.com/SuperTux/supertux/blob/18955515e0622fa8e16987aaac489fb3bebc16ea/src/badguy/yeti.cpp#L325-L328

erratic-c-programmer:
Let’s not Implicit fallthrough like that; not much gain I’d wager

HybridDog:
cppcheck (`cppcheck . --enable=warning,style,missingInclude --quiet`) is relatively fast and can also detect some style problems:

[object/player.cpp:1118]: (style) Condition 'm_physic.get_velocity_x()>=0' is always true
[object/player.cpp:1571]: (style) Condition 'm_physic.get_velocity_y()<=0' is always true
[...]

PR #2042, opened 2021-12-29
Improvements to flipping

### LINKED ISSUE #2002: Sector flip bug list
I don't expect these to be fixed before the release (the release is today, after all!) but I will keep it here for the future:

- Backgrounds/gradients don't flip along with the sector.
- Flipped doors don't adjust themselves to still appear on the ground.  This causes them to appear slightly off-ground when flipped.
- Similarly, platform nodes don't adjust to accomodate the height of the platform either, which causes weird clipping with poles that the platforms might have been placed with.  Either this could be fixed by the node offset PR (#1974) or the path nodes could be specifically adjusted for height when the level gets flipped.

There might be more, but these are what I think of right now.

Narre:
Fixes #2002

This PR fixes the offsets `PathObject`s had on levelflip (which could in some cases extend a sector's height just by flipping, because a flipped path-following tilemap got offset beyond the sector's boundaries). It also changes the hitbox of `door.sprite` (just slightly) to prevent doors from having a gap underneath them on levelflip.

It also makes backgrounds and gradients flip.

### DIFF
--- a/src/object/coin.cpp
+++ b/src/object/coin.cpp
@@ -317,6 +317,13 @@ Coin::after_editor_set()
   }
 }
 
+void
+Coin::on_flip(float height)
+{
+  MovingSprite::on_flip(height);
+  PathObject::on_flip();
+}
+
 ObjectSettings
 HeavyCoin::get_settings()
 {
--- a/data/images/objects/door/door.sprite
+++ b/data/images/objects/door/door.sprite
@@ -1,14 +1,14 @@
 (supertux-sprite
   (action
     (name "closed")
-    (hitbox 8 4 44 80)
+    (hitbox 8 4 44 92)
     (images 
       "door-0.png"
     )
   )
   (action
     (name "opening")
-    (hitbox 8 4 44 80)
+    (hitbox 8 4 44 92)
     (images 
       "door-0.png"
       "door-1.png"
@@ -22,14 +22,14 @@
   )
   (action
     (name "open")
-    (hitbox 8 4 44 80)
+    (hitbox 8 4 44 92)
     (images 
       "door-7.png"
     )
   )
   (action
     (name "closing")
-    (hitbox 8 4 44 80)
+    (hitbox 8 4 44 92)
     (images 
       "door-7.png"
       "door-6.png"
@@ -43,7 +43,7 @@
   )
   (action
     (name "locked")
-    (hitbox 8 4 44 80)
+    (hitbox 8 4 44 92)
     (images 
       "door-locked.png"
     )
--- a/src/badguy/willowisp.cpp
+++ b/src/badguy/willowisp.cpp
@@ -352,4 +352,11 @@ WillOWisp::move_to(const Vector& pos)
   set_pos(pos);
 }
 
+void
+WillOWisp::on_flip(float height)
+{
+  BadGuy::on_flip(height);
+  PathObject::on_flip(height);
+}
+
 /* EOF */
--- a/src/badguy/willowisp.cpp
+++ b/src/badguy/willowisp.cpp
@@ -356,7 +356,7 @@ void
 WillOWisp::on_flip(float height)
 {
   BadGuy::on_flip(height);
-  PathObject::on_flip(height);
+  PathObject::on_flip();
 }
 
 /* EOF */
--- a/src/badguy/willowisp.hpp
+++ b/src/badguy/willowisp.hpp
@@ -58,6 +58,8 @@ public:
   virtual ObjectSettings get_settings() override;
   virtual void move_to(const Vector& pos) override;
 
+  virtual void on_flip(float height) override;
+
   virtual void expose(HSQUIRRELVM vm, SQInteger table_idx) override
   {
     ExposedObject<WillOWisp, scripting::WillOWisp>::expose(vm, table_idx);
--- a/src/object/background.cpp
+++ b/src/object/background.cpp
@@ -17,9 +17,11 @@
 #include "object/background.hpp"
 
 #include <physfs.h>
+#include <utility>
 
 #include "editor/editor.hpp"
 #include "supertux/d_scope.hpp"
+#include "supertux/flip_level_transformer.hpp"
 #include "supertux/gameconfig.hpp"
 #include "supertux/globals.hpp"
 #include "util/reader.hpp"
@@ -48,7 +50,8 @@ Background::Background() :
   m_target(DrawingTarget::COLORMAP),
   m_timer_color(),
   m_src_color(),
-  m_dst_color()
+  m_dst_color(),
+  m_flip(NO_FLIP)
 {
 }
 
@@ -73,7 +76,8 @@ Background::Background(const ReaderMapping& reader) :
   m_target(DrawingTarget::COLORMAP),
   m_timer_color(),
   m_src_color(),
-  m_dst_color()
+  m_dst_color(),
+  m_flip(NO_FLIP)
 {
   reader.get("fill", m_fill);
 
@@ -293,6 +297,7 @@ Background::draw_image(DrawingContext& context, const Vector& pos_)
   const int end_y   = static_cast<int>(ceilf((cliprect.get_bottom() - (pos_.y + img_h/2.0f)) / img_h)) + 1;
 
   Canvas& canvas = context.get_canvas(m_target);
+  context.set_flip(context.get_flip() ^ m_flip);
 
   if (m_fill)
   {
@@ -365,6 +370,7 @@ Background::draw_image(DrawingContext& context, const Vector& pos_)
         break;
     }
   }
+  context.set_flip(context.get_flip() ^ m_flip);
 }
 
 void
@@ -478,6 +484,16 @@ Background::load_background(const std::string& image_path)
 
   new_path = default_dir + it->second;
   return Surface::from_file(new_path);
+}
+
+void
+Background::on_flip(float height)
+{
+  GameObject::on_flip(height);
+  std::swap(m_image_bottom, m_image_top);
+  m_pos.y = height - m_pos.y - m_image->get_height();
+  m_scroll_offset.y = -m_scroll_offset.y;
+  FlipLevelTransformer::transform_flip(m_flip);
 }
 
 
--- a/src/object/background.hpp
+++ b/src/object/background.hpp
@@ -24,6 +24,7 @@
 #include "supertux/timer.hpp"
 #include "video/blend.hpp"
 #include "video/drawing_context.hpp"
+#include "video/flip.hpp"
 #include "video/surface_ptr.hpp"
 
 class ReaderMapping;
@@ -48,6 +49,8 @@ public:
 
   virtual ObjectSettings get_settings() override;
   virtual void after_editor_set() override;
+
+  virtual void on_flip(float height) override;
 
   void set_image(const std::string& name);
   void set_images(const std::string& name_top, const std::string& name_middle, const std::string& name_bottom);
@@ -104,6 +107,8 @@ private:
 
   Timer m_timer_color;
   Color m_src_color, m_dst_color;
+
+  Flip m_flip;
 
 private:
   Background(const Background&) = delete;
--- a/src/object/coin.hpp
+++ b/src/object/coin.hpp
@@ -48,6 +48,8 @@ public:
 
   virtual void move_to(const Vector& pos) override;
 
+  virtual void on_flip(float height) override;
+
   void collect();
 
 private:
--- a/src/object/gradient.cpp
+++ b/src/object/gradient.cpp
@@ -14,6 +14,8 @@
 //  You should have received a copy of the GNU General Public License
 //  along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+#include <utility>
+
 #include "object/gradient.hpp"
 
 #include "editor/editor.hpp"
@@ -267,6 +269,14 @@ Gradient::is_saveable() const
 {
   return !(Level::current() &&
            Level::current()->is_worldmap());
+}
+
+void
+Gradient::on_flip(float height)
+{
+  GameObject::on_flip(height);
+  if (m_gradient_direction == VERTICAL || m_gradient_direction == VERTICAL_SECTOR)
+    std::swap(m_gradient_top, m_gradient_bottom);
 }
 
 /* EOF */
--- a/src/object/gradient.hpp
+++ b/src/object/gradient.hpp
@@ -47,6 +47,8 @@ public:
 
   virtual ObjectSettings get_settings() override;
 
+  virtual void on_flip(float height) override;
+
   void set_gradient(Color top, Color bottom);
   void fade_gradient(Color top, Color bottom, float time);
   Color get_gradient_top() const { return m_gradient_top; }
--- a/src/object/path_object.cpp
+++ b/src/object/path_object.cpp
@@ -115,4 +115,11 @@ PathObject::editor_set_path_by_ref(const std::string& new_ref)
   m_path_uid = path_obj->get_uid();
 }
 
+void
+PathObject::on_flip()
+{
+  m_path_handle.m_scalar_pos.y = 1 - m_path_handle.m_scalar_pos.y;
+  m_path_handle.m_pixel_offset.y = -m_path_handle.m_pixel_offset.y;
+}
+
 /* EOF */
--- a/src/object/path_object.hpp
+++ b/src/object/path_object.hpp
@@ -45,6 +45,8 @@ public:
   void editor_set_path_by_ref(const std::string& new_ref);
 
 protected:
+  void on_flip();
+
   PathWalker::Handle m_path_handle;
 
 private:
--- a/src/object/platform.cpp
+++ b/src/object/platform.cpp
@@ -181,4 +181,11 @@ Platform::move_to(const Vector& pos)
   set_pos(pos);
 }
 
+void
+Platform::on_flip(float height)
+{
+  MovingSprite::on_flip(height);
+  PathObject::on_flip();
+}
+
 /* EOF */
--- a/src/object/platform.hpp
+++ b/src/object/platform.hpp
@@ -46,6 +46,8 @@ public:
 
   virtual void editor_update() override;
 
+  virtual void on_flip(float height) override;
+
   const Vector& get_speed() const { return m_speed; }
 
   /** @name Scriptable Methods
--- a/src/object/tilemap.cpp
+++ b/src/object/tilemap.cpp
@@ -414,6 +414,7 @@ TileMap::on_flip(float height)
   Vector offset = get_offset();
   offset.y = height - offset.y - get_bbox().get_height();
   set_offset(offset);
+  PathObject::on_flip();
 }
 
 void


Narre:
Just for illustration

**Before:**

https://user-images.githubusercontent.com/14074789/147709662-0de16787-2172-4a18-bef1-5038140bfb33.mp4

**After:**

https://user-images.githubusercontent.com/14074789/147709742-7853ec3b-b2ef-4554-8a50-51e6b509db92.mp4

Semphriss:
Would it be possible to flip the texture as well?

Narre:
I compiled a list of what I think should and shouldn't be flipped. I won't promise to do everything in this PR but I will try to do most of it. Feedback please!

```
Platform - always
Coin - always
Heavy coin - never
Powerup - only if it disobeys gravity
Bonusblock - always
Infoblock - always
Hurting platform - always
Bicycle/pneumatic/circular platform - always
Unstable tile - always
Skull tile - always
Weak block - always
Dispenser - only if it disobeys gravity
Ispy - always
Darttrap - always
Brick - always
Heavy brick - always
Invisible block - always
Decal - always (already implemented)
Falling block - feedback needed; no idea
Stalactite - feedback needed; no idea
Scripted object - only if it disobeys gravity
Firefly(checkpoint) - always
Rock - always
Rublight - always
Trampoline(all kinds) - never
Door - always (already implemented)
Pushbutton - always (it's not usable when on the ceiling but at least it can look good; I suggest adding a flipped sprite to use in flipped sections)
Switch(lever) - always
Torch - urgh (I'd say always but the flame will have to be moved)
Candle - the sprite should probably have a separate "flipped" action so the candle is flipped but the flame still burns upwards
Lantern - never
Magic block - always
Spotlight - never (it rotates anyway, there would be no point flipping it)

Badguys not in the list - never
```

TL;DR Unless it obeys gravity, flip it (with a few special cases)

Narre:
I worked on this for many hours straight. I hope it's worth it.

https://user-images.githubusercontent.com/14074789/147796124-c84c9f1a-3b59-4b1e-9669-ded543415329.mp4

Semphriss:
Feature-wise, this looks really good!

I'll review code quality in a while, I can see this needed changes in plenty of files :D

src/badguy/dart.hpp
```diff
@@ -49,10 +52,15 @@ class Dart final : public BadGuy
   virtual void stop_looping_sounds() override;
   virtual void play_looping_sounds() override;
 
+  virtual void on_flip(float height) override;
+
 protected:
   const BadGuy* parent; /**< collisions with this BadGuy will be ignored */
   std::unique_ptr<SoundSource> sound_source; /**< SoundSource for ambient sound */
 
+private:
```
tobbi, line 61: I believe you can move the `m_flip` variable into a base class that all objects inherit from. Either as a public variable or with getters and setters.

Narre:
I moved the `m_flip` variable to the `MovingSprite` class, made the hitbox y-offset flip properly (removing the need to only use vertically symmetrical sprite hitboxes in flipping levels), made the mole flip (and throw rocks downwards when it's flipped), made the icecrusher flip (and attack upwards when it's flipped), and I made the ghosttree flip (+ its roots and their direction).

I haven't quite figured out how the icecrusher's eyes work so they are a bit off when flipped, but it's nothing major. Other that that, I'd say this PR is probably complete feature-wise. Any feedback will be welcome though, I have heard opposing opinions on the flipping of some objects (especially platforms).

Current (updated) showcase level (not everything's in it):

https://user-images.githubusercontent.com/14074789/147841730-f4d35969-bc92-4efd-9d93-a220690fe6db.mp4

src/object/ispy.cpp
```diff
@@ -86,16 +87,16 @@ Ispy::collision(GameObject& , const CollisionHit& )
...
   if (state == ISPYSTATE_IDLE) {
     // check if a player has been spotted
     Vector eye = m_col.m_bbox.get_middle();
     if (dir == Direction::LEFT) eye = Vector(m_col.m_bbox.get_left(), m_col.m_bbox.get_middle().y);
-    if (dir == Direction::RIGHT) eye = Vector(m_col.m_bbox.get_right(), m_col.m_bbox.get_middle().y);
-    if (dir == Direction::UP) eye = Vector(m_col.m_bbox.get_middle().x, m_col.m_bbox.get_top());
-    if (dir == Direction::DOWN) eye = Vector(m_col.m_bbox.get_middle().x, m_col.m_bbox.get_bottom());
+    else if (dir == Direction::RIGHT) eye = Vector(m_col.m_bbox.get_right(), m_col.m_bbox.get_middle().y);
+    else if (dir == Direction::UP || (dir == Direction::DOWN && m_flip != NO_FLIP)) eye = Vector(m_col.m_bbox.get_middle().x, m_col.m_bbox.get_top());
+    else if (dir == Direction::DOWN) eye = Vector(m_col.m_bbox.get_middle().x, m_col.m_bbox.get_bottom());
```
Semphriss, line 99: Switch-case? :wink:

src/object/firefly.hpp
```diff
@@ -36,10 +36,14 @@ class Firefly final : public MovingSprite
...
 
+  virtual void on_flip(float height) override;
+
 private:
   SpritePtr m_sprite_light;
   bool activated;
   Vector initial_position; /**< position as in level file. This is where Tux will have to respawn, as the level is reset every time */
+
+private:
   void reactivate();
```
Semphriss, line 47: (If we want to be strict with code quality, this function should go with the other private functions above, and the [non-copyable footer](https://github.com/SuperTux/supertux/blob/master/CODINGSTYLE.md#classes) is missing. But I won't put the responsibility on you, it's only if you feel motivated to do it :smile:)

src/object/firefly.cpp
```diff
@@ -78,7 +79,10 @@ Firefly::draw(DrawingContext& context)
 
   if (m_sprite_name.find("torch", 0) != std::string::npos && (activated ||
         m_sprite->get_action() == "ringing")) {
-    m_sprite_light->draw(context.light(), m_col.m_bbox.get_middle() - TORCH_LIGHT_OFFSET, 0);
+    if (m_flip == NO_FLIP)
+      m_sprite_light->draw(context.light(), m_col.m_bbox.get_middle() - TORCH_LIGHT_OFFSET, 0);
+    else
+      m_sprite_light->draw(context.light(), m_col.m_bbox.get_middle() + TORCH_LIGHT_OFFSET, 0);
```
Semphriss, line 85: What is the difference between those two lines?

src/object/fallblock.hpp
```diff
@@ -28,18 +28,20 @@ class FallBlock : public MovingSprite
...
   
   virtual std::string get_class() const override { return "fallblock"; }
   virtual std::string get_display_name() const override { return _("Falling Platform"); }
+
+  virtual void on_flip(float height) override;
   
 protected:
-  enum State
+  enum State
   {
```
Semphriss, line 45: ...what changed on that line...? o.O

<sub>(I hope it's not an LF changed to a CRLF...)</sub>

src/object/fallblock.hpp
```diff
@@ -11,8 +11,8 @@
 //  GNU General Public License for more details.
 //
 //  You should have received a copy of the GNU General Public License
-//  along with this program.  If not, see <http://www.gnu.org/licenses/>.
-
+//  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
```
Semphriss, line 15: (uhm, same question here :thinking:)

src/object/fallblock.cpp
```diff
@@ -98,21 +99,21 @@ FallBlock::update(float dt_sec)
...
-	  SoundManager::current()->play("sounds/cracking.wav");
-	  timer.start(0.5f);
-	}
-	return FORCE_MOVE;
+    state = SHAKE;
+    SoundManager::current()->play("sounds/cracking.wav");
+    timer.start(0.5f);
+  }
+  return FORCE_MOVE;
 }
```
Semphriss, line 117: Formatting improvements! :tada:

(Nothing to change, I just want to celebrate the fact that I'm not the only one who cares about it :laughing:)

src/object/coin.cpp
```diff
@@ -234,6 +235,12 @@ HeavyCoin::update(float dt_sec)
   m_col.set_movement(m_physic.get_movement(dt_sec));
 }
 
+void
+HeavyCoin::draw(DrawingContext& context)
+{
+  m_sprite->draw(context.color(), get_pos(), m_layer);
+}
+
```
Semphriss, line 243: What is the redefinition for? Is there a difference with `MovingSprite::draw`?

src/object/block.hpp
```diff
@@ -62,6 +63,9 @@ class Block : public MovingObject
   float m_bounce_offset;
   float m_original_y;
 
+private:
+  Flip m_flip;
+
```
Semphriss, line 68: ~~Same comment as my second review (I think) - would it be possible to place this variable in `MovingSprite` instead?~~

EDIT: I just saw it inherits from `MovingObject` and not `MovingSprite` - perhaps that should be changed? (If it makes sense and Block is consistent with MovingSprite)

src/badguy/root.cpp
```diff
@@ -75,7 +79,10 @@ Root::active_update(float dt_sec)
       base_sprite->set_action("vanishing", 2);
       base_sprite->set_animation_loops(2); // TODO: doesn't seem to work for loops=1
     }
-    set_pos(m_start_position + Vector(0, offset_y));
+    if (m_flip == NO_FLIP)
+      set_pos(m_start_position + Vector(0, offset_y));
+    else
+      set_pos(m_start_position - Vector(0, offset_y));
```
Semphriss, line 85: `(m_flip == NO_FLIP) ? offset_y : -offset_y` ?

src/badguy/root.cpp
```diff
@@ -65,7 +66,10 @@ Root::active_update(float dt_sec)
       offset_y = static_cast<float>(-m_sprite->get_height());
       mystate = STATE_SHRINKING;
     }
-    set_pos(m_start_position + Vector(0, offset_y));
+    if (m_flip == NO_FLIP)
+      set_pos(m_start_position + Vector(0, offset_y));
+    else
+      set_pos(m_start_position - Vector(0, offset_y));
```
Semphriss, line 72: Same here

src/badguy/mole.cpp
```diff
@@ -78,7 +79,11 @@ Mole::collision_squished(GameObject& )
 void
 Mole::throw_rock()
 {
-  float angle = math::radians(gameRandom.randf(90.0f - 15.0f, 90.0f + 15.0f));
+  float angle;
+  if (m_flip == NO_FLIP)
+    angle = math::radians(gameRandom.randf(90.0f - 15.0f, 90.0f + 15.0f));
+  else
+    angle = math::radians(gameRandom.randf(270.0f - 15.0f, 270.0f + 15.0f));
```
Semphriss, line 86: Same here - perhaps a `base_angle` variable to hold the 270 vs 90?

src/badguy/icecrusher.cpp
```diff
@@ -297,13 +357,23 @@ IceCrusher::update(float dt_sec)
   switch (state) {
     case IDLE:
       m_col.set_movement(Vector (0, 0));
-      if (found_victim_down() && !sideways)
+      if (found_victim_up() && !sideways && m_flip != NO_FLIP)
+        set_state(CRUSHING_UP);
+      else if (found_victim_down() && !sideways && m_flip == NO_FLIP)
         set_state(CRUSHING);
-		  if (found_victim_right() && sideways)
+      else if (found_victim_right() && sideways)
```
Semphriss, line 364: Careful - does this still allow players to walk atop a crusher? Many levels depend on that functionality, such as Ice in the Hole and a secret area in the Rooted Forest

src/badguy/icecrusher.cpp
```diff
@@ -220,6 +231,55 @@ IceCrusher::collision_solid(const CollisionHit& hit)
...
+              0, 90+10*j, 140, 260, Vector(0, 300),
+              1, Color(.6f, .6f, .6f), 4, 1.6f, LAYER_OBJECTS+1);
+            Sector::get().add<Particles>(
+              Vector(m_col.m_bbox.get_left() + static_cast<float>(j) * 8.0f + 4.0f,
+                     m_col.m_bbox.get_top()),
+              270+10*j, 360, 140, 260, Vector(0, 300),
+              1, Color(.6f, .6f, .6f), 4, 1.6f, LAYER_OBJECTS+1);
+          }
+        }
+        set_state(RECOVERING_UP);
```
Semphriss, line 281: There is a lot of code duplication here - would it be possible to combine them?

src/badguy/ghosttree.cpp
```diff
@@ -154,7 +160,10 @@ GhostTree::active_update(float /*dt_sec*/)
       /* TODO indicate root with an animation */
       auto player = get_nearest_player();
       if (player) {
-        Sector::get().add<Root>(Vector(player->get_bbox().get_left(), m_col.m_bbox.get_bottom()+ROOT_TOP_OFFSET));
+        if (m_flip == NO_FLIP)
+          Sector::get().add<Root>(Vector(player->get_bbox().get_left(), m_col.m_bbox.get_bottom()+ROOT_TOP_OFFSET), m_flip);
+        else
+          Sector::get().add<Root>(Vector(player->get_bbox().get_left(), m_col.m_bbox.get_top()-ROOT_TOP_OFFSET-ROOT_HEIGHT), m_flip);
```
Semphriss, line 166: `bool ? var1 : var2`

src/badguy/ghosttree.cpp
```diff
@@ -116,7 +118,11 @@ GhostTree::active_update(float /*dt_sec*/)
 
     if (willowisp_timer.check()) {
       if (willowisps.size() < WILLOWISP_COUNT) {
-        Vector pos = Vector(m_col.m_bbox.get_width() / 2, m_col.m_bbox.get_height() / 2 + willo_spawn_y + WILLOWISP_TOP_OFFSET);
+        Vector pos;
+        if (m_flip == NO_FLIP)
+          pos = Vector(m_col.m_bbox.get_width() / 2, m_col.m_bbox.get_height() / 2 + willo_spawn_y + WILLOWISP_TOP_OFFSET);
+        else
+          pos = Vector(m_col.m_bbox.get_width() / 2, m_col.m_bbox.get_height() / 2 - willo_spawn_y - WILLOWISP_TOP_OFFSET - 32.0f);
```
Semphriss, line 125: Same here

src/badguy/dispenser.cpp
```diff
@@ -313,8 +314,15 @@ Dispenser::launch_badguy()
...
-          spawnpoint.x -= 0.5f * object_bbox.get_width();
+          if (m_flip == NO_FLIP) {
+            spawnpoint = get_anchor_pos (m_col.m_bbox, ANCHOR_BOTTOM);
+            spawnpoint.x -= 0.5f * object_bbox.get_width();
+          }
+          else {
+            spawnpoint = get_anchor_pos (m_col.m_bbox, ANCHOR_TOP);
+            spawnpoint.y -= m_col.m_bbox.get_height();
+            spawnpoint.x -= 0.5f * object_bbox.get_width();
+          }
```
Semphriss, line 325: General comment, in SuperTux the formatting should be:
```cpp
if (...)
{
  ...
}
else
{
  ...
}
```

Semphriss (changes requested):
Quite a large PR!

I've found some places that would need improvements, and a few potential bugs; please have a look whenever possible.

src/object/weak_block.cpp
```diff
@@ -203,6 +204,13 @@ WeakBlock::spreadHit()
   }
 }
 
+void
+WeakBlock::on_flip(float height)
+{
+  MovingSprite::on_flip(height);
+  FlipLevelTransformer::transform_flip(m_flip);
+}
+
```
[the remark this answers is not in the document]
Narre: No, because that would flip all `MovingSprite`s, including those that shouldn't be flipped (like all gravity-affected objects) - badguys walking on their heads.

src/object/firefly.cpp
Narre: That 
image

src/object/fallblock.hpp
Narre: It's probably the other way round. I doubt gedit, on Linux, would put CRs there.

src/object/fallblock.cpp
Narre: Yeah, mixing tabs and spaces... :rage:

I personally prefer tabs since each coder can set the indentation size to their liking in their editor's preferences without changing the code, but SuperTux uses spaces so I want to see them everywhere. As much as I hate Python, I really want to say `Error: inconsistent use of tabs and spaces` or whatever it was :D

src/object/coin.cpp
Narre: I'll have a look when I have time. `HeavyCoin`, which is gravity-affected and doesn't flip, inherits from `Coin`, which does. I know I changed the way that is handled at least twice when I was coding this, so I will double-check again.

src/badguy/root.cpp
Narre: Will do

src/badguy/icecrusher.cpp
Narre: I think I checked that, but I will check again just to be sure.

src/badguy/dispenser.cpp
Narre: I wasn't sure about that. I saw both used in ST, also Hume told me `if`s use this format, but he hasn't touched the code for years. I'll change that when I have time.

src/object/firefly.hpp
Narre: There aren't any other private functions in `firefly.hpp`, I just separated the _one_ from variables. But I will at least move it up to be under the public functions rather than the private variables.

src/object/ispy.cpp
Narre: Okay, the ispy code is a mess. In the end, I decided to rewrite most of it. The `facing-down` property is useless, direction can be determined from, well, the _direction_. I'll make the direction be determined automatically, and I will add sprite actions for facing up. Instead of making the ispy flip on levelflip, I'll change its direction from DOWN to UP (and vice versa) and reload the sprite action. Please check the code quality again when I push it. I'll also probably delete the old images since they aren't used by the sprite or anywhere in the game, they just weren't removed then the new ispy images were added.

src/object/firefly.cpp
Narre: I'll replace it with a ternary conditional

src/object/coin.cpp
Narre: Turns out I did that to not pass the `m_flip` variable to the draw function (to make it default to NO_FLIP), but since I then changed it so `m_flip` never gets transformed for `HeavyCoin`s, it makes this override unneeded. I'll remove it.

src/object/block.hpp
Narre: I honestly don't feel _at home_ enough with the code yet to evaluate and change that. It seems it might be better to make `Block` inherit from `MovingSprite`, but I don't think I'm the right person for the job. Either way, that's not really related to flipping and should probably be done in a separate PR at some point.

src/badguy/icecrusher.cpp
Narre: I checked it, it works fine.

src/badguy/icecrusher.cpp
Narre: I'll try to clean `icecrusher.cpp` a bit (this included), but since Daniel is supposedly working on an icecrusher code revamp, I won't go into any kind of extra ultra mega tidying mode.

src/badguy/dispenser.cpp
Narre: This style of format is all over `dispenser.cpp`. Fixing that now.

src/badguy/dispenser.cpp
```diff
@@ -147,20 +149,21 @@ Dispenser::Dispenser(const ReaderMapping& reader) :
...
 {
-  if (m_broken){
+  if (m_broken)
     return;
-  }
-  if (m_autotarget && !m_swivel){ // auto cannon sprite might be wrong
+
+  if (m_autotarget && !m_swivel)
+  { // auto cannon sprite might be wrong
     auto* player = get_nearest_player();
```
Semphriss, line 164: Can this comment be on a line of its own? It's a small thing but it tickles me :sweat_smile:

src/badguy/dispenser.cpp
```diff
@@ -205,22 +205,22 @@ HitResponse
 Dispenser::collision(GameObject& other, const CollisionHit& hit)
 {
   auto player = dynamic_cast<Player*> (&other);
-  if (player) {
-    // hit from above?
-    if (player->get_bbox().get_bottom() < (m_col.m_bbox.get_top() + 16)) {
+  if (player)
+  { // hit from above?
+    if (player->get_bbox().get_bottom() < (m_col.m_bbox.get_top() + 16))
```
Semphriss, line 210: Same :D

src/badguy/dispenser.cpp
```diff
@@ -232,26 +232,34 @@ Dispenser::active_update(float dt_sec)
...
         Direction targetdir = (player->get_pos().x > get_pos().x) ? Direction::RIGHT : Direction::LEFT;
-        if ( m_dir != targetdir ){ // no target: swivel cannon
+        if ( m_dir != targetdir )
+        { // no target: swivel cannon
           m_swivel = true;
           m_dir = targetdir;
           m_sprite->set_action(m_dir == Direction::LEFT ? "swivel-left" : "swivel-right", 1);
-        } else { // tux in sight: shoot
+        } else
+        { // tux in sight: shoot
```
Semphriss, line 256: Same here (and the `} else` :D)

src/badguy/dispenser.cpp
```diff
@@ -267,20 +275,24 @@ Dispenser::launch_badguy()
...
-  if (!is_offscreen() && !Editor::is_active()) {
+  if (!is_offscreen() && !Editor::is_active())
+  {
     Direction launchdir = m_dir;
-    if ( !m_autotarget && m_start_dir == Direction::AUTO ){
+    if ( !m_autotarget && m_start_dir == Direction::AUTO )
+    {
       Player* player = get_nearest_player();
-      if ( player ){
+      if ( player )
```
Semphriss, line 284: (Those ones aren't your fault, but if you're passing around that code and feel motivated, the spaces inside the parentheses...)

src/badguy/ghosttree.cpp
```diff
@@ -153,9 +157,8 @@ GhostTree::active_update(float /*dt_sec*/)
     if (root_timer.check()) {
       /* TODO indicate root with an animation */
       auto player = get_nearest_player();
-      if (player) {
-        Sector::get().add<Root>(Vector(player->get_bbox().get_left(), m_col.m_bbox.get_bottom()+ROOT_TOP_OFFSET));
-      }
+      if (player)
+        Sector::get().add<Root>(Vector(player->get_bbox().get_left(), (m_flip == NO_FLIP ? (m_col.m_bbox.get_bottom()+ROOT_TOP_OFFSET) : (m_col.m_bbox.get_top()-ROOT_TOP_OFFSET-ROOT_HEIGHT))), m_flip);
```
Semphriss, line 161: (spaces around + and -)

src/object/coin.cpp
```diff
@@ -335,4 +344,12 @@ HeavyCoin::after_editor_set()
   MovingSprite::after_editor_set();
 }
 
+// Call on_flip from grandparent class HeavyCoin to avoid flipping of
+// gravity-affected object HeavyCoin
+void
+HeavyCoin::on_flip(float height)
+{
+  MovingSprite::on_flip(height);
+}
```
Semphriss, line 353: Detail, but many IDEs interpret comments *before* a function to be documentation; it would be preferable to have the comment inside the function's body.

src/object/fallblock.cpp
Semphriss: I had a reason to prefer spaces over tabs detailed [here](https://github.com/SuperTux/supertux/issues/1552#issuecomment-973093814), but that would be for another time anyways.

src/object/firefly.cpp
```diff
@@ -148,4 +149,11 @@ Firefly::get_settings()
   return result;
 }
 
+void
+Firefly::on_flip(float height)
+{
+  MovingSprite::on_flip(height);
+  FlipLevelTransformer::transform_flip(m_flip);
+}
+
```
Semphriss, line 158: I'm starting to wonder; how feasible and maintanable would it be to use a `virtual bool is_flippable() const` inside MovingSprite objects?

src/object/ispy.cpp
```diff
@@ -86,42 +85,98 @@ Ispy::collision(GameObject& , const CollisionHit& )
...
+    if (m_sprite->animation_done())
+    {
+      switch (m_dir)
+      {
+        case Direction::DOWN:  m_sprite->set_action("hiding-down",  1); break;
+        case Direction::UP:    m_sprite->set_action("hiding-up",    1); break;
+        case Direction::LEFT:  m_sprite->set_action("hiding-left",  1); break;
+        case Direction::RIGHT: m_sprite->set_action("hiding-right", 1); break;
+        default: break;
+      }
```
Semphriss, line 129: This pattern comes back often in this file; would it be possible to add a private member function to handle this, with the action name in `const std::string&` argument?

Semphriss (changes requested):
I've got a few more comments; I started by including comments regarding code quality, but as I saw most problems were already there, at some point I decided to skip most of code quality, I'll take care of that myself later.

src/object/coin.cpp
Narre: I only just noticed I wrote that the grandparent of HeavyCoin is HeavyCoin :D
I'll fix that too (probably this evening)

src/object/firefly.cpp
Narre: It would certainly be an option, but I'm not sure how readable that would be since some objects that are flippable would need that bool set to false if they need to handle the flipping differently (like objects where the flippability changes based on whether they are set to obey gravity or not, or Ispy, which changes directions between UP and DOWN instead of flipping, but it's still flippable).

The bool itself could be added (for possible use elsewhere in the future), but I'm not sure whether it'd be a good idea to use it to do the flip transforming inside the `MovingSprite` class - then the problem above happens and it wouldn't be obvious what the bool exactly means since flipping isn't a simple "yes flip" or "no flip" situation.

src/badguy/dispenser.cpp
Narre: Yes, I will do that (and all of the above). There was so much code mess in `dispenser.cpp` that I missed something.

src/object/ispy.cpp
```diff
@@ -86,42 +72,83 @@ Ispy::collision(GameObject& , const CollisionHit& )
...
+Ispy::on_flip(float height)
+{
+  MovingSprite::on_flip(height);
+  if (m_dir == Direction::UP)
+  {
+    m_dir = Direction::DOWN;
+    m_sprite->set_action("idle-down");
+  }
+  else if (m_dir == Direction::DOWN) {
+    m_dir = Direction::UP;
```
Semphriss, line 149: Bracket line!

Semphriss (approved):
One small code quality change, and after that it's good.

PR #2045, opened 2021-12-30
Steam edits

Semphriss:
This modifies the 0.6.3 release with some Steam-specific adjustments:
- Withdrawal of the donation button, per Steam's rules regarding off-site purchases (Donations are included in "purchases")
- Added automatic pausing when the Steam overlay is open (Shift+tab) and when a joystick is unplugged, per Steam's recommended best practices

I made a separate branch based on the v0.6.3 tag in order to keep the minimum amount of changes from the release; I would prefer to stay as close to the release as possible, as using nightlies might include buggy content - including those in the 13 or so PRs I merged a few days ago. I will update the Steam builds using only that branch; it should be kept with the strict minimum of additions to make sure no bugs are added.

This PR is there to be merged on master just before the next release, so changes are included in the base game again.

(Please note that there are more things I need to do, and with the next review Steam will most likely have more suggestions, so please do not merge too early so we don,t have to merge the branch multiple times)


Semphriss:
I do not plan to do any further edit on this branch that will need a Steam update; it's ready for merge now.

PR #2046, opened 2021-12-31
Followup to custom menus

weluvgoatz:
This PR makes the color-customizable menus more efficient, allows for more colors to customize (which should fix a crash one user was experiencing but that I could not reproduce) and allows the customization of non-neutral-colored text.

### DIFF
--- a/src/editor/button_widget.cpp
+++ b/src/editor/button_widget.cpp
@@ -43,12 +43,10 @@ ButtonWidget::draw(DrawingContext& context)
   }
 
   if (m_grab) {
-    context.color().draw_filled_rect(m_rect, Color(g_config->editorcolor.red + 0.4f, g_config->editorcolor.green + 0.4f,
-      g_config->editorcolor.blue + 0.4f, g_config->editorcolor.alpha + 0.1f), 4.0f,
+    context.color().draw_filled_rect(m_rect, Color(g_config->editorgrabcolor), 4.0f,
                                      LAYER_GUI-5);
   } else if (m_hover) {
-    context.color().draw_filled_rect(m_rect, Color(g_config->editorcolor.red + 0.4f, g_config->editorcolor.green + 0.4f,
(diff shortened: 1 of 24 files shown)


src/supertux/gameconfig.cpp
```diff
@@ -126,36 +134,37 @@ Config::load()
...
-  config_mapping.get("menuroundness", menuroundness, 16.f);
+  // menu colors
+
+  std::vector<float> menubackcolor_, menufrontcolor_, menuhelpbackcolor_, menuhelpfrontcolor_,
+    labeltextcolor_, activetextcolor_, hlcolor_, editorcolor_, editorhovercolor_, editorgrabcolor_;
+
+  boost::optional<ReaderMapping> interface_colors_mapping;
+  if (config_mapping.get("interface_colors", interface_colors_mapping))
+  {
+    menubackcolor = Color(interface_colors_mapping->get("menubackcolor", menubackcolor_) ?
```
tobbi, line 145: Doesn't the `mapping->get` method have an overload that takes a default parameter?

tobbi: It did not have that before, but it does now. Please rebase!

src/editor/button_widget.cpp
```diff
@@ -43,12 +43,10 @@ ButtonWidget::draw(DrawingContext& context)
   }
 
   if (m_grab) {
-    context.color().draw_filled_rect(m_rect, Color(g_config->editorcolor.red + 0.4f, g_config->editorcolor.green + 0.4f,
-      g_config->editorcolor.blue + 0.4f, g_config->editorcolor.alpha + 0.1f), 4.0f,
+    context.color().draw_filled_rect(m_rect, Color(g_config->editorgrabcolor), 4.0f,
```
Semphriss, line 46: At a few places you cann the `Color(Color&)` copy constructor; is there a reason? Wouldn't just `g_config->editorgrabcolor` work?

PR #2063, opened 2022-01-03
Make stalactite's layer more consistent

### LINKED ISSUE #1991: Icicle goes behind switch
I find it weird that the falling icicle is on a layer behind the switch, so it appears to go between the switch and its backing.  Just a small bug but it is a weird looking graphical error.

Semphriss:
Prior to this PR, the stalactite would be rendered on `LAYER_OBJECTS` (layer 50) when inactive, and, when shaking or falling, on its internal `m_layer`, which defaults to `LAYER_TILES - 1` (layer -1) and cannot be edited from the editor.

This PR makes it more consistent by:
- Using `m_layer` everywhere.
- Setting `m_layer` to `LAYER_OBJECTS`, which makes the stalactite render and fall on the same layer it used to render to when standing still.

Fixes #1991

### DIFF
--- a/src/badguy/stalactite.cpp
+++ b/src/badguy/stalactite.cpp
@@ -29,7 +29,7 @@ static const float SHAKE_TIME = .8f;
 static const float SHAKE_RANGE_Y = 400;
 
 Stalactite::Stalactite(const ReaderMapping& mapping) :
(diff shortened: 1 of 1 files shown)


Alasdairbugs:
no, because this makes stalactites appear over the bottom of icy background tiles when they are set to 0, interactive, which is the case for like. all levels ever made.

PR #2065, opened 2022-01-03
Show confirmation message when deleting level or world

mrkubax10:


src/supertux/menu/editor_delete_levelset_menu.cpp
```diff
@@ -68,18 +69,21 @@ EditorDeleteLevelsetMenu::refresh()
...
-      m_editor_levelset_select_menu->reload_menu();
-      refresh();
+      Dialog::show_confirmation(_("Are you sure?"), [this]()
+      {
+        std::vector<std::string>& contrib_worlds = m_editor_levelset_select_menu->get_contrib_worlds();
+        physfsutil::remove_with_content(contrib_worlds[m_item_id]);
+        m_editor_levelset_select_menu->reload_menu();
+        refresh();
+      });
     }
```
Semphriss, line 87: You can pass `id` by capture in the lambda, removing the need for a member variable.

Semphriss (changes requested):

src/supertux/menu/editor_delete_levelset_menu.cpp
```diff
@@ -72,14 +72,17 @@ EditorDeleteLevelsetMenu::menu_action(MenuItem& item)
...
       Dialog::show_message(_("You cannot delete the world that you are editing"));
     else
     {
-      physfsutil::remove_with_content(contrib_worlds[id]);
-      m_editor_levelset_select_menu->reload_menu();
-      refresh();
+      Dialog::show_confirmation(_("Are you sure?"), [this, id]()
+      {
+        std::vector<std::string>& contrib_worlds = m_editor_levelset_select_menu->get_contrib_worlds();
+        physfsutil::remove_with_content(contrib_worlds[id]);
```
Semphriss, line 82: It's probably the most complex logic about a variable's lifetime I've done in a while, but:

`contrib_worlds` can be passed as capture as well, by reference, like `[this, id, &contrib_worlds]`". Although `contrib_worlds` is a local variable, and by the time it'll be executed the variable will have ended its life, it can be noted that this is a reference, and the list it refers to will still exist when the lambda will be executed.

Also, `contrib_worlds` can be declared as `const`, and you may use `auto` to avoid typing `std::vector<std::string>` at long.

Semphriss (approved):

src/supertux/menu/editor_delete_level_menu.cpp
```diff
@@ -53,12 +53,15 @@ EditorDeleteLevelMenu::menu_action(MenuItem& item)
       Dialog::show_message(_("You cannot delete level that you are editing!"));
     else
     {
-      PHYSFS_delete(m_level_full_paths[id].c_str());
-      delete_item(id + 2);
-      m_level_full_paths.erase(m_level_full_paths.begin() + id);
-      m_level_select_menu->reload_menu();
-      if (!Editor::current()->is_level_loaded())
-        m_levelset_select_menu->reload_menu();
+      Dialog::show_confirmation(_("Are you sure?"), [this, id]()
```
Vankata453, line 56: Would it be possible to show the name of the level or levelset, that's being deleted, in the message?

mrkubax10: Yes

Vankata453 (approved):

src/supertux/menu/editor_delete_level_menu.cpp
```diff
@@ -53,7 +54,8 @@ EditorDeleteLevelMenu::menu_action(MenuItem& item)
       Dialog::show_message(_("You cannot delete level that you are editing!"));
     else
     {
-      Dialog::show_confirmation(_("Are you sure?"), [this, id]()
+      const std::string& level_name = get_item_by_id(id).get_text();
+      Dialog::show_confirmation(str(boost::format(_("You are about to delete level \"%s\". Are you sure?")) % level_name), [this, id]()
```
Semphriss, line 58: Getting the button's label is poor practice; there is no guarantee that it won't change in the future. I'll try to find an alternative.

src/supertux/menu/editor_delete_levelset_menu.cpp
```diff
@@ -76,7 +76,8 @@ EditorDeleteLevelsetMenu::menu_action(MenuItem& item)
       Dialog::show_message(_("You cannot delete the world that you are editing"));
     else
     {
-      Dialog::show_confirmation(_("Are you sure?"), [this, id, &contrib_worlds]()
+      const std::string& world_name = get_item_by_id(id).get_text();
+      Dialog::show_confirmation(str(boost::format(_("You are about to delete world \"%s\". Are you sure?")) % world_name), [this, id, &contrib_worlds]()
```
Semphriss, line 80: Same here

Semphriss (changes requested):

Semphriss (approved):

PR #2066, opened 2022-01-03
Add Open Directory option to FileSystemMenu

mrkubax10:


src/gui/menu_filesystem.cpp
```diff
@@ -108,6 +108,8 @@ FileSystemMenu::refresh_items()
     item_id++;
   }
 
+  add_hl();
+  add_entry(-2, _("Open Directory"));
   add_hl();
```
Semphriss, line 113: Which item has the id -1?

mrkubax10: Back

Semphriss: I did some testing, and it seems like the back button does not trigger the `menu_action` function at all; however, looking at other files in the `src/supertux/menu` folder, it seems like the ID is overridden anyways, or that a non-standard back button is used.

I retract my comment, then.

Semphriss (approved):

Issue #505, opened 2016-07-23
Level editor documentation


kneekoo:
An editor guide would be great to let the users know what hey have to cover for a working level and then some important hints here and there.

**Example:**
1. The goal: lay down a path for Tux to walk on, set obstacles and foes, place rewards and the level exit;
2. For the path, use this and that from _there_ and _the other place_;
3. The foes are _there_, _this_ is mandatory and _this_ is optional;
4. The rewards must be placed _like this_, and mind the size of your Tux, to place the breaking blocks and triggers at this and that height;
5. The exit should be set as _this_ or _that_;
6. Tips to add secrets;
7. If you can't figure out what the buttons do, here are some useful details/features to look out for.

tobbi:
IMO this is an important UX step for 0.5.0.

maxteufel:
This should replace the C# editor guide and FAQ in the wiki, since that editor is going to be replaced by the in-game editor and Flexlay. Documentation for Flexlay will then be hosted in the Flexlay wiki.

Karkus476:
Assigning @Hume2 because he, well, created the editor as we see it in 0.5.0.

maxteufel:
@brmbrmcar You are working on this, aren't you?

brmbrmcar:
Yes, I will be.

Ordoviz:
I’ve made a guide: https://forum.freegamedev.net/viewtopic.php?f=69&t=8121
But I really would like to see some ingame help.

Rusty-Box:
A guide has been added to our wiki: https://github.com/SuperTux/supertux/wiki/Level-Editor
While an ingame version would be nice to have, this is the best we can offer for now,

Alasdairbugs:
closed as guide exists

PR #2068, opened 2022-01-04
Add "Editor Guides" button to level editor menu

Vankata453:
Adds a button that redirects the user to the community-made level editor [guides](https://github.com/SuperTux/supertux/wiki/Level-Editor), located in the wiki. The button is positioned on the bottom of the menu. The browser redirection code for the donation button was used.

Aims to provide a temporary solution for #505.


mrkubax10:
It seems like you are missing `#include <emscripten.h>` and `#include <esmcripten/html5.h>`

Semphriss:
How complete and ready are those guides? Is it a good idea to redirect to GitHub rather than the main site?

Also, the button doesn't look like it opens a browser window. Opening a browser window unexpectedly is poor UX; there should at least be a confirmation dialogue in the user would like to open an external link.

I recommend we discuss all this in discussions before continuing this PR.

Vankata453:
Thanks for the response!

Vankata453:
Will be closing this PR, because of two reasons:

1. As @Semphriss pointed out, those current guides may not really be complete and ready and most of them are made for an older SuperTux version.
2. As mentioned in discussion #2075, made for this PR, it's clear that a new editor guide is developed, which is going to be up-to-date with the newest SuperTux versions and is more likely to be regularly updated and built upon. When this new official level editor guide is released, the addition of an "Editor Guides" button may be rethought.

PR #2072, opened 2022-01-05
Use show_confirmation for level sharing in editor

Vankata453:
Adds the use of a `show_confirmation` dialog to confirm level sharing in editor, which was not used before, but rather different code, which achieves the same thing.

### DIFF
--- a/src/supertux/menu/editor_menu.cpp
+++ b/src/supertux/menu/editor_menu.cpp
@@ -159,13 +159,9 @@ EditorMenu::menu_action(MenuItem& item)
 
     case MNID_SHARE:
     {
-      auto dialog = std::make_unique<Dialog>();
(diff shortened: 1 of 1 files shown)


Vankata453:
Also added Emscripten support, so the URL can be opened from WASM, too.

Semphriss (approved):

tobbi:
Shouldn't the WASM support be added inside FileSystem::open_path? Check if file toopen is a URL and if it is, then use the WASM thing.

Semphriss:
It could be better to wrap that together since that's what the CPP function is for, but **please please please please do NOT put it in *the same JS function!*** The JS function isn't only used there, it's used also at other places, and changing it can have unexpected consequences. Instead, keep the check in the CPP function where it's meant to be, and then select what JS code you want to execute based on that.

PR #2073, opened 2022-01-05
Replaced greenish Will-O-Wisps with white version

Semphriss:
This allows for better usage with the color property.


tobbi:
Could we set the color property by default to something that resembles the previous version of the sprite?

Semphriss:
It already defaults to green, which has been the case since the color property has been introduced.

Semphriss:
Nevermind the second part; I just added a new sprite for the editor.

weluvgoatz (approved):

Rusty-Box (approved):

PR #2078, opened 2022-01-12
Android support and CI build file

pelya:
You can run the generated .apk file on your phone, but you need a keyboard to play.
I will submit my touchscreen fixes and other Android changes in a separate patch

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -1,0 +1,134 @@
+#  SuperTux
+#  Copyright (C) 2021-2021 Sergii Pylypenko <x.pelya.x@gmail.com>
+#
+#  This program is free software; you can redistribute it and/or
+#  modify it under the terms of the GNU General Public License
+#  as published by the Free Software Foundation; either version 3
+#  of the License, or (at your option) any later version.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License
+#  along with this program; if not, write to the Free Software
+#  Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
+
+name: android
+on:
+  push:
+    branches:
+      - master
+    tags:
+      - '*'
+  pull_request: {}
+
+jobs:
+  build-android:
+    env:
+      # Revision of https://github.com/pelya/commandergenius.git - SDL2 repo is a submodule
+      # We check out a specific stable revision, so the build won't break if I modify my SDL repo
+      SDL_ANDROID_REV: ad753cbdd
+      # It takes 35 minutes per one architecture to build SuperTux, of which 20 minutes are for building Boost and ICU libraries
+      # Enable additional architectures when we have these libraries inside the cache, or create a separate Android script for each architecture
+      # ARCH_LIST: armeabi-v7a arm64-v8a x86 x86_64
+      ARCH_LIST: arm64-v8a
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 1
+          submodules: true
+
+      - name: Clone SDL Android repo
+        run: |
+          git clone --depth=100 https://github.com/pelya/commandergenius.git build-android
+          git -C build-android checkout ${SDL_ANDROID_REV}
+
+      - name: Clone SDL Android repo submodules
+        # There are many more submodules in this repo, we don't need them all
+        working-directory: build-android
+        run: >
+          git submodule update --init --recursive --depth=1
+          project/jni/boost/src project/jni/iconv/src
+          project/jni/sdl2 project/jni/sdl2_image
+          project/jni/sdl2_mixer project/jni/sdl2_ttf
+
+      - name: Symlink the application dir
+        run: |
+          rm -rf build-android/project/jni/application/supertux/supertux
+          ln -s `pwd` build-android/project/jni/application/supertux/supertux
+          ln -s supertux build-android/project/jni/application/src
+
+      - name: Set architectures list
+        working-directory: build-android
+        run: |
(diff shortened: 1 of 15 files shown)


Semphriss:
Oh my, this is a surprise, but a welcome one! Many thanks!

I'll try to give it a look ASAP. Would it be possible (and if yes, would you be open) to plug the CI directly to the Google Play app? I've had a few plans to setup automatic publishing, but before I start working on them, I'd like to see what's feasible and reasonable to do.

pelya:
I've never set up an automated deployment on Play Store, but it's possible.

The CI script generates an .apk file, which can be side-loaded on any phone
with no trouble.

It also does not strip debug info from the binaries, it's a bug in Android
SDK that I somehow fixed on my machine, but I have no idea how. I hope it
will go away with the new Android SDK version.

You generate .aab file for Play Store, which is a different thing and
cannot be installed right away, and you cannot distribute the .apk file
through Google Play because it's bigger than 150 MB.

All this can be solved of course, but I'll need another New Year vacation
for that.

I can also transfer my app listing on Play Store, if you have a dedicated
Android dev to build/test/deploy it. The reason I've neglected the app for
so long is because I've switched from my own SDL 1.2 Android port to the
official SDL2 port, which was painful.

Semphriss:
No worries, I'll try to see if I can get something working well. I'll reply again if I can get something working.

Again, nice job and thanks for the contribution!

Semphriss:
Heya - I've made a developer account for the team!

How would you like to proceed?

pelya:
Open Google Payments and find the transaction ID for $25 you've paid to
register your dev account.

https://payments.google.com/payments/home

Send it to me to ***@***.***

Do not post it publicly or put it into Github issue! It's used for various
important stuff like deleting your dev account.

More info about transferring apps:

https://support.google.com/googleplay/android-developer/answer/6230247

Semphriss:
I'll do that! Just one detail; GitHub redacted your email address automatically. Is it the same email address you put in the copyright header in the code? (The one surrounded by x. and .x)

pelya:
Yes, email address in my git commits is the same.

pelya:
SuperTux Android app had been transferred to the dev account created by Semphriss.
You may wish to edit "What's New" text in the Play Console to remove my snarky comment about upstream devs.
I don't know how are you intending to fix slow rendering, rolling back to the GLES1 video backend is not exactly possible. Maybe remove displacement_texture from the shader on Android, or force 200% magnification so that the game will render less tiles.

Semphriss:
@pelya Have you tried running the SDL renderer? It reportedly runs much smoother in browsers with SDL, perhaps it's the same with Android?

(SuperTux works fine on my phone when I compile it manually)

pelya:
11. Setup the project properties. <sub>TODO: What do these do and what value can
they take?</sub>

```
# Assuming CWD = build.android/project
echo "sdk.dir=$ANDROID_SDK_ROOT" > local.properties
echo "proguard.config=proguard.cfg;proguard-local.cfg" >> local.properties
```

This step is needed to strip the .so files inside .apk file. Gradle will search for llvm-objcopy binary to strip .so files inside NDK 21.4 installation. This NDK version seems to be hardcoded in Gradle, because Android Studio would download it each time when importing the project, even when NDK 23.1 is present. Without NDK 21.4 and setting sdk.dir manually in the config file, application.so inside SuperTux.apk becomes 80 MB in size because of debug info.
Normally, every SDK component would be inside your SDK installation directory, and Gradle would set sdk.dir path by itself.
In our case, Gradle downloads missing SDK components locally into build.android and does not set sdk,dir, and that's also why you need to accept the licenses twice.

pelya:
8. Assemble a release build. <sub>TODO: Check for debug builds</sub>

```
# Assuming CWD = build.android
cd project
./gradlew assembleRelease
```

This step is needed only to let Gradle download missing Android SDK components, this build command will fail at this step.
Normally, build.sh will run ./gradlew assembleRelease only after building all .so files to assemble the .apk file out of them.

pelya:
14. To generate a .apk, run `./create-apk-with-data.sh` from the
`build.android/project/jni/application/supertux` folder. The file will be
located at `build.android/SuperTux-with-data.apk`.

Normally, ./gradlew assembleRelease or ./build.sh would build an .apk file. However we do a post-processing step and put all SuperTux data inside the .apk file.
It would probably be cleaner to create a file data.zip with all data files inside, and put it into 'assets' directory inside .apk, but then we need to mount a data.zip archive inside an apk zip archive using physfs in SuperTux code, so physfs performs double unzipping. We do this with .aab file anyway, because I did not find a method to put data files outside of the 'assets' directory inside the .aab, and physfs cannot mount a path like supertux.apk/assets, the .apk or .zip file must be the last element of the path, so every file path that SuperTux reads would start with 'assets/...'

pelya:
> @pelya Have you tried running the SDL renderer? It reportedly runs much smoother in browsers with SDL, perhaps it's the same with Android?

No, I did not try SDL renderer. I will try it.
 
> (SuperTux works fine on my phone when I compile it manually)

That's good to hear.

pelya:
SDL renderer works on Android, with some glitches.

1. Menu cursor draws with graphical glitches.
2. You must rotate your screen in the landscape mode before running the app, it seems to handle the screen dimensions change event poorly and it does not re-initialize the screen from the vertical orientation. I have added Android-specific video init code to gl_video_system.cpp, but I did not modify sdl_video_system.cpp. Just copying the code inside `#ifdef __ANDROID__` should do the trick.
3. There are no Tux reflections in the ice blocks.
4. FPS seems to be better on the worldmap than with the GL renderer.
5. You will need to revert #2088, because I only implemented it for GL renderer.

To enable SDL renderer, edit `AndroidAppSettings.cfg`:
- Add `--renderer sdl` to `AppCmdline`
- Remove `supertux/src/video/sdl` from `AppBuildExclude`

You also need to increase `AppVersionCode` and `AppVersionName` when publishing a new release.

Semphriss:
I'll request the team for reviews to get this merged, but I haven't forgotten the edits that need to be done. I'm keeping an eye here; I' only going to try and get as much stuff merged as possible.

pelya:
I have reviewed and I approve this merge request 👍 
You can fix other things later.

pelya:
Finally! By the way, is IRC alive? There was no one in the room.

mrkubax10:
Discord is more active, also IRC channel was moved to libera.chat

PR #2080, opened 2022-01-12
Use SDL_OpenURL() to open URL cross-platformingly

pelya:
Including Android support


Semphriss:
Can `SDL_OpenURL()` also open local files? (`open_path` is also used to show up the system file manager, or certain files with their default app)

pelya:
It should be able to open local files

https://wiki.libsdl.org/SDL_OpenURL

Semphriss:
Reading the documentation, `SDL_OpenURL()` could open files only if they are prefixed with `file:///`; I'll conduct tests and research to see what's best to handle this.

Also, I believe the team decided to split the code that opens files from the code that opens URLs, given problems with WASM; it might be a good opportunity to increase the quality of the code.

Is it fine if I push changes to your branch directly, or would you prefer that I create a different branch?

pelya:
It's fine, I only care that the opening URLs works on Android.

Zwatotem:
Looks okay

PR #2084, opened 2022-01-13
Improved touch controls

pelya:
Touch button images are half-transparent and smaller, but touch button area is bigger.
Touch buttons will not send mouse click events to the underlying widgets.
Does not depend on touch device ID 0, on my phone the touchscreen has device ID 4.


pelya:
Fixed compiler warnings.

pelya:
Tux will not drop an object when releasing Action button on touchscreen

Holding the Action button on a phone makes it impossible to press the Jump button
Also prevents Tux from blasting a stunned Mr. Ice Block with a fireball when trying to pick him up
Using keyboard or gamepad will behave the same way as before, releasing Action button will drop the object

Semphriss:
I can't get the nightlies to work on Google Play but I'll merge this now to avoid having to rebase it later, plus plenty of people seem to ask for exactly those changes.

PR #2085, opened 2022-01-13
Remove get_pixel API

pelya:
On my old Galaxy Note 3 the FPS dropped to one frame per two seconds the moment a magic block was on screen.
GL drivers on Android does not handle reading pixels from the screen at all.
The magic blocks should work more or less the same, I have tested the block activation distance on Tower of Ghosts level.
Magic blocks are now activated only by lanterns. If there are any community levels that used Live Fire guy to activate magic blocks, they are now broken.


tobbi:
Not only Livefire, but also spotlights can activate blocks and that uses the get_pixel API...that's the case in the one forest level that uses magic blocks.

pelya:
Tower of Ghosts uses only lanterns, but Ghouls Lair uses a rotating
spotlight in a secret area. You still can finish the level if you don't go
there.
I've added spotlight support to the magic block update code.
There are a bunch of levels with magic blocks in bonus packs, I've tried
Castle of Light and Darkness, and I could not finish it with the vanilla
SuperTux, it just looks broken near the end.

swagtoy:
@Vankata453

PR #2086, opened 2022-01-13
Tux will not drop an object when releasing Action button

pelya:
You need to press the button again to drop the object
Holding the Action button on a phone makes it impossible to press the Jump button
Also prevents Tux from blasting a stunned Mr. Ice Block with a fireball when trying to pick him up

### DIFF
--- a/src/object/player.cpp
+++ b/src/object/player.cpp
@@ -189,6 +189,7 @@ Player::Player(PlayerStatus& player_status, const std::string& name_) :
   m_visible(true),
   m_grabbed_object(nullptr),
   m_grabbed_object_remove_listener(new GrabListener(*this)),
+  m_released_object(false),
   // if/when we have complete penny gfx, we can
   // load those instead of Tux's sprite in the
   // constructor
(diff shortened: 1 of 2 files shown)


mrkubax10:
Maybe do that only on mobile platforms

pelya:
I think it's better like this on desktop platforms too, it's not fun to hold the key while trying to run and jump, and the change is intuitive enough. I can put it under #ifdef if you insist.

Semphriss:
It would be better to implement this in the mobile interface in some way, either by improving it (I admit I'm not a mobile player, so the interface is more-than-likely unintuitive) or by implementing the keep-pressing mechanic on buttons.

I would not be in favor of changing the behavior entirely for two reasons:
- The users are already used to the way it currently works, and requiring them to change their habits is not a good idea unless the improvement is significant.
- Games similar to SuperTux which implement carrying objects (I think notably about New Super Mario Bros Wii) use a hold-to-carry mechanic, similar to how it currently works on master; since people who discover SuperTux today are likely to have played such games, it would be better if they did not have to learn a new way of working.

Also, when using #ifdef's, please keep in mind the other mobile platforms (such as Ubuntu Touch, or possible iOS someday - the App Store forbids the GPL, but it shouldn't be a source of technical restrictions). Generally, it would be preferable to either avoid #ifdef's entirely, or if they are strictly required and relate to a game feature (in opposition to an OS feature), it would be preferable to use one variable in CMakeLists.txt and to enable/disable it by default depending on the target platform.

pelya:
Okay, #ifdef and a variable it is. I'll update the patch in a day or two.
I am against adding a third touch button to pick up objects, it will be not used until Forest world, and other solutions like sliding finger from button to button are not very intuitive.
Similarly, people on the Android side are already used to not holding the button, the app is on Play Store for several years.

pelya:
The improved patch is integrated into #2084

PR #2087, opened 2022-01-13
Improvements to profile management + unlimited profiles

### LINKED ISSUE #485: Allow more profiles to be added and their names to change
Currently, there are 5 profiles, which in many cases is more than enough, but sometimes it could be necessary to increase it. And I think the names of them, including names of the folders. This would be useful when there are many players or simply different types of play.

Vankata453:
Makes additions to the profile management system:

1. Allows to create an unlimited amount of profiles, all of which are name-able. Profiles are created as folders in the format `profile{id}`, as well as added to the `profiles` section in the config file, where their respective IDs and names are stored. Names are limited to a maximum of 20 characters. The abilities to add and rename these profiles, as well as do so with a default name, were also added.
2. Aside from the "Reset" and "Reset All" options, "Delete" and "Delete All" were added, which allow the user to delete the currently selected profile, or all profiles. If all profiles are deleted, the next one to be created will be 1.

<details>
<summary>Screenshots</summary>

Main profiles menu
Adding a new profile
Renaming an existing profile
</details>

Closes #485.

### DIFF
--- a/src/supertux/menu/profile_menu.cpp
+++ b/src/supertux/menu/profile_menu.cpp
@@ -224,7 +224,7 @@ namespace savegames_util {
   std::vector<int> get_savegames()
   {
     std::vector<int> savegames;
-    char **rc = PHYSFS_enumerateFiles("profiles");
+    char **rc = PHYSFS_enumerateFiles("/");
     char **i;
     for (i = rc; *i != nullptr; i++)
     if (std::string(*i).substr(0, 7) == "profile") savegames.push_back(std::stoi(std::string(*i).substr(7)));
--- a/src/supertux/menu/profile_menu.cpp
+++ b/src/supertux/menu/profile_menu.cpp
@@ -170,7 +170,7 @@ ProfileMenu::menu_action(MenuItem& item)
       fmt::runtime(_("This will reset all game progress on the profile \"{}\".\nAre you sure?")),
       m_profile_names[g_config->profile - 1]);
 
-    Dialog::show_confirmation(message, [this]() {
+    Dialog::show_confirmation(message, []() {
       savegames_util::delete_savegames(g_config->profile, true);
     });
   }
--- a/src/supertux/menu/profile_menu.cpp
+++ b/src/supertux/menu/profile_menu.cpp
@@ -1,5 +1,6 @@
 //  SuperTux
 //  Copyright (C) 2008 Ingo Ruhnke <grumbel@gmail.com>
+//                2022 Vankata453
 //
 //  This program is free software: you can redistribute it and/or modify
 //  it under the terms of the GNU General Public License as published by
@@ -17,6 +18,7 @@
 #include "supertux/menu/profile_menu.hpp"
 
 #include <fmt/format.h>
+#include <physfs.h>
 #include <sstream>
 
 #include "gui/dialog.hpp"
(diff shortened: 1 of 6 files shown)


src/supertux/menu/profile_menu.cpp
```diff
@@ -91,50 +85,28 @@ ProfileMenu::menu_action(MenuItem& item)
...
-    else
-    {
-      conf_msg = "This will delete your profile, wiping all game progress on it and removing it from the list.\nAre you sure?";
-    }
+    std::string conf_msg = is_reset ? "Resetting your profile will delete all of your game progress on it.\nAre you sure?" :
+      "This will delete your profile, wiping all game progress on it and removing it from the list.\nAre you sure?";
     Dialog::show_confirmation(_(conf_msg), [this, is_reset]() {
       delete_savegames(g_config->profile, is_reset);
       if (!is_reset) g_config->profile = "default";
     });
```
mrkubax10, line 101: I will try to review to not overwork Semphris. Problem with this code is that string extraction script for po files will not catch this. You can try this instead:
```
std::string conf_msg = is_reset?_("Resetting your profile will delete all of your game progress on it.\nAre you sure?") :
      _("This will delete your profile, wiping all game progress on it and removing it from the list.\nAre you sure?");
Dialog::show_confirmation(conf_msg, [this, is_reset]() {
...
```

src/supertux/menu/profile_menu.cpp
```diff
@@ -91,50 +85,28 @@ ProfileMenu::menu_action(MenuItem& item)
...
-    {
-      conf_msg = "This will reset all of your profiles and game progress on them.\nAre you sure?";
-    }
-    else
-    {
-      conf_msg = "This will delete all of your profiles, wiping all game progress on them and removing them from the list.\nAre you sure?";
-    }
+    std::string conf_msg = is_reset ? "This will reset all of your profiles and game progress on them.\nAre you sure?" :
+      "This will delete all of your profiles, wiping all game progress on them and removing them from the list.\nAre you sure?";
     Dialog::show_confirmation(_(conf_msg), [this, is_reset]() {
```
mrkubax10, line 108: Same here

Semphriss:
As usual, I'm pointing out only negative points here, but it's because only those require intervention; what works well need not to be changed.

------------------------

#### Design issues:
- This is not retrocompatible with the old profile system, and offers no smooth upgrade path. Progress will be lost for all users. This is a particularly problematic issue given the Steam release, which shares save data across computers.
- The hardcoded charset restriction is inconvenient and penalises languages that use other characters.
- In addition to the above, the multiple restrictions on the name (maximum size, etc.) are not obvious until the corresponding error message pops up. This increases the complexity of a task that should be simple; it goes from a single click to a click + inputting a string + receiving an error message that requires manual intervention from the user for at most as many times are there are restrictions.

#### Usability & Appearance issues:
- There are too many category separators for user comfort; wouldn't it be better to have all action buttons in the same category?
- The error message for the restricted charset says only lowercase letters are accepted, but it accepts uppercase letters; also, uppercase letters are converted to lowercase without warning.

There are also code-related issues (handling FS errors, formatting, optimisations, etc.), but the issues listed above should be taken care of first, as they will probably require big changes in the code.

#### Proposed alternatives (for some on the issues only)
- Instead of using a name-based profile system, use a number-based system, like multiplayer does.
- Instead of using a name-based profile system, use a sprite-based system, similar to SMG2.
- If a name-based system is strictly required, use a system similar to world names: generate a folder name by parsing characters, and inside put an `info` file containing the real name of the profile (or put it in an already existing file). This, however, has the downside of not having a clear and recognizable filename for certain languages.

#### Additional note:
What is the use case which requires more than 5 profiles? I read the comments on the linked issue, but it fails to mention that use case. Although I am not strictly opposed to removing the hardcoded restriction of the amount of profiles, we should make sure the increase in complexity is not only optimal, but also better than the problem arising from the use case.

If the reason why having an unrestricted set of profiles is merely because, let's say, speedrunners don't want to reset all their profiles each 5 runs, then increasing the number of profiles to 10 would most likely be enough. Now, surely there are other use cases, but those use cases should be mentioned so that we can find the optimal solution. If we do not know what the use cases are, we cannot find an optimal solution.

Semphriss:
Has anything been decided about this? If my questions aren't considered relevant, it's all good to merge this PR as-is.

Vankata453:
This PR is still a WIP. When I'm ready for a review, I'll be sure to mark it.

src/supertux/menu/profile_menu.cpp
```diff
@@ -58,40 +145,107 @@ void
...
 }
 
+namespace savegames_util {
+  std::vector<int> get_savegames()
+  {
+    std::vector<int> savegames;
+    char **rc = PHYSFS_enumerateFiles("/");
+    char **i;
+    for (i = rc; *i != nullptr; i++)
+    if (std::string(*i).substr(0, 7) == "profile") savegames.push_back(std::stoi(std::string(*i).substr(7)));
```
mrkubax10, line 230: Please use brackets or at least indentation here.

mrkubax10 (approved):
Works as intended

weluvgoatz (approved):

PR #2088, opened 2022-01-13
Group repeating tiles to render long strips of tiles as a single GL rectangle

pelya:
You may have noticed that worldmap on Android renders terribly slowly.
That's because Android GL hardware cannot handle the amount of triangles that SuperTux pushes each frame.
This optimization will calculate the adjacent tiles into a bigger rectangle, and render them together as two GL triangles, with the repeating texture.
There is also a script update-tilecache.sh which will launch SuperTux with each level, which will calculate the big rectangles and save them into .local/share/supertux2/tilecache so the levels will load faster.
On the worldmap on "The Journey Begins" level this optimization reduces the amount of GL triangles from 8500 to 2860, and increases FPS from 6-7 to 43-48, when measured on my Poco X3 Pro, which is actually a gaming phone, but nowhere near a gaming laptop in terms of GL triangles output.
On the Forest worldmap on the level "Tower of Ghosts" FPS is 9-11, and 7-8 inside the level. With the optimization, FPS is 46-52 on the worldmap, and 39-42 inside the level.
Android app is compiled with LTO enabled. Without LTO, FPS drops miserably even with the GL optimization.
This is the last remaining Android patch, I've saved the biggest for last.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,6 +46,7 @@ supertux.*
 
 # Useful Customization Files
 data/levels/more/
+data/tilecache
 *.patch
 *.zip
 *.diff
--- a/data/shader/shader100.frag
+++ b/data/shader/shader100.frag
@@ -12,18 +12,19 @@ uniform vec2 animate;
 uniform vec2 displacement_animate;
 
 varying vec2 texcoord_var;
+varying vec4 texcoord_repeat_var;
 varying vec4 diffuse_var;
 
 void main(void)
 {
   if (backbuffer == 0.0)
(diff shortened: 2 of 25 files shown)


tobbi:
@pelya Could you rebase this, please? Also, is this ready for review?

pelya:
I'm in the army right now, I cannot provide any support.

Rusty-Box:
> I'm in the army right now, I cannot provide any support.

That's rough. Stay safe!

PR #2098, opened 2022-01-15
Detect touchscreen when enabling mobile controller

mrkubax10:
Fixes issue when mobile controller is enabled by default on devices without touchscreen

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -152,12 +152,9 @@ endif(WIN32)
 option(UBUNTU_TOUCH "Compile the project for an Ubuntu Touch target" OFF)
 
 # Mobile builds
-option(ENABLE_TOUCHSCREEN_SUPPORT "Enable on-screen controls and event detection for touchscreen devices" ON)
-if (UBUNTU_TOUCH OR ANDROID)
-  option(SHOW_TOUCHSCREEN_CONTROLS "Show mobile controls by default" ON)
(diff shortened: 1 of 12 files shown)


Semphriss (commented):
CMakeLists.txt defiles a `SHOW_MOBILE_CONTROLS` (or similar) which I somehow forgot to use here.

Semphriss:
On second though (or rather, 10th though after those two weeks), I believe it might indeed be wiser to consider if SDL detects any touch device rather than using a macro.

mrkubax10:
I will do that then

CMakeLists.txt
```diff
@@ -152,12 +152,9 @@ endif(WIN32)
 option(UBUNTU_TOUCH "Compile the project for an Ubuntu Touch target" OFF)
 
 # Mobile builds
-option(ENABLE_TOUCHSCREEN_SUPPORT "Enable on-screen controls and event detection for touchscreen devices" ON)
-if (UBUNTU_TOUCH OR ANDROID)
-  option(SHOW_TOUCHSCREEN_CONTROLS "Show mobile controls by default" ON)
+if (UBUNTU_TOUCH)
```
Semphriss, line 155: `UBUNTU_TOUCH OR ANDROID` here would be preferable

Semphriss (changes requested):

Semphriss (approved):

PR #2101, opened 2022-01-16
Fix ItemStringSelect::get_width() changing unexpectedly

Grumbel:


src/gui/item_stringselect.cpp
```diff
@@ -58,8 +58,14 @@ ItemStringSelect::draw(DrawingContext& context, const Vector& pos, int menu_widt
...
-ItemStringSelect::get_width() const {
-  return static_cast<int>(Resources::normal_font->get_text_width(get_text()) + Resources::normal_font->get_text_width(list[*selected])) + 64;
+ItemStringSelect::get_width() const
+{
+  float max_width = 0;
+  for (auto const& item : list) {
+    float const width = Resources::normal_font->get_text_width(item) + 64;
+    max_width = std::max(width, max_width);
+  }
+  return Resources::normal_font->get_text_width(get_text()) + max_width;
```
Semphriss, line 68: I'm unsure if I should recommend using `std::max_element` here - on one hand it gives cleaner code, on the other hand it results in calling `get_text_width()` multiple times for each item. For the latter, I'm unsure how performance could be affected; I have no idea how frequently `get_width()` can be called each frame.

One possibility would be to fill an array with the text widths and use that pre-computed array to perform the calculations, like this:
```c++
int
ItemStringSelect::get_width() const
{
  std::vector<float> width(list.size());

  std::transform(list.begin(), list.end(), width.begin(), [](const auto& s) {
    return Resources::normal_font->get_text_width(s);
  });

  float max_item_width = *std::max_element(width.begin(), width.end());
  float text_width = Resources::normal_font->get_text_width(get_text());

  return static_cast<int>(text_width + max_item_width + 64.f);
}
```

I'm not sure which method could be considered best in this case, but I leave the suggestion here anyways, in case it might be interesting and/or someone can optimise it further.

tobbi:
Please fix the conversion warnings.

Zwatotem:
Looks good to me

PR #2106, opened 2022-01-16
Add commandline option --test

Zwatotem:
This pull request will have to be merged by the end of this week, so I'll see how much more I can fit in here by then. Planned are:
- Get rid of audio during these tests
- Warp playthrough mode (disable delays between game logic steps)
- Support for single file testing (Either level and demo in the same file or pointing to these files)
- Hide this stuff behind a compiler switch, so it's not exposed in public builds.
- Scripting functions that work like test steps (e.g. all must fire, or any must fire, or neither etc.)
- Test manager executable, that can fire several of these tests in parallel.

The tests themselves are probably going to be introduced in a separate PR. Let me know ahead of time, where should they be and how to package them, so they don't end up like secret swimming test that's shipped with every copy of SuperTux right now.


tobbi:
Can you please fix the build errors?

Zwatotem:
@tobbi I'm trying to, but I have no idea, how to revert a submodule change on Github.

Semphriss:
Is it possible to add this to the test suite instead?

PR #2120, opened 2022-01-25
Add force-content option to bonus blocks

mrkubax10:


Semphriss:
There has already been discussions about this feature in #1298. This breaks consistency and causes poor user experience.

PR #2121, opened 2022-01-25
Use higher quality Retro-Brick

serano01:
Made pieces for it, too.


serano01:
Based upon RustyBox' work, but since he also made other new blocks, I thought this change would be appropriate, as well.

Rusty-Box:
I don't know about this. As these are meant for providing a more retro look to the RiR add-on an updated sprite should reflect more closely the old brick rather than the new brick design.

serano01:
Yeah, but the whole retro-tileset is brand new, so the old 0.0.4-graphics look odd. And for me, the old brick was, apart from the lower quality, just greener.

Rusty-Box:
I would to play around with it myself to better explain it. It is much brighter and and less detail which I think makes it suit the retro aspect much more

serano01:
retro_brick

Maybe more along the lines of this?

Rusty-Box:
Kinda, but more shiny as the orginal.

serano01:
I'm closing this, since RustyBox has an other solution in mind.

PR #2122, opened 2022-01-25
Delete unused Files

serano01:
I want to get rid of some of the old graphics, since they are no longer fitting to the newer tileset RustyBox made.


Semphriss:
I didn't even notice those were removed from Revenge in Redmond :smile:

Semphriss (approved):

serano01:
Thanks for checking.

Issue #2123, opened 2022-01-26
The RTL text is reversed


omeritzics:
**SuperTux version:** 0.6.3
**System information:** N/A

##### Expected behavior
The RTL text should not reverse.

##### Actual behavior
The RTL text has reversed. Additionally, the Arabic letters are not connected.

##### Steps to reproduce actual behavior
Use SuperTux with an RTL localization (Hebrew, Arabic, ...).

##### Screenshots
Arabic:
![supertux-arabic](https://user-images.githubusercontent.com/66558205/151213596-e997bf0d-7b83-4e39-a763-f5efce30c7a9.png)

Hebrew:
![supertux-hebrew](https://user-images.githubusercontent.com/66558205/151213686-15a88133-b844-4f3a-b063-69ba38c30d6b.png)

How the Hebrew version should be like:
![109401513-e5d52980-7957-11eb-8e54-11cae3f096b2](https://user-images.githubusercontent.com/66558205/151213753-2c7181f1-23fb-45b0-b687-ee109fa733e0.png)

##### Additional information
I'm sorry about my past 'contributions' on Github - which were more spamming than helping - from when I just started using Github; I'm also sorry about my English, which was not clear enough and made misunderstandings sometimes. Hope to make real contributions in the future.

Semphriss:
I've checked the CI builds and it seems Windows builds don't pickup libraqm; I believe this might be a Windows-only problem, although not OS-related.

Do you have a macOS/Linux device to verify if this issue also happens on other OSes?

tobbi:
I believe I've already fixed this, but after 0.6.3.

Please check the nightlies from http://download.supertux.org

omeritzics:
Hi, I'm sorry for the late response.
Unfortunately, I cannot verify that this bug has happened on other OSes, but this bug has been fixed, as Tobby wrote.

Thank you!!

PR #2127, opened 2022-01-28
CI cleanup

Semphriss:
Refactored the workflow files to separate each OS. Also fixes some FIXME's, such as adding GLBinding builds.


tobbi:
We're not installing libraqm in all the builds, it seems.

#2123

Semphriss:
The CIs that are missing libraqm don't provide libraqm: for Linux 32-bit, `apt` only provides libraqm for amd64; for WASM, vcpkg fails to build one of its components.

.github/workflows/gnulinux.yml
```diff
@@ -0,0 +1,280 @@
...
+            libfreetype6-dev:i386 \
+            libcurl4-openssl-dev:i386 \
+            libharfbuzz-dev:i386 \
+            libfribidi-dev:i386 \
+            libglm-dev \
+            zlib1g-dev
+
+          # Nethier GLEW nor glbinding exist in 32-bit for Ubuntu 20.04, so snatch the debs from 16.04 instead
+          wget archive.ubuntu.com/ubuntu/pool/main/g/glew/libglew1.13_1.13.0-2_i386.deb && sudo dpkg -i libglew1.13_1.13.0-2_i386.deb
+          wget archive.ubuntu.com/ubuntu/pool/main/g/glew/libglew-dev_1.13.0-2_i386.deb && sudo dpkg -i libglew-dev_1.13.0-2_i386.deb
```
Zwatotem, line 152: lol

src/video/gl/gl_texture.cpp
```diff
@@ -125,7 +125,8 @@ GLTexture::GLTexture(const SDL_Surface& image, const Sampler& sampler) :
       const int bpp = convert->format->BytesPerPixel;
       const int x = m_image_width - 1;
       const int y = m_image_height - 1;
-      Uint32 color = *reinterpret_cast<Uint32*>(static_cast<uint8_t*>(convert->pixels) + y * convert->pitch + x * bpp);
+      Uint32 color = 0;
+      memcpy(&color, static_cast<uint8_t*>(convert->pixels) + y * convert->pitch + x * bpp, bpp);
```
Zwatotem, line 129: Is this related? If so, approved.

.github/workflows/gnulinux.yml
Semphriss: It has been the case since 32-bit Linux builds existed :smile:

src/video/gl/gl_texture.cpp
Semphriss: This is to fix a warning with the OpenGL renderer on Ubuntu Touch (and possibly other older OSes as well).

PR #2128, opened 2022-01-28
Opening language packs from options only focuses on language packs

### LINKED ISSUE #2124: "Language packs" menu flatly opens the Add-Ons menu
**SuperTux version:** Latest @ master

From the main menu, Options -> Language packs will open the add-ons menu, including all packages not related to language, and will not even put focus on the language pack(s). This is a friction for the user as it causes confusion. The menu should be more intuitive.

Some suggestions:
- Add an intermediate dialog to tell the user they'll get redirected to the add-ons menu
- Automatically find and offer to download the language pack, without passing through the add-ons menu

If people would like to discuss more suggestions, this issue can be converted into a discussion.

Vankata453:
Closes #2124.

Makes "Language packs" from options only focus on language packs from the add-ons menu, when opened. Also makes some strings change, depending on where the menu has been opened from ("Add-ons" or "Language packs").

### DIFF
--- a/src/supertux/menu/addon_menu.cpp
+++ b/src/supertux/menu/addon_menu.cpp
@@ -83,12 +83,13 @@ std::string generate_menu_item_text(const Addon& addon)
 
 } // namespace
 
-AddonMenu::AddonMenu(bool auto_install_langpack) :
+AddonMenu::AddonMenu(bool auto_install_langpack, bool language_packs_only) :
   m_addon_manager(*AddonManager::current()),
   m_installed_addons(),
   m_repository_addons(),
   m_addons_enabled(),
-  m_auto_install_langpack(auto_install_langpack)
+  m_auto_install_langpack(auto_install_langpack),
+  m_langpacks_only(language_packs_only)
 {
   refresh();
   if (auto_install_langpack)
@@ -119,18 +120,18 @@ void
 AddonMenu::rebuild_menu()
 {
   clear();
-  add_label(_("Add-ons"));
+  add_label(m_langpacks_only?_("Language Packs") :_("Add-ons"));
   add_hl();
 
   if (m_installed_addons.empty())
   {
     if (!m_repository_addons.empty())
     {
-      add_inactive(_("No Add-ons installed"));
+      add_inactive(m_langpacks_only?_("No language packs installed") :_("No Add-ons installed"));
     }
     else
     {
-      add_inactive(_("No Add-ons found"));
+      add_inactive(m_langpacks_only?_("No language packs found") :_("No Add-ons found"));
     }
   }
   else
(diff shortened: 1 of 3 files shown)


mrkubax10:
Looks ok

src/supertux/menu/addon_menu.cpp
```diff
@@ -119,18 +120,18 @@ void
...
-  add_label(_("Add-ons"));
+  add_label(m_langpacks_only?_("Language Packs") :_("Add-ons"));
   add_hl();
 
   if (m_installed_addons.empty())
   {
     if (!m_repository_addons.empty())
     {
-      add_inactive(_("No Add-ons installed"));
+      add_inactive(m_langpacks_only?_("No language packs installed") :_("No Add-ons installed"));
```
tobbi, line 130: Code style: Please ad spaces around `?`

src/supertux/menu/addon_menu.cpp
```diff
@@ -119,18 +120,18 @@ void
...
     }
     else
     {
-      add_inactive(_("No Add-ons found"));
+      add_inactive(m_langpacks_only?_("No language packs found") :_("No Add-ons found"));
```
tobbi, line 134: Same here.

src/supertux/menu/addon_menu.cpp
```diff
@@ -186,17 +193,20 @@ AddonMenu::rebuild_menu()
...
+          }
         }
       }
       idx += 1;
     }
 
     if (!have_new_stuff && m_addon_manager.has_been_updated())
     {
-      add_inactive(_("No new Add-ons found"));
+      add_inactive(m_langpacks_only?_("No new language packs found") :_("No new Add-ons found"));
```
tobbi, line 209: Same here as well.

src/supertux/menu/addon_menu.cpp
```diff
@@ -119,18 +120,18 @@ void
 AddonMenu::rebuild_menu()
 {
   clear();
-  add_label(_("Add-ons"));
+  add_label(m_langpacks_only?_("Language Packs") :_("Add-ons"));
```
tobbi, line 123: Same here, spaces around `?`

Semphriss:
Wouldn't it be better to just re-use the ready-made `auto_install_langpack` instead?

(Late to the party, sorry, I should have reviewed this PR earlier)

Vankata453:
I wasn't sure if this boolean doesn't represent something different, like autoinstalling an addon or an action of that sort. Do you think it's good to be used for this current purpose as well?

Semphriss:
Reviving: does anybody have any input about this? I believe auto-installing langpacks directly from the option sounds acceptable, but if there is a preference to make it work this way specifically, I don't object.

Vankata453:
This way should also allow for installing other language packs, if available in the future.

Zwatotem (approved):
Good 🚀

tobbi:
Approval from me as well 🚨

PR #2129, opened 2022-01-29
Keep InfoBlock message inside screen

### LINKED ISSUE #1229: Help block text might get cut off

##### Expected behavior
All help block texts are fully visible on the screen.

##### Actual behavior
In the first Forest level, the help block text is not completely visible on my screen (1920×1080 fullscreen).

##### Workaround
When I press "scroll up", I can see the full text. But still, that's confusing!

1. Go to the first help block in the first Forest level
2. Activate it
3. Look at the text. Is it fully visible or cut off by the screen?

mrkubax10:
Fixes #1229

### DIFF
--- a/src/object/infoblock.cpp
+++ b/src/object/infoblock.cpp
@@ -20,6 +20,7 @@
 
 #include "editor/editor.hpp"
 #include "object/player.hpp"
+#include "object/camera.hpp"
 #include "supertux/info_box_line.hpp"
 #include "supertux/sector.hpp"
 #include "util/reader_mapping.hpp"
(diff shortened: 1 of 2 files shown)


Semphriss:
Would it be feasible (and if so, a positive addition) to preemptively also constrain the left, right and bottom borders?

mrkubax10:
Yes

mrkubax10:
After some research: it is impossible for message to show under screen, also constrain at edges of sectors is already in game

Semphriss (approved):
Looks good then

Zwatotem:
> After some research: it is impossible for message to show under screen, also constraining at edges of sectors is already in game

If this holds true than looks good to me as well.

PR #2043, opened 2021-12-29
Multiplayer

Semphriss:
'Nuff said.

Fixes #224
Fixes #852
Closes #1956

Additionally:
Fixes #2057
Fixes #1537 

Supports:
- Any number of player
- Control both joysticks/controllers and keyboard for all players
- Hot-plugging of controllers automatically spawn/de-spawn players if there is an active sector that isn't the title screen
- Manual configuration through the options menu
- Stats saved for all players, persistent even if a certain player doesn't play for a long time
- Adaptive camera to track all players on screen

https://user-images.githubusercontent.com/66701383/147707597-044a5718-3bab-4083-82f9-011c35f0cd58.mp4

(Sorry for the non-simultaneous movements, it's hard to play two players with two hands)

### Split screen?

Split screen has been considered and will likely be added in the future, but it requires some improvements to the rendering engine first so it supports cropping properly. I've already done that for another unreleased PR, so I know it's possible and I know how to do it, but I need to actually take the time to do it, so it won't be immediate.

PR #2137, opened 2022-02-05
Add controller rumble support

Epicalert:
When using a controller, rumble events will be sent with
`SDL_GameControllerRumble()`.

Rumble events are sent in the following situations:
- player death
- player hurt
- bad guy squished
- bad guy killed
- block hit
- buttjump hit

Tested on Arch Linux, Linux 5.16.3, Xbox One Controller (version 1)

### DIFF
--- a/src/badguy/badguy.cpp
+++ b/src/badguy/badguy.cpp
@@ -363,6 +363,7 @@ BadGuy::collision(GameObject& other, const CollisionHit& hit)
 
     // hit from above?
     if (player->get_bbox().get_bottom() < (m_col.m_bbox.get_top() + 16)) {
+      InputManager::current()->rumble_effect(RUMBLE_SQUISH);
       if (player->is_stone()) {
         kill_fall();
         return FORCE_MOVE;
@@ -408,6 +409,7 @@ HitResponse
 BadGuy::collision_player(Player& player, const CollisionHit& )
 {
   if (player.is_invincible()) {
+    InputManager::current()->rumble_effect(RUMBLE_SQUISH_INVINCIBLE);
     kill_fall();
     return ABORT_MOVE;
   }
--- a/src/control/game_controller_manager.cpp
+++ b/src/control/game_controller_manager.cpp
@@ -176,6 +176,42 @@ GameControllerManager::process_axis_event(const SDL_ControllerAxisEvent& ev)
 }
 
 void
+GameControllerManager::rumble_effect(int type)
+{
+  for (const auto& con : m_game_controllers)
+  {
+    switch (type) {
+      case RUMBLE_HURT:
+        SDL_GameControllerRumble(con, 0x9FFF, 0xBFFF, 400);
+        break;
+
+      case RUMBLE_KILL:
+        SDL_GameControllerRumble(con, 0x9FFF, 0xBFFF, 750);
+	break;
+
+      case RUMBLE_SQUISH:
+        SDL_GameControllerRumble(con, 0x2FFF, 0xBFFF, 200);
+	break;
+
+      case RUMBLE_SQUISH_INVINCIBLE:
+        SDL_GameControllerRumble(con, 0x2FFF, 0xBFFF, 150);
+        break;
+
+      case RUMBLE_BLOCK_HIT:
+        SDL_GameControllerRumble(con, 0x4FFF, 0x9FFF, 200);
+	break;
+
+      case RUMBLE_BUTTJUMP:
+        SDL_GameControllerRumble(con, 0xDFFF, 0xBFFF, 300);
+	break;
+
+      default:
+        break;
+    }
+  }
+}
+
+void
 GameControllerManager::on_controller_added(int joystick_index)
 {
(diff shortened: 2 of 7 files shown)


src/badguy/badguy.cpp
```diff
@@ -442,6 +444,7 @@ BadGuy::collision_badguy(BadGuy& , const CollisionHit& )
 bool
 BadGuy::collision_squished(GameObject& object)
 {
+
```
tobbi, line 447: Did you forget the rumble_effect code here?

Epicalert: No, it seems I accidentally added a newline and missed it when reviewing the diff. I'll amend the commit.

Semphriss:
This is in duplicate of #2043, which also handles rumbling.

Also, the rumbling effect might not be desirable for every user. There is a setting in the aforementioned PR that allows users to adjust this.

Is it possible to redirect this PR to [`Semphriss:multiplayer`](https://github.com/Semphriss/supertux/compare/multiplayer...Semphriss:multiplayer)? It would be better to handle the feature there.

If not, we can also await merging of #2043 and rebase this PR on top of it, and proceed from there.

Semphriss:
After a brief review, I recommend that this PR undergoes some discussion regarding its design phase.
<sub>Usual disclaimer, especially for new contributors: the amount of questions might look scary, but it's not just to be a hindrance, it's so we can actually find the best solution as early as possible to avoid having to take care of deeply-nested bugs and poor code quality in the future.</sub>

Basics:
- Do we want rumbling at all? What are the upsides and downsides of each possibility?
- If we want rumbling, how should we use it? #2043 uses rumbling as a method of identifying controllers. It is also common practice in many games to use rumbling as part of the gameplay. But under which circumstances should it be used? In which cases? This PR offers a handful of ready-made cases; are all of these good? Should we have more use cases? For example, should it really rumble when (I'm picking one at random) a crate is destroyed? Should it not rumble when, for example, leaving the stone state for the stone powerup?
- Related to above: How does rumbling affect the gameplay? What effect does it create for the player? What desirable and undesirable effect does it have? What significance can it carry? Which ones _should_ it carry?
- What features regarding this do we have? SDL offers us varying intensities and delays; how should these map to each of the use cases? Should it rumble at different intensities (and delay, and other potential factors) for each case? Why or why not? If it should, what variations should each case have, and why or why not? What effect will that create for the player?
- What can we build around it? Do we have a setting to disable rumbling? Should we have a slider to adjust the intensity?

Code:
- Which class should hold the code that executes rumbling? Right now, it is the GameController class, which seems reasonable to me, but has any other class been considered? Which ones, and what were the upsides and downsides of each one?
- There are gamepads but also joysticks, handled by the `src/control/joystick*.*pp` classes; is there a reason they don't use rumbling as well? Was it a simple oversight, or is it unsopported?
- Right now, the events of rumbling are handled by plenty of classes throughout the code; should they really all be taking care of that themselves? For example, it is the BadGuy class that handles rumbling for the specific case of rumbling when the player gets hit; shouldn't that be in the Player class instead, under the kill() function? That is just one case;  how should it be for all other cases?
- This code gives a rumbling to *all* controllers whevener a request is sent; should it really be the case? I suppose few people who play in single player have all controllers on hand at the same time; should we add something to ensure only the controller the player is using gets a rumble? If so, how do we detect that?
- How will this play with multiplayer? Should we add an argument to rumble only one player? (This is one of the reasons why I recommended either getting this into the multiplayer PR or to wait until it is merged before reviewing this)

cooljeanius:
Another question I'd add is, what about controllers with variable rumble settings? One of the Nintendo Switch marketing points was "HD Rumble" that allows rumbles to have varying degrees of intensity; if I connect my Switch Pro controller to my computer to play SuperTux, how strong would the rumbles on it come through with this PR? Personally I think if supporting HD Rumble is possible, it would be better to have a stronger rumble for a player death, and weaker rumbles for the other things listed in the OP.

Semphriss:
@cooljeanius The 4th point in "Basic" specifies intensity; is it about that?

Controller support (rumbling being part of it) is handled by SDL, which offers a "low frequency" and a "high frequency" rumble intensity values, in addition to the delay. There are also some extra rumblers specific to some PS5 controllers if I understand the documentation properly, although that may not be relevant for us.

Semphriss:
These questions weren't meant to receive flat answers, they were meant to initiate in-depth discussion with the development team.

Inside PR comments, one can only simplify the answers to the general idea, which defeats the point of discussing, which is meant to have a clear idea of all the details. If you wish to discuss the feature, [GitHub Discussions](https://github.com/SuperTux/supertux/discussions) are usually much better suited for this.

Semphriss:
Reviving from inactivity: does any team member intend to review this PR design-wise? If it's just about adding rumbling at all, I can take it upon myself to implement it directly in the multiplayer PR and let future maintainers adjust settings to their liking.

Semphriss:
Closing as I will add this in #2043 for ease of management; help would be welcome.

PR #2138, opened 2022-02-05
More forgiving explosions, less bouncy bombs

weluvgoatz:
This PR makes throwing bombs to activate bricks/weak blocks more forgiving by giving the explosions a larger affect range and making bombs bounce less intensely when hitting solids.

### DIFF
--- a/src/object/brick.cpp
+++ b/src/object/brick.cpp
@@ -62,6 +62,17 @@ Brick::hit(Player& player)
   try_break(&player);
 }
 
+void
+Brick::update(float dt_sec)
+{
+  for (auto& explosion : Sector::get().get_objects_by_type<Explosion>())
+  {
+    if (get_bbox().grown(8).contains(explosion.get_bbox()))
+      try_break(nullptr);
+  }
+  Block::update(dt_sec);
+}
+
 HitResponse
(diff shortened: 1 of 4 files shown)


src/object/brick.cpp
```diff
@@ -62,6 +62,17 @@ Brick::hit(Player& player)
...
+Brick::update(float dt_sec)
+{
+  for (auto& explosion : Sector::get().get_objects_by_type<Explosion>())
+  {
+    if (get_bbox().grown(8).contains(explosion.get_bbox()))
+      try_break(nullptr);
+  }
+  Block::update(dt_sec);
+}
+
```
Semphriss, line 75: This code belongs in the explosion, this should not be handled by bricks. Possible solutions would be to a) retain the current explosion size but diminish its lifetime or b) give the hurting sequence a smaller radius than normal.

Zwatotem (changes requested):
IMO this is very hacky and non-extensible solution. The propper and futureproof solution would be to add different radiuses to the explosion object, that affect different objects (e.i. player, enemies, other objects) separately for physics and for damage.
I propose to close this and readdress this design need in a new PR.

PR #2139, opened 2022-02-06
new level stat icons

Frostwithasideofsalt:
i made new level stat icons (see changed files to see new ones)


Semphriss:
Has there been any discussion regarding adding those? If not, I recommend submitting these in [Discussion](https://github.com/SuperTux/supertux/discussions/new?category=ideas) instead.

Zwatotem:
Designs for first and third one are okay for me, but the second looks like a snowball enemy with _cool_ glasses on. Aside from that I'd say all icons need at least slight anti-aliasing and bottom right corner of each looks like it has a chopped off part.

PR #2140, opened 2022-02-06
Hidden levelsets made visible with Developer Mode

Vankata453:
Using Developer Mode will show levelsets, that have `hide-from-contribs` set to `true` and are not visible to the regular user. This applies to the "Contrib Levels" menus, as well as the level editor. Also adds `(contrib-type "official")` to some official levelsets, so they show up as "Official Contrib Levels".

### DIFF
--- a/data/levels/misc/info
+++ b/data/levels/misc/info
@@ -1,6 +1,7 @@
 (supertux-world
   (title (_ "Misc"))
   (description (_ "How did you get here?"))
+  (contrib-type "official")
   (levelset #t)
   (hide-from-contribs #t)
 )
--- a/data/levels/test/info
+++ b/data/levels/test/info
@@ -1,6 +1,7 @@
 (supertux-level-subset
   (title (_ "Test Levels"))
   (description (_ ""))
+  (contrib-type "official")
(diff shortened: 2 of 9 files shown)


Semphriss:
Levels with "Hide from contribs" are not meant to be contib levels at all and should not be available in the contrib levels tabs, even with developer mode activated. One can (and should) access, for instance, the main worldmap by opening the Story Mode normally.

Vankata453:
This prevents a user from being able to edit a level from Story Mode, without having to modify files, for example. I made it allowed for Developer Mode, because I thought it may be easier for more advanced users to access these levelsets in such cases, while still having them not show up without Developer Mode (for regular users).

Vankata453:
Discussion #2142 was made for further discussing this PR, converting to draft in the meantime.

Semphriss:
Reviving this PR so we decide something about it. I retire any personal interest of mine in the discussion.

Semphriss:
Closing per [discussion](https://github.com/SuperTux/supertux/discussions/2142) (despite not too much activity); feel free to re-open if further discussion overturns this decision.

PR #2141, opened 2022-02-06
Allow vertical peeking even if Tux isn't standing on the ground

Narre:
I've got no idea why this wasn't always the case, if there's a reason and this PR breaks something, please let me know. The inability to peek up/down mid-air always bothered me a bit, but now swimming is a thing, and the inability to use this feature in water just creates a bad player experience.

This PR removes the check that prevented vertical peeking unless Tux was standing on solid ground.

### DIFF
--- a/src/object/player.cpp
+++ b/src/object/player.cpp
@@ -1105,25 +1105,20 @@ Player::handle_input()
   }
 
   /* Peeking */
-  if ( m_controller->released( Control::PEEK_LEFT ) || m_controller->released( Control::PEEK_RIGHT ) ) {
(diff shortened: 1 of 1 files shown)


Semphriss:
That's odd - the code shows that the restriction was intended. Is it known why that was put in place originally?

Narre:
No idea. It wasn't there for horizontal peeking and it doesn't seem to make any sense for vertical peeking. I didn't find a possible reason for it, but there probably(?) was one. If it was obscured or not at some point, I'm not sure. That's why I wrote the "if it breaks something" in the PR description. I only tested it briefly though, so some amount of testing should be done before this gets merged, just in case.

Semphriss (approved):

Zwatotem (approved):

PR #2144, opened 2022-02-07
Level time improvements

### LINKED ISSUE #2143: Time limit resets, when respawning on a checkpoint
### Discussed in https://github.com/SuperTux/supertux/discussions/2060

<div type='discussions-op-text'>

<sup>Originally posted by **Vankata453** January  2, 2022</sup>
**SuperTux version:** 0.6.3

##### Expected behavior
Time limit does not reset, when respawning on a checkpoint.
##### Actual behavior
Time limit resets, when respawning on a checkpoint.
Enter a level with a time limit, activate a checkpoint and die. Notice that the time limit resets.</div>

The most upvoted proposed solution (with 4 upvotes) is:

<div type='discussions-op-text'>

Do not change the time left in anyway when respawning. Keep the timers running through deaths; retain the remaining time at the moment of death and restore that same time upon respawning at a checkpoint. The time in a timer never goes up while playing the level, even at deaths.</div>

Semphriss:
Fixes #2143 

As discussed in #2060, this implements the most popular opinion: Level Time is retained throughout deaths involving a respawn at a checkpoint. The time will keep running down and never go back up until checkpoints are cleared (either as a result of running out of time, or by resetting the level manually).

Note that this design involves code both in the LevelTime object and in the GameSession class; this is necessary because taking the checkpoint coins is handled directly in player.cpp, which happens at the moment of dying, not resetting. I would welcome a proposal that would help fix this, if any.

This also fixes a bug with the timer running despite all players being either dead, dying of winning.

### DIFF
--- a/src/object/level_time.cpp
+++ b/src/object/level_time.cpp
@@ -60,6 +60,13 @@ LevelTime::update(float dt_sec)
 {
   if (!running) return;
 
+  int players_alive = Sector::current() ? Sector::current()->get_object_count<Player>([](const Player& p) {
+    return !p.is_dead() && !p.is_dying() && !p.is_winning();
+  }) : 0;
+
+  if (!players_alive)
+    return;
+
   int prev_time = static_cast<int>(floorf(time_left*5));
   time_left -= dt_sec;
   if (time_left <= 0) {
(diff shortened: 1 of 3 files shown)


Zwatotem:
> this is necessary because taking the checkpoint coins is handled directly in player.cpp,

We should at some point (hopefully in the near future) decide on some new design directions with coins and their meaning in gameplay. This is complicated a lot of code, and the more we keep it, the more complicated the redesign will ultimately become.

PR #2146, opened 2022-02-07
Keep music across sectors, if the name and path of music file match

Vankata453:
When there are two sectors and both have the same music, respawning in the other and noticing the music restarts from the beginning may be a bit unsatisfactory. This PR fixes that issue, by checking if the path to the music file, the sector's `MusicObject` uses, is equal to the one of the other sector.

Also modifies levels "Tux The Builder" and "Lost Village" from `world2`, to make them support this.

(@Zwatotem has requested a review for this PR, mentioning them here.)

### DIFF
--- a/data/levels/world2/lost_village.stl
+++ b/data/levels/world2/lost_village.stl
@@ -197,7 +197,7 @@
       (y 488)
     )
     (music
-      (file "forest\\shallow-green.music")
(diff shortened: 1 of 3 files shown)


mrkubax10:
Fine for me

Semphriss (approved):

Semphriss:
Does this conflict with #2151?

PR #2151, opened 2022-02-16
Fix music not playing

mrkubax10:
Fixes issue when some music doesn't want to play


serano01:
I think the "tropical" music is affected as well. It isn't just as obvious since no one uses it.

mrkubax10:
Yeah, it is known. This pull requests fixes the issue for every music

Semphriss (approved):
I didn't check for possible quality optimisations, but the whole thing lookd good.

Semphriss:
Does this conflict with #2146?

Zwatotem:
> Does this conflict with #2146?

No, it shouldn't.

PR #2153, opened 2022-02-17
Fix FreeBSD builds

Semphriss:
Related to [this upstream issue](https://github.com/vmactions/freebsd-vm/issues/38).


tobbi:
Can I merge this one-line commit or would you rather have 10 more people looking at it?

PR #2159, opened 2022-03-03
Make changing userdir affect console output files

mrkubax10:


Semphriss:
This breaks the initialisation order; the redirection must be setup before anything else. I probably should even have put the two init lines after that one.

PR #2163, opened 2022-03-08
Refactor CMakeLists.txt, split into smaller pieces, accept more system libraries

Grumbel:
Change CMakeLists.txt to look for system libraries whenever possible for squirrel, SDL2_ttf, tinygettext, etc. and use them instead of external/. 

Some -DUSE_SYSTEM_... options are still missing.


tobbi:
Are the fmt commits supposed to be in this PR?

Grumbel:
The fmt is in there because this is based on #2004, it will go away once #2004 is merged. Should have market this as draft, it's not done yet.

Grumbel:
This ended up being a bit bigger than originally intended. The `CMakeListst.txt` is now split up and the pieces are stored under `mk/cmake/SuperTux/`. Each dependency got their own file `Provide{NAME}.cmake` file. Instead of loose variables, the libraries and include directories get bundled into a single INTERFACE library target for each dependency. `option(USE_SYSTEM_...)` is provided to look for libraries in the system, before falling back to the `external/` copy.

PR #2168, opened 2022-03-13
Remove redundant _SQ64 define

Grumbel:
This is already defined in the Squirrel itself:

* external/squirrel/CMakeLists.txt:49


Grumbel:
On second look, we don't actually make use of the CMake target that defines `_SQ64` when using `external/squirrel`, that's only used for system installs. However `squirrel.h` contains:

```
#if (defined(_WIN64) || defined(_LP64))
#ifndef _SQ64
#define _SQ64
#endif
#endif
```

Which should accomplish the same thing, unless we have platforms where LP64 doesn't hold:

> _LP64
>
> These macros are defined, with value 1, if (and only if) the compilation is for a target where long int and pointer both use 64-bits and int uses 32-bit.

PR #2172, opened 2022-03-16
Fixed panning of ambient sounds.

Pheazant:
Fix panning for the ambient sound level object.

Before it used to come from the left more, most likely from the origin point.

This patch makes position-based stereo not apply to ambient sounds.

The change is most noticeable (and tested) in the starting area of `world2/lost_village.stl`.

### DIFF
--- a/src/object/ambient_sound.cpp
+++ b/src/object/ambient_sound.cpp
@@ -149,6 +149,7 @@ AmbientSound::start_playing()
 
     sound_source->set_gain(0);
     sound_source->set_looping(true);
+    sound_source->set_relative(true);
     currentvolume=targetvolume=1e-20f;
     sound_source->play();
   } catch(std::exception& e) {


Semphriss:
Looking at OpenAL code, this seems to be intended behavior - I even had to fix it to make it work properly in #2126. Perhaps an option should be added to disable positioning?

Pheazant:
The sound source's position doesn't seem to be set anywhere, though. I'm not sure we should pan it as it stands now, unless we set a parameter (and a reasonable default) for the source point.

Using relative positioning is the way thunder is handled - can we say AmbientSound is supposed to create a similar effect?

Semphriss:
The sound positioning is automatically set and calculated by OpenAL; that's how bombs that explode to the left of the screen actually sounds like it comes from the left. Disabling it would defeat the entire point of the feature. I believe it is either the camera or the player which is used as source point.

Pheazant:
Just to confirm we are experiencing the same thing, and it is not an issue on my end only: could you enter `world2/lost_village.stl` and check the first basement area on the master branch?

The issue I am experiencing is that the sound is noticeably louder in the left speaker/headphone. This cannot be changed in the level editor. A similar issue exists in a level of Bonus Island 4 (Spooky Mansion, I believe).

The way AmbientSound is set up right now seems to set the default source position incorrectly - i.e., the `set_position` call (present in the mentioned bomb [here](https://github.com/SuperTux/supertux/blob/f494411463f7fd81248c92bc2ee633491d5db1bc/src/badguy/bomb.cpp#L34)) is missing. Instead of toggling relative positioning on, should we rather set the sound source in the center of the trigger area? I can amend the PR if so.

Pheazant:
Done.

Semphriss:
Thanks, looks good!

PR #2136, opened 2022-02-04
Complete overhaul of major tilesets + extras

Rusty-Box:
This PR provides heavy changes to all major level assets from both **Icy Island** and **Rooted Forest**, as well as re-introduce the old ghost forest theme as the new **Corrupted Forest**.

---

**Initial changes include the following:**
- Revamped forest tileset
- Revamped underground forest tileset
- New corrupted forest tileset
- New corrupted underground forest tileset
- New Ghost Tree roots tileset
- Updated trees for both normal and corrupted forest
- Updated background assets for both normal and corrupted forest
- Updated spikevine graphics
- Autotiling for all major forest tilesets
- New goal houses for both normal and corrupted forest
- Remove all the unneeded/outdated graphics

**With patch 1, the following was also added:**
- Updated snow surface tileset
- Improved ice cave tileset
- Theme specific block overlays
- Updated castle tilesets
- New Ruins tileset
- Wall torch variant for torch object
- Floor tile for goal post
- Normal and corrupted flowers for the forest theme
- New sign to indicate the requirement of bombs
- Removal of more old/unneeded graphics

**With patch 2, the following was also added:**

- Updated visuals for the Pogonia plant
- Improved pipe graphics
- Improved chain tiles
- Added outlines to the Halloween tileset
- Add spikes for the Ruin tileset
- Improved looks + new animation for unstable tiles
- Add missing outlines to Retro tilesets
- Minor changes to the pumpkin house
- Add autotiling for Halloween, Ice Chunk and Snow Fort tileset
- Fixed error hiding corrupted plant tile
- Removed several tiles from tilegroups which are going to be heavily changed or removed entirely soon

**With patch 3, the following was also added:**

- Update visuals for weakblocks (i.e. meltbox and strawbox)
- Improve and tweaked igloo decals
- Update poles
- New goalpost variant for secret level exits
- New flag design + animation
- Improve and new crystal background elements
- New bridge tilesets (frozen and un-frozen)
- Add missing retro snow tiles for autotiling (will follow soon)
- Add outlines to grave decals
- Move waterfall tiles over to 'images/tiles/water' + adjust colors
- Update retro brick and boxes to better suit art-style
- Update clouds
- Update rope, pushbuttons, snow bumps and forest walljump tiles - thanks to @weluvgoatz 
- Better magicblock visuals
- Add proper rublight graphics (crystal themed)
- Remove more outdated/un-wanted assets

**With patch 4, the following was also added:**

- Improved and expanded snow + ice cave tileset
[...]

PR #2175, opened 2022-03-17
Proposing an alternate goalpost for secret exits.

Pheazant:
The way secret exits are currently handled in SuperTux is not obvious. The current paths are straightforward enough, but whether an exit after a secret area is a big shortcut or an alternate path should be more apparent to the player.

This change proposes an alternate goalpost design with an inoffensive placeholder graphic. The new goalposts are added to `world1/icy_valley.stl` and `world2/crumbling_path.stl`
 
Similar ways of marking secret exits can be observed in the [New Super Mario Bros. series](https://www.mariowiki.com/Goal_Pole_(secret)).

<details>
<summary>Screenshots</summary>

image

image
</details>


Semphriss:
Marking ending posts for secret endings would be a good idea, and indeed it would help make the special endings feel special.

However, I don't feel this specific appearance does the job; the green of the post clashes with the color palette of its environment (green doesn't fit really well in either scene IMO). I like the concept though, perhaps we can get a team artist to make new posts for secret endings?

Is it all good if I bump this to a [discussion](https://github.com/SuperTux/supertux/discussions)?

Pheazant:
It looks like a question of implementation. I would propose continuing it in #2136 instead, to keep momentum. There is a redesign of goalposts and exit houses in progress already, perhaps @Rusty-Box would be interested in taking the mantle up.

Or we could find a different colour to tint it :)

Semphriss:
Since it won't be merged as-is, but it'll be integrated with #2136, I'll close this, but I'll leave a mention on that PR so we keep it in mind.

PR #2176, opened 2022-03-17
Removed some unreachable enemies from Bonus Island II.

Pheazant:
These seem unintentional.

Allows the corresponding levels to be 100%ed.


Semphriss:
@serano01 Same as 2177, no issue on my end; leaving merging up to you

PR #2183, opened 2022-03-20
Fix rocks not hurting

mrkubax10:


tobbi:
Does the cast to `Badguy*` take into account Tux?

mrkubax10:
The spike hitboxes pull request by Daniel. It makes `collision_tile` check for near hurting tiles but since rock is not tile it doesn't work.

Issue #2173, opened 2022-03-16
OpenGL 2.1 hardware: SuperTux fails to load with "X connection to :0 broken (explicit kill or server shutdown)"


lorn10:
**SuperTux v0.6.2 (Ubuntu PPA), v0.6.3 (official AppImage)**
```
test@iMac-4:~$ inxi -b
System:    Host: iMac-4 Kernel: 5.13.0-35-generic x86_64 bits: 64 Desktop: KDE Plasma 5.18.8 
           Distro: Ubuntu 20.04.4 LTS (Focal Fossa) 
Machine:   Type: Desktop System: Apple product: iMac5,1 v: 1.0 serial: <superuser/root required> 
           Mobo: Apple model: Mac-F4228EC8 v: DVT serial: <superuser/root required> BIOS: Apple 
           v: IM51.88Z.0090.B09.0706270921 date: 06/27/07 
CPU:       Dual Core: Intel Core2 T7400 type: MCP speed: 1572 MHz min/max: 1000/2167 MHz 
Graphics:  Device-1: Advanced Micro Devices [AMD/ATI] RV530/M56-P [Mobility Radeon X1600] driver: radeon v: kernel 
           Display: x11 server: X.Org 1.20.13 driver: radeon FAILED: ati unloaded: fbdev,modesetting,vesa 
           resolution: 1680x1050~60Hz 
           OpenGL: renderer: ATI RV530 v: 2.1 Mesa 22.1.0-devel (git-177805c 2022-03-13 focal-oibaf-ppa) 
Network:   Device-1: Marvell 88E8053 PCI-E Gigabit Ethernet driver: sky2 
           Device-2: Broadcom and subsidiaries BCM4360 802.11ac Wireless Network Adapter driver: wl 
Drives:    Local Storage: total: 232.89 GiB used: 131.14 GiB (56.3%) 
Info:      Processes: 182 Uptime: 1d 21m 05m Memory: 2.90 GiB used: 1.07 GiB (36.8%) Shell: bash inxi: 3.0.38
```

##### Expected behavior
SuperTux should work also at **OpenGL 2.1** ~~and **OpenGL ES 1.0**~~ class hardware. 
**Update**: The Debian [SuperTux page](https://wiki.debian.org/Games/Supertux) containing the OpenGL ES 1.0 compatibility is **outdated**.

##### Actual behavior
[...]

PR #2186, opened 2022-03-21
Better OpenGL detection

mrkubax10:
This PR makes VideoSystem use OpenGL version information reported by OS when determining which OpenGL version to use.

### DIFF
--- a/src/video/gl/gl_video_system.cpp
+++ b/src/video/gl/gl_video_system.cpp
@@ -55,8 +55,10 @@ GLVideoSystem::GLVideoSystem(bool use_opengl33core, bool auto_opengl_version) :
   assert_gl();
 
 #if defined(USE_OPENGLES2)
+  m_context.reset(new GL33CoreContext(*this));
   m_use_opengl33core = true;
 #elif defined(USE_OPENGLES1)
+  m_context.reset(new GL20Context);
   m_use_opengl33core = false;
 #else
   if (auto_opengl_version)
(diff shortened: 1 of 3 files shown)


lorn10:
I mention that here too, - this "better OpenGL fallback" really works in my case. :+1: However, this is true for the **ubuntu-latest-64-gcc-Release-appimage** but not the **glbinding** version. More information can be found at my original bug report #2173.

Semphriss:
[Relevant:](https://stackoverflow.com/questions/25171460/which-version-does-glgetstringgl-version-get)

```cpp
GLint opengl_major_version;
glGetIntegerv(GL_MAJOR_VERSION, &opengl_major_version);
if (opengl_major_version > 3) {
 ...
```

mrkubax10:
According to this website: https://www.khronos.org/opengl/wiki/OpenGL_Context
> `glGetIntegerv(GL_MAJOR_VERSION, *);`
> `glGetIntegerv(GL_MINOR_VERSION, *);`
> These are available on OpenGL version 3.0 and above contexts. ...

lorn10:
Again a short side-note. Regarding the **Mesa** drivers on Linux this is absolutely no problem, the **GL_ARB_compatibility** feature is available. :+1: 

However, this may be an issue for (older?) Mac OS versions prior 10.7, see [OpenGL Capabilities Tables - Apple Developer](https://developer.apple.com/opengl/OpenGL-Capabilities-Tables.pdf).

src/video/gl/gl_video_system.cpp
```diff
@@ -61,7 +61,27 @@ GLVideoSystem::GLVideoSystem(bool use_opengl33core) :
...
+    sscanf(version_string, "%d", &major);
+
+    if (major >= 3)
+      m_use_opengl33core = true;
+    else if (major == 2)
+      m_use_opengl33core = false;
+    else
+    {
+      // OpenGL 2.0 or higher is unsupported, throw exception so SDL renderer will be used instead
+      throw std::exception();
```
Semphriss, line 78: There should be a message here that explains what happened

weluvgoatz (approved):

Zwatotem (approved):

PR #2190, opened 2022-04-01
Make objects carried by owl selected from list, fix issues with owl carrying objects other than skydive

mrkubax10:


Semphriss:
Owls don't properly support carrying anything other than a bomb; I advise commenting out the feature entirely until it's stable.

Semphriss:
Are all the fixes done? The title can be updated, and I'll do a quick review when ready.

mrkubax10:
It should be ready for review now.

PR #2192, opened 2022-04-16
Disabled option to change owls' carried object

Semphriss:
The feature was excessively broken and resulted in crashes or excessively weird behavior. Since it hasn't received any attention lately, I disabled it so people don't accidentally use a broken feature (and because it feels unclean to have an unpolished feature available). I did not remove code directly, to make it easier it someone, someday, wants to fix it and re-enable it.


mrkubax10:
#2190 also fixes issues with owl carrying other objects than skydive that I was able to find.

Semphriss:
Might be relevant to update its title then :stuck_out_tongue:

PR #2200, opened 2022-04-25
Fix walking in ground after standing up under pushbutton

### LINKED ISSUE #2197: Tux can walk in the ground after going under pushbuttons
SuperTux version:| 0.6.3

Expected behavior:| Big Tux shouldn't be able to stand under pushbuttons that are slightly below tux's height, instead tux should only be able to crouch / move while crouching under them because of how low they are when tux is standing next to them. 

Actual behavior:| Tux is able to walk under pushbuttons that are slightly below tux's height which causes tux to appear to be walking inside of the ground along with having a backflipping boost and being able to backflip into them and move for a few seconds which is strange behavior.

Steps to reproduce actual behavior:| below

Additional debugging information:|

https://user-images.githubusercontent.com/83178674/164959020-81d6ade6-daf9-4f33-a642-6b621beda4a2.mp4

mrkubax10:
Fixes #2197

### DIFF
--- a/src/object/player.cpp
+++ b/src/object/player.cpp
@@ -319,7 +319,7 @@ Player::adjust_height(float new_height, float bottom_offset)
   if (new_height > m_col.m_bbox.get_height()) {
     //Rectf additional_space = bbox2;
     //additional_space.set_height(new_height - m_col.m_bbox.get_height());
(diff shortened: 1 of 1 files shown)


mrkubax10:
I checked if there are any issues when player tries to stand up under flying platform but I was unable to find any.

PR #2214, opened 2022-05-21
Use an addon index for 0.7.x versions

Vankata453:
Releated to SuperTux/addons#37.

This pull request changes the main add-on index to the new `index-0_7.nfo` file, included in the linked pull request. The aim is having a new index file, which will contain add-on updates, ready for the SuperTux 0.7.0 release, such as the `contrib-type` property for add-ons (also included in the linked PR) and any future changes, without affecting any older SuperTux versions.


Semphriss:
It's a bit early for such a PR, isn't it? :^)

Vankata453:
It is possible for tileset changes (+ possibly any other, that may break levels) to happen soon and add-on owners may want to update their add-ons, so it would be good to have a new index to keep those changes, without affecting 0.6.x versions and below. Even if this PR isn't merged soon, in my opinion it would still be good to switch to a different index as soon as major changes like this happen. Thanks for the comment! 👍

Vankata453:
This should directly be committed, when it's time to change the index.

PR #2215, opened 2022-05-21
Fix missing argument in add_file in some places

### LINKED ISSUE #2213: Worldmap: Adding level dots misconfigures the level's file path
**SuperTux version:** Latest @ master

##### Expected behavior

When adding a level dot in a worldmap, and selecting the level file for that level dot, the game should automatically fill in the correct path.

##### Actual behavior

When adding a level dot in a worldmap, and selecting the level file for that level dot, the game duplicates the path folder. For example, for a level located in `/levels/myworldmap/level1.stl`, the game will put `/levels/myworldmap//levels/myworldmap/level1.stl`.

1. Create a new world.
2. Create a level in that world.
3. Create a worldmap for that world.
4. Edit the worldmap and add a level dot.
5. Edit that level dot, and set its "Level" property to the level created at step 2.
6. Position Tux over the level dot, or add a path on a solid tilemap between Tux and the dot.
7. Any of the following:
  - Test the worldmap
  - Save the worldmap, quit, and re-=open the worldmap from the "Contrib levels" menu.
8. Navigate to the level, and attempt to play it. The game will kick you out to the last screen, complaining that it can't find the level.

mrkubax10:
Fixes #2213

### DIFF
--- a/src/editor/object_settings.cpp
+++ b/src/editor/object_settings.cpp
@@ -241,7 +241,7 @@ ObjectSettings::add_level(const std::string& text, std::string* value_ptr, const
                           const std::string& basedir,
                           unsigned int flags)
 {
-  add_file(text, value_ptr, key, {}, {".stl"}, basedir, flags);
+  add_file(text, value_ptr, key, {}, {".stl"}, basedir, true, flags);
 }
 
 void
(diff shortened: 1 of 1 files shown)


Semphriss:
So there was a missing argument in the _middle_ of the function call, and compilers don't complain because the last argument has a default value, and the type of the argument given in the call can be implicitly cast to a boolean. The joys of programming :laughing: 

I gave the code a quick look and noticed there are plenty of other calls to `add_file` that are missing the special argument. Perhaps it would be wise to `grep -rn "add_file` and fix the ones that don't have a boolean as second-last argument.

PR #2216, opened 2022-05-21
Realign Decals in secret area

serano01:
Just a small tweak in the secret area of "Who Is This Gown?"


tobbi:
Was the path that got deleted needed for anything?

serano01:
Strangely, I didn't delete any path, and don't know why it shows it. But as far as I know, there was never any. Maybe Rusty knows more.

PR #2218, opened 2022-05-22
"No levels available" message in level deletion menu

Vankata453:
Shows a "No levels available" message, when the user is in the editor level deletion menu, and there are no levels to delete.


src/supertux/menu/editor_delete_level_menu.cpp
```diff
@@ -36,6 +36,9 @@ EditorDeleteLevelMenu::EditorDeleteLevelMenu(std::unique_ptr<Levelset>& levelset
 {
   add_label(_("Delete level"));
   add_hl();
+  if (levelset->get_num_levels() == 0) {
```
tobbi, line 39: Put the opening brace on a separate line.

PR #2225, opened 2022-05-31
Move fire and ice caps down to better match Tux

windowsdylan:
[ci skip]

During the stand frame, the fire and ice hats are higher on tux's head then they are during the other frames (best example is walking). The fire hat was 1 pixel too high, and the ice hat was 2 pixels too high compared to the walking animation. I'm guessing this was done by error, as tux's head actually moves from the stand position to the first walking frame. These files move the fire hat down by 1 pixel, and the ice hat down by 2, so these hats stay in the same consistent position.


Semphriss:
"[ci skip]" should be in commit messages :stuck_out_tongue: 

Since this is purely graphical, I'll leave that to @Rusty-Box - although I'm not sure whether this is superseded by #2136.

Rusty-Box:
As I'm working on improving the player sprites and power up costumes overall this isn't necessarily in need of fixing right now. If someone truly wants to do so in the meantime, this is fairly easy to fix. But with the future removal of hats (minus the christmas hat) I think this is ignorable.

ixPost:
And the miner helment(or earth cap) is too high.

ixPost:
> But with the future removal of hats (minus the christmas hat) I think this is ignorable.

What removal of hats!?

PR #2226, opened 2022-06-02
Fix triggers

mrkubax10:


Rusty-Box:
They work again - what a lifesaver you are!

PR #2227, opened 2022-06-03
Fix unportable trampoline sprite changing

### LINKED ISSUE #2220: Unable to change the sprite of unportable trampoline
**SuperTux version:**v0.6.3-529-g80198aff5
**System information:** irrelevant

##### Expected behavior
When I set the trampoline sprite to anything else, it will appear so.

##### Actual behavior
When the trampoline is unportable, it always uses the default sprite, regardless of what sprite it is set to.

1. Place a trampoline in the level editor.
2. Set portable to false
3. Change the sprite to anything else.
4. Play the level.
5. You will see the default sprite instead.

O_o

Maybe this bug was introduced because the sprite wasn't changed automatically when the portable flag was set to false. In such case, I suggest the sprite being changed automatically in the level editor when the portability is changed, not when the level is loaded.

mrkubax10:
Fixes #2220

### DIFF
--- a/src/util/reader_mapping.hpp
+++ b/src/util/reader_mapping.hpp
@@ -56,6 +56,7 @@ public:
 
   bool get(const char* key, sexp::Value& value) const;
 
+  bool has_item(const char* key) const;
   /** Read a custom data format, such an as enum. The data is stored
       as string and converted to the custom type using the supplied
       `from_string` convert function. Example:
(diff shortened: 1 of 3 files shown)


src/util/reader_mapping.hpp
```diff
@@ -56,6 +56,7 @@ class ReaderMapping final
 
   bool get(const char* key, sexp::Value& value) const;
 
+  bool has_item(const char* key) const;
```
Hume2, line 59: If I recall correctly, there is already a standard way to handle optional attribute. But using this instead doesn't mind me at all because it looks more straightforward.

PR #2229, opened 2022-06-12
Drag drop install addons

tobbi:
This is still kinda work in progress...there have been some crashes here and there if I can still remember.

I'd like for people to try this out though.

### DIFF
--- a/src/addon/addon_manager.cpp
+++ b/src/addon/addon_manager.cpp
@@ -17,12 +17,16 @@
 
 #include "addon/addon_manager.hpp"
 
+#include "boost/format.hpp"
+#include <boost/algorithm/string/predicate.hpp>
 #include <physfs.h>
 
 #include "addon/addon.hpp"
 #include "addon/md5.hpp"
+#include "gui/dialog.hpp"
 #include "physfs/util.hpp"
 #include "supertux/globals.hpp"
+#include "supertux/menu/menu_storage.hpp"
 #include "util/file_system.hpp"
 #include "util/gettext.hpp"
 #include "util/log.hpp"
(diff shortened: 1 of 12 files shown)


src/supertux/screen_manager.cpp
```diff
@@ -488,6 +489,13 @@ ScreenManager::process_events()
         {
           session->toggle_pause();
         }
+        break;
+
+      case SDL_DROPFILE:
+        char* filename = event.drop.file;
+        AddonManager::current()->install_addon_from_local_file(std::string(filename));
+        SDL_free(filename);
```
Semphriss, line 497: This is part of `event.drop.file`, which is managed by SDL; are you sure it needs to be freed manually?

Semphriss:
Will review the PR in full in a bit

src/supertux/screen_manager.cpp
mrkubax10: According to https://wiki.libsdl.org/SDL_DropEvent it has to be freed manually.

src/addon/addon_manager.cpp
```diff
@@ -402,6 +406,22 @@ AddonManager::install_addon(const AddonId& addon_id)
   }
 }
 
+void
+AddonManager::install_addon_from_local_file(const std::string& filename)
+{
+  const std::string& source_filename = FileSystem::basename(filename);
+  if(!boost::algorithm::ends_with(source_filename, ".zip"))
+    return;
```
mrkubax10, line 414: Maybe it would be better idea to check file magic instead (e.g. see if it starts with 'PK'). I'm not sure what impact it would have on performance, if it makes much difference it's fine how it is already.

mrkubax10 (approved):

Zwatotem (approved):
If it works, you're good to go.

Vankata453 (approved):

PR #2230, opened 2022-06-16
List object names within editor toolbox

Vankata453:
Adds the ability for an object's name to be shown in the editor toolbox, when the mouse cursor hovers over it.

image

**Potential issue:** Currently, the toolbox gets an object's name by using a new temporary instance of it, via `GameObjectFactory`. That may not be a good solution, because creating an object just to get its name is not optimal (additional messages from the new object can also be seen in the log, because of that). **[FIXED]**

### DIFF
--- a/src/badguy/angrystone.hpp
+++ b/src/badguy/angrystone.hpp
@@ -32,9 +32,12 @@ public:
   virtual void unfreeze(bool melt = true) override;
   virtual bool is_freezable() const override;
   virtual bool is_flammable() const override;
-  virtual std::string get_class() const override { return "angrystone"; }
-  virtual std::string get_display_name() const override { return _("Angry Stone"); }
+
   virtual std::string get_overlay_size() const override { return "3x3"; }
+  static std::string class_name() { return "angrystone"; }
+  virtual std::string get_class_name() const override { return class_name(); }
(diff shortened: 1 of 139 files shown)


Semphriss:
A potential solution to avoid having to create an object to get its name would be to refactor all GameObjects to add a static class function that would return the object's name and classname, and to have the virtual member (non-static) functions to fetch the names and classnames from their static counterparts. The GameObjectFactory class could then be refactored to allow calling the correct static function from the classes it holds, which could be used for this PR.

<sub>This is something I had planned doing for a certain fork.</sub>

-----------------------------

In short:
```cpp
...

class MyGameObject : public GameObject
{
public:
  static std::string classname() { return "mygameobject"; }
  static std::string name() { return "My GameObject"; }

...

  virtual std::string get_classname() const { return classname(); }
  virtual std::string get_name() const { return name(); }

...

};

...
```
(I don't remember how the currently existing functions are named, just make sure the static functions don't share names with the non-static ones)

As for the GameObjectFactory, I suppose there is a template somewhere; say the template is called `T`, you can invoke one of its static functions with `T::name();`.

weluvgoatz (approved):

mrkubax10 (approved):

PR #2231, opened 2022-06-17
Package addons as "Community Contrib Levels"

Vankata453:
Changes the `contrib-type` property of levelsets, packaged using the in-game level editor option "Package Add-on", to `community`, so new add-ons, packaged from there, show up in the "Community Contrib Levels" section in "Contrib Levels". Achieved by replacing the `info` file of the levelset with one where the property is set to `community`, before the add-on is archived, and afterwards reverting it back to `user`.

### DIFF
--- a/src/editor/editor.cpp
+++ b/src/editor/editor.cpp
@@ -943,6 +943,8 @@ Editor::pack_addon()
 {
   auto id = FileSystem::basename(get_world()->get_basedir());
 
+  get_world()->save(false, true); //Save with "contrib-type" set to "community".
+
   int version = 0;
   try
   {
@@ -985,6 +987,8 @@ Editor::pack_addon()
   info.end_list("supertux-addoninfo");
 
   *zip.Add_File(id + ".nfo") << ss.rdbuf();
+
+  get_world()->save(); //Revert saving with "contrib-type" set to "community".
 }
 
 /* EOF */
--- a/src/editor/editor.cpp
+++ b/src/editor/editor.cpp
@@ -943,7 +943,7 @@ Editor::pack_addon()
 {
   auto id = FileSystem::basename(get_world()->get_basedir());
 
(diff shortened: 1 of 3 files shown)


Semphriss:
IMO, there is a semantic problem with this PR; the point of the "user" category is, IIUC, to avoid mixing a user's self-generated add-ons with the ones they downloaded from the official repos. I see it as:
- "user", for everything the user created themselves;
- "official", for everything created by the team;
- "community", for everything created by other people and downloaded through the add-ons system.

With the current PR, a user's self-generated add-on would appear in "community", which would break its meaning.

I believe the reason of this PR is to make the job simpler for the team to redistribute contributed add-ons. I think it is better to handle that step manually: since it converts everything automatically to "community" add-ons, the ones that are meant to be "official" will need to be handled manually anyways (same thing if new categories are ever added in the future); also, the team should be reviewing and editing add-ons as necessary anyways, in all circumstances (whether it be for security, quality control, etc.), and the task of changing "user" to "community" or "official" is minor.

I would advise not doing anything unless long-term usage shows a real need for this feature; however, if (and/or when) it is decided to implement this, I would advise two modifications:
1. Have a menu that would allow the user to select what category they want to package their add-on as ("user", "community", "official" or any potential future category)
2. Make it export the resulting package in a temporary folder and offer the user to either upload the add-on automatically, or to open the temporary folder for manual intervention. The package would, by default, not show up at all in the "community" category, or any other, and it would be deleted automatically once the operation is complete. Optionally, there may be an option like "Test install add-on" (or any better wording) that would install the package into the user's add-ons, which would allow them to see how it would look when downloaded from the official add-on repository, if they wish to.

I'm not sure whether all of this is necessary, hence why I advise not doing anything for now, and wait to see if there is a need for it. If (and when) it is decided to do something, I would recommend following the two points above.

mrkubax10:
Given the comment above I advise to close this PR for now

PR #2238, opened 2022-07-06
Menu transitions for dialogs

### LINKED ISSUE #2071: Confirmation dialogs don't have transitions
Even if "Enable Transitions" is set to true, confirmation dialogs (the ones with "Yes" and "No" options) still do not have a transition, when entering, or quitting them.

Vankata453:
Adds the ability for non-passive dialogs to have menu transitions.

Fixes #2071.

### DIFF
--- a/src/gui/menu_manager.cpp
+++ b/src/gui/menu_manager.cpp
@@ -375,7 +375,7 @@ MenuManager::previous_menu() const
 }
 
 void
-MenuManager::transition(Menu* from, Menu* to, bool call_this)
+MenuManager::transition(Menu* from, Menu* to, bool call_this) // "call_this" -> calls this specific overload to prevent ambiguous calls
 {
   if (!from && !to)
   {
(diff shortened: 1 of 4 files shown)


src/gui/menu_manager.cpp
```diff
@@ -343,7 +375,7 @@ MenuManager::previous_menu() const
 }
 
 void
-MenuManager::transition(Menu* from, Menu* to)
+MenuManager::transition(Menu* from, Menu* to, bool call_this)
```
mrkubax10, line 378: Is this parameter needed?

Vankata453: WIthout this parameter, calls to the `transition()` function that may include `nullptr`, appear to be ambiguous.

mrkubax10: Alright, also yeah, adding comment describing that is a good idea.

weluvgoatz (approved):

mrkubax10 (approved):

Issue #936, opened 2018-11-19
Strings should allow for copy and paste, and text selection


Alzter:
**SuperTux version: 0.6.0 Alpha 3**
**System information:** *Windows 10*

At the moment strings (like script boxes in the level editor) don't let you select text or copy and paste. I would like it if you could select text, copy selected strings and paste in clipboard text into the string box. It would make copying and pasting scripts much easier.

Semphriss:
The new UI system does that and more - I'm leaving this issue open though, the legacy UI system still doesn't support that, and the editor still uses the legacy UI system.

PR #2246, opened 2022-07-11
Improvements to text field item and script menu

Vankata453:
Improves the text field and script line menu items:

* Support for going back in text and inserting/deleting characters from the current position. ("Backspace" for deleting characters in front of the cursor, "Delete" for deleting behind it.)
* Cut and copy support for whole lines. [Ctrl+X, Ctrl+C]
* Paste support (new-line characters are converted to spaces). [Ctrl+V]
* Undo/redo support (restricted to 1 step for both). [Ctrl+Z, Ctrl+Y]

Some tiny improvements take place in the script menu as well:

* Ability to duplicate lines. [Ctrl+D]
* Fix for a bug, where the menu doesn't change its size when a script line is deleted.

Other improvements included:

* The int field and float field items were refactored. They now inherit the text field item. A way to restrict them to only having positive values was also added.
* Objects can now have options, which support multi-line text editing. An example where this is applied is the info block's "Message" option.

This PR aims to provide a partial fix to #936. Partial cut/copy, and paste support are added, but not selecting text.

### DIFF
--- a/src/editor/object_option.cpp
+++ b/src/editor/object_option.cpp
@@ -240,6 +240,42 @@ StringObjectOption::add_to_menu(Menu& menu) const
   menu.add_textfield(get_text(), m_pointer);
 }
 
+StringMultilineObjectOption::StringMultilineObjectOption(const std::string& text, std::string* pointer, const std::string& key,
+                                       boost::optional<std::string> default_value,
+                                       unsigned int flags) :
+  ObjectOption(text, key, flags),
+  m_pointer(pointer),
+  m_default_value(std::move(default_value))
+{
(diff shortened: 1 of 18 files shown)


weluvgoatz (changes requested):
This PR looks good.  I *might* suggest that you add text selection, but it isn't that big of a deal with how it works currently and this PR can be merged without it.

One thing I hope to be added, which could fit in this PR, is multi-line entries for the infoblocks.

weluvgoatz (approved):

mrkubax10 (approved):

PR #2249, opened 2022-07-16
Creating Version reminder to tell which versions are up to date

CPa129:
This pr has a file that can be used to tell:  
1.If the version you are playing is supported.
2.Warn you if your version is not supported or is too old.
These are improvements that should be done by your side.
And also,This txt should open up to tell.


mrkubax10:
This PR contains default template text and I don't think that it's needed for now.

PR #2254, opened 2022-07-18
Icecrusher improvements

weluvgoatz:
This PR simplifies icecrushers and makes their behavior more consistent and sensible, especially when breaking bricks.  While I tried to fix icecrushers up as best I could, there are some bugs that seem to relate to our ever-broken collision system and as a result I couldn't fix those.  Those include:
- Rocks forcing icecrushers down despite the fact that they should definitely not be doing that
- Upon icecrusher hit, bricks with coins will release 1 coin and then break, creating an inconsistency but not necessarily a large one since those coins do not count for anything stats-wise.  I am aware it may because of the fake brick collision I had to implement, but I tried to create a check that prevented them from breaking that just would not work under any circumstances.
- Sideways icecrushers will force 3 contents out of bonus blocks for some reason, but normal icecrushers will force out the intended 1.


ixPost:
> Upon icecrusher hit, bricks with coins will release 1 coin and then break, creating an inconsistency but not necessarily a large one since those coins do not count for anything stats-wise. I am aware it may because of the fake brick collision I had to implement, but I tried to create a check that prevented them from breaking that just would not work under any circumstances.

The same happens with ice cubes.

mrkubax10 (approved):

PR #2255, opened 2022-07-18
Fix bug with swimming, walljumping and bricks

weluvgoatz:
There was a really annoying bug where bricks would repeat their sound forever if Tux swimjumped, walljumped, and then landed on them, so I fixed it.


weluvgoatz:
Merging already because it's a 1-line change that cannot possibly go wrong.

PR #2263, opened 2022-07-24
Remove hardcoded tile IDs for electrocution

### LINKED ISSUE #2261: [Tileset PR] New snow slope tiles might still have unintended attributes assigned
0.6.3-574-g11899ba2d
win10

image

Here, this tile, upon electrocution when submerged in solid water from lightning, turns into part of a rain cloud:

image

This is for Rusty's PR, #2136. might have some unintended attribute assigned? thought im not sure what tile changes (other than water) to a different tile upon being shocked.

Vankata453:
Removes hardcoded tile IDs, used for water electrocution in `thunderstorm.cpp`. They are now defined in the newly created `additional` category for additional attributes and properties in tileset files.

Importing the water electrocution tile IDs works by adding these data sections in the file, preferably at the end:

```
 ;; Additional attributes
  (additional
    (thunderstorm
      ;; Attributes for thunderstorm changing tiles
      (changing-tiles
        200  1421
        3419 3523
        3420 3524
        3421 3525
        [more pairs below...]
      )
    )
  )
```

Fixes #2261.

### DIFF
--- a/src/object/thunderstorm.cpp
+++ b/src/object/thunderstorm.cpp
@@ -19,7 +19,10 @@
 #include "audio/sound_manager.hpp"
 #include "editor/editor.hpp"
 #include "object/electrifier.hpp"
+#include "supertux/level.hpp"
 #include "supertux/sector.hpp"
+#include "supertux/tile_manager.hpp"
+#include "supertux/tile_set.hpp"
 #include "util/reader.hpp"
 #include "util/reader_mapping.hpp"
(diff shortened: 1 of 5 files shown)


weluvgoatz:
My suggestion is that we wait for the tileset PR (#2136) to be finished, and then after that, update tiles.strf in this PR to include the thunderstorm tile changes.  Otherwise, this looks good.

PR #2267, opened 2022-07-26
Add notification GUI item + check for new releases

### LINKED ISSUE #248: Update check for SuperTux

Vankata453:
Adds a new `Notification` GUI item, which shows up on the top-right of the screen with appearance, similar to the one of passive dialogs. It features "Do not show again" and "Close" buttons, which are useful if users want to ignore a notification. Clicking on a notification executes a given callback.

Other than that, this PR also adds the ability for the game to check which the latest SuperTux version is on startup, and if it's different from the current version, show a notification. Checking for new releases on game startup can be toggled from "Options". Clicking on the notification for the new release shows a dialog, indicating that the user could go to the SuperTux website and view more information about the new release there.

Preview of the dialog for new releases

Fixes #248.

### DIFF
--- a/src/gui/menu_manager.cpp
+++ b/src/gui/menu_manager.cpp
@@ -20,6 +20,7 @@
 #include "gui/dialog.hpp"
 #include "gui/menu.hpp"
 #include "gui/mousecursor.hpp"
+#include "gui/notification.hpp"
 #include "supertux/gameconfig.hpp"
 #include "supertux/globals.hpp"
 #include "supertux/menu/menu_storage.hpp"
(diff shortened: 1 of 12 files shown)


Vankata453:
The only part of this PR which has to be considered is the location where the file, which allows the game to perform release checks (`verinfo.nfo`), should be hosted. It should be updated on every major release of the game. Currently, this PR is set to add it to the root of this repository, but I don't think this is a good hosting location for this file.

weluvgoatz (approved):

mrkubax10 (approved):

PR #2275, opened 2022-08-04
Option to disable sound positioning

Semphriss:
This allows users to disable sound positioning entirely from a config entry.

Observations, remarks, caveats:
- Sounds that are played far away from the screen are very audible when the option is activated. This may give players a hint about activity far away from the screen, as well as when certain things are disabled because they are too far away to be worth updating every frame.
- If the option is changed from the pause menu while in-game, the sounds that are already playing won't be updated.
- The 'sssssssssss' of the bomb is excessively loud. This may need a fix.

### DIFF
--- a/src/audio/openal_sound_source.cpp
+++ b/src/audio/openal_sound_source.cpp
@@ -17,6 +17,8 @@
 #include "audio/openal_sound_source.hpp"
 
 #include "audio/sound_manager.hpp"
+#include "supertux/gameconfig.hpp"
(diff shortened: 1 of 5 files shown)


weluvgoatz (approved):

Vankata453:
Is there a need for such option? This would entirely break ambience, and seems to already have issues on its own.

Vankata453:
Closing this due to inactivity. Feel free to reopen if needed!

PR #2281, opened 2022-08-10
Swimming fish

weluvgoatz:
This PR is a collab between me and @Rusty-Box, with me doing the code and Rusty doing the graphics.  It adds 3 new swimming fish enemies so that Tux has some obstacles while in the water.  There are 3 new types of fish:
- **Swimming Fish**:  Just a normal, blue, fish that swims back and forth.
- **Harmless Fish**: A goldfish resembling the frozen fish that is harmless and deflects Tux off of it.
- **Chaser Fish**: A dopefish resembling the frozen dopefish found in the tiles that chases Tux when in range.


serano01:
So swimming and harmless fish are living decals and are not counting to the overall badguy-counter of the level, right? Regarding the jumping fish, he also could need an animation when in water, because he disappears in it, while everything else can clearly be seen.

mrkubax10:
Two things that I noticed after brief look at source code:
- Tabs are used instead of 2 spaces for indentation in some places
- Files `src/badguy/fish_swimming.hpp` and `src/badguy/fish_swimming.cpp` mention Semphris as author (unless he actually is author of these files).

PR #2282, opened 2022-08-11
Remove old unused Forest-Map music

serano01:
I think this one can be deleted, since it's old, unused, way too short and has already an proper replacement.


mrkubax10:
I think that before merging this pull request forestmap.ogg should be added to media repository.

serano01:
> I think that before merging this pull request forestmap.ogg should be added to media repository.

I would do it, but I have no rights to upload files to that repository.

mrkubax10 (approved):

PR #2284, opened 2022-08-13
Sliding + crawling

### LINKED ISSUE #1081: Sliding In SuperTux
Just like swimming, sliding in SuperTux is considered an essential penguin behaviour according to the Internal Chat.  Multiple things worth considering about sliding are:

- Does it work on all slopes? Or just special blocks?
- How does Tux attack enemies while sliding?
- What actions can Tux do?
- Would it break levels?

weluvgoatz:
Tux can now slide on slopes and on the ground.  He is a penguin AND a clone of Super Mario, both of which have this ability, so he should have it as well.  He gains the most speed from going down slopes, but can also gain some speed from falling onto normal terrain; this gives sliding more usage since slopes can't be used everywhere. Also, his acceleration depends on the intensity of the slope.  He can also slide up off of inclines, with the angle of slide-jumping also based off of the intensity of the slope.  Also added is "sniping" for when Tux hits certain enemies while sliding at a high speed.

Fixes #1081

### DIFF
--- a/src/badguy/badguy.cpp
+++ b/src/badguy/badguy.cpp
@@ -475,7 +475,7 @@ HitResponse
 BadGuy::collision_player(Player& player, const CollisionHit& hit)
 {
   if (player.is_invincible() ||
-    (is_snipable() && ((player.is_sliding() && glm::length(player.get_velocity()) > 320.f) || player.is_swimboosting()))) {
+    (is_snipable() && player.is_sliding())) {
     kill_fall();
     return ABORT_MOVE;
   }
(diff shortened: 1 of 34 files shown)


Rusty-Box:
Initial feedback/suggestions:
- Big Tux should be able to break crates while sliding (excluding metal crates)
- Bombs should explode when sliding into them
- Tux's sliding speed should have a tiny bit less friction. He stops a tad too soon at high speeds
- Badguys can hurt Tux right before he stops even though he is still in his sliding animation
- Perhaps instead of gaining speed after landing on terrain his speed should pick up whilst falling
- Tux should remain in his sliding mode whilst jumping during sliding - would allow for some nice parkour ;)

Other than that, this feel so good and is so much fun! A great and highly overdue gameplay addition!

Rusty-Box:
RE: Breaking Crates as Big Tux - It does indeed work, however like sideway Ice Crusher, Tux should blast through a whole line of them without getting stopped in his tracks after just one!

Rusty-Box (approved):
Sliding mechanic is really fun right now and all my suggestion have been addressed. Approved!

PR #2288, opened 2022-08-16
Allow buttons to be flipped with a toggle, fix positioning bug

Narre:
After #2208 was merged, the upside-down button sprite added with the flipping update was obsoleted, and it became impossible to have buttons that are flipped by default (meaning buttons that are on the ceiling in a non-flipped level). It also introduced a bug where a button would move a few pixels if pressed while the level was flipped. This PR introduces an option for buttons to be flipped by default and it also fixes the positioning bug.

image
_Previously, the buttons would end up in different positions based on whether they were pressed while flipped or not. This is now fixed._

image
_Buttons can now be upside-down by default using a toggle. The separate flipped sprite was removed._

The buttons behave as expected - they can always be pressed from the flat side.

One code quality note here. The way I wrote the sprite flip update is :vomiting_face:. But I genuinely don't know where else I can put this update. Please, before this gets merged, someone tell me how to do this properly and I'll do it.

image
_The sprite flip update has no place in the get_settings() function, better idea please..._

I'll leave this PR as a draft before the sprite update is improved.


src/object/pushbutton.cpp
```diff
@@ -51,8 +55,12 @@ PushButton::get_settings()
   ObjectSettings result = MovingSprite::get_settings();
 
   result.add_script(_("Script"), &script, "script");
+  result.add_bool(_("Upside down"), &m_upside_down, "upside-down");
+
+  result.reorder({"script", "upside-down", "x", "y"});
 
-  result.reorder({"script", "x", "y"});
+  if (m_upside_down && m_flip == NO_FLIP)
+    FlipLevelTransformer::transform_flip(m_flip);
```
mrkubax10, line 63: This code can be moved to `after_editor_set` method. Also another thing, apparently pushbutton remains flipped even if you uncheck the Upside down option. This can be fixed by changing `if (m_upside_down && m_flip == NO_FLIP)` to `if ((m_upside_down && m_flip == NO_FLIP) || (!m_upside_down && m_flip == VERTICAL_FLIP))`.

mrkubax10 (approved):

PR #2289, opened 2022-08-16
Fix translation error

jorgesumle:


mrkubax10:
Please use Transifex for translations

jorgesumle:
How to search that string on Transifex? I tried, but I didn't find it. I mean something like grep -R 'string' on Transifex

jorgesumle:
I've finally found the string on Transifex and fixed it. What a horrible GUI! Why not use something like [Weblate](https://weblate.org/), which is also free software?

PR #2290, opened 2022-08-26
Add Conveyor Belt object

Raoul1808:
This PR adds conveyor belts, a new object that shifts entities in a certain direction at a certain rate.

They have variable length and speed and can shift entities left or right. The conveyor can also be controlled through a script to start or stop the conveyor, change its speed or direction.

Some things may need more polish, like for example extending the texture/animation so it can be seamless over longer belts.

Any feedback is welcome, especially on the code. I'm very new to adding new objects into the game.


mrkubax10:
Could you fix compilation issues?

Raoul1808:
I just changed the way the conveyor belt shift objects, but there's still the issue of some enemies turning around at the end of the conveyor belt (it only affects enemies that stay on their platform), I don't really know how to tackle this

weluvgoatz (changes requested):
This is pretty well done and they are very fun to play with.  Conveyor belts could work good in World 4.  However I do have some requests:

1) The graphics are unfitting.  You can honestly get away with the current tiled approach and don't need a seamless texture but the graphic itself should fit better with our current graphic style.  Maybe get a graphic designer on this.
2) When you extend a conveyor belt in the editor, you cannot grab on to the extended part with the tooltip, only the original 1x1 tile part.  I think fixing this has something to do with after_editor_set() (which is or is similar to a function).
3) You already know this, but the stay-on-platform enemies linger weirdly on the platforms.  Maybe there should be an override in which they can be ejected from conveyor belts?

Overall good job!

mrkubax10 (changes requested):
I just noticed that files `src/object/conveyor_belt.hpp`, `src/object/conveyor_belt.cpp`, `src/scripting/conveyor_belt.hpp` and `src/scripting/conveyor_belt.cpp` have wrong indentation (4 spaces instead of 2)

mrkubax10 (approved):

src/object/conveyor_belt.cpp
```diff
@@ -0,0 +1,163 @@
...
+#include "sprite/sprite_manager.hpp"
+#include "supertux/direction.hpp"
+#include "math/util.hpp"
+#include "util/reader_mapping.hpp"
+
+ConveyorBelt::ConveyorBelt(const ReaderMapping &reader) :
+  MovingObject(reader), // TODO: sprite
+  ExposedObject<ConveyorBelt, scripting::ConveyorBelt>(this),
+  m_running(true),
+  m_direction(Direction::LEFT),
```
Vankata453, line 30: Direction member variables are mostly named `m_dir` in the code, so this one could be renamed to `m_dir` as well. Do this only if you wish to.

src/object/conveyor_belt.cpp
```diff
@@ -0,0 +1,163 @@
...
+  reader.get("length", m_length, 1);
+  if (m_length <= 0)
+    m_length = 1;
+
+  m_col.m_bbox.set_size(32.0f * static_cast<float>(m_length), 32.0f);
+
+  if (!m_running)
+    m_sprite->set_action("stopped");
+  else
+    m_sprite->set_action((m_direction == Direction::LEFT ? "left" : "right"));
```
Vankata453, line 55: `set_action(m_direction)`, also to be changed for other such calls in the file.

src/object/conveyor_belt.hpp
```diff
@@ -0,0 +1,91 @@
...
+//
+//  You should have received a copy of the GNU General Public License
+//  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#ifndef HEADER_SUPERTUX_OBJECT_CONVEYOR_BELT_HPP
+#define HEADER_SUPERTUX_OBJECT_CONVEYOR_BELT_HPP
+
+#include "scripting/conveyor_belt.hpp"
+#include "squirrel/exposed_object.hpp"
+#include "supertux/moving_object.hpp"
```
Vankata453, line 22: Move these two includes to the top, separated by an empty line from the current ones, as they are base classes for `ConveyorBelt`.

src/object/conveyor_belt.hpp
```diff
@@ -0,0 +1,91 @@
...
+
+  float m_frame;
+  int m_frame_index;
+
+  const float MAX_SPEED = 32.0f;
+
+  std::unique_ptr<Sprite> m_sprite;
+
+
+
```
Vankata453, line 82: Two redundant empty lines, remove if you wish to.

src/scripting/conveyor_belt.hpp
Vankata453, line 1: The scripting file is missing documentation as a whole, see other scripting files for reference.

src/supertux/game_object_factory.cpp
```diff
@@ -234,6 +235,7 @@ GameObjectFactory::init_factories()
   add_factory<Camera>("camera");
   add_factory<Candle>("candle");
   add_factory<CirclePlatform>("circleplatform");
+  add_factory<ConveyorBelt>("conveyor-belt");
   add_factory<CloudParticleSystem>("particles-clouds");
```
Vankata453, line 239: Swap these two lines to achieve alphabetical sorting.

Vankata453 (changes requested):

Vankata453 (approved):

weluvgoatz (approved):

PR #2299, opened 2022-09-28
Fix ItemIntField inserting 0, when "-" inserted in empty field

Vankata453:
Fixes an issue, where ItemIntField items insert "0", when a "-" is inserted, often for the creation of a negative number. Also removes one unneeded multiplication operation from both ItemIntField and ItemFloatField.

Related to #2246.


tobbi:
I think this maybe somewhat looks good, perhaps. So I might be merging it. If it causes any issues, I'll nonetheless blame you.

PR #2302, opened 2022-10-01
Separate dispenser sprite files + remove rocker launcher

Vankata453:
Separates dispenser sprites in 3 files, according to type:

- dropper.sprite
- canon.sprite
- invisible.sprite

Removes rocket launcher dispenser type, including code related to it.

Related to #2136.


Rusty-Box:
Okay, so I tested it and everything seems to work except for the iced-canon animations. Instead of playing the actual animation it just tints the normal canon sprite dark blue. Also they should play the freezing sound used for freezing adguys (i.e. sizzle.ogg)

PR #2308, opened 2022-10-07
Add "loop-frame" property for sprite actions

### LINKED ISSUE #1109: Add "loop postion" for looping animations
What I'm suggesting is that we add a "loop position" option to animations in .sprite files. The loop position is which frame to play the animation at when it loops. For example, I could make a running animation for Tux that has a transition into running for frames 1 - 3 and the actual running at frames 4 - 12. By setting the loop position to 4 once the animation loops Tux would continue running and skip the transition at the start of the animation for subsequent loops.

Vankata453:
Adds a new `loop-frame` property for sprite actions, which allows a sprite to loop from a specific frame. It accepts values from 1 to the total amount of frames available.

Closes #1109.

### DIFF
--- a/src/sprite/sprite.cpp
+++ b/src/sprite/sprite.cpp
@@ -125,7 +125,7 @@ Sprite::update()
 
   while (m_frameidx >= get_frames() && !animation_done()) {
     // Loop animation.
-    m_frameidx -= get_frames() - (m_action->loop_frame > 1 ? m_action->loop_frame - 1 : 0);
+    m_frameidx -= get_frames() - (m_action->loop_frame - 1);
     m_animation_loops--;
   }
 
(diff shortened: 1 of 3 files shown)


mrkubax10 (approved):

src/sprite/sprite.cpp
```diff
@@ -124,7 +124,8 @@ Sprite::update()
   }
 
   while (m_frameidx >= get_frames() && !animation_done()) {
-    m_frameidx -= get_frames();
+    // Loop animation.
+    m_frameidx -= get_frames() - (m_action->loop_frame > 1 ? m_action->loop_frame - 1 : 0);
```
tobbi, line 128: Since `loop_frame` is set to 1 if it isn't set in the sprite file, couldn't we all get rid of the whole if statement here?

m_action->loop_frame defaults to 1 which means that both cases of the if statement would result to 0.

PR #2316, opened 2022-10-11
Fix build when libdir is 'lib64'

h3xx:
If you specify `-DCMAKE_INSTALL_LIBDIR=lib64` it would break the build when trying to link to `libsquirrel_static.a`.

Because of `-DCMAKE_INSTALL_LIBDIR=lib64`, `libsquirrel_static.a` was being put in `<build dir>squirrel/ex/lib64/libsquirrel_static.a`.

This patch changes `-DCMAKE_INSTALL_LIBDIR=lib` just when building libsquirrel, causing the static libraries to be put in the `lib` dir, which is hard-coded inside the file.


tobbi:
Thanks 👍

PR #2320, opened 2022-10-19
Refactor undo/redo system to track separate object changes

### LINKED ISSUE #1108: Add option to hide undo/redo buttons in level editor
**SuperTux version: 0.6.0**

Can we add an option to hide and show the undo and redo buttons in the level editor?

### LINKED ISSUE #1688: UndoManager generates excessive amounts of lag in the editor
Currently, whenever a key or a mouse button is released, the UndoManager saves the entire level, regardless of whether or not something has actually changed.

With big levels, this can generate a small lag that significantly deteriorates the user experience, in addition to needlessly filling the undo manager with versions that contain no change at all.

I propose to change this to make it so it only the changes are pushed in the undo manager. The event would be called whenever an object is added, deleted or edited, and it would make sure something has changed before pushing on the undo stack. This will increase performance, reduce memory consumption and give and overall better experience.

### LINKED ISSUE #1954: Undo causes level editor to go back to tilemap 0
**SuperTux version:** 0.6.3 - Beta 1

##### Expected behavior
When the undo feature is used in the level editor and the selected tilemap is not 0, there shouldn't be any switching between tilemaps.
##### Actual behavior
When the undo feature is used in the level editor and the selected tilemap is not 0, the editor automatically switches to tilemap 0.
Open a worldmap or a level in the level editor, choose a tilemap, different than 0, place a tile or an object and undo.

### LINKED ISSUE #2297: Big performance issues with the editor.
**SuperTux version: SuperTux-v0.6.3-586-ge1e6fa8d0**
**System information:** *Windows 11 Amd Ryzen 5 with Vega 12*

When placing tiles, copying and pasting them around the RAM usage keeps increasing and it never goes down. This issue makes the editor obnoxiously laggy all the time and placing objects or doing anything in the editor can even take up a second.

https://user-images.githubusercontent.com/69918580/191545620-7b4557ff-b7e5-4f33-89f1-1dac45072b54.mp4

Vankata453:
`UndoManager`, previously used by the editor, has been removed, and the undo/redo functionality has been migrated to `GameObjectManager`, where the approach taken to save and reproduce object changes is to track each `GameObject`'s data separately.

The new object tracking system can only be enabled inside sectors, which take place in the editor, and can also manually be turned on/off by the user. The maximum amount of saved changes in the undo stack can also be changed by the user. The undo stack is cleaned up when pushing to it and when setting a new maximum limit to undo stack changes.

The virtual `save_state()` and `check_state()` `GameObject` functions can be overriden, meaning that markers or more complex objects can also save other objects' states, so they can be tracked.

Closes #1108.
Fixes #1688.
Closes #1954.
Fixes #2297.


mrkubax10 (approved):

src/editor/editor.cpp
```diff
@@ -872,31 +856,74 @@ Editor::check_save_prerequisites(const std::function<void ()>& callback) const
...
+void
+Editor::redo()
+{
+  BIND_SECTOR(*m_sector);
+  m_sector->redo();
+  perform_post_undo_actions();
+}
+
+void
+Editor::perform_post_undo_actions()
```
tobbi, line 921: Maybe change this name to reflect that this is run on both undo and redo?

src/editor/layers_widget.cpp
```diff
@@ -311,37 +322,13 @@ EditorLayersWidget::refresh()
   m_selected_tilemap = nullptr;
   m_layer_icons.clear();
 
-  bool tsel = false;
-  TileMap* first_tm = nullptr;
   for (auto& i : m_editor.get_sector()->get_objects())
```
tobbi, line 325: Can this be marked `const auto&` ? and can the variable name be changed to `obj`?

src/editor/layers_widget.cpp
```diff
@@ -353,6 +340,41 @@ EditorLayersWidget::refresh_sector_text()
...
+void
+EditorLayersWidget::refresh_layers()
+{
+  bool tsel = false;
+  TileMap* first_tm = nullptr;
+  for (auto& i : m_layer_icons)
+  {
+    auto* go = i->get_layer();
+    auto tm = dynamic_cast<TileMap*>(go);
+    if (tm)
```
tobbi, line 352: maybe reduce nesting?

if(!tm)
continue;

src/editor/layers_widget.cpp
```diff
@@ -353,6 +338,41 @@ EditorLayersWidget::refresh_sector_text()
   m_sector_text_width  = int(Resources::normal_font->get_text_width(m_sector_text)) + 6;
 }
 
+void
+EditorLayersWidget::refresh_layers()
+{
+  bool tsel = false;
```
tobbi, line 344: maybe rename this to `tilemap_selected`?

src/object/path_gameobject.cpp
```diff
@@ -49,26 +48,30 @@ PathStyle PathStyle_from_string(const std::string& text)
 } // namespace
 
 PathGameObject::PathGameObject() :
-  m_path(new Path),
+  m_path(new Path(*this)),
   m_style(PathStyle::NONE),
   m_edge_sprite(),
   m_node_sprite()
 {
+  m_track_undo = false;
```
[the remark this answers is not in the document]
Vankata453: `m_track_undo` is a member from `GameObject`, so I cannot.

PR #2321, opened 2022-10-19
Fix boost date time to 1.79 for wasm

tobbi:


Semphriss:
It would be good to consider if we should put all Boost libraries to the same version, to ensure proper compatibility.

tobbi:
My bad. Versioning only works in manifest mode:
https://github.com/microsoft/vcpkg/blob/master/docs/users/versioning.md

PR #2333, opened 2022-11-16
Tux and Penny Design Revamp

Rusty-Box:
After many years of SuperTux it is time to give our main character a desperatly needed glow up!

**This PR contains:**
- Fully re-drawn and re-animated sprites for both Small and Big Tux
- Many new animation for newly added mechanics (i.e. sliding, crawling etc.)
- A couple properly animated sprites for Penny

**List of coding changes required:**

- [x] Big + Small Tux - must call for the new **boost** action for swim boosting underwater
- [x] Big + Small Tux - must call for the new **slide** action for sliding
- [x] Big + Small Tux - must call for the new **slidejump** action for jumping while sliding
- [x] Big + Small Tux - call new **scratch** action before **idle** action and after **stand** action
_(Note: idle action should play endless until Tux moves again instead of reverting back to stand action mid-way!)_
- [x] Big + Small Tux - **climbing**, **floating**, **swimming** have been renamed to **climb**, **float**, **swim** in the `.sprite` file
---
- [x] Small Tux - remove sprite rotation for slidejump
- [x] Small Tux - must call for the new **slidegrow** action for getting an Egg powerup while sliding
- [x] Small Tux - must call for the new **climbgrow** action for getting an Egg powerup while climbing
_(Note: Tux should also not let go of climb-ables while getting a powerup!)_
---
- [x] Big Tux - must call for the new **stomp** action after touching the ground during a buttjump
- [x] Big Tux - remove sprite rotation for backflip
- [x] Big Tux - remove sprite rotation for slidejump
- [x] Big Tux - must call for the new **crawl** action for crawling

**_Note: Everything listed under `Big Tux` and `Big + Small Tux` also applies to Fire Tux, Ice Tux, Earth Tux and Air Tux!_**

Known bugs that need fixing:

1. [x] Small Tux's walk animation doesn't seem to load properly when in motion. It seems to work however when standing still first and during the main menu.

https://user-images.githubusercontent.com/20615048/202192920-a6ff5f17-2fef-4c19-aa6a-8fab53e8cab2.mp4

2. [x] Some animations (e.g. Big Tux idle animation) occasionally don't load immediately causing the previous animation to freeze temporarily instead of transitioning to the next animation immediately.

https://user-images.githubusercontent.com/20615048/202193431-1e8b9396-9d87-4b0a-9916-f9e480e1363f.mp4

---

[...]

PR #2338, opened 2022-11-28
Player behavior improvements

### LINKED ISSUE #1399: Earth flowers are underpowered
The earth flower has an ability (in the dark) to use a flashlight so you can see where you're going. So, in the rare case that you're actually in the dark (which i think is only like.. what, 1 level in the story mode so far?) it shines a poor, measly light in front of tux. This light should be bigger i think.

Also, his stone ability, while it is cool, i think it would be cooler if his stone ability maybe turned him into poison ivy or immune to ivy of some sort, and even maybe kill enemies when they touch him as he turns to stone. Would be cool. Also instead of freezing him maybe make him move slowly? or able to use his ability mid air? Some ideas.

I also think maybe he should have the ability to break tiles in some way, like tux can pull out a hatchet and start chopping at those wood blocks that act vertically, that would be cool.

### LINKED ISSUE #1901: Swimming velocity seems to override object-influenced velocity
I have noticed that when swimboosting, the power of the swimboost seems to override wind and trampolines, and as a result, Tux can't be affected by either while swimboosting.  I think that these velocity setters should simply add on to Tux's velocity already instead of setting it (which I think it currently does?), which might help them work better together.

### LINKED ISSUE #2260: Rock tux's idle position + crouch position appear when trying to stand up under stationary object
SuperTux version:| 0.6.3

Expected behavior:| Earth tux shouldn't be able to solidify under a stationary object that is 1 position above a tile. Instead, tux should only be able to crouch and move around while under a stationary object that is 1 position above a tile.

Actual behavior:| Earth tux can solidify under a stationary object and show 2 different positions with both being inside of each other, one being crouching with the earth hat and the other position being idle standing without the earth hat, which looks strange. 

Steps to reproduce actual behavior:| Place any stationary object 1 position above any tile and test the level using earth tux, then crouch under the stationary object as earth tux and solidify.

Additional debugging information:| video

https://user-images.githubusercontent.com/83178674/180323729-fa441d85-dfd5-4869-adf6-c0ef541bd389.mp4

### LINKED ISSUE #2335: Tux's sprite glitches when ducking after turning
**SuperTux version: 0.6.3-653-g13c8be990**
**System information:Windows 10** 

##### Expected behavior
Tux sprite changes correctly after pressing down key
##### Actual behavior
The sprite appears to be down a tile below tux's original position, with no damage to him
Press right or left key until Tux is running, then move to the opposite direction and right after that, press down. This bug can be performed more than once in a single turn.

https://user-images.githubusercontent.com/92935956/202869387-acc718df-673e-42cd-85e5-01cfa1c235ca.mp4

weluvgoatz:
This PR aims to provide multiple improvements to the Player object in order to finalize the players' behavior for the most part.  At the time of this first draft, I have included multiple features:

- Replaces Earth Tux’s ability with a controllable rolling rock (continuation of https://github.com/SuperTux/supertux/pull/1820)
- Glide time for Air Tux is infinite to stay in line with Earth Tux
- Made Air Tux’s trampoline bounce less crazy
- Made ghost mode a bit more like swimming (move using dir keys, boost using jump)
- Remove unneeded “m_shooting_timer” variable
- Fixes #1901 
- Fixes #2335
- Fixes #2260 (by making it redundant)
- Fixes #1399 (in a way?)

Also, I am aware that this may cause conflicts with the Tux resprite PR (https://github.com/SuperTux/supertux/pull/2333), so if that one is merged first, then adjustments will have to be made here!

A possibility for me to implement:
- [x] Stronger buttjumping!


Rusty-Box:
**Feedback to recent changes:**

- Rock Rolling doesn't revert to swimming immideatly when falling in water as a rock. That should be changed!
- When coming to a stop, to prevent easy access to Rock Rolling, Tux should revert back to ducking, so players need to get a headstart again to access Rock Rolling once more.
- I also believe that Rock Rolling needs a weaker "max. speed", atleast in terms of what the speed you gain from pressing left/right while Rock Rolling. You could gain more speed through inclines but friction should lower it till you're back at "max. speed".

Rusty-Box:
Another thing: Perhaps we don't want players to turn into a rock when pressing down while jumping as this prevents buttjumping!

Rusty-Box:
Final suggestion from my end:
The stronger buttjump seems a tad out of sync with the actual animation. Instead of having it based on a timer (from what I can tell), the strong push down should occour once the inital animation of the "buttjump" action has ended. That should give it a better flow with the visuals.

Rusty-Box:
Just something I thought of regarding swimming: I think it would be better if Tux eases into his max speed when boosting rather than instantaniously boosting along. Similar thing with him turning while boosting. Too harsh, should be more of "smooth" turn rather than a harsh zick zack

> Final suggestion from my end

So much on that statement :D

PR #2342, opened 2022-11-30
Badguy behavior improvements

weluvgoatz:
There is still much to-do, but this PR aims to drastically improve the behavior of badguys as a whole (minus bosses!)

Tasks this PR aims to complete:
- [x] Redesign Ghoul - now he moves more like Bloopers from Super Mario, can be frozen properly, and glows 
- [x] Snails can be thrown properly and do not get stuck
- [x] Bouncing snowballs bounce along walls properly
- [x] Enemies only disappear when their whole body is out of the sector
- [x] Igel's behavior should be redone to be the following: strolls at a normal pace, but rolls up in a ball to chase the player when within range.
- [x] Spidermite's movement should be changed to be more in-line with their upcoming new sprite, where it lunges at the player and bounces for a while.
- [x] Zeekling's behavior should be redesigned to be more fair, and its AI could possibly be made simpler
- [x] Enemies should interact with bumpers, trampolines, and water
- [x] Particles upon squishing enemies
- [x] Fix multiple squished-sprite issues
- [x] Add the fire death overlay for enemies
- [x] Improve explosion/icecrusher/generic particles
- [x] Remove randomness to enemies that have unnecessary randomness
Todo:
- [x] Update igel behavior slightly
- [x] More enemy interactions with environment, such as kamikaze snowballs/dart traps affecting bonus blocks, etc.

This PR might require graphical updates! Stay tuned...


Rusty-Box:
Could you also fix Bouncing Snowball only having a left facing squished action, while you're at it?

Rusty-Box:
**Feedback Time**

1) I noticed that in some instances there are to particles playin at once. For example when buttjumping a Mr. IceBlock it should only show the "ice" particles, not the "star" particles.
2) I don't like the fact that Igel seems to roll indefinitely. It feels especially odd when they fall on a lower platform yet still chase you non-stop even though you are no longer in range.
3) Why do bombs when squished show a bigger particle compared to other enemies? It looks weird. Way too big. Looks more like a rock particle.
4) The fading time on particles could be much longer. Right now (especially) the star particle feels way too static. I think all particles could do with some much stronger easing. Just a thought.
5) Using a fireball on enemies should get a particle sprite that looks more like smoke (I could provide that).
6) Spidermite needs a drop distance setting rather than a bounce distance setting so you can place them on various heights and adjust the distance they need to fall until in range to attack Tux. They drop down and then begin to bounce up and down for until Tux leaves/is out of range after which they slowly climb back up.
7) Enemies like Kamikaze Snowball and Leafshot's particles have the wrong motion when hitting a wall. They should behave like Mr. IceBlocks when slamming into walls!
8) I think mole actually needs gravity.

That's all for now. A lot but it is a big PR.

weluvgoatz:
This PR will be closed due to an enormous, annoyingly large number of conflicts, but is definitely still to be worked on.  It will possibly be split into multiple PRs when I come back to it.

PR #2346, opened 2022-12-02
Object behavior improvements

### LINKED ISSUE #1782: Implement objects sticking to solids
In the recent improvements to moving objects, the movement from the moving object is applied to those who stand on it. It would be interesting to add a similar thing but for the sides and bottom of the moving object.  This would only apply to objects with a set "sticky" attribute, and would simulate sticking objects to things.
  This could apply to dart traps, bumpers, levers, the upcoming crystal shards, and other objects who currently do not stick to the sides of solids and instead simply stand by themselves, which produces unexpected behavior.

weluvgoatz:
Includes the following:
- [x] Rocks now bounce on trampolines and bumpers, as well as lessen gravity underwater and slide on slick ice.
- [x] Refactor and fix bugs for fallblock
- [x] Rocks should not crush Tux
- [x] Wind particles should appear properly
- [x] New option for script triggers, that has a script that plays when Tux touches the trigger and leaves the trigger
- [x] Icicle should not go behind switch
- [x] Improve wind
- [x] Add stop_sound() function
Fixes #1782


weluvgoatz:
I will come back to this at a later time.

PR #2349, opened 2022-12-04
Display world screenshot previews and progress

Vankata453:
`WorldSetMenu` and `SortedContribMenu` now inherit the new `WorldPreviewMenu` class, which allows displaying profile-specific world previews and progress. A preview for a world is generated each time the user leaves its worldmap. Only worldmap levelsets support this feature.
Supported also are optional placeholder world previews, which have to be named `preview.png`, and stored in the relative to `levels` directory for the world.

image
image

### DIFF
--- a/src/gui/menu.cpp
+++ b/src/gui/menu.cpp
@@ -675,6 +675,8 @@ Menu::draw(DrawingContext& context)
                               Vector(m_pos.x, static_cast<float>(SCREEN_HEIGHT) - 48.0f - static_cast<float>(text_height) / 2.0f),
                               ALIGN_CENTER, LAYER_GUI);
   }
+
+  draw_additional(context);
 }
 
 MenuItem&
--- a/src/supertux/menu/sorted_contrib_menu.hpp
+++ b/src/supertux/menu/sorted_contrib_menu.hpp
@@ -1,5 +1,6 @@
 //  SuperTux
 //  Copyright (C) 2021 mrkubax10 <mrkubax10@onet.pl>
+//                2022 Vankata453
 //
 //  This program is free software: you can redistribute it and/or modify
 //  it under the terms of the GNU General Public License as published by
@@ -16,18 +17,20 @@
 
 #ifndef HEADER_SUPERTUX_SUPERTUX_MENU_SORTED_CONTRIB_MENU_HPP
 #define HEADER_SUPERTUX_SUPERTUX_MENU_SORTED_CONTRIB_MENU_HPP
-#include "gui/menu.hpp"
-#include "supertux/world.hpp"
-class SortedContribMenu : public Menu
+
+#include "supertux/menu/world_preview_menu.hpp"
+
+class SortedContribMenu final : public WorldPreviewMenu
 {
 public:
-  SortedContribMenu(std::vector<std::unique_ptr<World>>& worlds, const std::string& contrib_type, const std::string& title, const std::string& empty_message);
-  void menu_action(MenuItem& item) override;
-private:
-  std::vector<std::string> m_world_folders;
+  SortedContribMenu(std::vector<std::unique_ptr<World>>& worlds, const std::string& contrib_type,
+                    const std::string& title, const std::string& empty_message);
+
 private:
   SortedContribMenu(const SortedContribMenu&) = delete;
   SortedContribMenu& operator=(const SortedContribMenu&) = delete;
 };
+
 #endif
+
 /* EOF */
--- a/src/supertux/menu/world_preview_menu.cpp
+++ b/src/supertux/menu/world_preview_menu.cpp
@@ -1,0 +1,164 @@
+//  SuperTux
+//  Copyright (C) 2022 Vankata453
+//
+//  This program is free software: you can redistribute it and/or modify
+//  it under the terms of the GNU General Public License as published by
+//  the Free Software Foundation, either version 3 of the License, or
+//  (at your option) any later version.
+//
+//  This program is distributed in the hope that it will be useful,
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//  GNU General Public License for more details.
+//
+//  You should have received a copy of the GNU General Public License
+//  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#include "supertux/menu/world_preview_menu.hpp"
+
+#include <physfs.h>
+
+#include "gui/menu_manager.hpp"
+#include "gui/menu_item.hpp"
+#include "supertux/game_manager.hpp"
+#include "supertux/gameconfig.hpp"
+#include "supertux/globals.hpp"
+#include "supertux/resources.hpp"
+#include "supertux/menu/contrib_levelset_menu.hpp"
+#include "util/file_system.hpp"
+#include "util/gettext.hpp"
+#include "video/drawing_context.hpp"
+#include "video/surface.hpp"
+#include "video/video_system.hpp"
+#include "video/viewport.hpp"
+
+const Sizef WorldPreviewMenu::s_preview_size(426.f, 240.f);
+const float WorldPreviewMenu::s_preview_fade_time = 0.1f;
+
+WorldPreviewMenu::WorldPreviewMenu(float center_x_offset, float center_y_offset) :
+  m_world_entries(),
+  m_last_world_index(-1),
+  m_preview_fade_timer(),
+  m_preview_fade_active(false),
+  m_preview_fading_out(false)
+{
+  // Adjust center position to give space for displaying previews.
+  set_center_pos(static_cast<float>(SCREEN_WIDTH) / center_x_offset,
+                 static_cast<float>(SCREEN_HEIGHT) / center_y_offset);
+
+  m_preview_fade_timer.start(s_preview_fade_time);
+}
+
+SurfacePtr
+WorldPreviewMenu::find_preview(const std::string& preview_file, const std::string& basedir)
+{
+  std::string preview_path = FileSystem::join("profile" + std::to_string(g_config->profile), preview_file);
+  if (PHYSFS_exists(preview_path.c_str())) // A preview exists.
+  {
+    return Surface::from_file(preview_path);
+  }
+  else // Preview doesn't exist, so check for a placeholder.
+  {
+    preview_path = FileSystem::join(basedir, "preview.png");
+    if (PHYSFS_exists(preview_path.c_str())) // A preview placeholder exists.
+      return Surface::from_file(preview_path);
+  }
+
+  return nullptr;
+}
+
+void
+WorldPreviewMenu::draw_additional(DrawingContext& context)
+{
+  const int index = get_active_item_id();
+  bool valid_last_index = is_valid_index();
+
+  // Update fade.
+  if (index != m_last_world_index && !m_preview_fade_active) // Index has changed, there is no current fade.
+  {
+    if (valid_last_index) // Fade out only if the last index is valid.
+      m_preview_fade_timer.start(s_preview_fade_time);
+    m_preview_fading_out = true;
+    m_preview_fade_active = true;
+  }
+  float timeleft = m_preview_fade_timer.get_timeleft();
+  if (timeleft < 0 && m_preview_fade_active) // Current fade is over.
+  {
+    m_last_world_index = index;
+    valid_last_index = is_valid_index(); // Repeat valid last index check
+    if (m_preview_fading_out) // After a fade-out, a fade-in should follow up.
+    {
+      m_preview_fade_timer.start(s_preview_fade_time);
+      timeleft = m_preview_fade_timer.get_timeleft();
+      m_preview_fading_out = false;
+    }
+    else
+    {
+      m_preview_fade_active = false;
+    }
+  }
+
+  // Set alpha according to fade.
+  float alpha = 1.f;
+  if (timeleft > 0)
+  {
+    const float alpha_val = timeleft * (1.f / s_preview_fade_time);
+    alpha = m_preview_fading_out ? alpha_val : 1.f - alpha_val;
+  }
+  const Color color_white(1, 1, 1, alpha);
+
+  // Perform actions only if current index is a valid world index.
+  if (valid_last_index)
+  {
+    // Draw progress preview of current world.
+    Rectf preview_rect(Vector(context.get_width() * 0.73f - s_preview_size.width / 2,
+                            context.get_height() / 2 - s_preview_size.height / 2), s_preview_size);
+    if (m_world_entries[m_last_world_index].preview)
+    {
+      PaintStyle style;
+      style.set_alpha(alpha);
+      context.color().draw_surface_scaled(m_world_entries[m_last_world_index].preview, preview_rect, LAYER_GUI, style);
+    }
+    else // If preview is not available, draw a filled rect with text in its place.
+    {
+      context.color().draw_filled_rect(preview_rect, Color(0, 0, 0, alpha), LAYER_GUI);
+      context.color().draw_text(Resources::normal_font, _("No preview available."),
+                                preview_rect.get_middle(), ALIGN_CENTER, LAYER_GUI, color_white);
+    }
+
+    // Draw world progress.
+    const Savegame::Progress& progress = m_world_entries[m_last_world_index].progress;
+    if (progress.progress > -2) // Progress should be drawn.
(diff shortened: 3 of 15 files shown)


mrkubax10 (approved):

src/supertux/savegame.hpp
```diff
@@ -65,6 +66,24 @@ struct WorldmapState
...
+  public:
+    Progress() :
+      solved(),
+      total()
+    {}
+    Progress(int solved_, int total_) :
+      solved(solved_),
+      total(total_)
+    {}
+    uint32_t solved;
```
tobbi, line 81: m_solved
m_total

tobbi:
/home/runner/work/supertux/supertux/src/worldmap/worldmap_state.cpp:139:30: error: empty expression statement has no effect; remove unnecessary ';' to silence this warning [-Werror,-Wextra-semi-stmt]
  WORLDMAP_STATE_SECTOR_GUARD;
                             ^
/home/runner/work/supertux/supertux/src/worldmap/worldmap_state.cpp:172:30: error: empty expression statement has no effect; remove unnecessary ';' to silence this warning [-Werror,-Wextra-semi-stmt]
  WORLDMAP_STATE_SECTOR_GUARD;
                             ^
/home/runner/work/supertux/supertux/src/worldmap/worldmap_state.cpp:200:30: error: empty expression statement has no effect; remove unnecessary ';' to silence this warning [-Werror,-Wextra-semi-stmt]
  WORLDMAP_STATE_SECTOR_GUARD;
                             ^
/home/runner/work/supertux/supertux/src/worldmap/worldmap_state.cpp:256:30: error: empty expression statement has no effect; remove unnecessary ';' to silence this warning [-Werror,-Wextra-semi-stmt]
  WORLDMAP_STATE_SECTOR_GUARD;
                             ^
/home/runner/work/supertux/supertux/src/worldmap/worldmap_state.cpp:347:30: error: empty expression statement has no effect; remove unnecessary ';' to silence this warning [-Werror,-Wextra-semi-stmt]
  WORLDMAP_STATE_SECTOR_GUARD;
                             ^
/home/runner/work/supertux/supertux/src/worldmap/worldmap_state.cpp:361:30: error: empty expression statement has no effect; remove unnecessary ';' to silence this warning [-Werror,-Wextra-semi-stmt]
  WORLDMAP_STATE_SECTOR_GUARD;
                             ^
/home/runner/work/supertux/supertux/src/worldmap/worldmap_state.cpp:382:30: error: empty expression statement has no effect; remove unnecessary ';' to silence this warning [-Werror,-Wextra-semi-stmt]
  WORLDMAP_STATE_SECTOR_GUARD;
                             ^
/home/runner/work/supertux/supertux/src/worldmap/worldmap_state.cpp:410:30: error: empty expression statement has no effect; remove unnecessary ';' to silence this warning [-Werror,-Wextra-semi-stmt]
  WORLDMAP_STATE_SECTOR_GUARD;

src/gui/menu_filesystem.cpp
```diff
@@ -62,7 +65,13 @@ FileSystemMenu::refresh_items()
   m_files.clear();
   m_directory = FileSystem::normalize(m_directory);
 
-  add_label(m_directory);
+  // Make sure label doesn't get too long.
+  std::string title = m_directory;
+  const bool title_large = title.size() > s_title_max_chars;
+  while (title.size() > s_title_max_chars) title = title.substr(title.size() - s_title_max_chars);
```
mrkubax10, line 71: Maybe it would be better to do:
```cpp
if(title.size() > s_title_max_chars)
  title = title.substr(s_title_max_chars-3)+"...";
```
? It will show beginning of title instead.

Vankata453:
This PR is now ready for review.

src/supertux/savegame.hpp
Vankata453: I didn't think an `m_` prefix is needed here, since this struct is only used to store 2 variables together and has just one simple function to calculate percentage using them.

src/gui/menu.cpp
```diff
@@ -44,15 +44,18 @@
...
 #include "supertux/globals.hpp"
 #include "supertux/resources.hpp"
 #include "video/drawing_context.hpp"
 #include "video/renderer.hpp"
+#include "video/surface.hpp"
 #include "video/video_system.hpp"
 #include "video/viewport.hpp"
 
-#include "supertux/error_handler.hpp"
+const Sizef Menu::s_preview_size(426.f, 240.f);
```
MatusGuy, line 57: Hardcoding the preview size squishes pictures that are too big. Any reason as to why you're doing this?

MatusGuy (changes requested):
Is completing all levels normally supposed to count as 100%? What about getting all collectibles?

On small screens, the help text is not fully readable because it's not aligned to the screen but rather to the menu.
image

I don't see any animations when navigating a Filesystem menu.

Aside from that, nice work!

Vankata453:
> Is completing all levels normally supposed to count as 100%? What about getting all collectibles?

That is only supposed to indicate the progress through levels on the map. If the percentage isn't perceived well, I can remove it.

> On small screens, the help text is not fully readable because it's not aligned to the screen but rather to the menu.

True, it should stay centered regardless.

> I don't see any animations when navigating a Filesystem menu.

Do you have transitions enabled?

src/gui/menu.cpp
Vankata453: So the image and the menu could both fit on the screen. Maybe I should check the resolution before determining this though.

MatusGuy:
> Do you have transitions enabled?

Yes, but now that I think about it, this might be because the loading time overrides the animation. That can be something for another pr. Will do some further investigation later.

The preview still gets squished. I cloned this PR and am currently trying to write a fix for this

Vankata453:
> this might be because the loading time overrides the animation

Yes, if you go to the next item too quickly, there will be no time for animations.

> The preview still gets squished.

It does, but only if you change your resolution to be different from the one when the screenshot was taken.

src/gui/menu.cpp
```diff
@@ -617,9 +642,88 @@ Menu::draw(DrawingContext& context)
...
+  if (valid_last_index)
+  {
+    // Draw progress preview of current item.
+    const Sizef preview_size(static_cast<float>(SCREEN_WIDTH) / 2.5f, static_cast<float>(SCREEN_HEIGHT) / 2.5f);
+    SurfacePtr preview = m_items[m_last_preview_item]->get_preview();
+    const float width_diff = preview_size.width - static_cast<float>(preview->get_width());
+    const float height_diff = preview_size.height - static_cast<float>(preview->get_height());
+    // If the preview is smaller than the maximal size, make sure to draw it with its original size and adjust position to center.
+    Rectf preview_rect(Vector(static_cast<float>(context.get_width()) * 0.73f - preview_size.width / 2 + (width_diff > 0 ? width_diff / 2 : 0),
+                              static_cast<float>(context.get_height()) / 2 - preview_size.height / 2 + (height_diff > 0 ? height_diff / 2 : 0)),
```
MatusGuy, line 700: Use `std::max`
```suggestion
    Rectf preview_rect(Vector(static_cast<float>(context.get_width()) * 0.73f - preview_size.width / 2 + std::max(width_diff / 2, 0),
                              static_cast<float>(context.get_height()) / 2 - preview_size.height / 2 + std::max(height_diff / 2, 0)),
```

src/gui/menu.cpp
```diff
@@ -617,9 +642,88 @@ Menu::draw(DrawingContext& context)
...
+  {
+    const float alpha_val = timeleft * (1.f / s_preview_fade_time);
+    alpha = m_preview_fading_out ? alpha_val : 1.f - alpha_val;
+  }
+
+  // Perform actions only if current index is a valid preview index.
+  if (valid_last_index)
+  {
+    // Draw progress preview of current item.
+    const Sizef preview_size(static_cast<float>(SCREEN_WIDTH) / 2.5f, static_cast<float>(SCREEN_HEIGHT) / 2.5f);
```
MatusGuy, line 694: Use the context functions.
```suggestion
    const Sizef preview_size(static_cast<float>(context.get_width()) / 2.5f, static_cast<float>(context.get_height()) / 2.5f);
```

MatusGuy:
Hi back, I'm dad.

If I remember correctly, I remember that this PR didn't show the right aspect ratio for the source image when the screen resolution was smaller than the source picture... Something along those lines.

MatusGuy (approved):

weluvgoatz (approved):
This feature works well and I am satisfied with the current implementation.  One suggestion I have, though, is to make it so that the screenshot gets taken before the player presses the escape key, so that every screenshot doesn't have the escape menu in it.  For example, the player presses escape -> take screenshot -> show escape menu.

Vankata453:
It shows the menu? It's supposed to hide whenever the screenshot is being taken.

Frostwithasideofsalt:
image


my one thing with this pr is how it displays the stats when theirs no preview image. it's really ugly, unclear and overwhelming. it would be much preferred if the stats just showed up like this if there wasnt any image to go along with it

Frostwithasideofsalt:
image

because of that it also tends to send stuff off the screen. it also seems to show the wrong level count for some reason? like in this image it says the bonus island remake has *60 levels* when it only has 30. what i think whats happening is that it's counting the number of files in the level folder and using that for the level count, even if it's not a level file ...? im not too sure. but the number of levels it reports has so far been equal to the amount of files in the level folder. so yeah.

Frostwithasideofsalt (changes requested):
Even if that isn't the cause- there's for sure something inflating the level count by a lot. so the things im suggesting are:
-make level stats for a world without a screenshot display similarly to how their displayed when their is a screenshot
-fix the bug inflating level counts

swagtoy:
> because of that it also tends to send stuff off the screen.

I dont think there would be a genuine way to fix this without refactoring the menus a bit using a scrollbar system of some sort. iirc this behavior is pretty poorly baked in. Best hack of a fix for now would be truncating long lines if we just wanna ship it as is.

Frostwithasideofsalt:
whats the status on this addon?

Frostwithasideofsalt:
addon? lol. i mean pr

Alasdairbugs:
does this consider cutscene levels and not including those in the count?

Hume2:
Can we just remove the level counter, so we can finally get past this? There is currently no level counter, so introducing it isn't really necessary.

Vankata453:
I believe the level counter is a good addition to have. Also it's not the main issue preventing this PR right now. I'll be taking a look at it soon anyways.

Frostwithasideofsalt:
builds failing, cant test

tobbi:
Needs another rebase against master

Frostwithasideofsalt (approved):
All of my issues have been resolved.

provigz:
Only issue I see left with this PR is that the worldmap menu is shown in the world screenshot when using the OpenGL renderer on Windows. Considering the default renderer now is SDL, maybe we should merge this and maybe consider this issue for another time, as it's minor and likely directly related to the OpenGL video code.

The OpenGL bug in question

bruhmoent:
I'd say that if it ONLY happens with the OpenGL renderer, I'd let it slip by for now.

provigz:
From my testing, SDL was fine. I encourage you to test it.

tobbi:
Instead of "perfected", how about "perfect"?

bruhmoent (changes requested):
image
image

First image steps: Start game -> Contrib Levels -> Community Levels -> Back -> Back -> Start menu shifted
And when there are no add-ons installed, the contrib levels menu shouldn't shift

provigz:
Isn't it shifted in the first screenshot because of Story Mode?

bruhmoent:
There isn't a preview visible, as shown in the screenshot

provigz:
And there is no progression or anything from Story Mode that shows up on the right? Then sure.

Alasdairbugs:
"Perfected" is the proper english term for perfecting a level, so that should stay @tobbi

tobbi:
> "Perfected" is the proper english term for perfecting a level, so that should stay @tobbi

Ah, alright!

provigz:
This PR should be ready (review if you could), however I am getting a segfault/signal 6 when making a screenshot on the SDL renderer.

It might be a bug within `SDL_RenderReadPixels` on SDL2-compat. I'm also getting the following error when `SDLPainter::get_pixel()` runs, which also uses the function:
```
src/video/sdl/sdl_painter.cpp:635 failed to read pixels: Can't read outside the current viewport
```

src/gui/menu_filesystem.cpp
```diff
@@ -28,6 +28,9 @@
 #include "util/gettext.hpp"
 #include "util/string_util.hpp"
 
+static const size_t MAX_TITLE_CHARS = 30;
+static const std::vector<std::string> IMAGE_EXTENSIONS = { ".jpg", ".png", ".surface" };
```
swagtoy, line 32: std array or just a C style array. Try std stringview

src/gui/menu_filesystem.cpp
```diff
@@ -132,12 +145,20 @@ FileSystemMenu::has_right_suffix(const std::string& file) const
...
-    {
       return true;
-    }
-  }
+
+  return false;
+}
+
+bool
+FileSystemMenu::is_image(const std::string& file) const
```
swagtoy, line 156: smells like unsound logic anyway. sdl image might have something to check this? try running it through and seeing if a valid surface is returned if it doesnt.

swagtoy: granted, thats kinda rare. plus, what about .jpeg and whatever? maybe im looking at this wrong

src/gui/menu_item.cpp
```diff
@@ -50,6 +52,12 @@ MenuItem::set_help(const std::string& help_text)
   }
 }
 
+void
+MenuItem::set_preview(const std::string& preview_file)
+{
+  m_preview = Surface::from_file(preview_file);
```
swagtoy, line 58: error check

src/supertux/menu/options_menu.cpp
```diff
@@ -196,6 +196,9 @@ OptionsMenu::refresh()
       add_toggle(MNID_TRANSITIONS, _("Enable transitions"), &g_config->transitions_enabled)
         .set_help(_("Enable screen transitions and smooth menu animation"));
 
+      add_toggle(MNID_WORLD_PREVIEWS, _("Show world previews"), &g_config->show_world_previews)
+        .set_help(_("Show screenshot previews of the last worldmap state, when hovering over a world."));
```
swagtoy, line 200: remove comma

src/supertux/menu/sorted_contrib_menu.hpp
```diff
@@ -1,5 +1,6 @@
 //  SuperTux
 //  Copyright (C) 2021 mrkubax10 <mrkubax10@onet.pl>
+//                2022-2023 Vankata453
```
swagtoy, line 3: bump, change name to yours

src/supertux/menu/world_preview_menu.cpp
```diff
@@ -0,0 +1,121 @@
...
+    return false;
+
+  return !m_world_entries[index].progress_text.empty();
+}
+
+void
+WorldPreviewMenu::menu_action(MenuItem& item)
+{
+  const int index = item.get_id();
+  if (index >= 0 && index < static_cast<int>(m_world_entries.size())) // Valid world index.
```
swagtoy, line 107: flip logic and early return

src/supertux/savegame.cpp
```diff
@@ -102,6 +102,12 @@ LevelsetState::get_level_state(const std::string& filename) const
   }
 }
 
+uint32_t
+Savegame::Progress::get_percentage() const
```
swagtoy, line 106: why not int? also its a percentage, why not double?

src/supertux/savegame.hpp
```diff
@@ -68,6 +74,20 @@ struct WorldmapState
...
+  struct Progress final
+  {
+    Progress() :
+      solved(),
+      perfect(),
+      total()
+    {}
+    uint32_t solved;
+    uint32_t perfect;
+    uint32_t total;
```
swagtoy, line 86: just use unsigned

src/worldmap/worldmap.cpp
```diff
@@ -228,6 +281,36 @@ WorldMap::get_status() const
 }
 
 
+void
+WorldMap::take_preview_screenshot()
```
swagtoy, line 285: show a dialog and also log SDL_GetError (if applicable(

swagtoy:
We could probably use getpixel api instead of sdl screenshot. dunno how thatd work on the gl backend

PR #2354, opened 2022-12-09
Add keys and locked doors

weluvgoatz:
This PR adds keys, which can unlock the new locked doors.  Doors can also be locked.  This feature has been planned for quite a while, just lacking the time to actually implement it properly.  It allows the level designers to create new puzzles revolving around doors for the Rooted Forest.


Rusty-Box:
This works incredibly well and I only have a few minor complaints/nitpicks.

1. The particles when collecting a key are too "harsh". They disappear way too abrupt as well as losing speed way too fast after the inital burst.
2. Since they copy Tux's movement 1:1 it tends to look a bit off sometimes making me think that having them chase Tux more direct would be better

Outside of thsoe 2 things this is really cool and honestly for this PR to get merged my second point is less important. But the "collect key" particles do need some tweaking.

Rusty-Box (approved):
Particles feel much better now!

PR #2355, opened 2022-12-12
Add an option for crisp graphics

### LINKED ISSUE #1112: Sprites appear blurry
**SuperTux version: 0.6.0**
**System information:** *(details about your kernel, operating system, or distribution)*

Sprites seem to appear blurry. Do sprites render in between pixels or something, as sometimes they appear blurry. I'd prefer if they always rendered at a pixel precise location. Also iceblock has a border sometimes, which I think is caused by this. 
image

weluvgoatz:
Fixes #1112 by including the GL_NEAREST parameter for rendering, however this can appear poor on different magnifications because it renders to the nearest pixel, so I included the option to render either more "crisp" or with the more blurry (current) look.  Like other graphical options, it requires a restart to take effect.

I have not tested this against the entire game so it is possible something breaks... Let me know if that is the case.

### DIFF
--- a/src/supertux/menu/options_menu.cpp
+++ b/src/supertux/menu/options_menu.cpp
@@ -71,6 +71,7 @@ enum OptionsMenuIDs {
   MNID_RUMBLING,
   MNID_DEVELOPER_MODE,
   MNID_CHRISTMAS_MODE,
+  MNID_CRISP_GRAPHICS,
   MNID_TRANSITIONS,
   MNID_CONFIRMATION_DIALOG,
   MNID_PAUSE_ON_FOCUSLOSS,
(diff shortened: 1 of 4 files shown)


Semphriss:
The proper term for "crisp graphic" is [Nearest-pixel interpolation](https://en.wikipedia.org/wiki/Nearest-neighbor_interpolation). The "blurry" graphics is [linear interpolation](https://en.wikipedia.org/wiki/Linear_interpolation).

The difference is best shown by this image from [this website](https://learnopengl.com/Getting-started/Textures) under "Texture filtering":
image

The former ("nearest") gives a pixelized effect, whereas the latter ("linear") tries to smooth those edges out, making it less pixelized but blurry.

Note that the effect will be larger than this reference image in practice: slower movements will feel significantly more "blocky" with nearest interpolation, and certain scales close to 1 will duplicate certain rows of pixels, which stand out a lot.

If the end goal is to have a better image quality, instead of using nearest sampling, the solution would be to __make source images at higher resolution__ (2-4x larger), so that OpenGL has more room to pick a precise color.

weluvgoatz:
Thank you for the additional information, "crisp" is just a general term to refer to the GL_NEAREST since a lot of users might not know what that means.

mrkubax10 (approved):

src/supertux/menu/options_menu.cpp
```diff
@@ -454,6 +455,9 @@ OptionsMenu::OptionsMenu(bool complete) :
     add_toggle(MNID_CHRISTMAS_MODE, _("Christmas Mode"), &g_config->christmas_mode);
   }
 
+  if (g_config->video != VideoSystem::Enum::VIDEO_SDL)
```
Vankata453, line 458: There could be an overridable property for classes that inherit `VideoSystem`, which determines if the system supports crisp graphics, to be used for checks like this one.

mrkubax10:
@weluvgoatz Not sure if you will continue working on this PR.

weluvgoatz:
I will work on this at a later time.

PR #2357, opened 2022-12-14
Resource packs, add-on dependencies, "TransferStatusList"-s

Vankata453:
Adds a new add-on variant - resource packs. Resource packs are different from other add-on types, since they are allowed to override game files. This allows modifying images, sounds, music and more. Some essential game folders have been restricted from being overridable by resource packs and the user directory.

Along with that, add-on dependencies are also included. When an add-on, which has dependencies, is installed, its dependencies would be installed as well, if they are not already and are available. When uninstalling an add-on, on which other add-ons depend, a warning message would be thrown to indicate that.

A new `TransferStatusList` class is implemented, which contains a vector of `TransferStatus` pointers. All `TranferStatus`-es are updated and tracked by their parent `TransferStatusList`. A `TranferStatusList` executes the provided callback (with `success` = `false`), if any `TransferStatus` has failed, and executes the callback (with `success` = `true`), if all `TransferStatus`-es have finished successfully. `DownloadDialog` now supports and operates using `TranferStatusList`-s.


weluvgoatz:
Could this also possibly include pre-installed dependencies/resource packs, such as the old tiles from 0.6.3 and 0.6.0 as addons?  That way users can use these resource packs without having to install anything else.  It would prove the best solution to that tile ID fiasco, in my opinion.

Rusty-Box:
What is the status on this? Can this be merged?

mrkubax10:
> What is the status on this? Can this be merged?

First compilation errors have to be fixed.

src/supertux/menu/addon_preview_menu.cpp
```diff
@@ -209,12 +238,26 @@ AddonPreviewMenu::menu_action(MenuItem& item)
...
+      const std::vector<std::string> depending_addons = m_addon_manager.get_depending_addons(m_addon.get_id());
+      if (depending_addons.empty())
+      {
+        uninstall_addon();
+      }
+      else
+      {
+        // Other add-ons depend on the add-on that's being uninstalled.
+        const std::string dependency_message = fmt::format(fmt::runtime(_("NOTE: The add-on \"{}\" is a dependency of {} other installed {}.\nAre you sure you wish to uninstall?")),
+            m_addon.get_title(), depending_addons.size(), static_cast<int>(depending_addons.size()) <= 1 ? _("add-on") : _("add-ons"));
```
Zwatotem, line 255: Extract to an external helper function

src/supertux/menu/addon_preview_menu.hpp
```diff
@@ -17,18 +17,21 @@
 #ifndef HEADER_SUPERTUX_SUPERTUX_MENU_ADDON_PREVIEW_MENU_HPP
 #define HEADER_SUPERTUX_SUPERTUX_MENU_ADDON_PREVIEW_MENU_HPP
 
-#include <vector>
-
-#include "addon/addon_manager.hpp"
 #include "gui/dialog.hpp"
 #include "gui/menu.hpp"
 
+#include <vector>
```
Zwatotem, line 23: Move std dependency to the top

Zwatotem (changes requested):

Zwatotem (changes requested):

src/addon/downloader.cpp
```diff
@@ -152,9 +152,9 @@ TransferStatusList::push(TransferStatusListPtr statuses)
 
 // Called when one of the transfers completes.
 void
-TransferStatusList::on_transfer_complete(TransferStatusPtr this_status, bool success)
```
tobbi: Why not `successful`?

Vankata453: I later get a "variable shadows local variable error", that's what the last commit fixes.

Vankata453: This name seems better, will change.

src/addon/addon_manager.cpp
```diff
@@ -222,15 +223,15 @@ AddonManager::~AddonManager()
...
-  if (PHYSFS_exists(m_cache_directory.c_str()))
-  {
-    empty_cache_directory();
-    if (!PHYSFS_delete(m_cache_directory.c_str()))
-    {
-      log_warning << "Error deleting addon cache directory: PHYSFS_delete failed: " << PHYSFS_getErrorByCode(PHYSFS_getLastErrorCode()) << std::endl;
-    }
-  }
+
+  // Delete the addon cache directory, if it exists.
```
tobbi, line 227: add-on (unify this)

mrkubax10:
@Zwatotem do you still want to review this pull request?

Zwatotem:
> @Zwatotem do you still want to review this pull request?

Yes

Zwatotem (approved):

mrkubax10 (approved):

PR #2369, opened 2022-12-25
Elastic camera shake instead of linear

weluvgoatz:
This PR replaces the old, more "linear" way of camera shaking with an elastic way.  This proves for a smoother and more believable camera shake, and works well in instances where the camera must shake.  I noticed it looked better both in instances where the camera continuously shakes and instances in which it only shakes once.

Also, explosions now generate a camera shake.


mrkubax10:
Could you fix compilation errors?

mrkubax10 (approved):

PR #2375, opened 2023-01-08
Fix dead scripting objects when testing worldmaps

### LINKED ISSUE #1511: Testing worldmaps with Squirrel features from editor : tilemaps become dead objects
**SuperTux version:** 0.6.2
**System information:** Ubuntu Linux 18.04 LTS

##### Expected behavior

Making a worldmap with Squirrel features should work properly, similarly to ordinary levels.

##### Actual behavior

When running a script which attempts to acces a named scriptable object (for example, a tilemap, to fade it out), the following error will result :

```error: script is accessing a dead object: ``` [number]

It appears like the Squirrel references to tilemaps are always null when playtesting from the editor

1. Duplicate the world1 worldmap into the add-ons folder.
2. Open the game, go in the level editor, and open world1's worldmap
3. From the worldmap editor, test the worldmap.
4. Move Tux up to the entrance of the cavern. The error will be printed to the console.

##### Additional info

Playing the same worldmap from Start game -> Contrib levels will give a working worldmap. (Make sure to edit the "info" file and set "levelset" to false)

### PR DESCRIPTION
**Pull request by Vankata453**
Fixes #1511.

The reason this bug occurs is because instead of getting the current worldmap, the function `scripting::get_game_object_manager()` returns the current sector from the editor instead. This is what causes the objects to not be found, thus triggering the "dead objects" scripting error.

This PR swaps the check for a current worldmap with the one for the current sector, by prioritizing the current worldmap. Upon testing, it seems to be working fine with levels, loaded from the worldmap, worldmaps, loaded from the editor, and other cases similar to these. There may be a better way to solve this issue in general, though.

Another change featured in this PR is that instead of `scripting::GameObject`s storing UIDs of their exposed parents, and afterwards having to search through the whole current `GameObjectManager` for their parent object, they now store a pointer to their parent `GameObject`, which is more optimized.

### DIFF
--- a/src/scripting/game_object.hpp
+++ b/src/scripting/game_object.hpp
@@ -1,5 +1,6 @@
 //  SuperTux
 //  Copyright (C) 2018 Ingo Ruhnke <grumbel@gmail.com>
+//  Copyright (C) 2022 Vankata453
 //
 //  This program is free software: you can redistribute it and/or modify
 //  it under the terms of the GNU General Public License as published by
@@ -27,7 +28,7 @@
   auto object_ptr = get_object_ptr();                                   \
   if (object_ptr == nullptr) {                                          \
     log_fatal << "error: script is accessing a dead object: "           \
-              << m_uid << std::endl;                                    \
+              << std::endl;                                             \
     return;                                                             \
   }                                                                     \
   auto& object = *object_ptr
@@ -36,7 +37,7 @@
   auto object_ptr = get_object_ptr();                                   \
   if (object_ptr == nullptr) {                                          \
     log_fatal << "error: script is accessing a dead object: "           \
-              << m_uid << std::endl;                                    \
+              << std::endl;                                             \
     return {};                                                          \
   }                                                                     \
   auto& object = *object_ptr
@@ -45,7 +46,7 @@
   auto object_ptr = get_object_ptr();                                   \
   if (object_ptr == nullptr) {                                          \
     log_fatal << "error: script is accessing a dead object: "           \
-              << m_uid << std::endl;                                    \
+              << std::endl;                                             \
     return x;                                                           \
   }                                                                     \
   auto& object = *object_ptr
@@ -54,7 +55,7 @@
   auto object_ptr = GameObject<::OBJECT>::get_object_ptr();             \
   if (object_ptr == nullptr) {                                          \
     log_fatal << "error: script is accessing a dead object: "           \
-              << GameObject<::OBJECT>::m_uid << std::endl;              \
+              << std::endl;                                             \
     return;                                                             \
   }                                                                     \
   auto& object = *object_ptr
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: src/scripting/game_object.hpp [hunk 1]
```diff
@@ -54,7 +55,7 @@
   auto object_ptr = GameObject<::OBJECT>::get_object_ptr();             \
   if (object_ptr == nullptr) {                                          \
     log_fatal << "error: script is accessing a dead object: "           \
-              << GameObject<::OBJECT>::m_uid << std::endl;              \
+              << std::endl;                                             \
     return;                                                             \
```
Semphriss, line 59: **tl;dr** Regarding the last paragraph of the PR description, the presumed problem ("[...] having to search through the whole current GameObjectManager [...]") does not exist (e. g. it doesn't search), and the new proposed solution has a fatal flaw in its implementation (crash), and fixing it would not be worth it. Details below.

---------------------------

`object_ptr` here will never be null, because the pointer comes from the constructor and will never be modified. This makes the `if` dead code, but more than that, it shows the problem with using raw pointers: they don't become null when the object they point to gets destroyed.

Presuming the original code structure does indeed have a case where an object gets destroyed while a script still holds a reference to it, verifying that an object still exists is necessary. However, that can't be done with raw pointers: since the only thing you can do with them is compare with null and dereference them, that the former cannot be used as a way to detect the pointer-to object being destroyed and that the latter (dereferencing) is undefined behavior likely to result in a segfault, pointers cannot work here.

If one would insist on using pointers here, one would need to use a remove listener for each script reference to set the internal pointer to null when necessary, which is not only difficult to implement, but its implementation would take significantly more place in the source code, and the overhead that was meant to be avoided is added back.

Speaking of overhead: it may sound surprising, but **using pointers here is not an optimisation.**

The GameObjectManager holds the GameObject's in an `unordered_map`, which is implemented as a hash map. Accessing an object per ID in a hash map is [amortized constant complexity, or O(1)](https://cplusplus.com/reference/unordered_map/unordered_map/at/). That means that regardless of the amount of stuff in the map, it always takes the same amount of time to retrieve an object: Take the given ID, hash it, and access the location identified by that hash. It does not have to search through the whole current GameObjectManager. There are some edge cases which may increase complexity if a bad hash is used, hence ["amortized"](https://stackoverflow.com/questions/200384/what-is-constant-amortized-time), but since UID is based on a builtin type (size_t), we're using the default hash for that type, which we can assume is its best implementation.

Generally speaking: [beware of _premature optimisation_](https://stackify.com/premature-optimization-evil/). 1) Have we observed that accessing an object from a script takes too much time? If not, then we shouldn't waste time trying to optimize it. 2) Is the solution observably better than the original case, enough so that the increased complexity (presuming the remove listeners, needed for this to work, are added) is worth it? This should involve running benchmarks and compare the time taken by both methods. Doing this would've revealed that the complexity is not linear as assumed, but rather constant.

In short, this fixes a non-existent problem in a fundamentally flawed way, for which a fix would be way too complex. I advise keeping the original method with UIDs.

Vankata453:
A fix for this issue is included in #2377.

PR #2386, opened 2023-01-29
Store GameSession spawnpoints in a stack

Vankata453:
All `GameSession` spawnpoints are now stored in a stack. The first element is the start position/point, and the next are respawn positions/points. The way the spawnpoint stack works is the following:

* Restarting the level spawns the player at the set (or default with sector "main" and spawnpoint "main") start point, which is found first in the stack.
* When all players are dead, they respawn at the last set respawn point, which is found last in the stack.
* Using the "Restart From Checkpoint" option finds the last respawn point for a checkpoint in the stack and respawns the players there, as well as removes any custom spawnpoints that may follow the checkpoint one.

A fix for checkpoints not checking the current sector, when choosing if they should have the "ringing" action, is included.

Four new scripting functions for levels are included in this PR. Those include `set_start_point`, `set_start_pos`, `set_respawn_point` and `set_respawn_pos`. All of the functions that end in `point` require one sector and one spawnpoint name, while on the other hand, the functions ending in `pos` require one sector name and an `x, y` `float` position, where to spawn/respawn the players.


Rusty-Box (approved):

Vankata453:
This PR has another optimization I'd also want to make.

Vankata453:
@mrkubax10 Should be ready now.

mrkubax10 (approved):

PR #2407, opened 2023-03-15
Add Camera.get_position/size to Squirrel API.

tylerandari13:
This adds 4 functions to the Squirrel API.

The ability to get the Camera position. (Top left corner.)
`Camera.get_x()`
`Camera.get_y()`

The ability to get the size of the window. (In pixels)
`Camera.get_screen_width()`
`Camera.get_screen_height()`

### DIFF
--- a/src/object/camera.cpp
+++ b/src/object/camera.cpp
@@ -866,13 +866,13 @@ Camera::get_center() const
                               static_cast<float>(m_screen_size.height) / 2.0f);
 }
 
-Vector
+const Vector&
 Camera::get_position() const
 {
   return m_translation;
 }
 
-Size
+const Size&
 Camera::get_screen_size() const
 {
   return m_screen_size;
--- a/src/object/camera.cpp
+++ b/src/object/camera.cpp
@@ -866,6 +866,19 @@ Camera::get_center() const
                               static_cast<float>(m_screen_size.height) / 2.0f);
 }
 
+Vector
+Camera::get_position() const
+{
+  return m_translation;
+}
+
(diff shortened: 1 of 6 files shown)


Vankata453 (commented):
Comments about potential `const` reference usage in getters.

src/object/camera.cpp
```diff
@@ -866,6 +866,19 @@ Camera::get_center() const
                               static_cast<float>(m_screen_size.height) / 2.0f);
 }
 
+Vector
```
Vankata453, line 869: Could use a `const` reference (`Vector` -> `const Vector&`).

src/object/camera.cpp
```diff
@@ -866,6 +866,19 @@ Camera::get_center() const
...
+Camera::get_position() const
+{
+  return m_translation;
+}
+
+Size
```
Vankata453, line 875: Could also use a `const` reference (`Size` -> `const Size&`).

src/object/camera.hpp
```diff
@@ -100,6 +100,13 @@ class Camera final : public GameObject,
       camera */
   Vector get_center() const;
 
+  /** get the coordinates of the point directly in the top left of this
+      camera */
+  Vector get_position() const;
```
Vankata453, line 105: Could use a `const` reference (`Vector` -> `const Vector&`).

Vankata453 (commented):
Comments about potential `const` reference usage in getters.

Rusty-Box (approved):

Issue #2339, opened 2022-11-28
Game sometimes ignores hitboxes defined in `.sprite` files


Rusty-Box:
##### Expected behavior

If one defines a hitbox for an action in a `.sprite` file, this animation when called in game uses the specified hitbox in said action.

##### Actual behavior

Sometimes the game uses the proper hitbox for an action but sometimes it seems like either the hitbox of an enemy is hardcoded or always uses the hitbox from a specific action!

##### Steps to reproduce actual behavior

I discovered this issue on multiple occasions. Most recently in `fatbat.sprite` - For example, if you go change the hitbox of only Fatbat's "left" and "right" action, the game will not recognize these changes, instead you must change its "burning" actions' hitboxes as the game seems to only use those hitboxes in all instances, which is not how the whole hitbox defining should work! Another candiate seems to be `yeti.sprite`, I haven't figured out which hitbox from which action it persists on using instead of the actual ones...

Rusty-Box:
Another case where this is prevalent. Having a scripted object that uses `Tux.sprite` and applying the `[scriptedobject].set_action("big-walk-right");` (for example) the scripted object still uses the wrong hitbox (I believe it is still using the backflip hitbox as this is the default action for Tux it seems).

Rusty-Box:
Alright I found more info on this issue after trying out some stuff. For my experiment I used the `zeekling.sprite` file.

**What I did:**
1. Open `zeekling.sprite`
2. Change only the hitbox of the **"left"** action from x:31 y:30 to x:44 y:34
    -> When the **"left"** action plays, hitbox in-game still shows as x:31 y:30
4. Change the hitbox of the **"left"** action from x:31 y:30 to x:200 y:34
   -> When the **"left"** action plays, Hitbox shows it as x:200 y:34 even though the **"left"** action was defined as x:44 y:34 previously
5. Rename **"burning-left"** and **"burning-right"** actions to **"jurning-left"** and **"jurning-right"**
6. Check now alphabetically highest action in the .sprite file, which is diving-left with a x:30 y:31 hitbox
7. Change the hitbox of the **"diving-left"** action to x:400 y:100
   -> When the **"left"** action plays, Hitbox shows it as x:400 y:100 even though the **"left"** action was still defined as x:44 y:34 previously

**Conclusion:**
[...]

PR #2414, opened 2023-03-30
Fix scripted objects not changing hitbox according to sprite action

Vankata453:
Scripted objects no longer use a self-defined `set_action` function for changing the sprite action. Instead, they now use the function with the same name from the base `MovingSprite` class. `MovingSprite::set_action`, as opposed to the removed `ScriptedObject::set_action` function, changes the hitbox according to the one defined for the sprite action.

The now unneeded `ScriptedObject::get_action` function is also removed and replaced with a `public` one, added in `MovingSprite`.

Provides a fix for one of the issues mentioned in #2339.

### DIFF
--- a/src/object/moving_sprite.hpp
+++ b/src/object/moving_sprite.hpp
@@ -61,17 +61,20 @@ public:
   bool change_sprite(const std::string& new_sprite_name);
   void spawn_explosion_sprites(int count, const std::string& sprite_path);
 
-protected:
-  /** set new action for sprite and resize bounding box.  use with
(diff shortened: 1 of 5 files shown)


tobbi:
@Vankata453 After merging, what is still missing from the issue mentioned?

Vankata453:
@tobbi I couldn't reproduce the issues, mentioned in the initial comment of the issue. I cannot be sure if they are fixed for that reason.

Rusty-Box (approved):

PR #2416, opened 2023-03-31
Fix building on MSYS2

iahung:
The patch is based on the work of @Biswa96 of the MSYS2 project:

https://github.com/msys2/MINGW-packages/discussions/16526

### DIFF
--- a/mk/cmake/SuperTux/ProvideSquirrel.cmake
+++ b/mk/cmake/SuperTux/ProvideSquirrel.cmake
@@ -41,7 +41,7 @@ else()
     -DINSTALL_INC_DIR=include
     -DCMAKE_POSITION_INDEPENDENT_CODE=ON)
 
-  if(WIN32)
+  if(MSVC)
     add_library(LibSquirrel SHARED IMPORTED)
     set_target_properties(LibSquirrel PROPERTIES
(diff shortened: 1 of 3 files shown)


Biswa96:
This may break existing build workflow. I have already said this may not be appropriate for upstream. Also the mingw package is not complete--the paths do not follow typical *nix-like FHS.

Biswa96:
This may break existing build workflow. Also the mingw package is not complete--the paths do not follow typical *nix-like FHS.

mk/cmake/SuperTux/ProvideSquirrel.cmake
```diff
@@ -41,7 +41,7 @@ else()
     -DINSTALL_INC_DIR=include
     -DCMAKE_POSITION_INDEPENDENT_CODE=ON)
 
-  if(WIN32)
+  if(MSVC)
```
Biswa96, line 44: This is not required. cmake can find squirrel from the package.

iahung:
I tested both the nightly built using MSVC provided by SuperTux and my own version built using MSYS2. The MSYS2 version has weird bug in game play compared to the official nightly. I have no idea why it's this way.

PR #2440, opened 2023-04-22
Use Miniswig from external submodule

Vankata453:
Adds Miniswig as an external submodule from [WindstilleTeam/miniswig](https://github.com/WindstilleTeam/miniswig), which is more actively maintained.


src/scripting/wrapper.hpp
```diff
@@ -3,8 +3,8 @@
  *  'src/scripting/wrapper.interface.hpp'
  * DO NOT CHANGE
  */
-#ifndef HEADER_SUPERTUX_SCRIPTING_WRAPPER_HPP
-#define HEADER_SUPERTUX_SCRIPTING_WRAPPER_HPP
+#ifndef HEADER_SUPERTUX_SQUIRREL_WRAPPER_HPP
```
tobbi, line 6: Why was this changed?

Vankata453: "SCRIPTING" was inconsistent when using Miniswig in other projects. I settled for using "SQUIRREL" because it'd fit in all scenarios, but it could also be made so it's set by an argument.

Semphriss: I thought the include guards were meant to follow the file's pathname?

Vankata453: That's true. I could make it so it generates the header based on the path in `--output-hpp-include`.

mrkubax10 (approved):

PR #2472, opened 2023-05-14
Reduced width of crushers by one to fix bug get stuck

### LINKED ISSUE #2403: Ice crushers get stuck on blocks
**SuperTux version: SuperTux v0.6.3-659-g2fb84352d
**System information:** N/A

##### Expected behavior

Crushers go past objects if there is logical clearance

##### Actual behavior

Due to floating point error, crushers get stuck on blocks when going upwards

See video:
https://clipchamp.com/watch/BLx36RMzIwR

SharkyFlou:
Reduced the crushers width hitbox by one unit (64 -> 63, 128 -> 127).
I'm still afraid it has consequences I haven't thought of.

Closes #2403

### DIFF
--- a/data/images/creatures/crusher/corrupted/krosh_corrupt.sprite
+++ b/data/images/creatures/crusher/corrupted/krosh_corrupt.sprite
@@ -1,35 +1,35 @@
 (supertux-sprite
  (action
   (name "idle")
-  (hitbox 16 16 128 128)
+  (hitbox 16 16 127 128)
   (images "krosh_corrupt_main.png"
   )
  )
  (action
   (name "crushing")
-  (hitbox 16 16 128 128)
+  (hitbox 16 16 127 128)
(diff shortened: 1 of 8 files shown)


tobbi:
@Rusty-Box What was your decision regarding this? Or did someone else talk about it?

Rusty-Box:
I don't think this is the solution we're looking for. And I feel like this doesn't account for sideways moving Crushers
Crushers tend to get stuck often when they are passing by slopes.

One reason for that might be the fact they can be pushed from their original position y-position, or x-position for sidways Crushers. I wouldn't be suprised if that behavior is what's causing the issue. It would be interesting to see Crushers to not be push-able and then see if them getting stuck somewhere still applies.

Vankata453:
In my opinion, the proper solution would be to fix the issue coding-wise.

Closing due to inactivity. If there is interest in continuing work on this fix, please open a new PR!

Issue #2418, opened 2023-04-03
`CC-BY-SA-2.0` in `supertux2.appdata.xml` results in a proprietary software warning on KDE's Discover application


retcinder:
**SuperTux version: 0.6.3**
**System information:** *Arch Linux using KDE Plasma LTS (5.27.3)*

##### Expected behavior
SuperTux should have no warnings when viewed in Discover for both the Flatpak and distro-specific (e.g. pacman, apt) distribution methods.

##### Actual behavior
SuperTux has a Proprietary Software warning indicated by `CC-BY-SA-2.0` being orange along with an info button being next to `CC-BY-SA-2.0`.

This can be fixed by bumping the `CC-BY-SA` version from 2.0 to either 3.0 or 4.0 as these are valid project licenses as stated [here](https://freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-metadata_license).
This would require checking to see if the `CC-BY-SA` license can be used with the current `CC-BY-SA` graphics/audio being used.

##### Steps to reproduce actual behavior
1. Open Discover on a system using KDE Plasma 5.25 or above
2. Either find SuperTux on the home page or search for it
3. Open the page for SuperTux and there will be a warning next to `CC-BY-SA-2.0`

##### Screenshots
Warning next to `CC-BY-SA-2.0`.
![image](https://user-images.githubusercontent.com/63363329/229577255-523b506b-d76a-4c7e-8177-77c1ef4abfd5.png)

Proprietary Software warning (in German as that's my desktop's language)
![image](https://user-images.githubusercontent.com/63363329/229577045-0d3fb2d7-147c-409d-970d-67ea3965a2ae.png)

PR #2474, opened 2023-05-16
Bump `CC-BY-SA` to 4.0 in `supertux2.appdata.xml`

tulpenkiste:
This is a fix for #2418 by increasing the `CC-BY-SA` version to 4.0.
`CC-BY-SA-4.0` is considered a "free" license which results in the warning disappearing on Discover.

Before fix
image
After fix
image

### DIFF
--- a/supertux2.appdata.xml
+++ b/supertux2.appdata.xml
@@ -2,7 +2,7 @@
 <component type="desktop-application">
   <id>org.supertuxproject.SuperTux</id>
   <metadata_license>CC0-1.0</metadata_license>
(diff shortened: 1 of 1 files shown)


tobbi:
https://wiki.creativecommons.org/wiki/4.0_upgrade_guidelines

tulpenkiste:
@tobbi problem is:
- Some things in the AUTHORS file use CC-BY-SA-3.0
- Some do not specify a version at all
- For the single thing that uses CC-BY-SA-2.0, the link is down due to server error 500

PR #2479, opened 2023-05-30
New Tux run animation

Rusty-Box:
Tux walk animation alone doesn't look all that convincing at faster speeds so a run animation has been added (currently only accessible through scripting Tux as a Scripted Object, but hopefully that is to be changed later).

The idea is that Tux starts moving with his walking animation until he reaches max speed. Once the walk animation has played its last frame when at max speed. Tux's run animation will start playing until he gets slower. Exception would be moments like the "level end sequence" or his main menu appearence where Tux should always play the walking animation due to the speed at which he walks in so instances.

- [x] Things left to do is a transition frame from "standing" to "walking" and from "walking" to "running" for a smoother transition.

**A little video demonstration on how it would work (currently only done via scripting)**

https://github.com/SuperTux/supertux/assets/20615048/396ef956-167c-4724-9a04-def3bfdd2b24


Vankata453 (approved):

Frostwithasideofsalt:
it gets merged as im installing it to test lol

PR #2481, opened 2023-05-31
Add corrupted variant of Snail

Rusty-Box:
Graphics for the corrupted variant of the Snail badguy
### Behavior
Corrupted Snail will behave just like normal Snail, however every ~8 seconds they will stand still and reveal spiky roots poking out of their helmet. During that, jumping on them hurts Tux and leaves the badguy unharmed. Corrupted Snail will remain spiky for ~3 seconds until its spiky roots will retract and they continue walking along as normal.

When Corrupted Snail stops, the **shake** action will play once, followed by the **guard** action. As long as the **guard** action plays, Corrupted Snail will be harming Tux on touch and cannot be stomped. When they return to their normal state, the **retract** action will play once before Corrupted Snail begins to walk normally (like how their waking animatoin plays before walking along).

_I made a quick little demonstration of how I envision Corrupted Snail to work in-game (using a Scripted Object):_

https://github.com/SuperTux/supertux/assets/20615048/1930e7b3-1bdc-4865-a440-3943030a77cc


tobbi:
corrupted snail should be named "Darth Snailer".

autofly17:
I think the design looks a bit too creepy for Tux. Maybe consider adding pupils into their eyes so they look less possessed?

guard-0

PR #2493, opened 2023-06-12
Update several icons + update missing texture

Rusty-Box:
List of changes:
- Update player hud icons (includes coin counter and level timer)
- Make missing texture easier on the eyes

TODO:
- [x] Several level editor icons (Background object, Spawnpoint, Decal etc.)
- [x] Improve "level stats" icons
- [x] A few code adjustments regarding the HUD and states shown during the PAUSE menu


Rusty-Box:
Final things to be done:

- [ ] Adjust the powerup HUD counter, so it does not stick into the new coin counter
- [ ] Fix Timer HUD icon to not be affected by ambient light (i think someone mentioned it was due to the timer not being drawn properly in the game's code)

@MatusGuy Since you have write access to this branch, feel free to commit directly to it!

MatusGuy:
epic ci skip fail

tobbi:
I'm not so sure about the invisible wall one. Doesn't really convey its meaning...

Rusty-Box:
> I'm not so sure about the invisible wall one. Doesn't really convey its meaning...

If that one does not so does the invisible bonus block either, as it is using the same design choice.

tobbi:
> > I'm not so sure about the invisible wall one. Doesn't really convey its meaning...
> 
> If that one does not so does the invisible bonus block either, as it is using the same design choice.

Well, that one was a bad design choice as well, IMO.

Rusty-Box:
How else to make an icon for something "invisible" then?

tobbi:
> How else to make an icon for something "invisible" then?

You don't.

Rusty-Box:
Too bad. I'm gonna keep it then

tobbi:
How about this: Keep the transparency but have a brick wall as an icon?

mrkubax10 (approved):
Looks good code-wise.

MatusGuy:
> Too bad. I'm gonna keep it then

We should just not add an icon (its invisible).

i say it conveys its meaning + it reduces filesize

Rusty-Box:
> How about this: Keep the transparency but have a brick wall as an icon?

This is neither adds nor changes anything. That would be the same thing. Imo. that would even look "cheap" as it's just an image with reduced opacity.

src/supertux/statistics.cpp
```diff
@@ -319,10 +319,10 @@ Statistics::draw_ingame_stats(DrawingContext& context, bool on_pause_menu)
     if (!on_pause_menu)
       x_offset *= std::min(1.f, -std::abs(m_coins_time - 2.5f) + 2.5f);
 
-    Vector pos(context.get_width() - x_offset,
-               context.get_height() - height * 6.f - 20.f);
+    Vector pos(static_cast<float>(context.get_width()) - x_offset,
+               static_cast<float>(context.get_height()) - height * 6.f - 20.f - 16.f);
```
Vankata453, line 323: Same as above comment.

src/supertux/statistics.cpp
```diff
@@ -348,10 +348,10 @@ Statistics::draw_ingame_stats(DrawingContext& context, bool on_pause_menu)
     if (!on_pause_menu)
       x_offset *= std::min(1.f, -std::abs(m_badguys_time - 2.5f) + 2.5f);
 
-    Vector pos(context.get_width() - x_offset,
-               context.get_height() - height * 5.f - 10.f);
+    Vector pos(static_cast<float>(context.get_width()) - x_offset,
+               static_cast<float>(context.get_height()) - height * 5.f - 10.f - 8.f);
```
Vankata453, line 352: Here as well.

pazkero:
Regarding invisible wall, the proposed icon uses that [big block](https://github.com/SuperTux/supertux/raw/f90acacfc2656d4ee9a0cc707b21089e5af662a5/data/images/tiles/blocks/bigblock.png) which people use as a wall in levels, right? And with the usual transparency squares, it seems fairly well explained what it does.

The only downside imo is that you might need to strain your eyes a bit to understand what exactly that is. I probably would cut it in half (bottom half opaque and upper half semi-transparent like current one, with a red line in between)  and/or consider using red brick tiles, but that's more of an accessibility thing than anything else.

I believe there might be worse stuff in the game, at least in terms of "strain the eye to see" than this icon.

big block

Vankata453:
Honestly, I believe the texture is fine. If the user is not exactly sure what it is for, they can hover over it and the object name will be shown.

MatusGuy:
bruh!

src/supertux/statistics.cpp
```diff
@@ -333,8 +333,8 @@ Statistics::draw_ingame_stats(DrawingContext& context, bool on_pause_menu)
                                     : Statistics::perfect_color
                               );
     context.color().draw_surface_scaled(coin_icon,
-                                        Rectf(pos.x + width + 1.f, pos.y + 1.f,
-                                              pos.x + width + 16.f, pos.y + 16.f),
+                                        Rectf(pos.x + width + 3.f, pos.y - 5.f,
+                                              pos.x + width + 3.f + 32.f, pos.y - 5.f + 32.f),
```
Vankata453, line 337: `Rectf(Vector(pos.x + width + 3.f, pos.y - 5.f), Sizef(32.f, 32.f))`.

Also to be done for other such instances in the same file.

Vankata453 (approved):

PR #2505, opened 2023-06-27
Fix menu behaviour after deleting item

### LINKED ISSUE #2501: Menus are not resized properly after removing item from them
**SuperTux version:** latest master
**System information:** *N/A*

##### Expected behavior
Menu is resized properly

##### Actual behavior
Menu is not resized properly. There is empty space left left after deleted item.

1. Go to Level editor
2. Create levelset and few levels
3. Try deleting them one by one

Video of the issue: https://github.com/SuperTux/supertux/assets/63350514/5a7c6656-cfa5-451b-bd8a-2f916451d851

mrkubax10:
Fixes #2501

### DIFF
--- a/src/supertux/menu/editor_delete_level_menu.cpp
+++ b/src/supertux/menu/editor_delete_level_menu.cpp
@@ -28,18 +28,13 @@
 #include "editor/editor.hpp"
 #include "gui/menu_item.hpp"
 #include "gui/dialog.hpp"
-EditorDeleteLevelMenu::EditorDeleteLevelMenu(std::unique_ptr<Levelset>& levelset, EditorLevelSelectMenu* level_select_menu, EditorLevelsetSelectMenu* levelset_select_menu) : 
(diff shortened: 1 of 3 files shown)


src/gui/menu_item.hpp
```diff
@@ -22,6 +22,7 @@
 
 class MenuItem
 {
+  friend class Menu;
```
Vankata453, line 25: What is this needed for?

mrkubax10: Good catch, I forgot to remove it

src/supertux/menu/editor_delete_level_menu.cpp
```diff
@@ -67,7 +80,10 @@ EditorDeleteLevelMenu::menu_action(MenuItem& item)
       {
         PHYSFS_delete(m_level_full_paths[id].c_str());
         delete_item(id + 2);
-        m_level_full_paths.erase(m_level_full_paths.begin() + id);
+        m_level_full_paths[id].clear();
+        m_level_names[id].clear();
+        clear();
```
Vankata453, line 85: You could add `clear()` at the beginning of `refresh()`.

mrkubax10: `refresh` is also called in constructor and calling `clear` while menu is empty is redundant.

Vankata453 (approved):

PR #2510, opened 2023-06-29
Fixes and updates for Ubuntu Touch port

mrkubax10:
Updated to UT 20.04 and Clickable 7


.github/workflows/ubuntu-touch.yml
```diff
@@ -92,12 +92,12 @@ jobs:
...
-                                         --arch ${ARCH}
+      #- name: Publish to Open Store
+      #  if: ${{ github.ref == 'refs/heads/master' && matrix.build_type == 'Release' && matrix.opengl == 'sdl' }}
+      #  env:
+      #    ARCH: ${{ matrix.arch }}
+      #    OPENSTORE_KEY: ${{ secrets.OPENSTORE_KEY }}
+      #  run: |
+      #    ~/.local/bin/clickable publish "* $(git log -1 --pretty=%B |         \
+      #                                   head -1)" --apikey ${OPENSTORE_KEY}   \
+      #                                   --arch ${ARCH}
```
HybridDog, line 103: Is the disabling of this `Publish to Open Store` task intentional?

mrkubax10: Yes, because I think that only releases should be published. You can see at what state game currently is.

HybridDog: I agree.

HybridDog (approved):
The changes look fine for me. 
I know almost nothing about Ubuntu Touch, so my approval may not count for much.

mrkubax10:
This PR uses SDL2 from Ubuntu focal repositories which doesn't support Mir anymore and some devices still use it. To make UT builds at least compiling it should be merged. I will add Mir support to latest SDL2 and then create supplementary PR which will make use of that patched SDL2.

zubozrout:
Is the project currently buildable with clickable?
Simply running `clickable build --arch arm64` it ends up with the following for me when I try:
```
[100%] Built target tinygettext
Install the project...
-- Install configuration: "Release"
-- Up-to-date: supertux/build.clickable/tinygettext/lib/libtinygettext.a
CMake Error at cmake_install.cmake:58 (file):
  file cannot create directory: /tinygettext.  Maybe need administrative
  privileges.

make[3]: *** [Makefile:100: install] Error 1
make[2]: *** [CMakeFiles/tinygettext_project.dir/build.make:102: tinygettext_project-prefix/src/tinygettext_project-stamp/tinygettext_project-install] Error 2
make[1]: *** [CMakeFiles/Makefile2:169: CMakeFiles/tinygettext_project.dir/all] Error 2
make: *** [Makefile:156: all] Error 2
```

mrkubax10:
CI is building without issues so it should work.

zubozrout:
I didn't notice UT is also in the CI, thank you so much :). Still weird it fails building for me though - I can't see any difference quickly. But atm this is enough as my only goal was to install a Focal build on UT since it is no longer being updated in the OpenStore.

mrkubax10:
> But atm this is enough as my only goal was to install a Focal build on UT since it is no longer being updated in the OpenStore.

It will be uploaded to OpenStore when 0.7.0 is released.

PR #2512, opened 2023-07-01
[QoL] More climbable tweaks

Raoul1808:
This PR addresses the other climbable PR #1670 I made 2 years ago, as well as #1473 and #972, but partially.

Climbing zones now have a restriction to all 4 edges to prevent Tux from leaving without jumping. I also made it so you no longer leave without jumping by pressing the ACTION key, but rather by pressing DOWN + JUMP.

Possible improvement: checking for nearby climbable zones and disable boundaries that intersect directly with other zones.

### DIFF
--- a/src/object/player.cpp
+++ b/src/object/player.cpp
@@ -2204,7 +2204,8 @@ Player::collision(GameObject& other, const CollisionHit& hit)
   if (moving_object->get_group() == COLGROUP_TOUCHABLE) {
     auto trigger = dynamic_cast<TriggerBase*> (&other);
     if (trigger && !m_deactivated) {
-      if (m_controller->pressed(Control::UP))
+      auto climbable = dynamic_cast<Climbable*> (&other);
+      if ((climbable && m_controller->hold(Control::UP)) || (m_controller->pressed(Control::UP)))
         trigger->event(*this, TriggerBase::EVENT_ACTIVATE);
     }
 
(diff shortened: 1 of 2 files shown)


Raoul1808:
I just added an exception to triggers to make it so you can climb surfaces by holding the UP key instead of having to press it while being on top of one.

src/object/player.cpp
```diff
@@ -2523,18 +2523,19 @@ Player::handle_input_climbing()
 
   float vx = 0;
   float vy = 0;
-  if (m_controller->hold(Control::LEFT)) {
+  auto obj_bbox = m_climbing->get_bbox();
```
Vankata453, line 2526: Could be `const` reference (`const Rectf&`).

Vankata453 (approved):

Alasdairbugs:
> Possible improvement: checking for nearby climbable zones and disable boundaries that intersect directly with other zones.

IMO this is a must-have and should be included in this PR if we're thinking QOL fixes here.

mrkubax10 (approved):

Rusty-Box:
> I also made it so you no longer leave without jumping by pressing the ACTION key, but rather by pressing DOWN + JUMP

Hold on... so I can only leave a climbable when pressing Down + Jump? Or am I misunderstanding something here?
I might have to checkout this one myself

mrkubax10:
You can still leave Climbable by jumping.

PR #2514, opened 2023-07-06
Reduce flushing of game objects

tobbi:


src/supertux/sector_parser.cpp
```diff
@@ -101,6 +101,7 @@ SectorParser::parse_object(const std::string& name_, const ReaderMapping& reader
 void
 SectorParser::parse(const ReaderMapping& sector)
 {
+  GameObjectManager::s_is_loading = true;
```
Vankata453, line 104: The `s_is_loading` variable seems to be unneeded, since it's set to `true` in `false` in only 2 cases, which both happen within one function. This references all edits to this file.

src/supertux/game_object_manager.cpp
```diff
@@ -213,8 +230,14 @@ GameObjectManager::flush_game_objects()
...
     }
+    s_needs_flushing_gameobjects_added = false;
+  }
+
+  if(s_needs_flushing_tilemaps)
+  {
+    update_tilemaps();
+    s_needs_flushing_tilemaps = false;
   }
-  update_tilemaps();
```
Vankata453, line 217: Blocking `TileMap` updating to only when they are added is likely to cause issues, when a tilemap is dynamically set to solid, via the scripting function `set_solid()`.

src/supertux/game_object_manager.cpp
```diff
@@ -182,6 +188,15 @@ GameObjectManager::draw(DrawingContext& context)
...
+
+  if(!s_needs_flushing_gameobjects_added &&
+     !s_needs_flushing_gameobjects_removed &&
+     !s_needs_flushing_tilemaps)
+    return;
+
+  if(s_needs_flushing_gameobjects_removed)
   { // cleanup marked objects
     m_gameobjects.erase(
       std::remove_if(m_gameobjects.begin(), m_gameobjects.end(),
```
Vankata453, line 202: Are we sure not looping to look for objects which have been requested for removal every frame saves time?

src/supertux/game_object_manager.cpp
```diff
@@ -196,8 +211,10 @@ GameObjectManager::flush_game_objects()
                        }
                      }),
       m_gameobjects.end());
+      s_needs_flushing_gameobjects_removed = false;
   }
 
+  if(s_needs_flushing_gameobjects_added)
   { // add newly created objects
     // Objects might add new objects in finish_construction(), so we
     // loop until no new objects show up.
```
Vankata453, line 220: Don't believe this helps with optimization. `while (!m_gameobjects_new.empty())`, which is on the next line, already blocks this loop, if no new objects have been queued to be added.

Vankata453 (changes requested):

mrkubax10:
Could you fix merge conflicts?

PR #2520, opened 2023-07-08
(WIP) Improve OpenGL renderer performance

Raoul1808:
I noticed the OpenGL renderer for the game isn't very optimized, as it executes one draw call for every texture on screen (which is pretty bad).

This PR aims to improve the renderer by batching draw calls that use the same texture together.

### DIFF
--- a/src/video/gl/gl_painter.cpp
+++ b/src/video/gl/gl_painter.cpp
@@ -63,7 +63,12 @@ GLPainter::GLPainter(GLVideoSystem& video_system, GLRenderer& renderer) :
   m_video_system(video_system),
   m_renderer(renderer),
   m_vertices(),
-  m_uvs()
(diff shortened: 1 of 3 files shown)


NickerAban:
something is wrong with the graphs 


https://github.com/SuperTux/supertux/assets/71407963/dd323fe9-87b7-4db2-8890-15e0f9a0e313

mrkubax10:
> something is wrong with the graphs

No sh## Sherlock.

Raoul1808:
The more I look at the code and the graphical issues, the less it makes sense.

I have tried many local fixes but they lead nowhere. I'm closing this PR as I don't think this gets us anywhere (or at least for me).

I might get back to this, but if I do it will probably be either in the form of a bigger patch, or something completely different.

PR #2530, opened 2023-07-12
Unstable tile now refuses to respawn when tux is occupying original position

### LINKED ISSUE #2528: Unstable Tiles sometimes do not respawn
This one is a bug I noticed quite a while ago. So apparently (and I can't quite figure out what it is dependent on) sometimes in some levels, the unstable tiles do not respawn after getting destroyed. Considering they do technically not get destroyed on the spot, rather dropping to the ground with an invisible sprite as far as I know, I suspect something has to do with that, though I am not sure on this one specifically.

A recent example of such instance is the level Crumbling Path from the Rooted Forest, in Story Mode. Specifically this section right at the beginning.

https://github.com/SuperTux/supertux/assets/20615048/6ad48181-2bdb-42f5-a614-cdd04b23d582

MatusGuy:
Apparently #1998 might've tried to fix this?

This should also close #2528

### DIFF
--- a/src/object/unstable_tile.cpp
+++ b/src/object/unstable_tile.cpp
@@ -19,6 +19,7 @@
 
 #include "object/unstable_tile.hpp"
 
+#include "collision/collision.hpp"
 #include "object/explosion.hpp"
 #include "object/player.hpp"
 #include "sprite/sprite.hpp"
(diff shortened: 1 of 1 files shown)


Rusty-Box:
Works great, however, now you also break boxes on the ground without having to jump/gain y-velocity first which might be an issue.

Rusty-Box (approved):

PR #2535, opened 2023-07-13
Earth powerup stone form breaks bricks below without stopping

### LINKED ISSUE #2532: Earth Tux's rock form should break through boxes without stopping
Similar to how Tux's buttjump ability breaks all boxes in one swoosh, rather than stopping after each row, when Tux is in his rock form he should be doing the same when landing on both normal and heavy boxes.

MatusGuy:
This should close #2532


Rusty-Box:
> Update: I did, are metal boxes just not breaking for you?

If I roll right underneath a metal box and jump while in rock form, they do not break. The normal ones do.
Plus I noticed that rock Tux also does not keep his y-velocity when he jumps against boxes. He should.

Rusty-Box (approved):

src/object/player.cpp
```diff
@@ -745,12 +745,13 @@ Player::update(float dt_sec)
...
 
-  if (m_does_buttjump)
+  if (m_does_buttjump || (m_stone && m_physic.get_velocity_y() > 30.f))
   {
     Rectf downbox = get_bbox().grown(-1.f);
     downbox.set_bottom(get_bbox().get_bottom() + 16.f);
     for (auto& brick : Sector::get().get_objects_by_type<Brick>()) {
-      if (downbox.contains(brick.get_bbox()) && brick.get_class_name() != "heavy-brick") {
+      // stoneform breaks through any kind of bricks
+      if (downbox.contains(brick.get_bbox()) && (m_stone || brick.get_class_name() != "heavy-brick")) {
```
Vankata453, line 754: Not related to a change of yours, but it would be nice if `brick.get_class_name() != "heavy-brick"` is replaced with a `dynamic_cast`.

mrkubax10: Probably won't make much difference there but [string comparsion can be faster than dynamic_cast](https://stackoverflow.com/a/9778220).

Vankata453: We should keep it consistent throughout the code. Other places all use `dynamic_cast`.

src/object/player.cpp
```diff
@@ -761,6 +762,18 @@ Player::update(float dt_sec)
     }
   }
 
+  // break bricks above without stopping
+  if (m_stone && m_physic.get_velocity_y() < 30.f)
+  {
+    Rectf topbox = get_bbox().grown(-1.f);
+    topbox.set_top(get_bbox().get_top() - 16.f);
+    for (auto& brick : Sector::get().get_objects_by_type<Brick>()) {
+      if (topbox.contains(brick.get_bbox())) {
```
Vankata453, line 771: You can either follow the convention:

```
if (...)
{
}
else
{
}
```

Or even better for one-line statements, remove the curly brackets as a whole.
Applies to all changes.

mrkubax10 (approved):

PR #2539, opened 2023-07-15
Add a replace tool to the editor, clean up overlay_widget.cpp

Narre:
This tool was in the old Mono editor but has never been implemented in the in-game editor. It finds all instances of a tile in a tilemap and replaces them with the desired tile. This can be especially useful when a player creates their own tileset that doesn't have any autotiles for their custom tiles. With this feature, they can create a level using a theme that does have autotiles and then simply replace the tiles from the theme with corresponding tiles from their custom theme. It can also be used to change a level's theme in general.

https://github.com/SuperTux/supertux/assets/14074789/045059a2-b93f-416e-ac67-091128a56d0e


src/editor/overlay_widget.cpp
```diff
@@ -529,6 +529,23 @@ EditorOverlayWidget::fill()
   }
 }
 
+void
+EditorOverlayWidget::replace()
+{
+  auto tilemap = m_editor.get_selected_tilemap();
```
Vankata453, line 535: This function should utilize `tilemap->save_state()` in the beginning and `tilemap->check_state()` at the end, so the operation can be tracked by undo/redo.

src/editor/overlay_widget.cpp
```diff
@@ -447,62 +482,95 @@ EditorOverlayWidget::fill()
...
     pos_stack.pop_back();
   }
+  tilemap->check_state();
+}
+
+void
+EditorOverlayWidget::replace()
+{
+  auto tilemap = m_editor.get_selected_tilemap();
+  Uint32 replace_tile = tilemap->get_tile_id(static_cast<int>(m_hovered_tile.x), static_cast<int>(m_hovered_tile.y));
```
mrkubax10, line 555: Please note that `Uint32` is not the same thing as `uint32_t`

Narre: I used the same data type as is used in the fill() function. I assume it is used for a reason. If not, I can change it.

mrkubax10: Alright

src/editor/overlay_widget.cpp
```diff
@@ -150,56 +158,50 @@ EditorOverlayWidget::drag_rect() const
...
-  this->autotile(Vector(x + 1.0f, y + 1.0f), tile);
+  this->autotile(Vector(x - 1.0f, y - 1.0f), tile, save_state);
+  this->autotile(Vector(x       , y - 1.0f), tile, save_state);
+  this->autotile(Vector(x + 1.0f, y - 1.0f), tile, save_state);
+  this->autotile(Vector(x - 1.0f, y       ), tile, save_state);
+  this->autotile(Vector(x       , y       ), tile, save_state);
+  this->autotile(Vector(x + 1.0f, y       ), tile, save_state);
+  this->autotile(Vector(x - 1.0f, y + 1.0f), tile, save_state);
+  this->autotile(Vector(x       , y + 1.0f), tile, save_state);
+  this->autotile(Vector(x + 1.0f, y + 1.0f), tile, save_state);
```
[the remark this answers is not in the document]
Narre: save_state() and check_state() reverted to pre-this PR state, as discussed on Discord.

mrkubax10 (approved):

src/editor/overlay_widget.cpp
```diff
@@ -1311,11 +1412,10 @@ EditorOverlayWidget::draw_tile_grid(DrawingContext& context, int tile_size,
 void
 EditorOverlayWidget::draw_tilemap_border(DrawingContext& context)
 {
-  if ( !m_editor.get_selected_tilemap() ) return;
+  if (!m_editor.get_selected_tilemap()) return;
 
   auto current_tm = m_editor.get_selected_tilemap();
-  if (!current_tm)
-    return;
```
Vankata453, line 1318: I believe this is fine.

Vankata453 (approved):

src/editor/overlay_widget.cpp
Narre: It should be fine either way. The guidelines say that there are no hard rules for {} within functions, so I just went with short if() return on one line, consistently in the whole file. It was 50/50 before.

PR #2540, opened 2023-07-15
Refactor editor requests to call functions directly

tobbi:


tobbi:
For some reason, the object selects as well as tilegroup selects keep showing after selecting a tilegroup.

tobbi:
I'm gonna close this as it's too far behind to actually do anything useful with it.

PR #2542, opened 2023-07-16
Improvements to object type support, object versioning

### LINKED ISSUE #2502: Make Skull Tiles an option/type of Unstable Tiles
Since the **Skull Tile** behaves almost identical to an **Unstable Tile**, minus the fact that they don't respawn (which they should) and don't break immediately, fusing both into one wouldn't be such a bad idea. Let's say when placing an unstable tile you have setting for "Break immediately" (Y/N) or something along those lines or make the Skull Tile a type of the Unstable Tile objects.

Alternatively, if we don't wanna do that, give Skull Tiles a more neutral name based on what it does. Skull Tile as a name only works when it is using skulls as a sprite :/

Vankata453:
Object, utilizing object types to have their sprites changed, now change their default sprite, according to their type. This stops the game from having to save the object type's sprite to the level file, when it's supposed to be the default.

Also features code improvements, like the initial `on_type_change` call for `GameObject`s being moved to the `parse_type` function, and changing an object's sprite on type change being moved to `MovingSprite::on_type_change`.

Another major feature, included in this PR, is `GameObject` versioning. It aims to provide a simple way to deprecate features in an object, which would allow for retaining proper functionality in old levels, whilst providing new functionality for new levels. Object versioning is supposed to be used, when introducing changes to objects, which are incompatible with the previous way of functioning of the object.

**TODO:**

- [x] Make more objects support object types, rather than having hardcoded behaviour based on sprite, or containing other default sprites in a non-obvious way.

Fixes #2502.


Rusty-Box:
Some issues I found regarding certain types:
- The object types: Pumpkin, Bumpkin, BSOD, Laptop are immune to ice
- "Corrupted" type for Crusher shows Rock Crusher (the "Rock" type is missing entirely!)

Rusty-Box:
Minor request: Could the retro powerups in the selection be renamed to "Growth (retro)" "Star ("retro") and put after their respective powerup group (e.g.  Growth (retro) comes after the Growth line-up and Star ("retro") comes after the Star powerup)

Rusty-Box:
Also, add extra object types for Jumpy (i.e. Corruptjumpy - make sure to put it before retro and after woodjumpy) and Leafshot (Corrupted)

mrkubax10 (approved):
Looks alright code-wise but I'm not sure Laptop should be type of both Smartblock and Iceblock. They behave differently (turning around on edges) and it may cause inconsistency.

Rusty-Box:
> Looks alright code-wise but I'm not sure Laptop should be type of both Smartblock and Iceblock. They behave differently (turning around on edges) and it may cause inconsistency.

I agree. Laptop should only be an object type for Mr. Iceblock and not Smartblock. Just like BSoD is only an object type for Mr. Snowball and not Smartball.

Vankata453:
This must have been an oversight on my part, I forgot `SmartBlock` inherits `MrIceBlock`.

Rusty-Box (changes requested):
I just remembered tthat Bouncing Snowball actually has an object type as well that is missing! Fatbat is the object type for Bouncing Snowball.

Another thing: Rusty Trampoline seems to inherit code from Rocks hence they have the same object types, which is a mistake!

Also, didn't you make Ice the new default for new instances of weak block?

Rusty-Box:
Another bug I discovered too: you can squish forzen Bumpkin, Pumpkin and BSoDs

Vankata453:
@Rusty-Box Additionally, I also removed the "linked" property from weak blocks. Only "Hay" is linked, "Ice" is not. This would ensure consistency.

tobbi:
Does this fix #2502 ?

Vankata453:
It doesn't, but probably should.

Rusty-Box (approved):

PR #2546, opened 2023-07-18
Ability to set bonus block coin sprite for coin, coin rain and coin explosion

### LINKED ISSUE #2487: "Set coin sprite" option for Bonus Blocks
With the official Add-On "Revenge In Redmond" using the retro bonus blocks it would be nice to have the coins that come out of them look like the actual retro coins present in those levels instead of the normal gold coins.

Couldn't this be done by adding an option to set the sprite of the coin that comes out of the bonus block?

MatusGuy:
This should close #2487

### DIFF
--- a/src/object/bonus_block.cpp
+++ b/src/object/bonus_block.cpp
@@ -223,8 +223,6 @@ BonusBlock::get_settings()
                    "1up", "custom", "script", "light", "light-on", "trampoline", "portabletrampoline", "rain", "explode", "rock", "potion"},
                   static_cast<int>(Content::COIN), "contents");
   result.add_sexp(_("Custom Content"), "custom-contents", m_custom_sx);
-
-  // FIXME: Change this from text to path
   result.add_sprite(_("Coin sprite"), &m_coin_sprite, "coin-sprite", "images/objects/coin/coin.sprite");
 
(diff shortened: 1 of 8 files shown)


Rusty-Box (approved):

src/object/coin_explode.cpp
```diff
@@ -18,9 +18,11 @@
 
 #include "math/random.hpp"
 #include "object/coin.hpp"
+#include "sprite/sprite_manager.hpp"
```
Vankata453, line 21: Unused include.

Vankata453 (approved):

src/object/bonus_block.cpp
```diff
@@ -72,6 +72,7 @@ BonusBlock::BonusBlock(const Vector& pos, int tile_data) :
   m_custom_sx()
 {
   m_default_sprite_name = "images/objects/bonus_block/bonusblock.sprite";
+  m_coin_sprite = "images/objects/coin/coin.sprite";
```
Vankata453, line 75: The variable should rather be initialized in the member initialization list.

Vankata453 (changes requested):

mrkubax10 (approved):

Vankata453 (approved):

PR #2547, opened 2023-07-18
Fix worldmap sector spawnpoint issues, testing from worldmap spawnpoints

Vankata453:
Fixes issues, related to loading a worldmap with a forced spawnpoint. It can now be set both with `set_initial_spawnpoint` on `WorldMap`, and by providing it to the initializer of the class. The forced spawnpoint overrides the saved Tux position state on the worldmap sector.

This PR also adds the ability to initialize a worldmap with a forced sector, allowing for teleporting on a different sector, other than "main", on a different worldmap.

Another addition is the "Test from here" option for worldmap spawnpoints. The worldmap will be tested from the current sector, Tux being on the position of the chosen spawnpoint to test from.

### DIFF
--- a/src/editor/editor.cpp
+++ b/src/editor/editor.cpp
@@ -341,7 +341,7 @@ Editor::test_level(const std::optional<std::pair<std::string, Vector>>& test_pos
   }
   else
   {
-    GameManager::current()->start_worldmap(*current_world, "", m_autosave_levelfile);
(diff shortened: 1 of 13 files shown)


mrkubax10 (approved):

Rusty-Box:
There is a problem linked to savegames. If you have an active save game that savegames position of Tux on the worldmap will always overwrite your "test from here" position, basically nullifying it.

Rusty-Box (approved):

PR #2549, opened 2023-07-19
Fix oklab colour picker rendering when using SDL

### LINKED ISSUE #2548: Colour selection menu renders incorrectly on SDL renderer
**SuperTux version:** Latest `master`

The colour selection menu renders incorrectly when using the SDL renderer. Check the screenshot below. The whole machine is laggy while the menu is open.

Color selection menu rendering incorrectly

Upon testing, @Narre concluded the first stable release to feature this issue is v0.6.3, which leads to the cause of this issue likely being the new OKLab colour picker, introduced in #1816.

Narre:
Fixes #2548 

This PR aims to fix a problem with sdl_painter.cpp that causes this issue. <s>For now, it's only a dirty fix. I'll try to come up with something better later. I'd also appreciate if others looked into it since I'm not sure I'll be able to find a proper fix myself. PRs into this branch are welcome.

image

Problems:
a) it renders weird bars in the colour picker
b) this is a workaround, not a fix - it just doesn't recalculate coords when a gradient isn't at (0, 0), which, as far as I know, currently only applies to the oklab picker, but it's not a proper solution.</s>

**EDIT:**
 I figured it out. The maths in that function was all wrong, now it works properly, no dirty workarounds.

image

### DIFF
--- a/src/video/sdl/sdl_painter.cpp
+++ b/src/video/sdl/sdl_painter.cpp
@@ -282,7 +282,7 @@ SDLPainter::draw_gradient(const GradientRequest& request)
     {
       rect.x = static_cast<int>(region.get_left());
       rect.y = static_cast<int>(region.get_top());
-      rect.w = static_cast<int>(region.get_right() - region.get_left());
+      rect.w = static_cast<int>(region.get_right() - region.get_left() + 1);
       rect.h = static_cast<int>(region.get_bottom() - region.get_top());
     }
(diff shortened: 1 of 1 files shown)


Narre:
It turned out this was easier to fix than I expected. The maths in `SDLPainter::draw_gradient()` was all wrong and only worked for gradients at (0, 0). Now it should work for gradients at any coordinates and of any size.

mrkubax10 (approved):

PR #2556, opened 2023-07-22
Fix editor not updating direction of badguys, conveyor belts and bumpers

Narre:
This PR fixes the editor not updating sprite actions of badguys, bumpers and conveyor belts upon changing their direction. It also fixes the formatting in all the files I touched - indentation, `m_` prefix for member variables and CRLF → LF.

### DIFF
--- a/src/object/bumper.hpp
+++ b/src/object/bumper.hpp
@@ -40,7 +40,7 @@ public:
   virtual void after_editor_set() override;
   virtual void on_flip(float height) override;
 
-  Physic *get_physic();
+  Physic& get_physic();
 
 private:
   Physic m_physic;
--- a/src/object/bumper.hpp
+++ b/src/object/bumper.hpp
@@ -27,22 +27,23 @@ public:
   Bumper(const ReaderMapping& reader);
 
   virtual ObjectSettings get_settings() override;
-  
+
   virtual void update(float dt_sec) override;
   virtual HitResponse collision(GameObject& other, const CollisionHit& hit) override;
-  
+
   static std::string class_name() { return "bumper"; }
   virtual std::string get_class_name() const override { return class_name(); }
   static std::string display_name() { return _("Bumper"); }
   virtual std::string get_display_name() const override { return display_name(); }
 
+  virtual void after_editor_set() override;
   virtual void on_flip(float height) override;
 
-  Physic physic;
+  Physic m_physic;
 
 private:
-  bool left;
-	
+  bool m_facing_left;
+
 private:
   Bumper(const Bumper&) = delete;
   Bumper& operator=(const Bumper&) = delete;
--- a/src/object/bumper.hpp
+++ b/src/object/bumper.hpp
@@ -17,6 +17,7 @@
 #define HEADER_SUPERTUX_OBJECT_BUMPER_HPP
 
 #include "object/moving_sprite.hpp"
+
 #include "supertux/physic.hpp"
 
 class Player;
@@ -39,9 +40,10 @@ public:
   virtual void after_editor_set() override;
   virtual void on_flip(float height) override;
 
-  Physic m_physic;
+  Physic *get_physic();
 
 private:
+  Physic m_physic;
   bool m_facing_left;
 
 private:
--- a/src/object/conveyor_belt.cpp
+++ b/src/object/conveyor_belt.cpp
@@ -128,8 +128,8 @@ ConveyorBelt::after_editor_set()
 
   if (m_length <= 0)
     m_length = 1;
-  m_col.m_bbox.set_size(32.0f * static_cast<float>(m_length), 32.0f);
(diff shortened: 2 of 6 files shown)


mrkubax10 (approved):

src/object/bumper.hpp
```diff
@@ -1,53 +1,54 @@
...
+
+  static std::string class_name() { return "bumper"; }
+  virtual std::string get_class_name() const override { return class_name(); }
+  static std::string display_name() { return _("Bumper"); }
+  virtual std::string get_display_name() const override { return display_name(); }
+
+  virtual void after_editor_set() override;
+  virtual void on_flip(float height) override;
+
+  Physic m_physic;
```
Vankata453, line 42: This could be made private and have a `get_physic()` getter for it.

src/object/conveyor_belt.cpp
```diff
@@ -128,6 +128,8 @@ ConveyorBelt::after_editor_set()
 
   if (m_length <= 0)
     m_length = 1;
+  m_col.m_bbox.set_size(32.0f * static_cast<float>(m_length), 32.0f);
```
Vankata453, line 131: This is hardcoded and not dependent to the sprite hitbox. Conveyor belt overrides `update_hitbox()`, so that function should be called from here.

src/object/fallblock.hpp
```diff
@@ -1,69 +1,69 @@
...
+    LAND
+  };
+
+private:
+  State state;
+
+  Physic physic;
+  Timer timer;
+
+  bool found_victim_down() const;
```
Vankata453, line 60: Private functions should be in a separate `private:` from the member variables.

src/object/fallblock.hpp
```diff
@@ -1,69 +1,69 @@
...
+    SHAKE,
+    FALL,
+    LAND
+  };
+
+private:
+  State state;
+
+  Physic physic;
+  Timer timer;
```
Vankata453, line 58: Since this PR already aims to fix this, these variables could have the `m_` prefix.

src/object/fallblock.hpp
```diff
@@ -1,69 +1,69 @@
...
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//  GNU General Public License for more details.
+//
+//  You should have received a copy of the GNU General Public License
+//  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#ifndef HEADER_SUPERTUX_OBJECT_FALLBLOCK_HPP
+#define HEADER_SUPERTUX_OBJECT_FALLBLOCK_HPP
+
+#include "object/moving_sprite.hpp"
```
Vankata453, line 19: Same as previous comment.

src/object/bumper.hpp
```diff
@@ -1,53 +1,54 @@
...
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//  GNU General Public License for more details.
+//
+//  You should have received a copy of the GNU General Public License
+//  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#ifndef HEADER_SUPERTUX_OBJECT_BUMPER_HPP
+#define HEADER_SUPERTUX_OBJECT_BUMPER_HPP
+
+#include "object/moving_sprite.hpp"
```
Vankata453, line 19: Includes of a base class should be separated with an empty line from other includes.

src/object/conveyor_belt.cpp
```diff
@@ -128,6 +128,8 @@ ConveyorBelt::after_editor_set()
 
   if (m_length <= 0)
     m_length = 1;
+  m_sprite->set_action(dir_to_string(m_dir));
+  update_hitbox();
```
Vankata453, line 132: The `update_hitbox()` line here wouldn't be needed, if `set_action` is not called on the sprite, but rather on the `ConveyorBelt` itself (`MovingSprite::set_action` will call `update_hitbox()`).

Didn't notice earlier, sorry.

src/object/bumper.hpp
```diff
@@ -1,53 +1,56 @@
...
+
+  static std::string class_name() { return "bumper"; }
+  virtual std::string get_class_name() const override { return class_name(); }
+  static std::string display_name() { return _("Bumper"); }
+  virtual std::string get_display_name() const override { return display_name(); }
+
+  virtual void after_editor_set() override;
+  virtual void on_flip(float height) override;
+
+  Physic *get_physic();
```
Vankata453, line 43: This may be better with a reference. Also pointers should be added to the class name itself, for example: `Physic*`.

Narre: Ah, the fancy "++" side of C++. Done.

Vankata453 (approved):

PR #2554, opened 2023-07-21
Improvements to Dart and DartTrap

Narre:
Closes #2524 

This PR removes all hard-coded coordinates and offsets in `darttrap.cpp`, now the darts will always spawn in the centre of the hitbox's height (width for up and down darts). This makes it much more intuitive to create custom sprites for DartTrap. The default sprite is adjusted for this in a way that won't affect existing levels.
It also makes it possible to change the sprite of the dispensed Dart from DartTrap's menu in the editor.
Darts and DartTraps now also support UP and DOWN directions, in addition to LEFT and RIGHT.
This PR also cleans up `darttrap.cpp`/`hpp` according to the guidelines, namely adding the `m_` prefix to member variables.

image

PR #2558, opened 2023-07-23
Add a flip-action option for sprite actions

Narre:
This PR adds a "flip-action" option for sprite actions. It's a vertical version of the existing "mirror-action", which is used to remove the need for sprite images to be stored twice, once for each direction. This does the same for up and down.
As a result, UP images were removed for ispy, and replaced with flip-actions of down.

Note 1: The flip-action code is almost identical to the mirror-action code. I could probably squish them together in one block of code, but I think this way is a lot more readable and allows for separate edits to these two sprite options in the future. But I can change that if others think it would be better.

Note 2: <s>I'll leave this PR as a draft for now, since I also want to replace the hard-coded flipping of darts and darttraps for their UP once #2554 is merged. Then I'll replace it with the flip-action and un-draft this PR.</s> **DONE**.


data/images/creatures/crystallo/rcrystallo.sprite
```diff
@@ -10,34 +10,39 @@
...
-
+    (clone-action "left")
+  )
   (action
     (name "slowdown-right")
     (fps 5)
     (hitbox 8 1 32 25)
-    (clone-action "right"))
-
+    (clone-action "right"
```
Rusty-Box, line 36: You're missing a `)` here. Otherwise things work great

Rusty-Box (approved):

Vankata453 (approved):

PR #2559, opened 2023-07-25
Scared Gold Bomb

Rusty-Box:
Gold Bomb will panic when Tux gets too close or any type of bomb is about to explode nearby.

They will run away for as long as Tux and/or a triggered bomb is in close range. If they reach a bottomless pit they "freeze" quivering in fear for as long as the "threat" is nearby (if there is safe ground they will drop down). When the area is all-clear, after a couple seconds, they will slowly resume their original behavior again.

This is a proof-of-concept made through the use of scripting and a ScriptedObject:

https://github.com/SuperTux/supertux/assets/20615048/0892734e-d2f3-445a-93a3-20e90918311d


FILIPOKonGithup:
LOL Gold Bomb is now a Sacredy Cat

Linusintro:
How did you get the Sprites for this?

MatusGuy:
@Linusintro He drew them.

src/badguy/goldbomb.cpp
```diff
@@ -141,15 +162,113 @@ GoldBomb::active_update(float dt_sec)
...
+    // also check for STATE_CORNERED just so
+    // the bomb doesnt automatically turn around
+    cornered();
     return;
+  }
   WalkingBadguy::active_update(dt_sec);
+
+  MovingObject* obj = nullptr;
+  std::vector<MovingObject*> objs = Sector::get().get_nearby_objects(get_bbox().get_middle(), SAFE_DIST);
+  for (size_t i = 0; i < objs.size(); i++)
```
tobbi, line 179: Can you use an iterator? `for(const auto& obj : objs)` ?

src/badguy/goldbomb.cpp
```diff
@@ -29,14 +31,29 @@
...
+const float REALIZE_TIME = 0.5f;
+
+// SAFE_DIST >= REALIZE_DIST
+const float REALIZE_DIST = 32.f * 8.f;
+const float SAFE_DIST = 32.f * 10.f;
+
+const float NORMAL_WALK_SPEED = 80.0f;
+const float FLEEING_WALK_SPEED = 180.0f;
+const int NORMAL_MAX_DROP_HEIGHT = 16;
+const int FLEEING_MAX_DROP_HEIGHT = 600;
```
mrkubax10, line 44: Those constants can be marked as `static` since they are only used in this file.

MatusGuy: I've seen them wrapped around an anonymous namespace, should I do that aswell?

mrkubax10: No need to

src/badguy/goldbomb.cpp
```diff
@@ -141,15 +162,113 @@ GoldBomb::active_update(float dt_sec)
...
+  }
+
+  case STATE_REALIZING:
+    if (!m_realize_timer.check()) break;
+
+    flee(vecdist.x > 0 ? Direction::LEFT : Direction::RIGHT);
+    break;
+
+  default: break;
+
```
Vankata453, line 260: The insides of the switch statement should be indented two spaces further. The unneeded empty lines at the beginning and end could also be removed.

src/badguy/goldbomb.cpp
```diff
@@ -141,15 +162,113 @@ GoldBomb::active_update(float dt_sec)
...
 }
 
 void
 GoldBomb::draw(DrawingContext& context)
 {
   m_sprite->draw(context.color(), get_pos(), m_layer, m_flip);
+  //const Rectf realizerect(get_bbox().get_middle()-Vector(REALIZE_DIST, REALIZE_DIST), get_bbox().get_middle()+Vector(REALIZE_DIST, REALIZE_DIST));
+  //const Rectf saferect(get_bbox().get_middle()-Vector(SAFE_DIST, SAFE_DIST), get_bbox().get_middle()+Vector(SAFE_DIST, SAFE_DIST));
+  //context.color().draw_filled_rect(realizerect, Color::from_rgba8888(255, 0, 0, 100), realizerect.get_size().width/2, 100);
+  //context.color().draw_filled_rect(saferect, Color::from_rgba8888(0, 255, 0, 100), saferect.get_size().width/2, 100);
```
Vankata453, line 271: Since this code has only been used for debugging purposes, it could be removed.

Vankata453 (commented):
Some of the comments from my previous review are left unresolved.

MatusGuy:
i noticed that. im gonna tell rusty to fix it rq

src/badguy/goldbomb.cpp
```diff
@@ -139,15 +160,108 @@ GoldBomb::active_update(float dt_sec)
...
+  }
   WalkingBadguy::active_update(dt_sec);
+
+  if (m_frozen) return;
+
+  MovingObject* obj = nullptr;
+  std::vector<MovingObject*> objs = Sector::get().get_nearby_objects(get_bbox().get_middle(), SAFE_DIST);
+  for (MovingObject* currobj : objs)
+  {
+    obj = currobj;
```
tobbi, line 179: Why is the extra variable needed?

tobbi (changes requested):

src/badguy/goldbomb.cpp
MatusGuy: It's the object that the gold bomb will check. read the rest of the code if you haven't

tobbi: Can you not use the loop iterator variable for that?

MatusGuy: Will it stay in scope?

tobbi: You're not using this anywhere else except for the for loop, no?

MatusGuy: I do use this outside the loop

tobbi: Ah, sorry, the code logic was...copulating with my brain.

MatusGuy: i think it should also be tweaked. any suggestions?

mrkubax10 (approved):

src/badguy/goldbomb.cpp
```diff
@@ -92,12 +113,13 @@ GoldBomb::collision_player(Player& player, const CollisionHit& hit)
...
-      kill_fall();
-    return ABORT_MOVE;
-  }
+if (tstate == STATE_TICKING)
+{
+  if (m_physic.get_velocity() != Vector()) kill_fall();
+  return ABORT_MOVE;
+} else if (tstate != STATE_NORMAL) {
+  return FORCE_MOVE;
+}
```
Vankata453, line 122: Tabulation seems to be missing.

src/badguy/haywire.hpp
```diff
@@ -47,6 +47,8 @@ class Haywire final : public WalkingBadguy
   virtual std::string get_display_name() const override { return display_name(); }
   virtual bool is_snipable() const override { return true; }
 
+  inline bool is_exploding() const { return m_is_exploding; }
```
Vankata453, line 50: Unneeded `inline`.

Vankata453: If so, it could be a separate PR which does that for every getter in the code.

MatusGuy: that'd make sense

tobbi: inline has no benefit. Why did you think it would?

mrkubax10: Depends on optimization mode (and on compiler) but that what @MatusGuy said is mostly true.

MatusGuy: mrkubax knows anything therefore this should be a real pull request 😎

tobbi: So every getter should be inline

mrkubax10: Well it probably doesn't have impact worth the effort.

src/badguy/goldbomb.cpp
MatusGuy: @tobbi this is resolved is it?

tobbi (approved):

Vankata453 (approved):

PR #2564, opened 2023-07-31
Create objects from scripting

### LINKED ISSUE #366: Scripting badguys
I think scripting badguys could be fun. Here are some ideas:
- Changing sprites of badguys
- Spawning badguys
- Creating badguys yourself by script or lisp definition (?)

What does the rest of the team think?

### LINKED ISSUE #2445: Add the ability to manually spawn enemies/objects via scripting.
Pretty self explanatory, I'd also like the ability to do things like set the enemy's  name so it can be tampered with later.

tobbi:
Fixes #2445

### DIFF
--- a/src/scripting/sector.hpp
+++ b/src/scripting/sector.hpp
@@ -50,6 +50,20 @@ public:
    * @param float $gravity
    */
   void set_gravity(float gravity);
+
+  /**
+ * Adds a game object to the game
+ * 
+ * @param class_name GameObject's class
+ * @param name Name of the created object
+ * @param posX X position inside the current sector
+ * @param posY Y position inside the current sector
+ * @param direction Direction 
+ * @param data Additional data
+ */
+void add_object(const std::string& class_name, const std::string& name,
+                int posX, int posY, const std::string& direction,
+                const std::string& data);
 };
 
 } // namespace scripting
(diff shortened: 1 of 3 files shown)


tobbi:
Actually, as for #366 this only fixes the first two things.

Vankata453:
Wouldn't this be better suited for the `Sector` scripting class? That way it'd be accessible for the current sector from `sector.settings`.

tobbi:
Default arguments will not be possible anymore then:
```
function add_object(class_name, name, posX, posY, direction = "left", data = "")
{
  Level_add_object(class_name, name, posX, posY, direction, data);
}
```

mrkubax10 (approved):

src/scripting/sector.hpp
```diff
@@ -50,6 +50,20 @@ class Sector final : public GameObjectManager
...
+  /**
+ * Adds a game object to the game
+ * 
+ * @param class_name GameObject's class
+ * @param name Name of the created object
+ * @param posX X position inside the current sector
+ * @param posY Y position inside the current sector
+ * @param direction Direction 
+ * @param data Additional data
+ */
```
Vankata453, line 63: A few issues with this comment block:

1. It could be more descriptive of the fact that only moving objects could be added with this command.
2. It should have full stops at the end of the main and parameter descriptions.
3. The parameter formatting is wrong. It should rather be `@param type $name description`.
4. Everything below the line with `/**` should be indented 2 spaces further.

tobbi: https://www.doxygen.nl/manual/docblocks.html

Vankata453 (approved):

PR #2565, opened 2023-08-01
Use std::map for addons instead of list

tobbi:
Self-explanatory mostly, gets rid of some excess std::find_if code


Vankata453 (approved):

mrkubax10 (approved):

tobbi:
Look how fancy this looks ^^

PR #2572, opened 2023-08-05
Only show supported directions for each object that has Direction

Narre:
For most badguys, this is just `auto`, `left` and `right`. For some, `up` and `down` but not `auto`. For other objects, it varies - `Dart` has `auto`, `left`, `right`, `up`, `down`, `DartTrap` the same but without `auto`, `PushButton` has `up` and `down`, `Bumper` `left` and `right`. The `Switch` has `none`, `left`, `right`, `up` and `down` (I added `up` and `down` in this PR).

The bumper and the button used to have a `bool` instead of a `Direction`, I replaced those but made sure that levels saved with the old `bool` would load correctly and figure out the proper direction.

With switches, I replaced the three sprites with just one that supports directions. This does mean that old levels that used left or right switches will default to `switch.sprite` and show `Direction::NONE` switches, but that shouldn't matter much since those won't look out of place.

Also, code quality and guidelines in the files I touched.

https://github.com/SuperTux/supertux/assets/14074789/098f1708-b2e6-4117-9c15-69f9f59e9ef2

**NOTE:** This will require proper testing, the more people test, the better. Especially if all directions save and load properly.


src/object/bumper.cpp
```diff
@@ -32,30 +32,46 @@ const float BOUNCE_X = 700.0f;
...
+  if (reader.get("direction", dir_str))
+    m_dir = string_to_dir(dir_str);
+  else if (reader.get("left", old_facing_left) && old_facing_left)
+    m_dir = Direction::LEFT;
+  set_action("normal", m_dir);
   m_physic.enable_gravity(false);
+
+  auto allowed_directions = get_allowed_directions();
+
+  for (int i = 0; i < static_cast<int>(allowed_directions.size()); ++i)
```
tobbi, line 50: This code seems duplicated across multiple files. Would moving it to a base class work?

Vankata453: There's the idea of making an `ObjectOption` to manage custom directions, instead of the objects doing so.

Vankata453 (approved):

mrkubax10 (approved):

tylerandari13:
For certain objects such as the bonus block every time you placed one it would bring a warning like `invalid "direction" in bonus block`. Will this go away with this PR?

tobbi (approved):

Vankata453:
@tylerandari13 I am fixing how `BonusBlock` parse their data in my object types PR (#2542).

PR #2579, opened 2023-08-11
Add "show all files" option for filesystem menu

MatusGuy:
For example: choosing the sprite for a bonus block will show a gigantic list mostly of pngs when only the sprites matter

Fix this by only showing the pngs optionally via checkbox

### DIFF
--- a/src/editor/object_option.cpp
+++ b/src/editor/object_option.cpp
@@ -403,15 +403,11 @@ ScriptObjectOption::add_to_menu(Menu& menu) const
 FileObjectOption::FileObjectOption(const std::string& text, std::string* pointer,
                                    std::optional<std::string> default_value,
                                    const std::string& key,
-                                   std::vector<std::string> filter,
-                                   const std::string& basedir,
-                                   bool path_relative_to_basedir,
+                                   FileSystemMenu::MenuParams& params,
                                    unsigned int flags) :
   ObjectOption(text, key, flags, pointer),
   m_default_value(std::move(default_value)),
-  m_filter(std::move(filter)),
-  m_basedir(basedir),
-  m_path_relative_to_basedir(path_relative_to_basedir)
+  m_params(params)
 {
 }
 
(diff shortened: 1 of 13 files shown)


mrkubax10 (approved):

Vankata453:
Sometimes, `.png` and `.jpg` files are needed for sprites and surfaces. This PR makes it, so those file types could be selected using the "Show all files" option, but that will also show other file types which are not valid for selecting neither sprites or surfaces.

My proposal is that this PR makes it, so `FileSystemMenu` takes two vectors on initialization: One to indicate required file types to show, and another for additional file types. This way, less commonly used file types would be hidden by default, but still accessible, without invalid file types showing alongside them.

MatusGuy:
if only it COMPILED!! 😁😁😁😁😁

src/editor/object_option.cpp
```diff
@@ -410,13 +410,15 @@ FileObjectOption::FileObjectOption(const std::string& text, std::string* pointer
                                    std::vector<std::string> filter,
                                    const std::string& basedir,
                                    bool path_relative_to_basedir,
+                                   std::vector<std::string> additional_filter,
```
mrkubax10, line 415: Pass `additional_filter` as const reference.

src/editor/object_option.hpp
```diff
@@ -279,7 +279,8 @@ class FileObjectOption : public ObjectOption
                    std::vector<std::string> filter,
                    const std::string& basedir,
                    bool path_relative_to_basedir,
-                   unsigned int flags);
+                   std::vector<std::string> additional_filter = {},
```
mrkubax10, line 283: As above.

src/editor/object_option.cpp
MatusGuy: do same for filter?

mrkubax10: Yes.

mrkubax10:
Could you fix compilation issues?

PR #2580, opened 2023-08-13
Add `Tux.set_velocity()` to scripting API.

tylerandari13:
Pretty self explanatory.

We have Tux.get_velocity_x() and Tux.get_velocity_y() so why not add a velocity setter as well?


mrkubax10 (approved):

src/scripting/player.hpp
```diff
@@ -193,6 +193,11 @@ class Player final
    */
   float get_velocity_y() const;
 
+  /**
+   * Sets the velocity of the player to a programmable/variable speed.
+   */
```
tobbi, line 198: Add documentation for params.

tylerandari13: I noticed `Tux.set_pos()` doesn't have params either, so to keep things consistent i didn't add params here either.

mrkubax10: It should have param documentation for scripting reference.

tylerandari13: ok then, ill add params to the function

PR #2584, opened 2023-08-17
Fix corrupted crusher not spawning roots when hit.top

Narre:
When the corrupted Crusher was added, for some reason, it was only programmed to spawn CrusherRoots when hitting bottom, left or right, but not top. This prevented it from spawning the roots during levelflip. The roots didn't even have a sprite action for that direction.

Also, CRLF→LF in corrupted Crusher and CrusherRoot sprite files.

Left = before, right = this PR

https://github.com/SuperTux/supertux/assets/14074789/6a32fde0-4590-47b7-9e06-e18db7bc485a


src/badguy/crusher.cpp
```diff
@@ -694,6 +705,11 @@ CrusherRoot::start_animation()
     m_sprite->set_animation_loops(1);
     break;
 
+  case Crusher::Direction::UP:
+    set_action("upwards");
+    m_sprite->set_animation_loops(1);
```
tobbi, line 710: the calls to "set_animation_loops" can be moved out of the 'case's

tobbi (approved):

mrkubax10 (approved):

PR #2587, opened 2023-08-17
Do not count harmless fish towards fragging + no collision

Narre:
It's not an enemy, it doesn't hurt Tux and there's no reason to kill it.

EDIT: After a Discord discussion, I also disabled its collision, making it an ambient entity.


bruhmoent:
I agree.

Vankata453 (approved):

PR #2590, opened 2023-08-18
Fix missing (or rather, not loading) editor sprite actions

Narre:
Right now, some badguys in the editor show the wrong sprite action (mostly for `Direction::AUTO`, but fish with "`swim`" actions for all directions). There was also a bug that the editor checked `m_dir` instead of `m_start_dir` for the badguys with `Direction::AUTO`. (`m_dir` is the current direction when the level is running, `m_start_dir` is the direction saved in the `.stl` level file that the enemy starts in).
Some would show facing in the wrong direction, some would show the wrong sprite action (eg. woken up sleeping spiky instead of sleeping, burning toads, etc.)

This PR fixes that.

### DIFF
--- a/src/badguy/badguy.cpp
+++ b/src/badguy/badguy.cpp
@@ -1081,18 +1081,19 @@ BadGuy::after_editor_set()
       action_set = true;
       break;
     }
-    else if (m_sprite->has_action(direction_str))
-    {
-      set_action(direction_str);
-      action_set = true;
-      break;
-    }
   }
 
   if (!action_set)
   {
+    if (m_sprite->has_action(direction_str))
+    {
+      set_action(direction_str);
+    }
+    else
+    {
     log_warning << "Couldn't find editor sprite action for badguy direction='"
                 << dir_to_string(m_start_dir) << "': " << get_class_name() << "." << std::endl;
+    }
   }
 }
 
--- a/src/badguy/badguy.cpp
+++ b/src/badguy/badguy.cpp
@@ -16,8 +16,6 @@
 
 #include "badguy/badguy.hpp"
 
-#include <list>
-
 #include "audio/sound_manager.hpp"
 #include "badguy/dispenser.hpp"
 #include "editor/editor.hpp"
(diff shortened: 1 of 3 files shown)


mrkubax10 (approved):

src/badguy/badguy.cpp
```diff
@@ -1064,7 +1064,7 @@ BadGuy::after_editor_set()
 {
   MovingSprite::after_editor_set();
 
-  if (m_dir == Direction::AUTO)
+  if (m_start_dir == Direction::AUTO)
   {
     if (m_sprite->has_action("editor-left")) {
```
tobbi, line 1069: Could this monstrosity of an if/else statement not be converted to a list and a foreach?

MatusGuy: Lmfao

src/badguy/badguy.cpp
```diff
@@ -1064,59 +1064,35 @@ BadGuy::after_editor_set()
 {
   MovingSprite::after_editor_set();
 
-  if (m_dir == Direction::AUTO)
+  const std::string direction_str = m_start_dir == Direction::AUTO ? "left" : dir_to_string(m_start_dir);
+  bool action_set = false;
+  for (std::string action_str : {"editor", "normal", "idle", "flying", "walking", "standing", "swim"})
```
tobbi, line 1069: const auto& action_str

and move the actions strings into a variable

src/badguy/badguy.cpp
```diff
@@ -1066,7 +1068,8 @@ BadGuy::after_editor_set()
 
   const std::string direction_str = m_start_dir == Direction::AUTO ? "left" : dir_to_string(m_start_dir);
   bool action_set = false;
-  for (std::string action_str : {"editor", "normal", "idle", "flying", "walking", "standing", "swim"})
+  const std::list<std::string> actions = {"editor", "normal", "idle", "flying", "walking", "standing", "swim"};
```
tobbi, line 1071: List was not to be taken literally, I think you can do fine with just std::string[], no?

tobbi (approved):

PR #2591, opened 2023-08-20
Fix memory leaks and code quality in autotile code

mrkubax10:
Fixes various code quality related issues in autotile code. Also fixes all memory leaks found by AddressSanitizer. Game doesn't use heap allocation anymore where it's not neccessary and uses smart pointers where possible/required.

This is a follow-up to old PR #2133 which was reverted.

### DIFF
--- a/src/supertux/autotile.cpp
+++ b/src/supertux/autotile.cpp
@@ -14,11 +14,11 @@
 //  You should have received a copy of the GNU General Public License
 //  along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-#include <bitset>
-
 #include "supertux/autotile.hpp"
 
-//#include "supertux/autotile_parser.hpp"
+#include <bitset>
+
+#include "util/log.hpp"
 
 // AutotileMask.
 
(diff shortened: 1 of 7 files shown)


tobbi (approved):

src/supertux/autotile.cpp
```diff
@@ -116,35 +104,28 @@ uint8_t
...
-
-
 // AutotileSet.
 
-std::vector<AutotileSet*>* AutotileSet::m_autotilesets = new std::vector<AutotileSet*>();
+std::vector<std::unique_ptr<AutotileSet>> AutotileSet::m_autotilesets;
 
-AutotileSet::AutotileSet(std::vector<Autotile*> tiles, uint32_t default_tile, std::string name, bool corner) :
-  m_autotiles(std::move(tiles)),
+AutotileSet::AutotileSet(const std::vector<Autotile*>& tiles, uint32_t default_tile, const std::string& name, bool corner) :
```
Vankata453, line 115: `Autotile` still seems to be used as a raw pointer.

mrkubax10: This is intentional. If I changed it to `std::unique_ptr` I wouldn't be able to initialize `m_autotiles` class member.

Vankata453: What if `AutotileParser::parse_autotile` returns an `Autotile` object and the vector is just `std::vector<Autotile>`?

mrkubax10: `Autotile` class is not-copyable.

Vankata453: Is there a reason for it not to be?

mrkubax10: It has quite many members so copying object of that class will definitely take longer than only copying pointer. `AutotileMask` is small enough that it can be copied.

Vankata453: Okay.

Vankata453 (approved):

PR #2593, opened 2023-08-20
Use only default keybinds, when navigating menus

### LINKED ISSUE #2577: Text inputs will treat letters like navigation controls when using WASD for example
**SuperTux version: gb1844bd2c**

##### Expected behavior
Text inputs should not treat letters like navigation controls and should let you edit without interruptions.

##### Actual behavior
They do.

- Set your keybindings to W - Up, A - Left, S - Down, D - Right. (or any letters)
- Go to any text input. This can be a script edit or a name input.
- Type W. As you can see, cursor moves up.

Vankata453:
Previously, the game would allow using custom key binds as a fallback to the default navigation key binds, when navigating menus. This however causes some issues, for example, `ItemTextField` being unselected, if a custom key bind has been pressed, when typing.

I believe this feature is redundant, and since the game already aims to use default keys for menu navigation, it would likely be better off without the custom key bind fallback.

I am leaving this open for discussion, in case there are any objections.

Fixes #2577.

### DIFF
--- a/src/control/keyboard_manager.cpp
+++ b/src/control/keyboard_manager.cpp
@@ -219,15 +219,7 @@ KeyboardManager::process_menu_key_event(const SDL_KeyboardEvent& event)
       control = Control::REMOVE;
       break;
     default:
-      if (m_keyboard_config.m_keymap.count(event.keysym.sym) == 0)
-        return;
-
-      // Forbid events from players other than the first in menus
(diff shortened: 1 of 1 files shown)


mrkubax10 (approved):

Vankata453:
Oh wait, I don't believe it will fix that issue. I've only modified the code for the keyboard configuration in this PR.

PR #2596, opened 2023-08-21
Preserving players and grabbed objects between sectors

### LINKED ISSUE #2191: Carryable objects and doors
At the moment, when going through a door while carrying an object ,like lantern, trampolines etc. Tux will come out the other side with empty hand and the object slowly floating towards him and back into his hand. This looks very odd looking and very tedious at long distances.

Carryable objects should immediately teleport alongside Tux and remain in his hand when using doors.

Vankata453:
Previously, each sector would contain its own instances of the `Player` object for each available player. This, however, makes it tedious to preserve player data across sectors, for example, the currently grabbed object.

`GameObjectManager` now supports moving objects to another `GameObjectManager`.

Now, the `Level` instantiates all `Player`s in the "main" sector, and each sector, on activation, moves those players to itself, so it can manage them, essentially making players global for the level. This allows retaining the full state of all players, between sectors.

Any grabbed/carried object by a player is also moved between sectors, together with the player.

Additionally, a fix for the slow moving animation of grabbed objects, when teleporting while carrying them, is included. Carried objects now teleport together with the player.

Fixes #2191.


Vankata453:
This pull request seems to break trampoline collision. I noticed the trampoline doesn't have `hit.top` set to `true` in the `CollisionHit` it recieves, when colliding with the player, even though the collision does occur on the top. In fact, `left`, `right` and `bottom` of the provided `CollisionHit` are `false`, too.

The only time the `top` collision is detected is when the player collides with the left or right side of the trampoline, otherwise nothing is detected. I have no idea what causes this.

https://github.com/SuperTux/supertux/assets/78196474/e0bd042d-05bb-48ab-87ff-796bb50f2eae

mrkubax10 (approved):

src/object/player.cpp
```diff
@@ -311,6 +310,21 @@ Player::do_scripting_controller(const std::string& control_text, bool pressed)
   }
 }
 
+void
+Player::move_to_sector(Sector& other)
```
tobbi, line 314: Can `other` be `const Sector&` ?

Vankata453: No, because `other` is given as the second argument to `move_object`, which takes a non-`const` reference, because it's modified (the object is moved there).

Rusty-Box (approved):
Missing keys, otherwise works great!

Rusty-Box (approved):
Keys work too!

MatusGuy (approved):
after a quick look, it's completely fine

PR #2597, opened 2023-08-22
Fix SDL gradient overlaps + SDL rendering of sector-wide gradients

Narre:
This was discovered by FrostC.
Due to the conversion from floating point to integer values, there was a "build-up" of rounding errors when splitting gradients to single-colour rectangles for rendering, which caused that some of the rectangles would overlap by one pixel. This PR adds a check that ensures that the next step always starts at previous step + previous height/width (both of those are already integers), plus it always shortens the height/width by one when the rounding error exceeds one (checks the actual position of the rectangle calculated by adding the width/height of the previous one to its starting position against the position calculated directly using the floats), so the gradient isn't stretched. This is the point at which the overlap would previously occur.
This bug was introduced in #2549 . I didn't notice it because the overlap can only be seen when the overlapping rectangles change the rendering outcome, i.e. when using the Additive blend mode.

before:
image
now:
image

**EDIT:**
While testing this PR, I found another bug. This one would cause sector-wide gradients (VERTICAL_SECTOR and HORIZONTAL_SECTOR) to render incorrectly, so I also fixed that (see the video in the fourth comment).

### DIFF
--- a/src/video/sdl/sdl_painter.cpp
+++ b/src/video/sdl/sdl_painter.cpp
@@ -309,23 +309,45 @@ SDLPainter::draw_gradient(const GradientRequest& request)
       next_step += rect.w;
     }
 
-    float p = static_cast<float>(i+1) / static_cast<float>(n);
+    float p = static_cast<float>(i) / static_cast<float>(n == 1 ? n : n - 1);
     Uint8 r, g, b, a;
 
-    if ( direction == HORIZONTAL_SECTOR || direction == VERTICAL_SECTOR)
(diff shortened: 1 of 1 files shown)


tobbi:
Is this something that can occur for the horizontal version as well?

Narre:
image
yes

tobbi:
Sorry, I misread the code. It's all good!

Narre:
Alright, I also fixed this (which has been in the game before, it's not #2549 's fault):

https://github.com/SuperTux/supertux/assets/14074789/5f9b0d84-1a26-41c7-83e2-b2e28d7a4cf9

tobbi (approved):

PR #2599, opened 2023-08-22
Worldmap support for adding objects through scripting

Vankata453:
The `add_object` scripting function was moved to the `GameObjectManager` scripting class. It calls `add_object_scripting` on the parent `GameObjectManager`, which is virtual, so `WorldMapSector` can add additional checks for its implementation.

The function always accepts tile positions (sector position / 32), when adding objects to `WorldMapSector`s, no matter if the object being added isn't a `WorldMapObject`.

### DIFF
--- a/src/worldmap/worldmap_sector.cpp
+++ b/src/worldmap/worldmap_sector.cpp
@@ -30,6 +30,7 @@
 #include "supertux/debug.hpp"
 #include "supertux/fadetoblack.hpp"
 #include "supertux/game_manager.hpp"
+#include "supertux/game_object_factory.hpp"
 #include "supertux/game_session.hpp"
 #include "supertux/gameconfig.hpp"
 #include "supertux/level.hpp"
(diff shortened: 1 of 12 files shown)


src/worldmap/worldmap_sector.cpp
```diff
@@ -353,6 +354,26 @@ WorldMapSector::update(float dt_sec)
       // tux->set_direction(input_direction);
     }
   }
+
+  flush_game_objects();
+}
+
+
```
tobbi, line 361: remove extra line

Vankata453: I originally added those double lines to separate functions in the class, just like how they are separated in the `.hpp` file. I am not sure whether to remove it, because on one side, this file does this in many places, but otherwise, the rest of the code does not seem to do this.

tobbi: nvm then

mrkubax10 (approved):

tobbi (approved):

PR #2607, opened 2023-08-24
Fix camera zooming in "manual" mode

### LINKED ISSUE #2405: Broken Camera Zoom
I have noticed some problems with the Camera scripts, the scale scripts in particular.
Setting the camera mode to `manual`, usually used for cutscenes, or when using `Camera.set_pos(x,y);` prior, the camera zoom does not work properly and rather "freaks out".
 

https://user-images.githubusercontent.com/20615048/224091345-273914f8-5f1a-4066-a398-f90f43568539.mp4

Something else I noticed (you also see in the video above), if used along `Effect.sixteen_to_nine(0);` you can see that the black bars do not fully cover the screen and that the Camera zoom overall seems rather jittery than smooth, especially at slower speeds which is not ideal.

Vankata453:
Camera zooming in "manual" mode now accepts an anchor point, and is properly managed. The camera progresses to the anchor point, relative to the current position, simultaneously with the zooming progression.

Target anchor point can be provided for zooming on other modes, but does not serve a purpose.

All zooming-related scripting functions have new corresponding functions with `_anchor` prefixes in their name, which allow for setting the target anchor point. If they are not used for zooming on "manual" mode, the default anchor point (`ANCHOR_MIDDLE`) will be used.

Fixes #2405.


src/scripting/camera.cpp
```diff
@@ -100,12 +100,24 @@ Camera::set_scale(float scale)
   ease_scale(scale, 0, "");
 }
 
+void
+Camera::set_scale_target(float scale, float centerX, float centerY)
```
tobbi, line 104: Just wondering: Squirrel doesn't like default arguments, right?

Vankata453: I tried. The issues I got were with how Miniswig defines wrapper functions. Not sure about Squirrel.

tobbi: right, *that* issue again.

Rusty-Box (approved):

tobbi (approved):
I tried checking out how it works but couldn't. Nonetheless approving b/c fuck it

PR #2609, opened 2023-08-25
Extend sub-pixel rendering support

### LINKED ISSUE #2422: Weird issue with zooming in and out when using SDL video system
**SuperTux version:**
v0.6.3-690-g3aecbc25c (nightly)
**System information:**
-Os: Windows 10
-Cpu: 11th gen intel i5
-Gpu: Intel iris Xe Graphics
##### Expected behavior
(screenshot taken with openGL33)
opengl

##### Actual behavior
(screenshot taken with sdl)
Sdl

**both of these images were taken with the camera scale set to 0.6. This effect is not limited to the camera scale being set at 0.6**

Vankata453:
Implements sub-pixel rendering support for the SDL2 renderer. Achieved by replacing drawing functions in `SDLPainter` with their `float` equivalents.

Additionally, the `DrawingContext` size getters now return floating-point values, since before, they would always return an integer value, which didn't account for possible floating-point scale values of the current transform. This helps fix various other graphical glitches.

Fixes #2422.


Rusty-Box (approved):

mrkubax10:
Does this have any impact on performance? Floating point arithmetic tends to be slower.

Vankata453:
@mrkubax10 To me it made no difference. Either way, in my opinion, we shouldn't sacrifice proper rendering for performance.

mrkubax10:
It should be fine then.

mrkubax10 (approved):

PR #2610, opened 2023-08-25
Display languages localized

### LINKED ISSUE #291: Language names should be displayed in their language instead of in English

tobbi:
This allows to display the language names in the options menu in localized form. Also includes some previously missing font files.

Fixes #291


src/gui/menu.hpp
```diff
@@ -78,7 +79,7 @@ class Menu
 
   ItemHorizontalLine& add_hl();
   ItemLabel& add_label(const std::string& text);
-  ItemAction& add_entry(int id, const std::string& text);
+  ItemAction& add_entry(int id, const std::string& text, FontPtr font = Resources::normal_font);
```
Vankata453, line 82: There shouldn't be a need to provide the font in the `Menu` constructor function. Could be provided on the returned `ItemAction`.

Vankata453 (changes requested):

Vankata453:
image

Noticing these two issues with the language menu, everything else looks fine.

tobbi:
Did you check out the submodule changes as well?

Vankata453:
@tobbi That fixed it, I forgot.

Vankata453 (approved):

PR #2612, opened 2023-08-25
Add scriptable earthquake effect to Camera

### LINKED ISSUE #2611: Proper "earthquake" script
Using the `Camera.shake()` script is very annoying to set up to actually simulate an earthquake type camera shake.
It would be nice to have a script that will shake the camera constantly (vertically and hot horizontally at a specified strength until disabled.

Vankata453:
Implements an earthquake effect to `Camera`, which vertically shakes the camera at a certain minimal delay, with a given average strength.

This feature can be managed through the `start_earthquake` and `stop_earthquake` scripting `Camera` functions.

Fixes #2611.

### DIFF
--- a/src/object/camera.cpp
+++ b/src/object/camera.cpp
@@ -19,6 +19,7 @@
 #include <math.h>
 #include <physfs.h>
 
+#include "math/random.hpp"
 #include "math/util.hpp"
 #include "object/player.hpp"
 #include "supertux/level.hpp"
@@ -146,6 +147,11 @@ Camera::Camera(const std::string& name) :
   m_shakespeed(),
   m_shakedepth_x(),
   m_shakedepth_y(),
+  m_earthquake(false),
+  m_earthquake_strength(),
+  m_earthquake_delay(),
+  m_earthquake_last_offset(0.f),
+  m_earthquake_delay_timer(),
   m_scroll_from(0.0f, 0.0f),
   m_scroll_goal(0.0f, 0.0f),
   m_scroll_to_pos(),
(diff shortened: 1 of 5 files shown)


src/object/camera.cpp
```diff
@@ -412,6 +453,33 @@ Camera::shake()
...
+      m_earthquake_last_offset = m_earthquake_strength * graphicsRandom.randf(-2, 2);
+      m_translation.y += m_earthquake_last_offset;
+    }
+    else
+    {
+      m_translation.y -= m_earthquake_last_offset;
+      m_earthquake_last_offset = 0.f;
+    }
+
+    m_earthquake_delay_timer.start(m_earthquake_delay + static_cast<float>(graphicsRandom.rand(0, 1)));
```
tobbi, line 475: `graphicsRandom.randf` ?

src/object/camera.cpp
```diff
@@ -412,6 +453,33 @@ Camera::shake()
...
+  }
+  else if (!m_earthquake_delay_timer.started())
+  {
+    m_earthquake_delay_timer.start(m_earthquake_delay + static_cast<float>(graphicsRandom.rand(0, 1)));
```
tobbi, line 479: `graphicsRandom.randf` ?

src/object/camera.cpp
Vankata453: Adds variation to the delay.

tobbi: I mean: use randf instead of casting to float?

Vankata453: Tried but that makes the delay be longer most of the time and the earthquake is slower than it should be. So I chose either slow it down by 1 second more or none.

Rusty-Box (approved):

tobbi (approved):
My cat's breath smells like catfood.

Frostwithasideofsalt:
good to know, tobbi.

mrkubax10 (approved):

PR #2618, opened 2023-08-29
Fix build warnings

Narre:


src/gui/menu_manager.hpp
```diff
@@ -46,9 +46,9 @@ class MenuManager final : public Currenton<MenuManager>
 private:
   template<typename T>
   struct SettableItem {
-    std::unique_ptr<T> current;
-    std::unique_ptr<T> next;
-    bool has_next;
+    std::unique_ptr<T> current = nullptr;
+    std::unique_ptr<T> next = nullptr;
+    bool has_next = false;
```
mrkubax10, line 51: I am not sure about variable initalization in header files.

Vankata453: Same.

Narre: I tried everything I could think of, everything else either didn't work or at least didn't get rid of the warnings. Do you know how to do this properly?

mrkubax10: I suppose that it should be fine like it is.

PR #2624, opened 2023-09-05
Zeekling behaviour improvements

### LINKED ISSUE #533: Zeeklings: Curved flight path + screech

MatusGuy:
idea is to remove the randomness and make the movement smoother

fixes #533


src/badguy/zeekling.cpp
```diff
@@ -119,57 +121,34 @@ Zeekling::collision_solid(const CollisionHit& hit)
...
-  if (player) {
-    last_player_pos = player->get_pos();
-    last_self_pos = m_col.m_bbox.p1();
-  }
+  const Vector& rangeend = {eye.x + ((plrmid.y - eye.y) *
+                                     (m_dir == Direction::LEFT ? -1 : 1)),
+                            plrmid.y};
 
-  return false;
+  // FIXME: Give the actual object that hit the raycast, or at least the hitbox
```
MatusGuy, line 148: In which pull request should this function appear?

src/badguy/zeekling.cpp
```diff
@@ -119,57 +121,34 @@ Zeekling::collision_solid(const CollisionHit& hit)
...
+  // Left/rightmost point of the hitbox.
+  Vector eye;
+  const Rectf& bbox = get_bbox().grown(1.f);
+  eye = bbox.get_middle();
+  eye.x = m_dir == Direction::LEFT ? bbox.get_left() : bbox.get_right();
 
-    // Do not dive if we would not descend faster than the player.
-    float relSpeed = vy - player_mov.y;
-    if (relSpeed <= 0) return false;
+  const Vector& plrmid = get_nearest_player()->get_bbox().get_middle();
```
tobbi, line 135: That name is not very verbose.

src/badguy/zeekling.cpp
```diff
@@ -204,7 +183,7 @@ Zeekling::active_update(float dt_sec) {
       break;
   }
 
-  BadGuy::active_update(dt_sec);
+    BadGuy::active_update(dt_sec);
```
tobbi, line 186: Why the indentation?

MatusGuy: soz qt creator messed up

src/badguy/zeekling.cpp
MatusGuy: player middle. i thinks its pretty ok. what are your suggestions

Vankata453: `player_middle`, `player_center`?

MatusGuy: sure

mrkubax10:
Depends on #2631

tobbi:
Can you fixed unused fields?

MatusGuy:
will in the next commit

Rusty-Box (commented):
Initial review:

- Zeekling seems to dive way to early when standing still.



https://github.com/SuperTux/supertux/assets/20615048/2c7c6295-ee78-4096-b95e-04632a2842f3


- In some instamces they could clearly dive at me but they don't.



https://github.com/SuperTux/supertux/assets/20615048/c2260f2a-64d5-45a0-be55-7283f21bf7f3



- Zeekling should not need to dive at Tux when not higher than Tux himself (or below two tiles above the ground). It looks rather goofy



https://github.com/SuperTux/supertux/assets/20615048/8463ceb5-6390-4c05-af0e-cfed383f61fa




- You can crash the game by doing this:


https://github.com/SuperTux/supertux/assets/20615048/a80277e2-6f27-47e8-93f6-3c33af951a58



- Also, I personally think that after they reached their lowest diving position they should begin moving back up pretty soon after, if not immidiatly. Right now it takes way too long and makes it look rather stiff and clunky

MatusGuy:
Thanks for testing rusty, i wouldve never found these on my own.

> Zeekling seems to dive way to early when standing still.

My original idea was to not make zeekling as dependant on player velocity, because the old one tried to predict where you were gonna be using the velocity. Looks like the way I'm doing it will require me to find a way to depend on velocity, but this time, control it a bit, and make it not as unfair as the old one.

> In some instamces they could clearly dive at me but they don't.

Huh. I'll turn on drawing for the detection line.

> Zeekling should not need to dive at Tux when not higher than Tux himself

It does actually do this but it looks like I didn't make it inclusive (doesn't handle same y position)

> You can crash the game by doing this

Wtf.

> Also, I personally think that after they reached their lowest diving position they should begin moving back up pretty soon after

Alright I'll tweak it later

> You can make Zeekling bonk his head weirdly 

This is because the red line below the zeekling didn't touch the ground. Looks like I'll have to put it in the edge of the bounding box.

MatusGuy:
Thanks for testing rusty, i wouldve never found these on my own.

> Zeekling seems to dive way to early when standing still.

My original idea was to not make zeekling as dependant on player velocity, because the old one tried to predict where you were gonna be using the velocity. Looks like the way I'm doing it will require me to find a way to depend on velocity, but this time, control it a bit, and make it not as unfair as the old one.

> In some instamces they could clearly dive at me but they don't.

Huh. I'll turn on drawing for the detection line.

> Zeekling should not need to dive at Tux when not higher than Tux himself

It does actually do this but it looks like I didn't make it inclusive (doesn't handle same y position)

> You can crash the game by doing this

Wtf.

> Also, I personally think that after they reached their lowest diving position they should begin moving back up pretty soon after

Alright I'll tweak it later

> You can make Zeekling bonk his head weirdly 

This is because the red line below the zeekling didn't touch the ground. Looks like I'll have to put it in the edge of the bounding box.

Rusty-Box:
Damn, I am scratching my head on this one. It is more consisten but at the price of accuracy. I don't know how tho approach this issue yet. The only quick thoughts I have been juggling around are:
- Drastically decrease the wind-up time and motion between flying and diving.

_I feel like this delay between the actual attack had forced us into this weird velocity calcualtion causing to either attack too early or too late. If it could attack faster we might be able to decrease the gap._

- Increase the actual diving speed

_Controversial but something that might help if it's done carefully, lowering the chances of outrunning Zeekling before it even had a chance to strike_

One of the core issues I still see is how easily one could just run past Zeekling without properly engaging. Tight and small platforms shouldn't be the only way to insure that. Even flat ground should leave Zeekling a potential threat

MatusGuy:
> Damn, I am scratching my head on this one. It is more consisten but at the price of accuracy

Same.

> Drastically decrease the wind-up time and motion between flying and diving.

Could you rephrase this or explain in more detail?

> Tight and small platforms shouldn't be the only way to insure that. Even flat ground should leave Zeekling a potential threat.

I think so too. Remind me how you were gonna use zeekling in the new story mode?

Rusty-Box:
> Could you rephrase this or explain in more detail?

So right now, I think the time it takes for Zeekling to charg up for their dive and them actually diving down takes way too long which I believe makes it so easy to out run them. This whole "calculating the players walk speed" is trying to work aorund that but to me the core issue should be fixed instead. If that actually fixes it, that is :/

> I think so too. Remind me how you were gonna use zeekling in the new story mode?

A mix of both. Zeekling tries to "snipe" you out of the air during platforming thus you must take great care when to do your jumps to not get hit mid-air but also across relatively even stretches of land here and there.

Rusty-Box (approved):
Behavior's complete and feels much more fair and predictable than before

src/badguy/zeekling.cpp
```diff
@@ -17,28 +17,36 @@
...
 {
   m_physic.enable_gravity(false);
+  m_physic.set_velocity_x(FLYING_SPEED);
 }
 
 void
 Zeekling::initialize()
 {
-  m_physic.set_velocity_x(m_dir == Direction::LEFT ? -speed : speed);
+  m_physic.set_velocity_x(m_physic.get_velocity_x() * (m_dir == Direction::LEFT ? -1 : 1));
```
tobbi, line 49: I'm confused: When m_dir is not left, this becomes a no op. Can this not be moved to the constructor?

src/badguy/zeekling.cpp
```diff
@@ -54,149 +62,211 @@ Zeekling::collision_squished(GameObject& object)
...
-      }
+Zeekling::on_bump_horizontal()
+{
+  m_dir = (m_dir == Direction::LEFT ? Direction::RIGHT : Direction::LEFT);
+  set_action(m_dir);
+  m_physic.set_velocity_x(std::abs(m_physic.get_velocity_x()) * (m_dir == Direction::LEFT ? -1 : 1));
+  std::cout << m_state << " " << m_dir << std::endl;
+
+  switch (m_state)
+  {
```
tobbi, line 73: You have a switch with only one case. Can you not convert it to an if statement?

MatusGuy: Did that to keep consistency between this function and the one below, but yeah I can do that

src/badguy/zeekling.cpp
MatusGuy: No. When using the auto direction, auto gets turned into left or right before the initialize function is called. See https://github.com/MatusGuy/supertux/blob/zeekling_behavior/src/badguy/badguy.cpp#L828

src/badguy/zeekling.cpp
```diff
@@ -54,149 +62,205 @@ Zeekling::collision_squished(GameObject& object)
...
-    const Vector player_mov = (player_pos - last_player_pos);
-    const Vector self_pos = m_col.m_bbox.p1();
-    const Vector self_mov = (self_pos - last_self_pos);
+  const Vector& plrmid = player->get_bbox().get_middle();
 
-    // New vertical speed to test with.
-    float vy = 2*fabsf(self_mov.x);
+  // Do not dive if we are too close to the player.
+  float height = player->get_bbox().get_top() - get_bbox().get_top();
+  if (height <= 32.f * 4.5f)
```
tobbi, line 143: I don't really like magic numbers here.

src/badguy/zeekling.cpp
```diff
@@ -54,149 +62,205 @@ Zeekling::collision_squished(GameObject& object)
...
+    return false;
 
-    // Do not dive if we are not above the player.
-    float height = player_pos.y - self_pos.y;
-    if (height <= 0) return false;
+  // Do not dive if we are too far above the player.
+  if (height > 512)
```
tobbi, line 147: same for this. Is 512 just an arbitrary number?

src/math/util.hpp
```diff
@@ -38,6 +38,12 @@ const T& clamp(const T& val, const T& min, const T& max)
   }
 }
 
+template<class T>
+bool in_bounds(const T& val, const T& min, const T& max)
+{
+  return !(val < min) && (val < max);
```
tobbi, line 44: Why not 
```
val > min
```

?

Also, this is inconsistent: min is inclusive while max is exclusive.

src/supertux/sector.cpp
```diff
@@ -720,21 +720,28 @@ Sector::get_gravity() const
...
 Player*
-Sector::get_nearest_player (const Vector& pos) const
+Sector::get_nearest_player(const Vector& pos) const
 {
-  Player *nearest_player = nullptr;
+  auto players = get_objects_by_type_index(typeid(Player));
+  if (players.size() == 1)
+  {
+    Player* player = static_cast<Player*>(players[0]);
+    return (player->is_dying() || player->is_dead() ? nullptr : player);
```
tobbi, line 729: Something I don't know: Does `||` take preference over `?` If not, this needs brackets.

src/supertux/sector.cpp
```diff
@@ -720,21 +720,28 @@ Sector::get_gravity() const
...
-    Player& player = *static_cast<Player*>(player_ptr);
-    if (player.is_dying() || player.is_dead())
+    Player* player = static_cast<Player*>(player_ptr);
+    if (player->is_dying() || player->is_dead())
       continue;
 
-    float dist = player.get_bbox ().distance(pos);
+    float dist = player->get_bbox().distance(pos);
 
     if (dist < nearest_dist) {
```
tobbi, line 743: Put this opening brace a new line.

tobbi (changes requested):
Sorry for being so...meticulate.

Vankata453 (changes requested):
On the topic of zeekling behaviour, I noticed that if it's going left and the player is walking on the ground right, the zeekling just goes in the air and goes back up. It has no chance to catch him.

Basically, if you're going in the opposite direction of the zeekling, it doesn't dive properly.

image
image

MatusGuy:
From what I understand, Zeekling is not supposed to be placed in areas with flat surfaces. Zeekling works best when placed in sections with small platforms, where it's tricky to run right past it.

I'd wait for Rusty's input.

Vankata453:
In my opinion it should work in any case. Also, when going in the same direction it is going in, it works. So that seems like a bug.

MatusGuy:
If you run straight ahead while ignoring the zeekling, you stay out of its trajectory. But if you're big tux, you will get hit because of your bigger hitbox, unless you slide at the last minute.

Small Tux's hitbox is one of its advantages. You can die in one hit and get a more limited moveset but have more chance of survival.

- Black - ground
- Red - Zeekling
- Blue - Tux's direction
- Green - Tux's destination

image

But if you go in the same direction, you will be interfering with zeekling's trajectory, which will hurt you.

image

So this isn't really a bug, but a feature.

Rusty-Box:
@Vankata453 Wait, what is the "bug" in question? I'm confused here. He seemed to work fine on my end. Is his diving curve messed up? What does the red line mean you drew on your screenshot?

MatusGuy:
Rusty i think the red line is a question mark

Vankata453:
> Small Tux's hitbox is one of its advantages. You can die in one hit and get a more limited moveset but have more chance of survival.

Yes, but it seems like it's hard to die as small Tux. Personally I like how the old zeekling, regardless of direction, always poses a threat to the player and your task is to be careful around it. Additionally, if you want to kill it while on smooth terrain as small Tux, I don't think you would be able to, because you're not able to jump that high, according to my testing.

> Is his diving curve messed up?

The diving curve simply poses no threat to small Tux when the directions of Tux and the zeekling differ.

> What does the red line mean you drew on your screenshot?

Just a question mark, you can ignore it.

Rusty-Box:
> Additionally, if you want to kill it while on smooth terrain as small Tux, I don't think you would be able to, because you're not able to jump that high, according to my testing.

I don't get this one? Are trying to kill it when it flying in the air? Just wait for it to come down.

> The diving curve simply poses no threat to small Tux when the directions of Tux and the zeekling differ.

They do not pose as much of a threat on flat ground but why would they have to. Even as Big Tux a simply slight is all you need and you are in the clear. Flat ground has already several enemies and obstacles that you can use to enhance difficulty. Zeekling is an airborne threat trying to snipe you during platforming and are much more efficiently used on less flat ground that rquired tight platforming. Like how a Haywire is pretty much uninteresting on an empty flat long surface or a chasing fish pretty easy to dodge in a large, open pool of water, even Crushers are very harmless on flat ground, as you can just run straight and they will never harm you unless they are like one tile above Tux - certain enemies have circumstances they perform better in or worse. It comes up to the level designer to create a proper enviornment to bring the best out of each element used in it.

Vankata453:
Okay, did some platforming testing. Now that you explained it it makes sense. It seems challenging enough when platforming and also the reason that it doesn't catch Tux when directions differ is now clear too, seeing MatusGuy's response.

Being used to the old zeekling, I just didn't initially expect this. Only count on the code-side of my review then.

> Are trying to kill it when it flying in the air?

I was trying to kill it as small Tux while running in the other direction. Seems like, however, just standing there without moving makes it drop on Tux, giving you the chance. It is speedy though, so the player should react fast on that.

src/badguy/zeekling.cpp
```diff
@@ -54,149 +69,205 @@ Zeekling::collision_squished(GameObject& object)
...
+  }
+  else if (hit.left || hit.right)
+  {
+    on_bump_horizontal();
   }
 }
 
-/** Linear prediction of player and badguy positions to decide if we should enter the DIVING state. */
 bool
 Zeekling::should_we_dive()
```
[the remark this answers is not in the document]
MatusGuy: This function edits m_target_y

src/badguy/zeekling.cpp
```diff
@@ -54,149 +69,205 @@ Zeekling::collision_squished(GameObject& object)
...
+    case DIVING:
+      if (m_timer.check())
+      {
+        recover();
+      }
+      else
+      {
+        float dist = m_target_y - m_start_position.y;
+        double progress = CubicEaseIn(static_cast<double>(1.f - m_timer.get_progress()));
+        float value = m_target_y - (static_cast<float>(progress) * dist);
```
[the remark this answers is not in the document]
mrkubax10: States intent clearly.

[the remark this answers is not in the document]
Vankata453: Const-correctness. I believe it's good to follow, so that's why I just comment on it. Maybe I shouldn't directly say "do this" in reviews, though.

Frostwithasideofsalt (approved):
this makes the zeekling feel so much better to fight against. like wow.

src/badguy/zeekling.cpp
```diff
@@ -54,149 +69,204 @@ Zeekling::collision_squished(GameObject& object)
...
-    float estFrames = height / relSpeed;
+  auto* resultobj = std::get_if<CollisionObject*>(&result.hit);
+  
+  if (result.is_valid && resultobj &&
+      *resultobj == player->get_collision_object())
+  {
+    m_target_y = player->get_bbox().get_top() + CATCH_OFFSET;
+    return true;
+  }
+  else
```
tobbi, line 169: You can omit the else and just write 
```
return false;
```

at the end of the file.

src/badguy/zeekling.cpp
```diff
@@ -54,149 +69,204 @@ Zeekling::collision_squished(GameObject& object)
...
+      }
+      else
+      {
+        const float dist = m_target_y - m_start_position.y;
+        const double progress = CubicEaseIn(static_cast<double>(1.f - m_timer.get_progress()));
+        const float value = m_target_y - (static_cast<float>(progress) * dist);
+        const Vector pos(get_pos().x, value);
+
+        set_pos(pos);
+
```
tobbi, line 244: Remove the empty line here. not really important just saying.

src/badguy/zeekling.cpp
```diff
@@ -54,149 +69,204 @@ Zeekling::collision_squished(GameObject& object)
...
+      else
+      {
+        const float dist = m_catch_pos - m_start_position.y;
+        const double progress = QuadraticEaseInOut(static_cast<double>(m_timer.get_progress()));
+        const float value = m_catch_pos - (static_cast<float>(progress) * dist);
+        const Vector pos(get_pos().x, value);
+
+        set_pos(pos);
+      }
+
```
tobbi, line 262: Remove the empty line here.

tobbi (changes requested):
Sometimes I prefer getting run over by a train to reviewing PRs, but here we go: Just fix my review comments and it's all good.

weluvgoatz (commented):
Ok, this behavior is kind of weird.  It's not really better or worse than the old behavior.  Just weirder.  It does a weird "U" movement, and in most situations, isn't a threat to Tux at all.  I understand that it's supposed to go until 1 tile above the ground, but in many cases, it stops around 3 tiles above the ground, so it doesn't endanger any Tux variant in any way.

weluvgoatz:
Also:
"also sometimes it just straight up doesnt dive at all when tux is obviously within reach"

src/badguy/zeekling.cpp
```diff
@@ -54,149 +69,199 @@ Zeekling::collision_squished(GameObject& object)
...
+      }
+      break;
+
+    case RECOVERING:
+      if (m_timer.check())
+      {
+        fly();
+      }
+      else
+      {
```
tobbi, line 250: Maybe you can get rid of the else block by `break;`-ing early in the RECOVERING case.

MatusGuy: thanks for pointing that out

src/badguy/zeekling.cpp
```diff
@@ -54,149 +69,202 @@ Zeekling::collision_squished(GameObject& object)
...
 Zeekling::should_we_dive()
 {
+  using RaycastResult = CollisionSystem::RaycastResult;
+
   if (m_frozen)
     return false;
 
-  const auto player = get_nearest_player();
-  if (player && last_player && (player == last_player)) {
+  Player* player = get_nearest_player();
```
tobbi, line 134: *Maybe* use `auto`? It's not that important, but wanted to mention it anyway.

src/supertux/sector.cpp
```diff
@@ -720,21 +720,29 @@ Sector::get_gravity() const
...
 Player*
-Sector::get_nearest_player (const Vector& pos) const
+Sector::get_nearest_player(const Vector& pos) const
 {
-  Player *nearest_player = nullptr;
+  auto players = get_objects_by_type_index(typeid(Player));
+  if (players.size() == 1)
+  {
+    Player* player = static_cast<Player*>(players[0]);
+    return ((player->is_dying() || player->is_dead()) ? nullptr : player);
```
tobbi, line 729: Remind me to remind you to use player->is_alive() once my player method PR is merged.

src/supertux/sector.cpp
```diff
@@ -720,21 +720,29 @@ Sector::get_gravity() const
...
+  Player* nearest_player = nullptr;
   float nearest_dist = std::numeric_limits<float>::max();
 
-  for (auto player_ptr : get_objects_by_type_index(typeid(Player)))
+  for (auto player_ptr : players)
   {
-    Player& player = *static_cast<Player*>(player_ptr);
-    if (player.is_dying() || player.is_dead())
+    Player* player = static_cast<Player*>(player_ptr);
+    if (player->is_dying() || player->is_dead())
```
tobbi, line 738: Same for this here.

PR #2628, opened 2023-09-06
Set background color on lightning

### LINKED ISSUE #1243: Level "Shocking" is a bit unfair w/o sound

##### The problem
In the level Shocking from the Forest World, the water occassionally turns deadly because of lightning.

There is a thunder sound played shortly before the water hurts you.

But what if sound is turned off? Then there is no such warning. That's a problem.

tobbi:
Fixes #1243

### DIFF
--- a/src/object/thunderstorm.cpp
+++ b/src/object/thunderstorm.cpp
@@ -48,9 +48,7 @@ Thunderstorm::Thunderstorm(const ReaderMapping& reader) :
   flash_display_timer(),
   changing_tiles(TileManager::current()->get_tileset(Level::current()->get_tileset())->m_thunderstorm_tiles),
   last_ambient_color(),
-  last_layer(),
-  flash_layer(),
-  flash_alpha()
+  last_layer()
 {
   reader.get("running", running);
   reader.get("interval", interval);
@@ -100,14 +98,8 @@ Thunderstorm::update(float )
     lightning();
     time_to_thunder.start(interval);
   }
-  if(flash_display_timer.started()) {
-    flash_alpha = 1.0f;
-    if(flash_display_timer.get_timegone() > 0.4f)
-      flash_alpha = 1.0f - (flash_display_timer.get_timegone() / flash_display_timer.get_timeleft() - 0.4f);
-  }
-
-  if(flash_alpha <= 0.0f) {
-    flash_display_timer.stop();
+  if(flash_display_timer.check()) {
+    log_warning << "Reset ambient light" << std::endl;
     Sector::current()->get_singleton_by_type<AmbientLight>()
       .set_ambient_light(last_ambient_color);
   }
@@ -117,11 +109,23 @@ void
 Thunderstorm::draw(DrawingContext& context)
 {
   if (!flash_display_timer.started()) return;
-  auto color = Color(1, 1, 1, flash_alpha);
+
+  float alpha = 1.0f;
+  if(flash_display_timer.get_timegone() > 0.4f)
+     alpha = 1.0f - (flash_display_timer.get_timegone() / flash_display_timer.get_timeleft() - 0.4f);
+  
+  auto color = Color(1, 1, 1, alpha);
+  auto _layer = Sector::current()->get_foremost_layer() + 1;
+
+  if(alpha < 0.0f) {
+    flash_display_timer.stop();
+    Sector::current()->get_singleton_by_type<AmbientLight>()
+      .set_ambient_light(last_ambient_color);
+  }
 
   context.push_transform();
   context.set_translation(Vector(0, 0));
-  context.color().draw_filled_rect(context.get_rect(), color, flash_layer, Blend::BLEND);
+  context.color().draw_filled_rect(context.get_rect(), color, _layer, Blend::BLEND);
   context.pop_transform();
 }
 
@@ -161,7 +165,6 @@ Thunderstorm::lightning()
   }
 
   change_background_colors(true);
-  flash_layer = Sector::current()->get_foremost_layer() + 1;
   Sector::current()->get_singleton_by_type<AmbientLight>().set_ambient_light(Color::WHITE);
 }
 
--- a/src/object/thunderstorm.cpp
+++ b/src/object/thunderstorm.cpp
@@ -126,6 +126,7 @@ Thunderstorm::draw(DrawingContext& context)
 
   context.push_transform();
   context.set_translation(Vector(0, 0));
+  context.transform().scale = 1.f;
   context.color().draw_gradient(m_flash_color, m_flash_color, 500, GradientDirection::HORIZONTAL, context.get_rect(), Blend::ADD);
   context.pop_transform();
 }
--- a/src/object/thunderstorm.cpp
+++ b/src/object/thunderstorm.cpp
@@ -18,6 +18,7 @@
 
 #include "audio/sound_manager.hpp"
 #include "editor/editor.hpp"
+#include "object/ambient_light.hpp"
 #include "object/background.hpp"
 #include "object/electrifier.hpp"
 #include "supertux/level.hpp"
@@ -95,6 +96,10 @@ Thunderstorm::update(float )
     lightning();
     time_to_thunder.start(interval);
   }
+  if(flash_display_timer.check()) {
+    Sector::current()->get_singleton_by_type<AmbientLight>()
+      .set_ambient_light(last_ambient_color);
+  }
 }
 
 void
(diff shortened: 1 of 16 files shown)


tobbi:
https://github.com/SuperTux/supertux/assets/583169/20867e9b-7244-4ccc-b9fb-a32b7b13845e

tobbi:
https://github.com/SuperTux/supertux/assets/583169/59f2b905-d870-4d70-985a-aca4a1151211

Rusty-Box:
Not sure how far we can go in terms of brighting up the screen, but the flash should be very bright! Perhaps working with something on an "additive layer" would help out. I also think that it looks much nicer if the flash appears immideatly upon lighting strike but then fades out rather than dissappearing instantly again.

Rusty-Box:
Ideally would be to having something like this:

lightning_before_after

tobbi:
https://cdn.discordapp.com/attachments/1033726844660351026/1195497037022646393/2024-01-12_23-37-31.mp4?ex=65b43460&is=65a1bf60&hm=b6bf7468679146e4310aab59ec96144d00245653cb7985e0d269b30d8295c7ed&

Vankata453:
Personally I find the thunder too bright, making it uneasy on the eyes. Would there be an option to turn this off?

src/object/thunderstorm.cpp
```diff
@@ -141,6 +153,10 @@ Thunderstorm::lightning()
   if (!m_strike_script.empty()) {
 	  Sector::get().run_script(m_strike_script, "strike-script");
   }
+
+  change_background_colors(true);
+  Sector::current()->get_singleton_by_type<AmbientLight>().set_ambient_light(Color::LIGHTNING_HIT_COLOR);
+  layer = std::numeric_limits<int>::max();
```
MatusGuy, line 159: There's an enum for layers. It's in layer.hpp. You should use, use LAYER_LIGHTMAP.

src/video/color.cpp
```diff
@@ -26,6 +26,7 @@ const Color Color::CYAN(0.0, 1.0, 1.0);
 const Color Color::MAGENTA(1.0, 0.0, 1.0);
 const Color Color::YELLOW(1.0, 1.0, 0.0);
 const Color Color::WHITE(1.0, 1.0, 1.0);
+const Color Color::LIGHTNING_HIT_COLOR(0.6f, 0.6f, 0.6f, 0.5f);
```
MatusGuy, line 29: why is this here??? This goes in thunderstorm.cpp

MatusGuy (changes requested):

tobbi:
> Personally I find the thunder too bright, making it uneasy on the eyes. Would there be an option to turn this off?

This was Rusty's idea. Battle with him.

Rusty-Box:
It appears the white surface drawn when lightning strikes isn't even colored white. It looks rather gray. Even when the ambient light is set to r=1, g=1, b=1. Is its alpha value set to 1 when it flashes?

Rusty-Box:
Ideally, if we get the flash to be white we may not even have to tamper with the ambient light at all during the flash as the blend mode "add" brightens it all up anyways (unless I am mistaking, though that's how it works with a pure white gradient set to "add" and a z-layer of 500).

tobbi:
It starts at 1 and then quickly levels off to 0.

Rusty-Box:
Huh, it doesn't look like it though at all. I did notice something else though too. You should remove the stuff related to `const Color Color::LIGHTNING_HIT_COLOR(0.6, 0.6, 0.6, 0.5);` because it actually makes a fully bright sector darker when lightning strikes

Rusty-Box:
Not sure if that fixes all of it though since the white surfaces appears as not white even in a fully bright sector too.

Rusty-Box:
This is how it is supposed to look like. To give you a visual reference!
_Note: I used a scripted Gradient object for this!_




https://github.com/SuperTux/supertux/assets/20615048/86e72405-97d5-4da0-95d9-184b5f44761a

MatusGuy (approved):

tobbi:
https://github.com/SuperTux/supertux/assets/583169/ec82e7d8-983e-4ef2-ac3b-ce26f21b373f

Rusty-Box:
I'm still convied that the blending mode didn't really apply. For comparison I set up a Gradient Object with the same values as the drawn lighning flash you added (and on the same z-layer -> 500). But as you can see it is not the same. I don't know why and rn my only assuption is that the code may not support blending modes for you methode of drawing it.

Gradient set to r:1, g:1, b:1, a:0.9 and blend mode "Additive":
screenshot000011

Lighning Flash in this PR, which is supposedly also "Additive":
screenshot000012

src/object/thunderstorm.cpp
```diff
@@ -30,9 +32,8 @@
 namespace {
 
 const float LIGHTNING_DELAY = 2.0f;
-const float FLASH_DISPLAY_TIME = 0.1f;
+const float FLASH_DISPLAY_TIME = 2.0f;
```
MatusGuy, line 35: two seconds? What did this mean again?

src/object/thunderstorm.hpp
```diff
@@ -70,18 +70,22 @@ class Thunderstorm final : public GameObject,
...
-  
+
   std::string m_strike_script;
 
   Timer time_to_thunder; /**< counts down until next thunder */
   Timer time_to_lightning; /**< counts down until next lightning */
   Timer flash_display_timer; /**< counts down while flash is displayed */
 
   std::map<uint32_t, uint32_t> changing_tiles; /**< preserves the tiles which an electrocution should change */
+  Color flash_color;
```
MatusGuy, line 88: add m_ prefix

MatusGuy (approved):
I think this code could be much better, but this works

tobbi:
> I think this code could be much better, but this works

In what way could the code be better?

src/object/thunderstorm.cpp
tobbi: flash_display_timer; /**< counts down while flash is displayed */ So, yeah, that is the time that the flash is displayed.

tobbi:
https://cdn.discordapp.com/attachments/790381979128168448/1213525190949077073/2024-03-02_17-35-20.mp4?ex=65f5ca67&is=65e35567&hm=a34f7ed66a5cd0bedd383441bfc2674334a9631fcbb9bcfa7d2d77d371618ea5&

Rusty-Box (approved):
There we go.

mrkubax10 (approved):

src/object/thunderstorm.cpp
```diff
@@ -94,19 +96,35 @@ Thunderstorm::update(float )
...
 void
 Thunderstorm::draw(DrawingContext& context)
 {
   if (!flash_display_timer.started()) return;
 
-  float alpha = 0.33f;
   context.push_transform();
   context.set_translation(Vector(0, 0));
-  context.color().draw_filled_rect(context.get_rect(), Color(1, 1, 1, alpha), layer);
+  context.color().draw_gradient(m_flash_color, m_flash_color, 500, GradientDirection::HORIZONTAL, context.get_rect(), Blend::ADD);
```
Vankata453, line 126: Before the gradient is drawn, the `DrawingContext` transform scale should be set to 1:

```
context.transform().scale = 1.f;
```

Doing so will prevent #2770.

src/object/thunderstorm.cpp
```diff
@@ -94,19 +96,35 @@ Thunderstorm::update(float )
     lightning();
     time_to_thunder.start(interval);
   }
+
+  if(flash_display_timer.started()) {
```
Vankata453, line 100: Per `CODINGSTYLE.md`:

```
if (foo)
{
  long();
  complex();
  series();
  of();
  calls();
}
```
(first brace on separate line)

src/video/canvas.cpp
```diff
@@ -264,17 +264,19 @@ Canvas::draw_gradient(const Color& top, const Color& bottom, int layer,
 
 void
 Canvas::draw_filled_rect(const Rectf& rect, const Color& color,
-                         int layer)
+                         int layer, const Blend& blend)
```
Vankata453, line 267: This argument should be removed, since:

1. It's not passed, so it's unneeded.
2. Even if it was passed, it'd be better to use the other, more advanced `Canvas::draw_filled_rect()` overload for providing a blend.

Vankata453 (changes requested):

src/object/thunderstorm.cpp
```diff
@@ -18,6 +18,8 @@
 
 #include "audio/sound_manager.hpp"
 #include "editor/editor.hpp"
+#include "object/ambient_light.hpp"
```
Vankata453, line 21: Is this include used?

tobbi: Not anymore it ain't.

Vankata453 (approved):
Codewise looks good.

PR #2631, opened 2023-09-08
Raycast function returns what it hit

MatusGuy:


pazkero:
LGTM, for what little good it does.

MatusGuy:
little good??? I need this.

HybridDog:
If I understand this change correctly, from a user perspective it does nothing. Why do you need this change?

MatusGuy:
I'm trying to change zeekling's behaviour in order to make it more predictable: #2624 
and for that im doing raycasting

pazkero:
> little good??? I need this.

No, I mean my opinion does you little good, you need mrkubax or Vankata or Tobbi or any other C++ coder (up to and including Semphriss, I assume) to this go ahead, the fact I approve is not going to help you.

MatusGuy:
ah sorry for misunderstanding ur right

src/collision/collision_system.cpp
```diff
@@ -690,8 +690,11 @@ CollisionSystem::is_free_of_movingstatics(const Rectf& rect, const CollisionObje
   return true;
 }
 
-bool
-CollisionSystem::free_line_of_sight(const Vector& line_start, const Vector& line_end, bool ignore_objects, const CollisionObject* ignore_object) const
+CollisionSystem::RaycastResult
+CollisionSystem::get_first_line_intersection(const Vector &line_start,
+                                             const Vector &line_end,
```
Vankata453, line 695: `Vector&`

src/collision/collision_system.cpp
```diff
@@ -690,8 +690,11 @@ CollisionSystem::is_free_of_movingstatics(const Rectf& rect, const CollisionObje
...
+                                             bool ignore_objects,
+                                             const CollisionObject *ignore_object) const
```
Vankata453, line 697: `CollisionObject*`

src/collision/collision_system.hpp
```diff
@@ -56,6 +56,20 @@ class CollisionSystem final
...
   bool is_free_of_movingstatics(const Rectf& rect, const CollisionObject* ignore_object) const;
+
+  struct RaycastResult
+  {
+    const bool is_valid; /**< true if raycast hit something */
+    const bool is_tile = true; /**< true if raycast hit a tile */
+    const CollisionObject* object = nullptr; /**< object that the raycast hit, nullptr if is_tile is true */
+    const Tile* tile = nullptr; /**< tile that the raycast hit, nullptr if is_tile is false */
+    const Rectf tile_box = {}; /**< hitbox of tile, empty if is_tile is false */
+  };
```
Vankata453, line 67: Move the struct separately from the functions. For example, in a `public:` field right in the beginning of the class.

src/collision/collision_system.hpp
```diff
@@ -56,6 +56,20 @@ class CollisionSystem final
   bool is_free_of_tiles(const Rectf& rect, const bool ignoreUnisolid = false, uint32_t tiletype = Tile::SOLID) const;
   bool is_free_of_statics(const Rectf& rect, const CollisionObject* ignore_object, const bool ignoreUnisolid) const;
   bool is_free_of_movingstatics(const Rectf& rect, const CollisionObject* ignore_object) const;
+
+  struct RaycastResult
+  {
+    const bool is_valid; /**< true if raycast hit something */
+    const bool is_tile = true; /**< true if raycast hit a tile */
+    const CollisionObject* object = nullptr; /**< object that the raycast hit, nullptr if is_tile is true */
+    const Tile* tile = nullptr; /**< tile that the raycast hit, nullptr if is_tile is false */
```
mrkubax10, line 65: Since ray can hit either tile or object but not both at the same time, this can be replaced with `union`.

MatusGuy: How would I use this with initializer lists?

mrkubax10: You probably can't. Just initialize struct instance the normal way.

mrkubax10:
Apparently designated initializers (i.e. `.tile = tile`) are available since C++20.

MatusGuy:
huh???? ive been doing c++17??? How do i initialize then? Every other way has resulted in compiler error. Any examples?

mrkubax10:
You are probably using GCC which is much less pedantic about standard than for example Clang or MSVC.

mrkubax10:
> Any examples?

First define the `union` as not anonymous:
```cpp
union RaycastHit
{
  const CollisionObject* object;
  const Tile* tile;
};

struct RaycastResult
...
```
Then use it in `struct RaycastResult`:
```cpp
RaycastHit hit;
```
Initialize it the normal way (for example):
```cpp
RaycastHit hit;
hit.tile = tile;
```
And use in return value:
```cpp
return {true, hit, tilebox};
```

MatusGuy:
> You are probably using GCC which is much less pedantic about standard than for example Clang or MSVC (though it tends to add it's own weird features lol).

I believe i was using clang at the time

mrkubax10:
> I see. Do they get initialized as nullptr by default?

Since all `union` members use the **same** memory (they are overlapping), only one of them is valid which means that you can't initialize other members.

src/collision/collision_system.cpp
```diff
@@ -690,10 +690,15 @@ CollisionSystem::is_free_of_movingstatics(const Rectf& rect, const CollisionObje
...
 {
   using namespace collision;
+  using RaycastResult = CollisionSystem::RaycastResult;
```
Vankata453, line 700: This is unneeded, since this is function is in the `CollisionSystem` class.

src/collision/collision_system.hpp
```diff
@@ -35,6 +35,17 @@ class Sector;
...
+  {
+    bool is_valid; /**< true if raycast hit something */
+    union
+    {
+      const CollisionObject* object;
+      const Tile* tile;
+    } hit; /**< tile/object that the raycast hit */
+    Rectf box = {}; /**< hitbox of tile/object */
+  };
+
```
Vankata453, line 48: Add another `public:` down here to separate from functions.

Vankata453 (approved):

mrkubax10 (approved):

PR #2633, opened 2023-09-09
Make object option generic

### LINKED ISSUE #2632: Ability to get internal pointer from an object option
for example i want to get the contents of a ScriptObjectOption and i dont see any way to do this

tobbi:
"Sponges grow in the ocean. That just kills me. I wonder how much deeper the ocean would be if that didn't happen." - Steve Wright (from https://www.brainyquote.com/authors/steven-wright-quotes)

Fixes #2632

### DIFF
--- a/src/editor/object_option.hpp
+++ b/src/editor/object_option.hpp
@@ -53,9 +53,7 @@ class BaseObjectOption
 {
 public:
   BaseObjectOption(const std::string& text, const std::string& key, unsigned int flags);
-  virtual ~BaseObjectOption()
-  {
-  }
+  virtual ~BaseObjectOption() = default;
 
   virtual void save(Writer& write) const = 0;
   virtual std::string to_string() const = 0;
@@ -80,9 +78,7 @@ class ObjectOption: public BaseObjectOption
 {
 public:
   ObjectOption(const std::string& text, const std::string& key, unsigned int flags, T* pointer = nullptr);
-  virtual ~ObjectOption() override
-  {
-  }
+  virtual ~ObjectOption() override = default;
 
   virtual T* get_value() const { return m_value_pointer; }
 
@@ -94,7 +90,7 @@ private:
   ObjectOption& operator=(const ObjectOption&) = delete;
 };
 
-class BoolObjectOption : public ObjectOption<bool>
+class BoolObjectOption final : public ObjectOption<bool>
 {
 public:
   BoolObjectOption(const std::string& text, bool* pointer, const std::string& key,
@@ -113,7 +109,7 @@ private:
   BoolObjectOption& operator=(const BoolObjectOption&) = delete;
 };
 
-class IntObjectOption : public ObjectOption<int>
+class IntObjectOption final : public ObjectOption<int>
 {
 public:
   IntObjectOption(const std::string& text, int* pointer, const std::string& key,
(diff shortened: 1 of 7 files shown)


src/editor/object_option.hpp
```diff
@@ -68,12 +68,29 @@ class ObjectOption
   const std::string m_key;
   const unsigned int m_flags;
 
+private:
+  BaseObjectOption(const BaseObjectOption&) = delete;
+  BaseObjectOption& operator=(const BaseObjectOption&) = delete;
+};
+
+template <typename T>
```
MatusGuy, line 76: [You can use default template arguments.](https://stackoverflow.com/questions/15373823/how-to-avoid-specifying-arguments-for-a-class-template-with-default-template-arg)
You might need to change in other places, not sure

MatusGuy:
tobbi has the best commit messages

src/editor/object_option.hpp
```diff
@@ -68,12 +70,31 @@ class ObjectOption
...
+protected:
+  T* const m_value_pointer;
+
 private:
   ObjectOption(const ObjectOption&) = delete;
   ObjectOption& operator=(const ObjectOption&) = delete;
 };
 
-class BoolObjectOption : public ObjectOption
+class BoolObjectOption : public ObjectOption<bool>
```
Vankata453, line 97: Just a suggestion, but since this PR edits all object options, `final` could be added to the class declarations.

src/editor/object_option.hpp
```diff
@@ -49,11 +49,13 @@ class Rectf;
...
-class ObjectOption
+class BaseObjectOption
 {
 public:
-  ObjectOption(const std::string& text, const std::string& key, unsigned int flags);
-  virtual ~ObjectOption();
+  BaseObjectOption(const std::string& text, const std::string& key, unsigned int flags);
+  virtual ~BaseObjectOption()
+  {
+  }
```
Vankata453, line 58: Put this in the `.cpp`, or remove it.

src/editor/object_option.hpp
```diff
@@ -68,12 +70,31 @@ class ObjectOption
...
+};
+
+template <typename T = void>
+class ObjectOption: public BaseObjectOption
+{
+public:
+  ObjectOption(const std::string& text, const std::string& key, unsigned int flags, T* pointer = nullptr);
+  virtual ~ObjectOption() override
+  {
+  }
```
Vankata453, line 85: Same as previous comment.

src/editor/object_option.hpp
tobbi: I can't, otherwise compilation errors.

src/editor/object_option.hpp
tobbi: I can't, otherwise compilation errors.

src/editor/object_option.hpp
Vankata453: Could you then put the `{}` right next to the declaration's line?

mrkubax10: `virtual ~BaseObjectOption() = default;` in .hpp

MatusGuy (approved):

mrkubax10 (commented):
Apply @Vankata453's suggestions and it will be good to merge.

Vankata453 (approved):

mrkubax10 (approved):

PR #2637, opened 2023-09-15
Add + Overlay new bomb tick sound (4s) onto existing fizz.wav

Alasdairbugs:
Added bomb tick sound (plays for 4s), and gradually but systematically increases in tempo to alert the player of an impending explosion from the bomb enemy.

This is to be overlaid/played in conjunction with the first instance of fizz.wav being played (This needs coding). Idk how to do that, so some small help would be appreciated 👍


bruhmoent:
Since this is a file name change, I'm certain it will require code changes.

Alasdairbugs:
https://github.com/SuperTux/supertux/assets/41154567/8cedff9f-54d7-433e-bba2-1d1830633e12

- fizz.wav plays once (unsure how I feel about this, give opinion via video)
- the tick sound ends about 0.4s short (does anyone know how long the bomb is fused for before exploding?). I can adjust this.

to do:

1. make the tempo for the tick sound to be longer to better fit the fuse length,
2. discuss if the fizz.wav should be continuously played during the tick noise (i think it should, im pretty sure this is what it used to be).

Rusty-Box:
Would it not be smarter to make the sound file be a single “beep” and play it several times depending on how long the ticking phase is?

Alasdairbugs:
I think the several beeps, which give a sense of tempo, makes it more intuitive as to when the bomb is about to explode. Single beeps would not be as useful of information to the player, just indicating when the animation is changing in frequency.

But now that you mention it, I could fit the tempo changes to when the animation increases in frequency - can you give details on the length of each phase and how many flashes there are?

mrkubax10:
Playing short sound multiple times may (and definitely will) introduce some sort of lag.

Rusty-Box:
Currently not at home. I can tell you later. Perhaps making two sound files then (one for each flash intensity) and divide the action in two? That way we can play each sound based on each phase but also have control over how long either phase should be (should changes be made there) are and the sound still lining up

Alasdairbugs:
tick.ogg is one sound file - unsure how this will create lag? besides, we're not talking about increasing the amount of short sounds - i could decrease the tempo, which would actually mean less "short sound[s]" are played within a space of time.

making two sound files is probably best - more bomb types are being added to the game, which im assuming have differing fuse lengths with different length phases...? hmm

mrkubax10:
What I meant is playing **one short file** multiple times, not that sound data contains short beeps lol.

Rusty-Box:
> making two sound files is probably best - more bomb types are being added to the game, which im assuming have differing fuse lengths with different length phases...? hmm

Yes. Most likely Rusty Bomb and, perhaps, Old Bomb get a much shorter ticking time. Also Haywire and Hellbent don’t have a faster ticking phase so having two sound files also works great with them

Alasdairbugs:
I'm gonna move this to 0.8/0.9 roadmap when all bomb types are fully added with their blink cycles outlined more clearly, as each bomb is going to have unique ticking phases.

PR #2638, opened 2023-09-15
Fix awkward words

maldron0309:
"The words were technically correct but felt a bit awkward as they were direct translations. Therefore, we have decided to revert them back to their original English expressions. For instance, '세계지도' has been changed to '월드맵', and '세계' has been changed to '월드'."


Vankata453:
Thank you! However, please use [Transifex](https://www.transifex.com/arctic-games/supertux/) to fix or report translation related issues.

PR #2642, opened 2023-09-22
Delete *unnecessary* hpp file.

tylerandari13:
~~yes i made an entire pull request to delete one file~~

This deletes `src/scripting/worldmap.hpp`. The file serves no purpose and only creates an *unnecessary* entry in the scripting reference.


Vankata453:
I was just trying to delete it through the web GitHub interface. Will merge this shortly.

PR #2646, opened 2023-09-25
Check if `MovingSprite` type change call is initial

### LINKED ISSUE #1342: Spring sprite changing back in editor

### LINKED ISSUE #2701: Unable to set custom sprites to springs
0.7 Nighties
Expected Behavior - 
Spring will apply the texture given in the editor to the spring in-game.
Actual Behavior -
Spring will automatically revert back to the default sprite depending on object type, AKA the sprite change does not apply in game.

provigz:
Removes an old `MovingSprite::has_found_sprite()` hack, which allowed getting whether a custom sprite has been found only the first time it's called. After the first call, `m_sprite_found` always equaled to `false`.

`MovingSprite::on_type_change` now checks if the call is initial, and if so, only sets the default sprite for the type, if a custom sprite hasn't just been loaded.

Closes #1342. Fixes #2701.

### DIFF
--- a/src/badguy/corrupted_granito.cpp
+++ b/src/badguy/corrupted_granito.cpp
@@ -34,6 +34,13 @@ CorruptedGranito::CorruptedGranito(const ReaderMapping& reader) :
   SoundManager::current()->preload(CORRUPTED_GRANITO_SOUND);
 }
 
+CorruptedGranito::CorruptedGranito(const ReaderMapping& reader, int type) :
+  CorruptedGranito(reader)
+{
+  m_type = type;
+  on_type_change(TypeChange::INITIAL);
+}
+
 void
 CorruptedGranito::initialize()
 {
--- a/src/badguy/corrupted_granito.hpp
+++ b/src/badguy/corrupted_granito.hpp
@@ -29,6 +29,7 @@ public:
 
 public:
   CorruptedGranito(const ReaderMapping& reader);
+  CorruptedGranito(const ReaderMapping& reader, int type);
 
   virtual void initialize() override;
   virtual void collision_solid(const CollisionHit& hit) override;
--- a/src/supertux/game_object.cpp
+++ b/src/supertux/game_object.cpp
@@ -207,7 +207,7 @@ GameObject::parse_type(const ReaderMapping& reader)
   {
     try
     {
-      set_type(type_id_to_value(type));
+      m_type = type_id_to_value(type);
     }
     catch (...)
     {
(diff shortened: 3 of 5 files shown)


src/object/moving_sprite.hpp
```diff
@@ -53,11 +53,11 @@ class MovingSprite : public MovingObject
 
   virtual ObjectSettings get_settings() override;
   virtual void after_editor_set() override;
-  virtual void on_type_change(int old_type = -1) override;
+  virtual void on_type_change(int old_type) override;
```
tobbi, line 56: Technically, you can leave the default. IDK...

tobbi: The whole semantic of that function is...weird

Vankata453: Removed the default so someone editing the code can't accidentally call it with `old_type` as `-1`, which it will detect as an initial call.

tobbi: Pain!

Could you not add an m_type property to MovingObject, initialize that to -1 and check against that, then pass the current type as a value?

Semantically, this would make more sense.

Vankata453: You mean `MovingSprite`? Honestly, don't think adding another variable just for this, which is a copy of `GameObject::m_type` would make sense, neither would it be better for memory. What I'm thinking of is I could add a comment somewhere indicating values for `old_type`, which are below 0, should only be passed on object initialization?

tobbi: Can we check against m_type?

Vankata453: Another solution is to maybe initialize `GameObject::m_type` as -1 and then in `parse_type` set that to 0? Though that would also be a problem, since some objects call `parse_type` multiple times, because of inheritance.

MatusGuy (approved):

tobbi:
Can we move -1 to a constant with a proper name?

Vankata453:
> Can we move -1 to a constant with a proper name?

I've replaced -1 with the `TypeChange::INITIAL` enumerator.

Vankata453:
@tobbi Can you re-review this one?

src/badguy/crusher.cpp
```diff
@@ -115,7 +115,7 @@ Crusher::on_type_change(int old_type)
       break;
   }
 
-  MovingSprite::on_type_change();
+  MovingSprite::on_type_change(old_type);
```
tobbi, line 118: Can you move this up to line 101 for consistency?

mrkubax10 (approved):

PR #2647, opened 2023-09-25
Unisolid hitboxes for objects

### LINKED ISSUE #2644: Lack of unisolid hitboxes for GameObjects
I know the collision system is at a quite... weird state... not sure if we can add major features properly. But...

The current system lacks unisolid hitboxes for `GameObject`s.
This could be used, for example, in #2635, to add the ability to stand on top of the Granito.

Vankata453:
`CollisionObject` now contains a unisolid property. If it's `true`, only bottom constraints to the top of the bounding box would be applied to objects, colliding with that object.

The property is also settable from sprite file actions with the name `unisolid`. Default is `false`.

Fixes #2644.

### DIFF
--- a/src/collision/collision_object.cpp
+++ b/src/collision/collision_object.cpp
@@ -27,6 +27,7 @@ CollisionObject::CollisionObject(CollisionGroup group, CollisionListener& listen
   m_group(group),
   m_movement(0.0f, 0.0f),
   m_dest(),
+  m_unisolid(false),
   m_objects_hit_bottom(),
   m_ground_movement_manager(nullptr)
 {
--- a/src/sprite/sprite_data.cpp
+++ b/src/sprite/sprite_data.cpp
@@ -35,6 +35,7 @@ SpriteData::Action::Action() :
   y_offset(0),
   hitbox_w(0),
   hitbox_h(0),
+  hitbox_unisolid(false),
   fps(10),
   loops(-1),
   loop_frame(1),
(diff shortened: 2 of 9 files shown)


src/sprite/sprite_data.cpp
```diff
@@ -127,6 +128,7 @@ SpriteData::parse_action(const ReaderMapping& mapping)
         throw std::runtime_error("hitbox should specify 2/4 coordinates");
     }
   }
+  mapping.get("unisolid", action->hitbox_unisolid);
```
tobbi, line 131: Does this need a default?

Vankata453: It's already `false`, check the `SpriteData::Action` constructor.

src/collision/collision_object.cpp
```diff
@@ -27,6 +27,7 @@ CollisionObject::CollisionObject(CollisionGroup group, CollisionListener& listen
   m_group(group),
   m_movement(0.0f, 0.0f),
   m_dest(),
+  m_unisolid(false),
```
tobbi, line 30: No need for 'false' here, it's the default anyway.

src/sprite/sprite_data.cpp
```diff
@@ -35,6 +35,7 @@ SpriteData::Action::Action() :
   y_offset(0),
   hitbox_w(0),
   hitbox_h(0),
+  hitbox_unisolid(false),
```
tobbi, line 38: Same here.

tobbi (changes requested):
You need to leave a comment indicating the requested changes.

src/collision/collision_object.cpp
Vankata453: Usually I prefer to specify, so you can see it's the default. But I can remove it too.

tobbi: fine.

tobbi (approved):

PR #2650, opened 2023-09-30
Add `srand()` to scripting API.

tylerandari13:
`srand()` allows you set the seed that `rand()` will use.

Also its not called something like `set_seed()` because its called srand in Squirrel. (see http://squirrel-lang.org/squirreldoc/stdlib/stdmathlib.html#srand)

### DIFF
--- a/src/scripting/functions.cpp
+++ b/src/scripting/functions.cpp
@@ -444,6 +444,11 @@ int rand()
   return gameRandom.rand();
 }
 
+void srand(int v)
+{
+  return gameRandom.seed(v);
+}
+
 void set_game_speed(float speed)
 {
   if (speed < 0.05f)
--- a/src/scripting/functions.hpp
+++ b/src/scripting/functions.hpp
@@ -287,6 +287,11 @@ void set_gamma(float gamma);
 int rand();
 
 /**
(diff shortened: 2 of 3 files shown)


tobbi:
I guess it makes sense. Just wondering why we don't use the Squirrel implementation of that in the first place.

Vankata453:
We don't use the Squirrel implementation probably, because randomizing from scripts is intended to be a part of the game's randomization flow? So when you replay a level which uses `rand()` with the same seed as before, you get the same randomizations.

Vankata453:
Also, on the topic of this PR, I'm not really sure what the purpose for this function would be, because it would essentially make randomizations be the same, if set with the same seed. Of course, the seed could be set from time (did Squirrel provide a function for that?), though the game does that on startup anyways, so I don't know why you'd want to do that.

An additional thing to point out is that this function would set the randomization seed globally for the game, which means it will affect other levels as well. However, that wouldn't be a problem, I believe.

tylerandari13:
Closing this. Discussed this on the Discord and tbh theres not much to gain from an srand function, and theres another issue regarding the rand function bringing the same random numbers every time you start the game, so i might (emphasis on *might*) look into that in another PR.

PR #2651, opened 2023-09-30
Fix setting random seed to current time

Vankata453:
Setting the random seed to 0 or lower of a `Random` generator will now set it to the UNIX timestamp of the current time.

### DIFF
--- a/src/math/random.cpp
+++ b/src/math/random.cpp
@@ -31,11 +31,8 @@ void
 Random::seed(int v)
 {
   if (v <= 0)
-  {
-    // Use the UNIX timestamp of the current time as a seed.
-    m_generator.seed(static_cast<unsigned int>(std::time(nullptr)));
-    return;
-  }
+    v = static_cast<int>(std::time(nullptr)); // Use the UNIX timestamp of the current time as a seed.
+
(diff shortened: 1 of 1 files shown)


tobbi:
```
Users/runner/work/supertux/supertux/src/math/random.cpp:36:22: error: implicit conversion loses integer precision: 'time_t' (aka 'long') to 'std::mersenne_twister_engine<unsigned int, 32, 624, 397, 31, 2567483615, 11, 4294967295, 7, 2636928640, 15, 4022730752, 18, 1812433253>::result_type' (aka 'unsigned int') [-Werror,-Wshorten-64-to-32]
    m_generator.seed(std::time(nullptr));
```

mrkubax10 (approved):

tobbi (approved):
Yeah, although you could've just assigned to v directly without that `return`.

mrkubax10:
> Yeah, although you could've just assigned to v directly without that `return`.

True

Vankata453:
> Yeah, although you could've just assigned to v directly without that `return`.

Indeed, realized that later, but I thought it's for better to directly cast to `unsigned int`.

PR #2653, opened 2023-10-02
Avoid drawing unnecessary things when we're about to quit

tobbi:
Also, don't update integrations for every tick, that's way too often and wastes CPU cycles.

### DIFF
--- a/src/supertux/screen_manager.cpp
+++ b/src/supertux/screen_manager.cpp
@@ -146,7 +146,8 @@ ScreenManager::ScreenManager(VideoSystem& video_system, InputManager& input_mana
   m_speed(1.0),
   m_actions(),
   m_screen_fade(),
-  m_screen_stack()
+  m_screen_stack(),
+  m_running()
 {
 }
 
@@ -200,6 +201,7 @@ ScreenManager::quit(std::unique_ptr<ScreenFade> screen_fade)
     m_screen_fade = std::move(screen_fade);
   }
   m_actions.emplace_back(Action::QUIT_ACTION);
+  m_running = false;
 }
 
 void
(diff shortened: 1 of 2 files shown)


Vankata453:
Is there a case when `ScreenManager` isn't running, other than when the game is being quit?

tobbi:
> Is there a case when `ScreenManager` isn't running, other than when the game is being quit?

Not that I know of.

Vankata453:
Additionally, while the game quit screen fade is being performed, wouldn't not further drawing the UI elements make them disappear immediately before the fade starts, causing for a visual inconsistency? I will test this soon to check, this is just an observation from looking at the code.

Vankata453:
Testing shows that now quitting the game with transitions enabled makes the UI disappear before the fade out happens. Considering this didn't happen before the change, and keeping in mind `ScreenManager` is only not running (shouldn't draw UI elements, process sound or events) when quitting the game, I'm not sure whether the changes (other than the integration update ones) provide a significant benefit.

Vankata453:
What is the optimization achieved in this PR? Is it significant enough to warrant not drawing UI elements on screen fade-out?

tobbi:
It's drawing them now

Vankata453:
> It's drawing them now

It's only drawing the menu on fade-out, still hides FPS counter, player position, etc... There should be some benefit from not drawing those UI items to warrant changing their behaviour, and I'm not sure what it is.

tobbi:
Fine, you want the pr closed, I'll close the pr.

Vankata453:
Sorry, I only wanted to ask a question regarding the PR, so I can justify a review. For example, the integration change seems beneficial. For this part I genuinely didn't know what benefit it serves, so I'm asking to hear why you decided to implement it.

tobbi:
Reason I implemented this was because shutdown always felt slow to me

Vankata453:
Did it feel slow when the menu was hidden?

PR #2660, opened 2023-10-13
Expose `get_filename` to scripting under `worldmap.settings`.

tylerandari13:
Think of this as a getter for the worldmap selection thing so you can tell which worldmap youre on. :snowsmiley:

Ive been having an issue with using a script to save info in the state table. Because its all saved to the same table it loads stuff where its not supposed to and it would be really helpful if i could get a string unique to each worldmap to save the data in different tables.

### DIFF
--- a/src/scripting/worldmap_sector.cpp
+++ b/src/scripting/worldmap_sector.cpp
@@ -59,6 +59,13 @@ WorldMapSector::move_to_spawnpoint(const std::string& spawnpoint)
   m_parent->move_to_spawnpoint(spawnpoint);
 }
 
+std::string&
+WorldMapSector::get_filename()
+{
+  SCRIPT_GUARD_WORLDMAP;
+  return worldmap.get_filename();
+}
+
 } // namespace scripting
 
 /* EOF */
--- a/src/scripting/worldmap_sector.cpp
+++ b/src/scripting/worldmap_sector.cpp
@@ -59,8 +59,8 @@ WorldMapSector::move_to_spawnpoint(const std::string& spawnpoint)
   m_parent->move_to_spawnpoint(spawnpoint);
 }
 
-const std::string&
-WorldMapSector::get_filename()
+std::string
+WorldMapSector::get_filename() const
 {
   SCRIPT_GUARD_WORLDMAP;
   return worldmap.get_filename();
--- a/src/scripting/worldmap_sector.cpp
+++ b/src/scripting/worldmap_sector.cpp
@@ -59,7 +59,7 @@ WorldMapSector::move_to_spawnpoint(const std::string& spawnpoint)
   m_parent->move_to_spawnpoint(spawnpoint);
 }
 
-std::string&
+const std::string&
 WorldMapSector::get_filename()
 {
   SCRIPT_GUARD_WORLDMAP;
--- a/src/scripting/worldmap_sector.hpp
+++ b/src/scripting/worldmap_sector.hpp
@@ -87,7 +87,7 @@ public:
   /**
    * Gets the path to the worldmap file. Useful for saving worldmap specific data.
    */
-  const std::string& get_filename();
+  std::string get_filename() const;
 
 };
 
--- a/src/scripting/worldmap_sector.hpp
+++ b/src/scripting/worldmap_sector.hpp
@@ -83,6 +83,12 @@ public:
    * @param string $spawnpoint
    */
   void move_to_spawnpoint(const std::string& spawnpoint);
+
+  /**
+   * Gets the path to the worldmap file. Useful for saving worldmap specific data.
+   */
+  std::string& get_filename();
+
 };
 
 } // namespace scripting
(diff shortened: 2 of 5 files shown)


src/scripting/worldmap_sector.hpp
```diff
@@ -83,6 +83,12 @@ class WorldMapSector final : public GameObjectManager
    * @param string $spawnpoint
    */
   void move_to_spawnpoint(const std::string& spawnpoint);
+
+  /**
+   * Gets the path to the worldmap file. Useful for saving worldmap specific data.
+   */
+  std::string& get_filename();
```
Vankata453, line 90: Make the reference `const`, as well as the method.

src/worldmap/worldmap.hpp
```diff
@@ -86,6 +86,8 @@ class WorldMap final : public Currenton<WorldMap>
   void set_sector(const std::string& name, const std::string& spawnpoint = "",
                   bool perform_full_setup = true);
 
+  std::string& get_filename();
```
Vankata453, line 89: Here as well.

Vankata453 (changes requested):

Vankata453:
On the topic of this, having worldmap-sector-scoped state tables may serve as a more universal solution.

tylerandari13:
> On the topic of this, having worldmap-sector-scoped state tables may serve as a more universal solution.

Tbh i dont hate the idea. Would this be a seperate table or would it replace the state table entirely tho?

Vankata453:
> Tbh i dont hate the idea. Would this be a seperate table or would it replace the state table entirely tho?

No, it could be a separate table just scoped to the sector. Will that work instead of scoping it for worldmaps, as you aim to do with this PR?

tylerandari13:
> > Tbh i dont hate the idea. Would this be a seperate table or would it replace the state table entirely tho?
> 
> No, it could be a separate table just scoped to the sector. Will that work instead of scoping it for worldmaps, as you aim to do with this PR?

You mean like a `sector.scoped_table_thing`? If it can allow me to seaprate data from one worldmap to the next then i think itll work.

src/scripting/worldmap_sector.cpp
```diff
@@ -59,6 +59,13 @@ WorldMapSector::move_to_spawnpoint(const std::string& spawnpoint)
   m_parent->move_to_spawnpoint(spawnpoint);
 }
 
+const std::string&
```
tobbi, line 62: "Mom, can we have `const` correctness?"
- "No, son, we have `const` correctness at home"
Const correctness at home:

tobbi: What @tylerandari13 did, has its purpose, even though it's something I rarely see in this form.

a) Returning a reference is potentially less expensive in terms of memory usage and "effort" (a reference uses less memory than a string afaik) but if the thing we return goes out of scope (not applicable in this case) or is garbage-collected (object gets destroyed) (likely), then we potentially have a problem since the reference points to invalid memory / garbage. 

b) `const correctness` is more of a convention than anything. By adding `const` to the end of the function declaration, you tell the compiler that the function does not modify the state of the object in any way (has no side effects). It's something that most compilers already check when compiling the program and if a class tries to modify its state inside a const method, then the compiler will produce an error.

Sorry, this is awfully technical. If you want, I could maybe break it down for you even further.

tobbi:
@tylerandari13 Change the method definition from

`const std::string& get_filename();` to `std::string get_filename() const;` in both the *.cpp and *.hpp file.

src/scripting/worldmap_sector.cpp
Vankata453: > a) Returning a reference is potentially less expensive in terms of memory usage and "effort" (a reference uses less memory than a string afaik) but if the thing we return goes out of scope (not applicable in this case) or is garbage-collected destroyed (set to null) (likely at some point), then we potentially have a problem since the reference points to invalid memory / garbage.

In this case, this can only go out-of-scope, when the worldmap is destroyed, which happens when it's exited. If any code needs to preserve this, in my opinion, it should do the copying, and the function should return a `const` reference, so it's less expensive when it doesn't have to be.

tobbi:
@tylerandari13 Do you understand const correctness now?

tylerandari13:
> @tylerandari13 Do you understand const correctness now?

of course not lmao
attachment

Vankata453:
Wouldn't it be better to return a const reference either way, which would be copied only when necessary?

tylerandari13:
> Wouldn't it be better to return a const reference either way, which would be copied only when necessary?

to this i respond with "i barely know what im doing lmao"

Vankata453:
> to this i respond with "i barely know what im doing lmao"

I'm wondering in general.

src/worldmap/worldmap.cpp
```diff
@@ -334,6 +334,12 @@ WorldMap::set_sector(const std::string& name, const std::string& spawnpoint,
     m_sector->move_to_spawnpoint(spawnpoint);
 }
 
+const std::string&
+WorldMap::get_filename()
+{
```
tobbi, line 339: Please fix it here as well.

Vankata453 (approved):

mrkubax10 (approved):

tobbi (approved):

tylerandari13:
:snowsmiley:

PR #2667, opened 2023-10-24
World-specific custom title screen levels

Vankata453:
Worlds now contain the "title-screen" property, which determines the title screen level, which should play, if this world is the last entered world on the current profile.

Additionally, the title screen level property of a world can be overriden for a certain savegame using the `worldmap.settings.set_title_level(filename)` scripting function.

Custom title screens can be toggled from the "Extras" options category.

Changing profiles or toggling the custom title screens option will now automatically change the title level in the background, if needed, applying a fade transition.

Conditional of #2666.


tobbi:
src/supertux/profile_manager.cpp:47:14: style: Variable 'result' can be declared as reference to const [constVariableReference]
    Profile& result = *profile;

Rusty-Box (approved):

mrkubax10 (approved):

Issue #1898, opened 2021-11-18
Dispenser enemy duplication


weluvgoatz:
In some cases of using a dispenser that has a set max concurrent enemies, the amount of enemies that come out are 1 more than there should be... This only happens sometimes, too; I've played levels where it doesn't happen and ones where it does.

Busybody64:
I think the game just spawns it before it realizes it's already at the max amount of enemies it can spawn.

mrkubax10:
In which levels?

Alasdairbugs:
reproduced. i think this is because the dropper is being loaded offscreen, then reloaded and it thinks it hasnt spawned an object yet?

Alasdairbugs:
additionally, droppers off-screen, revisiting them after a while shows like 20 enemies being spawned or something.

PR #2669, opened 2023-10-30
Fix enemy death script running twice if Tux slides into it.

tylerandari13:
If you slide into an enemy and kill it the death script triggers twice, once when you first hit him and another when hes falls into the void. This should fix the issue by removing the death script the first time the enemy is hit.

### DIFF
--- a/src/badguy/badguy.cpp
+++ b/src/badguy/badguy.cpp
@@ -645,10 +645,14 @@ BadGuy::kill_fall()
 void
 BadGuy::run_dead_script()
 {
+  if(m_is_active_flag) {
+    m_is_active_flag = false;
+  } else {
+    return;
+  }
+
   if (m_countMe)
     Sector::get().get_level().m_stats.increment_badguys();
-
-  m_countMe = false;
 
   if (m_parent_dispenser != nullptr)
   {
@@ -658,7 +662,6 @@ BadGuy::run_dead_script()
   // Start the dead-script.
   if (!m_dead_script.empty()) {
     Sector::get().run_script(m_dead_script, "dead-script");
-    m_dead_script = "";
   }
 }
 
--- a/src/badguy/badguy.cpp
+++ b/src/badguy/badguy.cpp
@@ -658,6 +658,7 @@ BadGuy::run_dead_script()
   // Start the dead-script.
   if (!m_dead_script.empty()) {
     Sector::get().run_script(m_dead_script, "dead-script");
+    m_dead_script = "";
   }
 }
 
--- a/src/badguy/badguy.cpp
+++ b/src/badguy/badguy.cpp
@@ -647,9 +647,8 @@ BadGuy::run_dead_script()
 {
   if(!m_is_active_flag) {
      return;
-  } else {
-     m_is_active_flag = false;
-  }
+
+  m_is_active_flag = false;
 
   if (m_countMe)
     Sector::get().get_level().m_stats.increment_badguys();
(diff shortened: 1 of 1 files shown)


tobbi:
Hang on! Does this only apply to the death script or does the whole function run multiple times?

tylerandari13:
> Hang on! Does this only apply to the death script or does the whole function run multiple times?

Probably the whole function, but there seems to be stuff in the function to stop things from repeating twice.

tobbi:
The proper way is to introduce a new bool variable `m_death_whatever_thingy_run` and set that to true. If you do that, m_countMe doesn't even need to be set to false anymore.

The way I see it, `notify_dead` could still be run multiple times.

tobbi:
Actually, what calls run_dead_script() and why does that call it multiple times?

tylerandari13:
> The proper way is to introduce a new bool variable `m_death_whatever_thingy_run` and set that to true. If you do that, m_countMe doesn't even need to be set to false anymore.
> 
> The way I see it, `notify_dead` could still be run multiple times.

Too bad, im rebellious and improper.

tylerandari13:
> The way I see it, `notify_dead` could still be run multiple times.

I just looked in the code and `notify_dead`  doesnt seem to be related to `m_countMe`. lmao

src/badguy/badguy.cpp
```diff
@@ -645,11 +645,15 @@ BadGuy::kill_fall()
 void
 BadGuy::run_dead_script()
 {
+  if(m_is_active_flag) {
```
tobbi, line 648: If this does end up staying, please invert the control flow.

tylerandari13: I can do that soon, whats the purpose tho?

tobbi: Satisfies my OCD.

tylerandari13: Bruh

tobbi:
```cpp
if(!m_active_flag)
   return;

m_active_flag = false;
```

is easier to read IMO.

tylerandari13: i did the thing maybe

tylerandari13: > ```c++
> if(!m_active_flag)
>    return;
> 
> m_active_flag = false;
> ```
> 
> is easier to read IMO.

somehow this did not load on either the Discord or the Github, give me one more second

tobbi:
...was gonna say

tylerandari13:
> ...was gonna say

:snowsmiley:

tylerandari13:
I was fooling around and this might fix the bug where Mr IceBlock spawns too many times out of a dispenser.

tobbi:
> I was fooling around and this might fix the bug where Mr IceBlock spawns too many times out of a dispenser.

> "The way I see it, notify_dead could still be run multiple times, the way it is currently."

tylerandari13:
> > I was fooling around and this might fix the bug where Mr IceBlock spawns too many times out of a dispenser.
> 
> > "The way I see it, notify_dead could still be run multiple times, the way it is currently."

tbh thats probably what caused it lmao

tobbi:
Does it have an issue number on Github.

tylerandari13:
> Does it have an issue number on Github.

#1898 seems to relate

tobbi (approved):

mrkubax10 (approved):

Vankata453 (approved):

PR #2462, opened 2023-05-05
Remap new tiles to compatible mapping

Semphriss:
image

Fixes #2306

The new tiles have been remapped according to their old IDs.

TODO:
- [x] Remap forest tiles
- [x] Remap ghost forest tiles

PR #2671, opened 2023-11-09
Backwards-compatible tile mappings, tile converters, tile and texture deprecation

### LINKED ISSUE #2306: Recent tileset changes broken
The tileset changes from 2022-10-06 are seemingly broken:

screenshot cleaned

Vankata453:
[Continuation of the now abandoned #2462]

**Reverts the tile IDs** in `tiles.strf` to ones, compatible with the old tile mappings, before the full tile overhaul. Any old tiles, which do not have a respective new version, are now marked as deprecated.

**Tile converters** can now be used to convert all tiles on all tilemaps in a level, from the editor. Contains two stock tile converters (0.5.x Crystal Tiles, Nightly Tiles -> Present), and a new tile conversions menu (accessible from the main level editor menu). The "Nightly Tiles" converter allows for converting from the current Nightly Builds mapping, to this new, final one. The tiles of the main menu and credits levels are converted in this PR.

Introduces **tile deprecation** and various editor locks to prevent using such tiles, as well as show whether/where such tiles are present in a level.

Additionally introduces **texture deprecation**. Any image files, containing `.deprecated` right before the extension in their filename will be used, whenever the main texture (the one, which doesn't contain the `.deprecated` part in its filename) does not exist, taking its place. Deprecated files do not show up on the file browsing menu.

-----

Huge thanks to [Semphris](https://github.com/Semphriss) for all the initial work on this, including quite a lot of tile re-mapping, organizing the full workflow, as well as the initial process of auto-generating new tile IDs and creating converters, and the implementation of the tile conversion algorithm.

Fixes #2306.


tylerandari13 (approved):

src/supertux/menu/editor_converters_menu.cpp
```diff
@@ -0,0 +1,100 @@
...
+}
+
+void
+EditorConvertersMenu::menu_action(MenuItem& item)
+{
+  if (item.get_id() != MNID_CONVERT_TILES)
+  {
+    log_warning << "Invalid item with ID " << item.get_id() << " pressed." << std::endl;
+    return;
+  }
```
mrkubax10, line 85: For cases that shouldn't happen you it is better to use `assert`.

Vankata453: True, though a lot of menus currently have the check like this, so this should be saved for a different commit.

src/supertux/menu/editor_menu.cpp
```diff
@@ -198,6 +225,28 @@ EditorMenu::menu_action(MenuItem& item)
       Editor::current()->m_quit_request = true;
       break;
 
+    case MNID_CHECKDEPRECATEDTILES:
+      editor->check_deprecated_tiles();
+      if (editor->has_deprecated_tiles())
+      {
+        if (g_config->editor_show_deprecated_tiles)
+        {
+          Dialog::show_message(_("Deprecated tiles are still available in the level."));
```
mrkubax10, line 234: Maybe "Deprecated tiles are still present in the level."?

Frostwithasideofsalt (approved):
Autotiles seem to be the old ones but other then that it's all good. the autotiles could probably just be fixed later anyway.

mrkubax10 (approved):

Issue #2617, opened 2023-08-29
.music files should have an "author" entry.


bruhmoent:
Presently, the identity of the composer behind the game's soundtrack might pose as an enigma. Simplifying matters, it would be highly convenient to incorporate an author designation. This could take the form of something like:

(supertux-music
  (file "forest-cave.ogg")
  (author "Marek Moeckel")
  (loop-begin 0)
  (loop-at    149)
)

Vankata453:
In my opinion, this attribution should show up when hovering over `.music` files in the `FileSystemMenu`, if it's available.

tobbi:
I'm not sure. `.music` files are only used to denote start and end.

tylerandari13:
While im not sure if it should be in the file itself there should at least be a spot in the credits for all the music and who made it.


> In my opinion, this attribution should show up when hovering over `.music` files in the `FileSystemMenu`, if it's available.

If it can be pulled off i think it would be cool.

Vankata453:
> I'm not sure. `.music` files are only used to denote start and end.

Should they have to only note that? If necessary, they could provide more details about the music file.

mrkubax10:
> They could provide more details about the music file.

True.

jamescdericco:
I started work on this issue in the branch [2617-music-author](https://github.com/jamescdericco/supertux/tree/2617-music-author) of my fork.

jamescdericco:
I finished making a UI mockup for this. I would like your feedback before starting implementation. Do you all like this or is there anything you prefer added, subtracted or adjusted?

### UI Design Mockup
In the design below, we display music metadata, including author, when hovering over music files in the hover "help text" below the menu:

![Screenshot from 2023-10-26 11-33-52](https://github.com/SuperTux/supertux/assets/26558265/15a4e838-8593-4bf9-9077-5a7e2ce2b9c9)

This "help text" area is currently used by the SuperTux Options submenus which display additional explanation when hovering over options. I'm thinking that we could also use this help text area for displaying music metadata.

I also want to propose displaying the original titles of songs in addition to their authors, where the info is available, so that users can find the original song online, and so authors can display the original name of their work. We can store the title and author metadata as a "title" and "author" fields each ".music" files, and each of these are optional: when the information is not in the .music file, SuperTux will just not display that field in the UI.

If the "title" field is not available, but only "author" is, the hover info could look like this:
[...]

PR #2674, opened 2023-11-11
Display music metadata in the level editor

jamescdericco:
Fixes issue #2617.

When navigating or hovering over menu items in the music settings file browser, information about a song's author, license, and title will now be displayed in the help text at the bottom of the screen.


src/editor/object_option.cpp
```diff
@@ -440,7 +442,7 @@ FileObjectOption::to_string() const
 void
 FileObjectOption::add_to_menu(Menu& menu) const
 {
-  menu.add_file(get_text(), m_value_pointer, m_filter, m_basedir, m_path_relative_to_basedir);
+  menu.add_file(get_text(), m_value_pointer, m_filter, m_basedir, m_path_relative_to_basedir, -1, m_generate_help_text_for_file);
```
tobbi, line 445: `m_generate_help_text_for_file` somehow suggests that this is boolean. No idea if I can think of another name, though.

src/gui/menu_filesystem.cpp
```diff
@@ -98,7 +102,13 @@ FileSystemMenu::refresh_items()
 
   for (const auto& item : m_files)
   {
-    add_entry(item_id, item);
+    ItemAction& entry = add_entry(item_id, item);
+    std::string help_text = m_generate_help_text_for_file ? m_generate_help_text_for_file(FileSystem::join(m_directory, item)) : "";
+
+    if (!help_text.empty()) {
+      entry.set_help(help_text);
+    }
```
Vankata453, line 110: Redundant `.empty()` check.

src/gui/menu.cpp
```diff
@@ -594,10 +597,12 @@ Menu::draw(DrawingContext& context)
     const int text_width = static_cast<int>(Resources::normal_font->get_text_width(m_items[m_active_item]->get_help()));
     const int text_height = static_cast<int>(Resources::normal_font->get_text_height(m_items[m_active_item]->get_help()));
 
+    const float margin_y = 16.0f;
```
Vankata453, line 600: This variable can be moved as a `static` constant on the top of the `.cpp` file.

`static const float HELP_MARGIN_Y = 16.f;`

src/editor/object_settings.cpp
```diff
@@ -293,7 +296,35 @@ ObjectSettings::add_music(const std::string& text, std::string* value_ptr,
...
+    }
+
+    std::string title_or_filename = title.empty() ? filename : "\"" + title + "\""; // assumes path is just a filename
+
+    std::string help_text =
+        title_or_filename + (author.empty() ? "" : "\n by " + author)
+        + (license.empty() ? "" : "\nLicense: " + license);
+
+    return help_text;
+  };
```
Vankata453, line 325: Wouldn't it be better to have the music file info detection code in `MenuFilesystem` itself? It seems like a task the menu should always take care of.

src/editor/object_settings.cpp
```diff
@@ -293,7 +296,35 @@ ObjectSettings::add_music(const std::string& text, std::string* value_ptr,
...
+    std::unique_ptr<SoundFile> sound_file;
+    try {
+      sound_file = load_sound_file(path);
+    } catch (...) {
+      return "";
+    }
+
+    std::string& author = sound_file->m_author;
+    std::string& license = sound_file->m_license;
+    std::string& title = sound_file->m_title;
```
Vankata453, line 311: `const std::string&`

src/editor/object_settings.cpp
```diff
@@ -293,7 +296,35 @@ ObjectSettings::add_music(const std::string& text, std::string* value_ptr,
...
+
+    bool hasMusicMetadata = !title.empty() || !author.empty() || !license.empty();
+    if (!hasMusicMetadata) {
```
Vankata453, line 314: Redundant variable.

src/editor/object_settings.cpp
```diff
@@ -293,7 +296,35 @@ ObjectSettings::add_music(const std::string& text, std::string* value_ptr,
...
+      return "";
+    }
+
+    std::string title_or_filename = title.empty() ? filename : "\"" + title + "\""; // assumes path is just a filename
+
+    std::string help_text =
+        title_or_filename + (author.empty() ? "" : "\n by " + author)
+        + (license.empty() ? "" : "\nLicense: " + license);
```
Vankata453, line 322: `fmt::format` should be used here to support translations.

``fmt::format(fmt::runtime(_("{} by {}")), title_or_filename, author)`` in one variable, if an author is provided. If not, only copy the title to this variable (by the way, if no info is provided, maybe nothing should be shown at all?).

``"\n" + fmt::format(fmt::runtime(_("License: {}")), license)`` in another variable, if license is provided. If not, keep empty.

Return the two variables concatenated.

src/gui/menu_filesystem.cpp
jamescdericco: Thanks for the feedback. I don't see how this condition could be removed and also avoid calling entry.set_help(string) when string is empty. The string is empty when m_generate_help_text_for_file is nullptr (not defined) or if m_generate_help_text_for_file(string) returns the empty string, which could happen in the event of a failure to load the file or when it finds no data for a file. We already have a conditional for the former condition on the previous line, but the !help_text.empty() is needed to handle the later condition. Are you recommending that we call set_help(string) when string is empty?

src/gui/menu.cpp
jamescdericco: Thanks, I pushed a commit that changes this.

src/editor/object_settings.cpp
jamescdericco: Thanks, I pushed a commit that changes this for all local vars in this pull request.

src/editor/object_settings.cpp
jamescdericco: Thanks, I pushed a commit that changes this.

src/editor/object_settings.cpp
```diff
@@ -293,7 +296,35 @@ ObjectSettings::add_music(const std::string& text, std::string* value_ptr,
...
+    std::string& title = sound_file->m_title;
+
+    bool hasMusicMetadata = !title.empty() || !author.empty() || !license.empty();
+    if (!hasMusicMetadata) {
+      return "";
+    }
+
+    std::string title_or_filename = title.empty() ? filename : "\"" + title + "\""; // assumes path is just a filename
+
+    std::string help_text =
```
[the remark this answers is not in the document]
jamescdericco: Thanks, I pushed a commit that changes this for all local vars in this pull request.

src/gui/menu_filesystem.cpp
Vankata453: Yes, wouldn't the item just not have any help text set if `string` is empty?

jamescdericco:
Thanks for all your feedback @tobbi and @Vankata453! I pushed commits that fix all the easy code style changes, I will work on resolving your remaining feedback tomorrow.

MatusGuy:
i have enabled workflows in this pr

src/gui/menu_filesystem.cpp
jamescdericco: You're right, MenuItem::set_help(string) behaves well even in the empty string case. I pushed a commit that implements your suggested change. Thanks!

jamescdericco:
> What is the current state of this PR?

@mrkubax10 I am resolving remaining feedback on this PR. In particular, right now I am working on supporting translations.

src/editor/object_settings.cpp
jamescdericco: Hi @Vankata453, good catch: I updated the code to support translations. However, while I wrote the code, I discovered that there's an issue preventing "{}" substitutions from working in the current Git master version. I reported this issue in issue #2700.

I am currently working on code changes based on your recommendation. There are some differences I would lke to make to your suggestions. Notably, I would like to translate just the word License with `_("License") + ": {}"` to take advantage of an existing .po file msgid "License" translations. 

There may be an issue with _("{} by {}") msgid for some languages like Japanese, where it seems that the order of the first two arguments should be swapped, relative to the English message. For example, '"Piano Man" by Billy Joel' is Birī Joeru no `piano man' in Japanese (according to Google Translate). I'll lookup how to support argument swapping with tinygettext.

Currently, there is a newline separating the TITLE and the "by AUTHOR" parts. Do you know if _("{}\nby {}") works?

Vankata453: > Notably, I would like to translate just the word License with `_("License") + ": {}"` to take advantage of an existing .po file msgid "License" translations.

Seems fine.

> Currently, there is a newline separating the TITLE and the "by AUTHOR" parts. Do you know if _("{}\nby {}") works?

I believe it should.

tobbi:
Looking good! My only style complaint would be that `m_generate_help_text_for_file` sounds like it's a bool parameter, not a function. But that's not important.

MatusGuy:
May I suggest something like `help_text_for_file_generator` or `help_text_generator`

src/gui/menu_filesystem.cpp
```diff
@@ -111,6 +114,9 @@ FileSystemMenu::refresh_items()
     if (in_basedir && m_item_processor)
       m_item_processor(menu_item);
 
+    const std::string help_text = m_generate_help_text_for_file ? m_generate_help_text_for_file(FileSystem::join(m_directory, item)) : "";
+    menu_item.set_help(help_text);
+
```
Vankata453, line 119: To achieve this, an item processor in `m_item_processor` can be provided.

Vankata453 (changes requested):

tobbi:
What still needs to get done here?

jamescdericco:
> What still needs to get done here?

# To Do
- **WIP**: Finish refactor to use item_processor
- Plan how to remove English phrases from the author field
- Add music metadata for music added to SuperTux master recently (since December)

This feature was functionally complete, but refactoring it to use the Vankata's item_processor parameter (instead of the separate specific generate_help_text_for_file) and rebasing it on the latest master is still a work in progress. Also, there are English phrases within the author field data that should be translated (some author values include a lot of info, including roles different contributors had for a given song). Finally, there are a few new music files added to SuperTux since December that need the metadata fields added to their .music files.

# Current Progress on Refactoring
Last week I worked on the refactoring, and I am planning on continuing the refactor this week. My current work is held up by a tricky illegal instruction exception issue when the item_processor callback returns in my local refactored code, which I am going to try solving this week by rebuilding and updating my development container to fix potential ABI compatibility issues. Last week I tried using Google's AddressSanitizer to search for memory corruption, but it didn't report any issues in my testing.

bruhmoent:
Is this done?

jamescdericco:
> Is this done?

This is work in progress. All functionality and translations are complete. Still left to do are updating music metadata, like title, author and license, for recent music added to the game, and refactor to extract music help text item processor function out of object_settings.cpp. I'm planning on resuming my work for this PR this week.

tobbi:
Thank you! 👍

src/gui/menu_filesystem.cpp
jamescdericco: I agree! I refactored the code to use `m_item_processor`.

jamescdericco:
This pull request is ready for code review and merging! I completed all the remaining work that I mentioned in my last status update 3 weeks ago. The code now uses `m_item_processor`, resolving feedback provided by @Vankata453. My testing passed for the level editor, as well as regression tests for story mode. It looks good to merge by me, but it would be good for this to get reviewed by other contributors.

src/gui/menu_filesystem_item_processor_music_help.cpp
Vankata453, line 1: I don't believe this item processor function needs its own file. It could be put directly in `object_settings.cpp` under the `add_file()` function.

Noting this, because:

* The source code doesn't really use separate files to solely put functions to only be referenced/called once. (An exception to this would be the `register_scripting_api` function under "src/squirrel/supertux_api.*pp" but that is done so it can register all the anonymous scripting functions in "src/squirrel/supertux_api.cpp".)
* The "src/gui" directory only consist of files used to define classes, making the "menu_filesystem_item_processor_music_help.*pp" files seem out of place.

src/supertux/menu/editor_converters_menu.cpp
```diff
@@ -62,15 +62,17 @@ EditorConvertersMenu::EditorConvertersMenu() :
...
+            if (in_basedir) {
+              std::string basename = FileSystem::basename(file_path);
+              auto it = m_converters.find(basename);
+              if (it == m_converters.end())
+                return;
+
+              item.set_text("\"" + it->second.title + "\"");
+              item.set_help(it->second.description + (it->second.author.empty() ? "" :
+                            "\n\n" + fmt::format(fmt::runtime(_("By: {}")), it->second.author)));
+            }});
```
Vankata453, line 75: Needs one space more in tabulation so it's 2 spaces per code style:

```suggestion
             if (in_basedir) {
               std::string basename = FileSystem::basename(file_path);
               auto it = m_converters.find(basename);
               if (it == m_converters.end())
                 return;

               item.set_text("\"" + it->second.title + "\"");
               item.set_help(it->second.description + (it->second.author.empty() ? "" :
                            "\n\n" + fmt::format(fmt::runtime(_("By: {}")), it->second.author)));
             }});
```

Vankata453 (commented):
Works great! You can see all important information about a music file:

image

Left 2 minor code-related comments. If you agree with any of them, I am also up to make the changes listed (if you want me to, of course). Offering that, so you don't have to do any code-related work on this no more if you don't want to, seeing the effort you put in throughout all this time, encouraged by all the previous comments.

src/gui/menu_filesystem_item_processor_music_help.cpp
jamescdericco: Sure, we can do it this way. The downside is that this adds sound_file, format, and file_system dependencies to the object_settings file, which could get bloated with dependencies in the future. But specifying the item_processor as an anonymous function is consistent with how it is done in `editor_converters_menu.cpp` for the "Select Tile Conversion File" menu. I pushed a commit to resolve this.

jamescdericco:
> Works great! You can see all important information about a music file:
> 
> image
> 
> Left 2 minor code-related comments. If you agree with any of them, I am also up to make the changes listed (if you want me to, of course). Offering that, so you don't have to do any code-related work on this no more if you don't want to, seeing the effort you put in throughout all this time, encouraged by all the previous comments.

Thanks for the feedback: I fixed both of them! I enabled "Allow edits by maintainers" for this PR so you are welcome to add  commits to the PR repo if you prefer. However, I am happy to fix any additional feedback you mention here on GitHub.

src/audio/sound_file.cpp
```diff
@@ -243,6 +286,20 @@ std::unordered_map<std::string, std::string> fallback_paths = {
...
+  {"fortress_old.music", "/music/retro/fortress_old.music"},
+  {"fortress_old.ogg", "/music/retro/fortress_old.ogg"},
+  {"ice_music.music", "/music/retro/ice_music.music"},
+  {"ice_music.ogg", "/music/retro/ice_music.ogg"},
+  {"worldmap_old.music", "/music/retro/worldmap_old.music"},
+  {"worldmap_old.ogg", "/music/retro/worldmap_old.ogg"},
+  {"saharan_penguin.music", "/music/tropical/saharan_penguin.music"},
+  {"saharan_penguin.ogg", "/music/tropical/saharan_penguin.ogg"},
+  {"tropicalbreeze.music", "/music/tropical/tropicalbreeze.music"},
+  {"tropicalbreeze.ogg", "/music/tropical/tropicalbreeze.ogg"},
```
Hypernoot, line 302: Do we actually need to update this list, as it is used only for legacy music to work correctly?

jamescdericco: Based on my notes, without `fallback_paths`, the music hover help text does not work. A while ago I reviewed the code execution path with a debugger, and found that `fallback_paths` was always used for loading SuperTux `.music` files in the level editor. I updated `fallback_paths`, but a better solution is to replace it with something more maintainable: music contributors should not have to update this hardcoded table in order for this feature to work. I think the music file paths can be found at runtime instead of hardcoded at compile time.

Hypernoot: Can we fix the behaviour instead, so the `fallback_paths` are used if and only if the actual file isn't found?

PR #2678, opened 2023-11-15
Editor toolbox scrollbar, adjusted scroll speed

### LINKED ISSUE #2675: make scroll wheel for the menu in tile selector adjustable/ stronger
image

currently, when scrolling in this menu, its very slow, and quite useless.

the alternative to scrolling is to use the mouse and awkwardly hover it over an unmarked location to move quicker.

this is finnicky, better UI would have either:

1. the scroll be stronger, or adjustable
2. or the menu have a decal designated to hovering your mouse over to scroll.

Vankata453:
The `EditorToolboxWidget` now features a `ControlScrollbar`, which appears on the right and can be controlled using the mouse. The scroll speed has been raised.

https://github.com/SuperTux/supertux/assets/78196474/4d7c1f6a-26c2-416f-b59d-e47b56d2f3dd

`ControlScrollbar` has been brought into a working state, where its main purpose is managing a scroll progress variable.

Fixes #2675.


mrkubax10 (approved):

src/interface/control_scrollbar.hpp
```diff
@@ -22,48 +23,67 @@
...
+  float m_total_region;
 
   /** The length (height) of the viewport for the region */
-  int m_covered_region;
+  float m_covered_region;
 
   /** The length (height) between the beginning of the viewport and the beginning of the region */
-  int m_progress;
-  
+  float& m_progress;
```
MatusGuy, line 78: Why a float reference?

MatusGuy (approved):
Looks completely fine

src/interface/control_scrollbar.hpp
Vankata453: As the description says, it manages a scroll progress variable. That variable is an outside one, usually owned by the object which contains the scrollbar.

MatusGuy: Ah. Nice

Alasdairbugs (commented):
image

think the scroll bar should go on the left and be a bit bigger (about half a tile width). the length is fine. this helps people who are on 4k monitors as the tool bar in 4k is quite small.

currently it being on the right sometimes results in you missing the task bar when your mouse hits the end of the screen, and clicking then tabs you out entirely.

the scroll speed adjustment is fine

Vankata453:
I don't think it would like having the scrollbar on the left, personally. Having it wider is fair.

tylerandari13:
Why is the screenshot so chonky lmao.

In all seriousness we if we did have it on the left we could make some sort of accessibility option, but it feels a little... *unnecessary*. Im not against making it wider tho.

MatusGuy:
Having the scrollbar on the left is extremely weird, considering every GUI application puts the scroll bar on the right, just look at your browser. The width change is ok.

Alasdairbugs (approved):

PR #2681, opened 2023-11-18
Modifying bonus block custom object from editor

Vankata453:
Adds the ability to edit a bonus block custom object directly from the editor. Previously, this had to be done via editing the file.

Additionally fixes a segmentation fault when dropping more than 1 custom object.


mrkubax10 (approved):

src/editor/object_option.cpp
```diff
@@ -513,8 +524,9 @@ ObjectSelectObjectOption::to_string() const
 void
 ObjectSelectObjectOption::add_to_menu(Menu& menu) const
 {
-  menu.add_entry(get_text(), [pointer = m_value_pointer, parent = m_parent]() {
-    MenuManager::instance().push_menu(std::make_unique<ObjectSelectMenu>(*pointer, parent));
+  menu.add_entry(get_text(), [pointer = m_value_pointer, get_objects_param = m_get_objects_param,
```
MatusGuy, line 527: Wh... Heh... Ha... Wha???

mrkubax10: What do you mean?

MatusGuy (approved):
After a quick look without testing, looks completely fine.

src/editor/object_option.cpp
MatusGuy: What's with the square brackets?

mrkubax10: Lambda function capture.

MatusGuy: Ohhh those are aliases. Ok

PR #2683, opened 2023-11-22
Fix smart BadGuy movement on e.g. rocks

### LINKED ISSUE #2194: Plants and totems do not behave properly when placed on rocks that are atop of trampolines
SuperTux version:| 0.6.3

Expected behavior:| Totems and plants should be able to walk on rocks like other enemies can.

Actual behavior:| The plant in the video below appears to clip into the rocks after the falling rocks and comes back out, and it seems to be in its moving animation/position even though it itself isn't moving. The totem touches the rock(s) and doesn't move around; instead it acts glitchy by repeating its left and right walking animations constantly / consistently.

Steps to reproduce actual behavior:|
Video below

Additional debugging information:|

https://user-images.githubusercontent.com/83178674/164649289-54f11452-b9b8-4497-b9ab-a24e6ae18023.mp4

AidenIsik:
This fixes a bug where "smart" badguys can't move when walking on rocks.

I just modified the BadGuy::might_fall() function to also check for "movablestatics" in the return, which makes the smart
badguys no longer act like they are about to walk off of an edge when they're really just walking on rocks (maybe this also fixes other things badguys can walk on too).

Closes #2194

### DIFF
--- a/src/badguy/badguy.cpp
+++ b/src/badguy/badguy.cpp
@@ -775,7 +775,9 @@ BadGuy::might_fall(int height) const
     x1 = m_col.m_bbox.get_right();
     x2 = m_col.m_bbox.get_right() + 1;
   }
-  return Sector::get().is_free_of_statics(Rectf(x1, y1, x2, y2));
+  const Rectf rect = Rectf(x1, y1, x2, y2);
+
+  return Sector::get().is_free_of_statics(rect) && Sector::get().is_free_of_specifically_movingstatics(rect);
 }
 
 Player*
--- a/src/collision/collision_system.cpp
+++ b/src/collision/collision_system.cpp
@@ -714,6 +714,22 @@ CollisionSystem::is_free_of_movingstatics(const Rectf& rect, const CollisionObje
   return true;
 }
 
+bool
+CollisionSystem::is_free_of_specifically_movingstatics(const Rectf& rect, const CollisionObject* ignore_object) const
+{
+  using namespace collision;
+
+  for (const auto& object : m_objects) {
+    if (object == ignore_object) continue;
+    if (!object->is_valid()) continue;
+    if ((object->get_group() == COLGROUP_MOVING_STATIC)
+        && (intersects(rect, object->get_bbox())))
+      return false;
+  }
+
+  return true;
+}
+
 CollisionSystem::RaycastResult
 CollisionSystem::get_first_line_intersection(const Vector& line_start,
                                              const Vector& line_end,
--- a/src/collision/collision_system.hpp
+++ b/src/collision/collision_system.hpp
@@ -65,6 +65,7 @@ public:
   bool is_free_of_tiles(const Rectf& rect, const bool ignoreUnisolid = false, uint32_t tiletype = Tile::SOLID) const;
   bool is_free_of_statics(const Rectf& rect, const CollisionObject* ignore_object, const bool ignoreUnisolid) const;
   bool is_free_of_movingstatics(const Rectf& rect, const CollisionObject* ignore_object) const;
+  bool is_free_of_specifically_movingstatics(const Rectf& rect, const CollisionObject* ignore_object) const;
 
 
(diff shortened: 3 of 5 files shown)


AidenIsik:
This seems to allow smart badguys to walk on tux and enemies if they are one tile down and next to the level where they are walking. I need to revise this slightly.

tobbi (approved):
Seems easy enough to just outright approve.

MatusGuy:
Shouldn't we wait for someone to test this?

tobbi:
Just from the code logic, this seems reasonable, but fine...

tobbi:
Testing my ass

AidenIsik:
Please don't merge yet I need to change it a bit later:

is_free_of_movingstatics includes COLGROUP_MOVING. I need to make a slightly modified version of that function excluding COLGROUP_MOVING and use that instead.

Rusty-Box:
> Testing my ass

That's not how this works.

tobbi:
> Please don't merge yet I need to change it a bit later:
> 
> is_free_of_movingstatics includes COLGROUP_MOVING. I need to make a slightly modified version of that function excluding COLGROUP_MOVING and use that instead (perhaps calling it is_free_of_only_movingstatics)

I've marked it as a draft for you.

AidenIsik:
Thanks. I'll get this done in a few hours.

MatusGuy:
Tobbi I've seen stuff break because a little insignificant thing was merged. Just don't be like that

MatusGuy:
> Just from the code logic, this seems reasonable, but fine...

If it was an or (`||`) it'd be fine but you better be careful with the ands

tobbi:
Yeah okay that went over my head

tobbi:
> > Just from the code logic, this seems reasonable, but fine...
> 
> If it was an or (`||`) it'd be fine but you better be careful with the ands

&& is just fine. That function shold return true when both conditions are.

AidenIsik:
Done.

tobbi (approved):

MatusGuy:
Yo tobbi did you actually test this time?

src/collision/collision_system.cpp
```diff
@@ -714,6 +714,22 @@ CollisionSystem::is_free_of_movingstatics(const Rectf& rect, const CollisionObje
...
+bool
+CollisionSystem::is_free_of_specifically_movingstatics(const Rectf& rect, const CollisionObject* ignore_object) const
+{
+  using namespace collision;
+
+  for (const auto& object : m_objects) {
+    if (object == ignore_object) continue;
+    if (!object->is_valid()) continue;
+    if ((object->get_group() == COLGROUP_MOVING_STATIC)) {
+      if (intersects(rect, object->get_bbox())) return false;
```
mrkubax10, line 726: These could be merged to single `if`.

AidenIsik: Do you want me to do that for the other functions with redundant if statements, or just this new one?

mrkubax10: Only these that you added, changing all of them would be out of this PR scope.

src/badguy/badguy.cpp
```diff
@@ -775,7 +775,7 @@ BadGuy::might_fall(int height) const
     x1 = m_col.m_bbox.get_right();
     x2 = m_col.m_bbox.get_right() + 1;
   }
-  return Sector::get().is_free_of_statics(Rectf(x1, y1, x2, y2));
+  return Sector::get().is_free_of_statics(Rectf(x1, y1, x2, y2)) && Sector::get().is_free_of_specifically_movingstatics(Rectf(x1, y1, x2, y2));
```
Vankata453, line 778: The `Rectf(x1, y1, x2, y2)` rectangle could be created in a `const Rectf` variable above and then copied, saving another initialization.

mrkubax10 (approved):

tobbi:
Thank you for your contribution!

PR #2688, opened 2023-12-01
Refactor AmbientSound

mrkubax10:
Implements new SoundObject object that is hearable in entire Sector. It is possible to change it's volume and looping sound play interval. Attempts to fix the long-broken AmbientSound.


src/object/ambient_sound.cpp
```diff
@@ -252,22 +126,94 @@ AmbientSound::collision(GameObject& other, const CollisionHit& hit_)
...
+      player_distance = player_center.x<get_bbox().get_left() ? get_bbox().get_left()-player_center.x : player_center.x-get_bbox().get_right();
+    else if (player_center.x<=get_bbox().get_left() && player_center.y<=get_bbox().get_top())
+      player_distance = glm::distance(player_center, get_bbox().p1());
+    else if (player_center.x>=get_bbox().get_right() && player_center.y<=get_bbox().get_top())
+      player_distance = glm::distance(player_center, get_bbox().p1()+Vector(get_bbox().get_width(), 0));
+    else if (player_center.x<=get_bbox().get_left() && player_center.y>=get_bbox().get_bottom())
+      player_distance = glm::distance(player_center, get_bbox().p1()+Vector(0, get_bbox().get_height()));
+    else if (player_center.x>=get_bbox().get_right() && player_center.y>=get_bbox().get_bottom())
+      player_distance = glm::distance(player_center, get_bbox().p2());
+    m_sound_source->set_gain(std::max(m_radius_in_px-player_distance, 0.0f)/m_radius_in_px*m_volume);
```
Vankata453, line 179: Spaces between comparators and operators like `<=` and `-` would be nice for readability.

src/object/ambient_sound.cpp
```diff
@@ -48,175 +46,51 @@ AmbientSound::AmbientSound(const ReaderMapping& mapping) :
...
-  silence_distance(),
-  maximumvolume(vol),
-  targetvolume(),
-  currentvolume()
+  m_sample(file),
+  m_sound_source(),
+  m_radius(radius),
+  m_radius_in_px(m_radius*32.0f),
+  m_volume(vol),
+  m_first_update(true)
```
Vankata453, line 65: `m_has_played_sound` perhaps?

src/object/ambient_sound.cpp
```diff
@@ -252,22 +126,94 @@ AmbientSound::collision(GameObject& other, const CollisionHit& hit_)
...
+    return;
+
+  m_sound_source->play();
+}
+
+void
+AmbientSound::update(float dt_sec)
+{
+  if (Editor::is_active())
+    return;
```
Vankata453, line 154: `update()` isn't called in the editor either way. Instead, `editor_update()` is.

src/object/ambient_sound.hpp
```diff
@@ -87,21 +65,18 @@ class AmbientSound final : public MovingObject,
...
+  float m_radius;
+  float m_radius_in_px;
+  float m_volume;
+  bool m_first_update;
 
-  float maximumvolume; /// maximum volume
-  float targetvolume;  /// how loud we want to be
-  float currentvolume; /// how loud we are
+private:
+  void prepare_sound_source();
```
Vankata453, line 79: Functions before member variables (coding style).

src/object/sound_object.cpp
```diff
@@ -0,0 +1,164 @@
...
+}
+
+void
+SoundObject::update(float dt_sec)
+{
+  if (m_scheduled_for_removal)
+  {
+    remove_me();
+    return;
+  }
```
Vankata453, line 69: `remove_me()` could be called directly on the object, whenever it should be scheduled for removal. `GameObjectManager` will take care of it.

src/scripting/sound_object.hpp
```diff
@@ -0,0 +1,85 @@
...
+#include "scripting/game_object.hpp"
+
+class SoundObject;
+#endif
+
+namespace scripting {
+
+/**
+ * @summary An ""SoundObject"" that was given a name can be controlled by scripts.
+ * @instances An ""SoundObject"" is instantiated by placing a definition inside a level.
```
Vankata453, line 30: **A** ""SoundObject""

Vankata453 (approved):
Looks good code-wise.

Rusty-Box (commented):
The "Sound Object" (imo. it's name should be displayed as just "Sound" when hovering over it in the editor, since the music one is also is simply called "Music") has a problem with looping a sound. If the interfall is set to anything above 0 it only plays the sound once. Set it to 0 or a negative and after the sound has played fully once, it will bug out (I tested it with "sounds/rain.wav" which is 25 seconds long).

mrkubax10:
This seems to be some sort of OpenAL issue because it works fine with short sounds. Apparently after playing once sound source drops the attached buffer data? Not sure.

Rusty-Box:
Seems like the sound looping remains the only major issue with this one. Otherwise the rest seems be fine, aside the display name change for consistency

mrkubax10:
Maybe dropping interval setting and using internal OpenAL looping feature would be a good enough solution?

Rusty-Box:
Would be worth a try, I suppose.

Rusty-Box (approved):
Looking good. That seems to have fixed the looping issue - approve!

src/object/ambient_sound.cpp
```diff
@@ -252,22 +126,91 @@ AmbientSound::collision(GameObject& other, const CollisionHit& hit_)
...
+void
+AmbientSound::update(float dt_sec)
+{
+  const Player* const nearest_player = Sector::get().get_nearest_player(get_bbox().get_middle());
+  if (!nearest_player)
+    return;
+  const Rectf& player_bbox = nearest_player->get_bbox();
+  const Vector player_center = player_bbox.get_middle();
+
+  if (collision::intersects(player_bbox, get_bbox()))
```
Vankata453, line 159: `player_bbox.contains(get_bbox())` might be better, since this is an old collision method, which is a duplicate of `Rectf::contains()` that, in my opinion, should be removed.

src/object/ambient_sound.cpp
```diff
@@ -252,22 +126,91 @@ AmbientSound::collision(GameObject& other, const CollisionHit& hit_)
...
+  {
+    m_sound_source = SoundManager::current()->create_sound_source(m_sample);
+    if (!m_sound_source)
+      throw std::runtime_error("file not found");
+
+    m_sound_source->set_gain(0);
+    m_sound_source->set_looping(true);
+    m_sound_source->set_relative(true);
+  }
+  catch(std::exception& e)
```
Vankata453, line 208: `const std::exception&`

src/object/sound_object.cpp
```diff
@@ -0,0 +1,124 @@
...
+    m_sound_source->set_gain(0);
+    m_sound_source->set_looping(true);
+    m_sound_source->set_relative(true);
+    m_sound_source->set_gain(m_volume);
+
+    m_sound_source->play();
+  }
+  catch(const std::exception& e)
+  {
+    log_warning << "Couldn't load '" << m_sample << "': " << e.what() << "" << std::endl;
```
Vankata453, line 111: `<< ""` is unneeded.

src/object/ambient_sound.cpp
```diff
@@ -252,22 +126,91 @@ AmbientSound::collision(GameObject& other, const CollisionHit& hit_)
...
+  {
+    log_warning << "Couldn't load '" << m_sample << "': " << e.what() << "" << std::endl;
```
Vankata453, line 210: `<< ""` is unneeded.

src/object/ambient_sound.cpp
mrkubax10: Yeah, though according to FIXME in src/math/rectf.hpp it should be renamed to `overlaps` because name is misleading.

Vankata453: That would be good to do while on it as well.

Vankata453 (approved):

PR #2690, opened 2023-12-02
Improve `DrawingRequest` API

Vankata453:
1. `DrawingRequest` no longer preserves a variable indicating its type, but rather utilizes a `virtual` function for this purpose.
2. `DrawingRequest` now initializes some of its variables from the current `DrawingTransform`.
2. The viewport rectangle is now stored in `DrawingTransform`, instead of `DrawingContext`. bb18239df3d1ed34840e78fc074200ecbb07ee15 has added per-request viewport cropping. Utilizing it now is more convenient, because the old viewport wouldn't have to be stored in a separate variable in a draw function, so it could be restored later. A new transform could be pushed and popped instead.

### DIFF
--- a/src/video/canvas.cpp
+++ b/src/video/canvas.cpp
@@ -21,6 +21,7 @@
 #include "supertux/globals.hpp"
 #include "util/log.hpp"
 #include "util/obstackpp.hpp"
+#include "video/drawing_context.hpp"
 #include "video/drawing_request.hpp"
 #include "video/painter.hpp"
 #include "video/renderer.hpp"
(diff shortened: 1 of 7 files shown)


src/video/canvas.cpp
```diff
@@ -121,14 +124,11 @@ Canvas::draw_surface(const SurfacePtr& surface,
      position.y + static_cast<float>(surface->get_height()) < cliprect.get_top())
     return;
 
-  auto request = new(m_obst) TextureRequest();
```
MatusGuy, line 124: What??

MatusGuy (approved):

src/video/canvas.cpp
MatusGuy: Sorry. What is this syntax? `new(m_obst) TextureRequest()`

Vankata453: Creating pointer with `obstack`.

Vankata453:
> For shadowing fields, couldn't the member variables be prefixed with `m_`?

About the `DrawingRequest` struct, I believe it is more convenient to have its member variables not prefixed with `m_`, because the purpose it serves is to group variables together, not for the class to perform any actions on its members, which would then make prefixing necessary.

mrkubax10 (approved):

PR #2695, opened 2023-12-10
Fix `wait()` segfaulting on the worldmap.

tylerandari13:
`wait()` checks for the current level in the gamesession to check if it can skip the cutscene/if ones running, but if you run this on the worldmap then it segfaults because there isnt a current gamesession (sometimes, idk why but sometimes there is a gamesession and i dont know why). Basically this makes it where it doesnt check for cutscene specific stuff if it cant find a gamesession.

~~I know it might seem a little strange to have 2 `GameSession::current() != nullptr`s but i tried to make it just 1 and it didnt work.~~

### DIFF
--- a/src/scripting/functions.cpp
+++ b/src/scripting/functions.cpp
@@ -137,7 +137,7 @@ bool check_cutscene()
 
 void wait(HSQUIRRELVM vm, float seconds)
 {
-  auto session = GameSession::current()
+  auto session = GameSession::current();
 
   if(session && session->get_current_level().m_skip_cutscene)
   {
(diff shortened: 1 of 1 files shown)


Vankata453:
I believe you can wrap the two conditional statements in one `if (GameSession::current())` and return whenever either one succeeds. Then remove the `else` from the last code.

MatusGuy (approved):

tylerandari13:
> I believe you can wrap the two conditional statements in one `if (GameSession::current())` and return whenever either one succeeds. Then remove the `else` from the last code.

I tried this and it didnt work. For some reason the game just waits forever which is why i had to do it like i did. I did try a different soloution tho.

MatusGuy (approved):

Vankata453 (approved):

PR #2702, opened 2023-12-14
Make light block use it's texture before being used

### LINKED ISSUE #2670: Light Block disguised as Bonus Block
Placing a Light Block (a Bonus Block that emits light when hit) always looks identical to the other Bonus Blocks before hit once. This should be changed as it makes more sense to tell the player immediately which of these blocks is one that emits light. In dark areas/levels this is vital information!

Light Blocks already have an "off"-sprite which should be the default when placed in levels rather than looking like a normal Bonus Block.

Miner34dev:
Closes #2670


Rusty-Box:
I'll check this one out in just a moment!

Rusty-Box (approved):
Working a-okay! Thank you - approve

MatusGuy (approved):

PR #2719, opened 2023-12-30
Cloning paths in editor, fix various undo/redo issues

Vankata453:
Cloning path objects in the editor now clones their path as well. This allows for more robust manipulation of path objects.

Additionally, various undo/redo issues, related to path objects, have also been fixed.

Depends on #2713.

### DIFF
--- a/src/object/path_object.cpp
+++ b/src/object/path_object.cpp
@@ -21,6 +21,7 @@
 #include "editor/editor.hpp"
 #include "object/path_gameobject.hpp"
 #include "supertux/d_scope.hpp"
+#include "supertux/game_object_factory.hpp"
 #include "supertux/sector.hpp"
 #include "util/log.hpp"
 #include "util/reader_mapping.hpp"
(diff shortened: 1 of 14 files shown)


mrkubax10 (approved):

src/object/path_object.cpp
```diff
@@ -63,8 +64,8 @@ PathObject::init_path(const ReaderMapping& mapping, bool running_default)
   else if (mapping.get("path-ref", path_ref))
   {
     d_gameobject_manager->request_name_resolve(path_ref, [this, running](UID uid){
-        m_path_uid = uid;
-        m_walker.reset(new PathWalker(uid, running));
+        if (!m_path_uid) m_path_uid = uid;
```
tobbi, line 67: Maybe line break after if statement?

Vankata453: Could add it, although a lot of the code does 1-line `if` statements, alongside 2-line ones.

PR #2722, opened 2023-12-31
Instantiable `Text` and `TextArray` objects

Vankata453:
`TextObject` and `TextArrayObject` can now be instantiated from scripting. This allows for more possibilities in scripting texts.

The former `Text` and `TextArray` instances, which are always added to a `Sector`, have been retained to allow for compatibility. The default `Text` object is used by `TextArea` triggers.

Additionally, a function to get/set the text wrap width of a `Text` or `TextArray` text object is added.


tobbi (approved):
Preliminary approval.

It can't be that I can't find anything wrong with this code.

mrkubax10 (approved):

PR #2726, opened 2024-01-01
Pass reference to the player, who has triggered a script

Vankata453:
Previously, there was no way to know which player executed a script.

Squirrel weak references to entries under a `SquirrelEnvironment` table in `sector.triggers.{trigger_object}.{name_of_reference}` are now supported. For example, a reference to the player who has triggered a script trigger, can be found under `sector.triggers.{script_trigger_name}.Tux`. The object must have a name to support trigger info. Before running a script of an object, the `triggers` table of that object is deleted.

References to the player who has activated a script are supported in `PushButton`, `ScriptedObject` hit script, `ScriptTrigger` and `Switch`.

An idea for the future would be to provide such player references from `BadGuy` death scripts as well.

### DIFF
--- a/src/supertux/sector_base.cpp
+++ b/src/supertux/sector_base.cpp
@@ -16,8 +16,6 @@
 
 #include "supertux/sector_base.hpp"
 
-#include "util/log.hpp"
-
 namespace Base {
 
 Sector::Sector(const std::string& type) :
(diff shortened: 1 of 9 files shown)


tylerandari13:
actually so real

tried it out and it appears to work fine

src/supertux/sector_base.cpp
```diff
@@ -33,6 +31,29 @@ Sector::run_script(const std::string& script, const std::string& sourcename)
...
+        try
+        {
+          vm.get_table_entry("triggers");
+          vm.delete_table_entry(object.get_name().c_str());
+        }
+        catch (...) {}
+      });
+
+    for (auto it = triggers.begin(); it != triggers.end(); it++)
+      m_squirrel_environment->create_reference(it->first, it->second, "triggers", object.get_name());
```
mrkubax10, line 51: Since we are using C++17 you can use structured binding:
```cpp
for (const auto& [name, class_name] : triggers)
  m_squirrel_environment->create_reference(name, class_name, "triggers", object.get_name());
```

Vankata453:
I will try to make more triggers support this.

PR #2732, opened 2024-01-12
Mr. Iceblock falls off any ledge

MatusGuy:
I wanted to leave this up to discussion so i made a pull request instead of committing directly.


tobbi:
🇱🇹 🐔

Vankata453:
I think this would make sense, since the smartblock exists so it doesn't fall of ledges.

mrkubax10:
Yeah, me too.

CentrawhileYT:
Well, I'd prefer the AI where the iceblock falls off ledges safe to land on, similar to the spiky's AI.

PR #2734, opened 2024-01-13
Rename `Rectf::contains(const Rectf&)` to `Rectf::overlaps(const Rectf&)`

mrkubax10:
As the title suggests. Additionally remove `collision::intersects` to avoid code duplication because it's exactly the same like `Rectf::overlaps`.


tobbi:
Fix tests, please.

PR #2737, opened 2024-01-17
Improve haywire jumping

### LINKED ISSUE #2716: Improve Haywire jumping
As far as I know, while chasing, Haywire will jump depending on Tux's y-position, relative to their own - If Tux is above Haywire as they approach a gap, Haywire will attempt a jump. If Tux is below Haywire, they will not jump and instead simply drop.

I suggest to have Haywire always attempt a jump when approaching a gap (such is the case with enemies like Captain Snowball), as often times Haywire will not jump even though they should, making them a pain to use for things such as luring them from point A to point B.

Brockengespenst:
Let haywire always attempt to jump over gaps, not only if Tux is above.

Fixes #2716

### DIFF
--- a/src/badguy/haywire.cpp
+++ b/src/badguy/haywire.cpp
@@ -151,15 +151,14 @@ Haywire::active_update(float dt_sec)
       }
       else
       {
-        // Jump over gaps if Tux isn't below.
+        // Jump over gaps
         Rectf gap_box = get_bbox();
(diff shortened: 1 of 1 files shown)


MatusGuy (approved):
Looks fine code-wise

Rusty-Box:
I'll give this a look tomorrow. Hold on tight

Rusty-Box (approved):

MatusGuy:
Thanks for the contribution!

Brockengespenst:
No problem. I hope I'll be be able to contribute more in the future.

PR #2739, opened 2024-01-24
Have Tux swim instead of floating when going up/down

n0toose:
This is a small detail that I mostly made so as to get used to the
codebase.

If the vertical velocity is under a specific threshold, then the
floating animation will be used. Otherwise, if v is too low but
vy is too high, show the swimming animation instead.

I feel like the change in itself makes sense, even if it won't
be noticed by any player that isn't willing to move around in squares.


n0toose:
(Maybe the commit message is a bit misleading.)

PR #2740, opened 2024-01-24
Do not let Tux shoot when ducking

n0toose:


Frostwithasideofsalt:
Personally i would be against this being merged, as tux being able to shoot when ducking isn't really an issue.

n0toose:
I don't really have any other proper arguments for it other than vibes, the lack of an animation that makes this feature ultimately feel clunky and naming Mario games as an example where you can't do that.

tylerandari13:
> I don't really have any other proper arguments for it other than vibes, the lack of an animation that makes this feature ultimately feel clunky and naming Mario games as an example where you can't do that.

To be fair theres no animation for him throwing a fireball when standing either.

n0toose:
> To be fair theres no animation for him throwing a fireball when standing either.

yeah, uh, I didn't think about that, but it looks like the fireball is spawning out of thin air! Would you be able to throw a bowling ball like that? 🎳

MatusGuy:
It's not a bad nerf to the fire flower.

Vankata453:
Tux shooting while ducked is essential for passing certain sections in levels. For example, it allows shooting fireballs more precisely, which makes some close-up one-tile pathway sections with ice passable.

> preparing for a backflip also comes with the tradeoff that you are leaving yourself somewhat exposed, and throwing a fireball feels like cheating your way away from that danger.

I don't agree, it's not really cheating the way through danger. In fact, if a player were to backflip/slide and danger immediately approached, I'd be impressed from it being pulled off, considering the danger should have come pretty fast towards Tux (otherwise the player would've already eliminated it before preparing to backflip/slide), and the player would have been focused on performing a different action in the meantime.

Personally I don't see why this mechanic should be taken away from the player. Additionally, before sliding was implemented, Tux was able to do the so-called "Duck Jumping" when ducked, which I'd personally love to see come back.

n0toose:
You have a point, I think this should be closed.

Issue #2741, opened 2024-01-24
Small "shaky" pause when dying


n0toose:
Mario games and other platformers like Celeste or Kirby have this thing where the screen and the entire world shake a little bit, a music pause, a vibration or a screen shake when "dying".

Deaths currently feel more like as if you're being teleported back to the checkpoint / beginning of the level, and I begin to realize that halfway through the death animation.

The game already seems to have some features that could be put together to make something like this, such as the screen shake when buttjumping, as well as controller rumble.

I watched this video from Masahiro Sakurai close to a year ago, and this was one of the things that always felt somewhat off but could not exactly point out why - I did play Kirby's Adventure when I was younger and that is an effect that I'd like to see imitated: https://youtube.com/watch?v=OdVkEOzdCPw

n0toose:
> This was in the game as of recently, but was removed due to it being generally disliked by the community.

Oh, a "damned if I do, damned if I don't" situation. Is it possible that there may be a compromise / that the previous implementation was somehow faulty in a way that could be salvaged and that the shake needed to be combined with something else?

On 0.6.3, it feels like there's an explosion happening when I die (which was fixed here: https://github.com/SuperTux/supertux/pull/2575). I will take a look at https://github.com/SuperTux/supertux/pull/2522 on a custom branch and return with feedback.

n0toose:
Okay, the shake is kind of intense and doesn't really involve a pause - I think that could be possibly improved by reducing the intensity and the duration, and combining it with something else.

But if that's a `wontfix`, I can respect that.

Vankata453:
> Oh, a "damned if I do, damned if I don't" situation - sorry for the duplicate. Is it possible that there may be a compromise / that the previous implementation was somehow faulty in a way that could be salvaged and that the shake needed to be combined with something else?

The only way I can think of to have this feature is to make it optional via a setting. That setting could also be combined with the aforementioned controller rumble, if it's added as well.

n0toose:
> The only way I can think of to have this feature is to make it optional via a setting. 

I think this is a good approach. I think I'll start working on it after my workload lightens up a little bit.

Just one more question: What do you think of this death screen? https://yewtu.be/watch?v=hF8moB-ONRY

Vankata453:
> I think I'll start working on it after my workload lightens up a little bit.

Nice! However, I'm thinking this may need more discussion before it's actually re-implemented.

tylerandari13:
[...]

PR #2742, opened 2024-01-24
Overload GameControllerManager::rumble to make it configurable

n0toose:
The default vibration duration was also decreased from 300ms to 100ms.


n0toose:
Wait, I completely forgot about joysticks.

n0toose:
Should hopefully be good to go? (Note: I don't own a joystick, so the changes affecting the joystick component are untested - even if they are presumed to be fine.)

src/control/game_controller_manager.cpp
```diff
@@ -325,6 +325,12 @@ GameControllerManager::has_corresponding_game_controller(int player_id) const
 
 int
 GameControllerManager::rumble(SDL_GameController* controller) const
+{
+  return this->rumble(controller, 0xFFFF, 0xFFFF, 100);
```
mrkubax10, line 329: No need to access method through `this`.

src/control/game_controller_manager.hpp
```diff
@@ -17,12 +17,13 @@
 #ifndef HEADER_SUPERTUX_CONTROL_GAME_CONTROLLER_MANAGER_HPP
 #define HEADER_SUPERTUX_CONTROL_GAME_CONTROLLER_MANAGER_HPP
 
+#include "control/controller.hpp"
```
mrkubax10, line 20: Local headers come after system headers. See [CODINGSTYLE.md](https://github.com/SuperTux/supertux/blob/master/CODINGSTYLE.md).

src/control/joystick_manager.cpp
```diff
@@ -327,17 +328,23 @@ JoystickManager::has_corresponding_joystick(int player_id) const
 }
 
 int
-JoystickManager::rumble(SDL_Joystick* controller) const
+JoystickManager::rumble(SDL_Joystick* joystick) const
+{
+  return this->rumble(joystick, 0xFFFF, 0xFFFF, 100);
```
mrkubax10, line 333: Same as above.

src/control/joystick_manager.hpp
```diff
@@ -18,11 +18,13 @@
...
 
-#include <SDL.h>
+#include "control/controller.hpp"
+
+#include <stdint.h>
 #include <vector>
 #include <unordered_map>
 
-#include "control/controller.hpp"
+#include <SDL.h>
```
mrkubax10, line 27: Same as previously too.

mrkubax10:
Besides my previous comments maybe it would make sense to add rumble configuration to options menu? Because currently this change seems little bit pointless.

n0toose:
> Because currently this change seems little bit pointless.

I haven't completed my work in trying to get `GameControllerManager` to communicate with `Player` yet - but I thought that I should send this very small refactor (which was the only working thing I got after completing my coding session) so that it's easier to pick it up again later or let someone else pick it up if I abandon it.

> Besides my previous comments maybe it would make sense to add rumble configuration to options menu?

I do not know if I should do that yet, as that would require me to answer the following: Should I use the `GameControllerManager` class for storing these settings, instead of `Player`? I do not have a perfect understanding of how these components work together, and would appreciate some assistance.

Later on, I'd like to use information such as the direction that a player was hit from or whether the player got a fire flower more easily later.

MatusGuy:
I think you should trigger rumble when there's any type of screenshake, which does not require player

n0toose:
> I think you should trigger rumble when there's any type of screenshake, which does not require player

Hm, would be a good start and one could immediately combine it with https://github.com/SuperTux/supertux/issues/2741.

n0toose:
> I think you should trigger rumble when there's any type of screenshake, which does not require player

I'll look at whether I can get the Camera class to talk to GameControllerManager.

MatusGuy:
it's not a singleton?

PR #2747, opened 2024-01-26
Properly handle SoundObject play time

### LINKED ISSUE #2744: Sound Object plays sounds of all sectors at once
**Example:**
A level with 3 sectors - each of which has one Sound Object present, playing a sound - will play all of these three sounds at the same time in one sector instead of only playing the sound from the Sound Object in that specific sector.

Another thing (since I forgot about it last time), you can hear the sound of the Sound Objects during the level intro screen, which should be fixed also!

mrkubax10:
Fixes two issues:
- OpenAL source not properly stopping via using `alSourcePause` instead of `alSourceStop`
- `SoundObject`s playing from currently inactive `Sector`s via playing OpenAL source in `update()`

Fixes #2744

### DIFF
--- a/src/audio/sound_source.hpp
+++ b/src/audio/sound_source.hpp
@@ -30,6 +30,7 @@ public:
 
   virtual void play() = 0;
   virtual void stop(bool unload_buffer = true) = 0;
+  virtual void pause() {}
   virtual bool playing() const = 0;
 
   virtual void set_looping(bool looping) = 0;
--- a/src/audio/sound_source.hpp
+++ b/src/audio/sound_source.hpp
@@ -30,7 +30,7 @@ public:
 
   virtual void play() = 0;
   virtual void stop(bool unload_buffer = true) = 0;
-  virtual void pause() {}
+  virtual void pause() = 0;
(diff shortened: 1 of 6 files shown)


Rusty-Box (approved):

src/audio/sound_source.hpp
```diff
@@ -30,6 +30,7 @@ class SoundSource
 
   virtual void play() = 0;
   virtual void stop(bool unload_buffer = true) = 0;
+  virtual void pause() {}
```
Vankata453, line 33: Maybe this function should be pure like the others?

MatusGuy (approved):

src/audio/sound_source.hpp
mrkubax10: I don't think that there will be any improvement after this change. It will require overriding that method in `DummySoundSource` too.

Vankata453: There wouldn't be a functional improvement, but will ensure any future sound sources will override this method. Also allows for consistency.

Vankata453 (approved):

PR #2748, opened 2024-01-26
Background ".sprite" support

### LINKED ISSUE #1089: Allow us to set .sprite files as backgrounds
Allows for animated backgrounds. You could also allow us to set the background's sprite action through scripting (like what is possible with FloatingImages), e.g. the level background is a volcano, once the player flips a lever the volcano in the background begins to erupt

Vankata453:
Backgrounds can now load ".sprite" files for top, middle and bottom images. The current action for any or all of those images can be set via scripting.

Backgrounds are not animated in the editor to prevent inconvenience.

Closes #1089.


src/object/background.cpp
```diff
@@ -356,15 +360,21 @@ Background::draw_image(DrawingContext& context, const Vector& pos_)
...
-              canvas.draw_surface(m_image_bottom, p, 0.f, m_color, m_blend, m_layer);
+              m_image_bottom->set_color(m_color);
+              m_image_bottom->set_blend(m_blend);
+
+              m_image_bottom->draw(canvas, p, m_layer);
             }
             else
             {
-              canvas.draw_surface(m_image, p, 0.f, m_color, m_blend, m_layer);
+              m_image->draw(canvas, p, m_layer);
```
tobbi, line 377: Aren't you forgetting setting color and blend here?

Vankata453: Lines 303, 304 have already set them.

mrkubax10 (approved):

src/object/background.cpp
```diff
@@ -459,15 +469,25 @@ std::unordered_map<std::string, std::string> fallback_paths = {
 
 } // namespace
 
-SurfacePtr
+SpritePtr
```
tobbi, line 472: Can this method be `const`?

Vankata453: In fact, I believe it could be `static`.

tobbi:
Can you rebase this and mark the method static or whatever?

src/object/background.cpp
tobbi: Okay, so mark this static and the one after that.

src/object/background.cpp
```diff
@@ -396,94 +505,32 @@ Background::draw(DrawingContext& context)
...
+Background::set_bottom_image_action(const std::string& action)
+{
+  if (m_image_bottom)
+    m_image_bottom->set_action(action);
+}
 
-  new_path = default_dir + it->second;
-  return Surface::from_file(new_path);
+void
+Background::set_image_actions(const std::string& action)
```
tobbi, line 529: I'm not sure about the name of this method. It's too similar to the other one.

Frostwithasideofsalt (approved):
image
works

src/object/background.cpp
Vankata453: Not sure how else to name it better. Also, if someone has to debug something in the code, I believe they'd notice the additional `s` at the end.

weluvgoatz (approved):
Looks good for me but let's still wait for Tobbi's requests to be addressed.

MatusGuy (commented):
Looks fine. Will test pretty soon.

src/object/background.cpp
```diff
@@ -31,7 +33,105 @@
...
+  {"oiltux.jpg", "misc/oiltux.jpg"},
+  {"transparent_up.png", "misc/transparent_up.png"},
+  {"nightsky.png", "nightsky/nightsky.png"},
+  {"nightsky_bottom.png", "nightsky/nightsky_bottom.png"},
+  {"nightsky_middle.png", "nightsky/nightsky_middle.png"},
+  {"nightsky_para.png", "nightsky/nightsky_para.png"},
+  {"nightsky_top.png", "nightsky/nightsky_top.png"},
+};
+
+static SpritePtr load_background_sprite(const std::string& image_path)
```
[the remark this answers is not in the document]
Vankata453: They don't have to be, only "background.cpp" uses them.

MatusGuy (dismissed):
I really don't like the step option in the config menu. First time mobile port users will be extremely confused when they see the slider going from min to max instantly when using the left/right buttons because they don't know what a "slider step" is (and they're not supposed to).

When using the slider with the mouse on the volume controls, it lags the game. If you want I will investigate this.

I gotta say. I really like how the mouse can be used for precision and the arrows can be used for bringing up quick values. But I think the roles should be switched. When using the mouse, the value should be snapped to the defined step, and the arrow keys should be used for precision values, because if the mouse is somewhat defective, that "precision" is lost.

I'm excited for this!

src/object/background.cpp
```diff
@@ -31,7 +33,105 @@
...
+  if (image_path.empty())
+    return nullptr;
+
+  if (PHYSFS_exists(image_path.c_str()))
+    // No need to search fallback paths.
+    return SpriteManager::current()->create(image_path);
+
+  // Search for a fallback image in fallback_paths.
+  const std::string& default_dir = "images/background/";
+  const std::string& default_dir2 = "/images/background/";
```
mrkubax10, line 108: These shouldn't be references and also could be marked as `static`.

MatusGuy (changes requested):
Seems to work fine. Just change the name of that function and I'll approve this.

MatusGuy (approved):

mrkubax10 (approved):

Issue #2711, opened 2023-12-26
Improving + balancing the "Chasing Fish" Enemy


Rusty-Box:
Here are some proposed tweaks that should be done to the enemy "Chasing Fish" as it can be a bit annoying to deal with right now.

- Slightly decreasing the **Losing Distance** for easier avoidance
- No longer detect Tux when Fish is not facing Tux, allowing a bit of "stealth"
- Ease into max speed rather than switching from 0 to a 100 in an instant
   - If need be even decrease **Chasing Speed** ever so slightly
- Abort chasing when Tux is no longer in water or its view is obstructed by a solid wall
- Make **Tracking, Loosing** and **Chase Speed** no longer adjustable to prevent inconsistencey

Rusty-Box:
Another thing:

- Chase Fish should kill other badguys while in chase mode (like Haywire)
  - Useful when luring them into a Dive Mine to defeat themselves

weluvgoatz:
I agree with the easing aspect, as well as the aborting of chasing when Tux isn't in water, and the fish killing others while in chase mode, however I do disagree with some of the others:
- While designing, having the losing distance being slightly shorter significantly affected the gameplay in my opinion as the fish would stop chasing Tux almost immediately.  However, we should definitely still look into this.
- I prefer having the tracking, loosing, and chase speed be adjustable so that in different circumstances (such as a winding tunnel or a large open area) the fish can still be a threat.  
Also, the one asking so that the fish can't detect Tux if he's not facing it is weird, because then wouldn't that mean that the fish would stop chasing Tux as soon as he starts swimming away?

Rusty-Box:
>       Also, the one asking so that the fish can't detect Tux if he's not facing it is weird, because then wouldn't that mean that the fish would stop chasing Tux as soon as he starts swimming away?

This only goes for initializing the chase. once they are in "chase mode" they will not be affected by this. This is purely to give the player the chance to avoid getting chase altogether if playing carefully around them!

weluvgoatz:
Oh ok, thanks for clarifying

Brockengespenst:
I had a look a this issue yesterday and already played around a bit on the following branch: https://github.com/Brockengespenst/supertux/tree/improve_chasing_fish
I added the following:

- No longer detect Tux when Fish is not facing Tux, allowing a bit of "stealth"
    -  Feels indeed more natural that chasing fish only spots Tux when it is facing
- Abort chasing when Tux is no longer in water
[...]

PR #2749, opened 2024-01-26
Improve chasing fish

Brockengespenst:
I tried to implement all suggestions from #2711:
- Kill badguy on collision if in chase mode
- Only spot Tux if it is in water and fish is facing Tux
- Stop chasing if player leaves the water
- Do not track player if obstacles are in the way
- Enable gravity until fish is completely in water
- Use linear acceleration at beginning of chase

Feel free to test and leave comments.

### DIFF
--- a/src/badguy/fish_chasing.cpp
+++ b/src/badguy/fish_chasing.cpp
@@ -25,11 +25,16 @@ static const float TRACK_DISTANCE = 200.0f;
 static const float LOST_DISTANCE = 400.0f;
 static const float CHASE_SPEED = 300.0f;
 static const float REALIZATION_TIME = 0.7f;
+static const float ACCELERATION_TIME = 0.2f;
+static const float DECELERATION_TIME = 0.4f;
 
 FishChasing::FishChasing(const ReaderMapping& reader) :
   FishSwimming(reader, "images/creatures/fish/ice/greenfish.sprite"),
   m_chase_state(ChaseState::NORMAL),
   m_realization_timer(),
+  m_acceleration_timer(),
+  m_is_accelerating(false),
+  m_last_chase_velocity(),
   m_track_distance(),
   m_lost_distance(),
   m_chase_speed()
@@ -57,7 +62,12 @@ FishChasing::active_update(float dt_sec) {
   // Perform basic updates.
   BadGuy::active_update(dt_sec);
   m_in_water = !Sector::get().is_free_of_tiles(get_bbox(), true, Tile::WATER);
-  m_physic.enable_gravity((!m_frozen && m_in_water) ? false : true);
+
+  Rectf top_of_fish = get_bbox();
+  top_of_fish.set_height(1.0f);
+
+  const bool is_completely_in_water = !Sector::get().is_free_of_tiles(top_of_fish, true, Tile::WATER);
+  m_physic.enable_gravity(m_frozen || !is_completely_in_water);
 
   // Handle beached state when the fish is in water and beached_timer is active.
   if (m_in_water && m_beached_timer.started())
@@ -73,16 +83,19 @@ FishChasing::active_update(float dt_sec) {
   // Behavior - chase the nearest player.
   auto player = get_nearest_player();
   if (!player) return;
-  Vector p1 = m_col.m_bbox.get_middle();
-  Vector p2 = player->get_bbox().get_middle();
-  Vector dist = (p2 - p1);
+  const Vector p1 = m_col.m_bbox.get_middle();
+  const Vector p2 = player->get_bbox().get_middle();
+  const Vector dist = (p2 - p1);
+  const bool is_player_in_water = player->is_swimming() || player->is_swimboosting() || player->is_water_jumping();
+  const bool is_facing_player = (m_dir == Direction::LEFT && dist.x <= 0.0f) || (m_dir == Direction::RIGHT && dist.x >= 0.0f );
+  const bool can_see_player = Sector::get().free_line_of_sight(p1,p2, true, this);
 
   switch (m_chase_state)
   {
   case NORMAL:
     FishSwimming::active_update(dt_sec);
 
-    if (glm::length(dist) <= m_track_distance && m_in_water && !m_frozen)
+    if (glm::length(dist) <= m_track_distance && m_in_water && !m_frozen && is_player_in_water && is_facing_player && can_see_player)
     {
       m_realization_timer.start(REALIZATION_TIME);
       m_chase_state = ChaseState::FOUND;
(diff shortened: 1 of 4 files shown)


Rusty-Box:
This is looking great! Chasing Fish is now much easier to avoid or deal with. Good stuff!

Rusty-Box:
Only suggestion I have is whether or not you could make it so that badguys specifically do not obstruct the Chasing Fish's view. Or at the very least during "chase mode"

src/badguy/fish_chasing.cpp
```diff
@@ -57,7 +60,12 @@ FishChasing::active_update(float dt_sec) {
   // Perform basic updates.
   BadGuy::active_update(dt_sec);
   m_in_water = !Sector::get().is_free_of_tiles(get_bbox(), true, Tile::WATER);
-  m_physic.enable_gravity((!m_frozen && m_in_water) ? false : true);
+
+  Rectf top_of_fish = get_bbox();
+  top_of_fish.set_height(1.0f);
+
+  const bool is_completely_in_water { !Sector::get().is_free_of_tiles(top_of_fish, true, Tile::WATER) };
```
mrkubax10, line 67: `=` will probably look better. No need to use initializer.

src/badguy/fish_chasing.cpp
```diff
@@ -73,16 +81,24 @@ FishChasing::active_update(float dt_sec) {
...
   auto player = get_nearest_player();
   if (!player) return;
-  Vector p1 = m_col.m_bbox.get_middle();
-  Vector p2 = player->get_bbox().get_middle();
-  Vector dist = (p2 - p1);
+  const Vector p1 = m_col.m_bbox.get_middle();
+  const Vector p2 = player->get_bbox().get_middle();
+  const Vector dist = (p2 - p1);
+  const bool is_player_in_water { player->is_swimming() || player->is_swimboosting() || player->is_water_jumping()};
+  const bool is_facing_player { (m_dir == Direction::LEFT && dist.x <= 0.0f) || (m_dir == Direction::RIGHT && dist.x >= 0.0f ) };
```
mrkubax10, line 88: Same as above.

src/badguy/fish_chasing.cpp
```diff
@@ -73,16 +81,24 @@ FishChasing::active_update(float dt_sec) {
...
+
+  const bool is_direct_view_to_player_possbible { obj && *obj == player->get_collision_object() };
 
   switch (m_chase_state)
   {
   case NORMAL:
     FishSwimming::active_update(dt_sec);
 
-    if (glm::length(dist) <= m_track_distance && m_in_water && !m_frozen)
+    if (glm::length(dist) <= m_track_distance && m_in_water && !m_frozen && is_player_in_water && is_facing_player && is_direct_view_to_player_possbible)
```
mrkubax10, line 101: Typo, `possible` is correct.

src/badguy/fish_chasing.cpp
```diff
@@ -57,7 +60,12 @@ FishChasing::active_update(float dt_sec) {
   // Perform basic updates.
   BadGuy::active_update(dt_sec);
   m_in_water = !Sector::get().is_free_of_tiles(get_bbox(), true, Tile::WATER);
-  m_physic.enable_gravity((!m_frozen && m_in_water) ? false : true);
+
+  Rectf top_of_fish = get_bbox();
+  top_of_fish.set_height(1.0f);
+
+  const bool is_completely_in_water { !Sector::get().is_free_of_tiles(top_of_fish, true, Tile::WATER) };
+  m_physic.enable_gravity((!m_frozen && is_completely_in_water) ? false : true);
```
Vankata453, line 68: `m_physic.enable_gravity(m_frozen || !is_completely_in_water);`

src/badguy/fish_chasing.cpp
Brockengespenst: I guess it was late when I wrote this. Thanks for pointing out!

src/badguy/fish_chasing.cpp
Brockengespenst: Definitiely makes more sense. Thanks.

Brockengespenst:
> Only suggestion I have is whether or not you could make it so that badguys specifically do not obstruct the Chasing Fish's view. Or at the very least during "chase mode"

Good point which I did not had in mind, to be honest. I will have a look at it.

src/badguy/fish_chasing.cpp
```diff
@@ -73,16 +81,24 @@ FishChasing::active_update(float dt_sec) {
...
+  const Vector p2 = player->get_bbox().get_middle();
+  const Vector dist = (p2 - p1);
+  const bool is_player_in_water = player->is_swimming() || player->is_swimboosting() || player->is_water_jumping();
+  const bool is_facing_player = (m_dir == Direction::LEFT && dist.x <= 0.0f) || (m_dir == Direction::RIGHT && dist.x >= 0.0f );
+
+  const CollisionSystem::RaycastResult player_collision_check { Sector::get().get_first_line_intersection(p1, p2, false, get_collision_object()) };
+
+  const auto* obj = std::get_if<CollisionObject*>(&player_collision_check.hit);
+
+  const bool is_direct_view_to_player_possible { obj && *obj == player->get_collision_object() };
```
Vankata453, line 94: This variable name is a bit too long. Maybe it can be renamed to something like `can_see_player`?

Additionally, `=` should be used.

src/badguy/fish_chasing.cpp
```diff
@@ -73,16 +81,24 @@ FishChasing::active_update(float dt_sec) {
...
-  Vector p1 = m_col.m_bbox.get_middle();
-  Vector p2 = player->get_bbox().get_middle();
-  Vector dist = (p2 - p1);
+  const Vector p1 = m_col.m_bbox.get_middle();
+  const Vector p2 = player->get_bbox().get_middle();
+  const Vector dist = (p2 - p1);
+  const bool is_player_in_water = player->is_swimming() || player->is_swimboosting() || player->is_water_jumping();
+  const bool is_facing_player = (m_dir == Direction::LEFT && dist.x <= 0.0f) || (m_dir == Direction::RIGHT && dist.x >= 0.0f );
+
+  const CollisionSystem::RaycastResult player_collision_check { Sector::get().get_first_line_intersection(p1, p2, false, get_collision_object()) };
```
Vankata453, line 90: Use `=` to set the variable.

Brockengespenst:
Other objects should be ignored now

> Only suggestion I have is whether or not you could make it so that badguys specifically do not obstruct the Chasing Fish's view. Or at the very least during "chase mode"

I slightly adapted the collision system to support checking for a specific object. With that change, Tux is seen by chasing fish also if other objects (maybe not necesarrily badguys, but I guess in water there may not be many objects that will affect this) are in the way. I my tests this change felt pretty good.

Please let me know if the change in the collision system is Ok. I did not want to duplicate code, that's why I added an optional parameter that is ignored by default so that there is no API break for existing code. Please let me know if anyone has a better solution.

mrkubax10:
I have noticed that initializers are used instead of simple `=` in other places too. Could you change that?

Rusty-Box:
Fantastic! One last thing - could you also make it so the Chasing Fish slowly decelerates when aborting the chase instead of coming to an instant halt? Makes the transition between chasing to not-chasing just as so smooth.

Brockengespenst:
> I have noticed that initializers are used instead of simple `=` in other places in your PR too. Could you change that?

Sure, done.

> I slightly adapted the collision system to support checking for a specific object
Nevermind, if existing API is used properly, this can be achieved much easier. I have reverted the changes.

> One last thing - could you also make it so the Chasing Fish slowly decelerates when aborting the chase instead of coming to an instant halt?
Makes sense. I will check.

Brockengespenst:
I added a deceleration phase which is a bit longer than the acceleration phase. This gives the fish a kind of "being exhausted from the chase" feeling.

mrkubax10 (approved):

Rusty-Box (approved):

MatusGuy (approved):

Vankata453 (approved):
Looks good code-wise.

Additional code-style suggestion you could apply if you want to, is to change:

```
if (...) {
} else {
}
```

to:

```
if (...)
{
}
else
{
}
```

for any such statement in the modified code.

Brockengespenst:
> Additional code-style suggestion you could apply if you want to, is to change:

I hope you don't mind that I also changed the one-liners to the same style (in my changes), so it looks more uniform.

Vankata453:
> I hope you don't mind that I also changed the one-liners to the same style (in my changes), so it looks more uniform.

It's fine. It's also fine without the brackets as a whole.

PR #2752, opened 2024-01-29
Misc fixes for Icy Island levels (story mode)

Miner34dev:
- Fix the notexture in Picnic With Penny instead of the Supertux title (caused by #2667)
- Fix switches direction in A fork in the road
- Fix krosh size in Ice in the Hole (caused by #2423)
- Fix krosh size and switch direction in The Crystal Mine


MatusGuy:
didn't someone in irc tell u that the levels were being remade?

PR #2753, opened 2024-01-30
Linked sprite support, custom `MovingSprite` light sprites

Vankata453:
Custom light sprites can now be added to all objects, which inherit `MovingSprite`. Setting a custom color for the light sprite is also possible, as long as the object is not designed to use foreign color values (for example, from an option).

To add a light sprite with an RGB color of `0.5, 0.2, 0.25`, in a `.sprite` file:

```
(linked-sprites
  (light "path/to/light/sprite" 0.5 0.2 0.25)
)
```

Keep in mind that specifying color values is optional. The entry can also end after specifying the file.

**Additionally:**

* Objects which utilize more than 1 sprite are now able to get the files of their additional sprites from the main one's `linked-sprites` list (the same block where `light` sprites can also be added, as shown above). This allows for expanded sprite customization for objects.

To add linked sprites, in a `.sprite` file:

```
(linked-sprites
  (key "path/to/sprite")
  (key2 "another/sprite/file")
)
```

Sprites, which the specific object does not support, will be ignored.

* Sprite actions now support the `flip-offset` property, which determines how much the sprite should be vertically offset when flipped vertically.

### DIFF
--- a/src/object/bonus_block.cpp
+++ b/src/object/bonus_block.cpp
@@ -138,7 +138,7 @@ BonusBlock::BonusBlock(const ReaderMapping& mapping) :
   if (m_contents == Content::LIGHT || m_contents == Content::LIGHT_ON)
   {
     SoundManager::current()->preload("sounds/switch.ogg");
-    m_lightsprite = m_sprite->get_linked_sprite("light");
+    m_lightsprite = m_sprite->get_linked_sprite("on-light");
     if (m_contents == Content::LIGHT_ON)
       set_action("on");
     else
--- a/src/object/bonus_block.cpp
+++ b/src/object/bonus_block.cpp
@@ -44,7 +44,6 @@
 #include "util/reader_mapping.hpp"
 #include "util/writer.hpp"
 #include "video/drawing_context.hpp"
-#include "video/surface.hpp"
 
 namespace {
 
@@ -139,7 +138,7 @@ BonusBlock::BonusBlock(const ReaderMapping& mapping) :
   if (m_contents == Content::LIGHT || m_contents == Content::LIGHT_ON)
   {
     SoundManager::current()->preload("sounds/switch.ogg");
-    m_lightsprite = Surface::from_file("/images/objects/lightmap_light/bonusblock_light.png");
+    m_lightsprite = m_sprite->get_linked_sprite("light");
     if (m_contents == Content::LIGHT_ON)
       set_action("on");
     else
@@ -163,6 +162,18 @@ BonusBlock::set_object(std::unique_ptr<GameObject> object)
 
   m_objects.clear();
   m_objects.push_back(std::move(object));
+}
+
+std::vector<MovingSprite::LinkedSprite>
+BonusBlock::get_linked_sprites()
+{
+  if (m_contents == Content::LIGHT || m_contents == Content::LIGHT_ON)
+  {
+    return {
+      { "light", m_lightsprite }
+    };
+  }
+  return {};
 }
 
 GameObjectTypes
@@ -688,7 +699,7 @@ BonusBlock::draw(DrawingContext& context)
   {
     Vector pos = get_pos() + (m_col.m_bbox.get_size().as_vector() - Vector(static_cast<float>(m_lightsprite->get_width()),
                                                                    static_cast<float>(m_lightsprite->get_height()))) / 2.0f;
-    context.light().draw_surface(m_lightsprite, pos, 10);
+    m_lightsprite->draw(context.light(), pos, 10);
   }
 }
 
@@ -748,7 +759,7 @@ BonusBlock::preload_contents(int d)
     case 6: // Light.
     case 15: // Light (On).
       SoundManager::current()->preload("sounds/switch.ogg");
-      m_lightsprite=Surface::from_file("/images/objects/lightmap_light/bonusblock_light.png");
+      m_lightsprite = m_sprite->get_linked_sprite("light");
       break;
 
     case 7:
--- a/src/object/bonus_block.cpp
+++ b/src/object/bonus_block.cpp
@@ -164,7 +164,7 @@ BonusBlock::set_object(std::unique_ptr<GameObject> object)
   m_objects.push_back(std::move(object));
 }
 
-std::vector<MovingSprite::LinkedSprite>
+MovingSprite::LinkedSprites
 BonusBlock::get_linked_sprites()
 {
   if (m_contents == Content::LIGHT || m_contents == Content::LIGHT_ON)
--- a/src/object/bullet.cpp
+++ b/src/object/bullet.cpp
@@ -30,7 +30,7 @@ Bullet::Bullet(const Vector& pos, const Vector& xm, Direction dir, BonusType typ
   physic(),
   life_count(3),
   sprite(),
-  lightsprite(SpriteManager::current()->create("images/objects/lightmap_light/lightmap_light-small.sprite")),
+  lightsprite(),
   type(type_)
 {
   physic.set_velocity(xm);
@@ -38,6 +38,7 @@ Bullet::Bullet(const Vector& pos, const Vector& xm, Direction dir, BonusType typ
   switch (type) {
     case FIRE_BONUS:
       sprite = SpriteManager::current()->create("images/objects/bullets/firebullet.sprite");
+      lightsprite = sprite->get_linked_sprite("light");
       lightsprite->set_blend(Blend::ADD);
       lightsprite->set_color(Color(0.3f, 0.1f, 0.0f));
       break;
@@ -61,12 +62,19 @@ void
 Bullet::update(float dt_sec)
 {
   // Cause fireball color to flicker randomly.
-  if (graphicsRandom.rand(5) != 0) {
-    lightsprite->set_color(Color(0.3f + graphicsRandom.randf(10) / 100.0f,
-                                 0.1f + graphicsRandom.randf(20.0f) / 100.0f,
-                                 graphicsRandom.randf(10.0f) / 100.0f));
-  } else
-    lightsprite->set_color(Color(0.3f, 0.1f, 0.0f));
+  if (lightsprite)
+  {
+    if (graphicsRandom.rand(5) != 0)
+    {
+      lightsprite->set_color(Color(0.3f + graphicsRandom.randf(10) / 100.0f,
+                                   0.1f + graphicsRandom.randf(20.0f) / 100.0f,
+                                   graphicsRandom.randf(10.0f) / 100.0f));
+    }
+    else
+    {
+      lightsprite->set_color(Color(0.3f, 0.1f, 0.0f));
+    }
+  }
 
   if (life_count <= 0)
   {
@@ -92,9 +100,8 @@ void
 Bullet::draw(DrawingContext& context)
 {
   sprite->draw(context.color(), get_pos(), LAYER_OBJECTS);
-  if (type == FIRE_BONUS){
+  if (lightsprite)
     lightsprite->draw(context.light(), m_col.m_bbox.get_middle(), 0);
-  }
 }
 
 void
--- a/src/object/bullet.cpp
+++ b/src/object/bullet.cpp
@@ -38,9 +38,6 @@ Bullet::Bullet(const Vector& pos, const Vector& xm, Direction dir, BonusType typ
   switch (type) {
     case FIRE_BONUS:
       sprite = SpriteManager::current()->create("images/objects/bullets/firebullet.sprite");
-      lightsprite = sprite->get_linked_sprite("light");
-      lightsprite->set_blend(Blend::ADD);
-      lightsprite->set_color(Color(0.3f, 0.1f, 0.0f));
       break;
 
     case ICE_BONUS:
@@ -53,6 +50,10 @@ Bullet::Bullet(const Vector& pos, const Vector& xm, Direction dir, BonusType typ
       sprite = SpriteManager::current()->create("images/objects/bullets/firebullet.sprite");
       break;
   }
+
+  lightsprite = sprite->get_linked_light_sprite();
+  if (lightsprite)
+    lightsprite->set_blend(Blend::ADD);
 
   m_col.m_bbox.set_pos(pos);
   m_col.m_bbox.set_size(sprite->get_current_hitbox_width(), sprite->get_current_hitbox_height());
--- a/src/object/explosion.cpp
+++ b/src/object/explosion.cpp
@@ -28,7 +28,6 @@
 #include "object/weak_block.hpp"
 #include "supertux/sector.hpp"
 #include "sprite/sprite.hpp"
-#include "sprite/sprite_manager.hpp"
 #include "supertux/sector.hpp"
 
 Explosion::Explosion(const Vector& pos, float p_push_strength,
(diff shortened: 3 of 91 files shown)


Rusty-Box:
Do the light sprites work per action or is it one for all?

Vankata453:
> Do the light sprites work per action or is it one for all?

One for all, currently. I think per-action could be done, though.

Rusty-Box:
> One for all, currently. I think per-action could be done, though.

That would be nice to have for sure

Vankata453:
> One for all, currently. I think per-action could be done, though.

I'm thinking it could be done by having the same block available in `action`, and if a sprite with the same key as in the global block is available, then prioritize the one of the action.

mrkubax10 (approved):

MatusGuy:
What was this trying to fix?

Rusty-Box:
@Vankata453 My brain explodes by trying to form this words so I'll do it like this. Can I:

- Make a light `.sprite` file with 2+ actions
- Link that `.sprite` file to the object's .sprite file
- Assign individual actions of light sprite to each action of the object .sprite if needed

The way I originally was hoping we could achieve that was through a special tag within an action.
Here is an example `Dive Mine.sprite` file using the theoretical `([key name]-action "[action name]")` tag:

```(supertux-sprite
 (action
  (name "left")
  (fps 12.0)
  (hitbox 14 19 32 32)
  (ticking-glow-action "idle")
  (images "left-0.png"
          "left-1.png"
          "left-2.png"
          "left-3.png"
		  "left-4.png"
		  "left-5.png"
		  "left-6.png"
		  "left-7.png"
		  "left-8.png"
		  "left-9.png"
		  "left-10.png"
		  "left-11.png"))

 (action
  (name "right")
  (fps 12.0)
  (hitbox 14 19 32 32)
  (ticking-glow-action "idle")
  (mirror-action "left"))
 
 (action
  (name "iced-left")
  (hitbox 5 8 32 32)
  (ticking-glow-action "idle")
  (images "left-0.png"))
 
 (action
  (name "iced-right")
  (hitbox 5 8 32 32)
  (ticking-glow-action "idle")
  (mirror-action "iced-left"))

 (action
  (name "ticking-left")
  (fps 15.0)
  (hitbox 14 19 32 32)
  (ticking-glow-action "ticking")
  (images "ticking-0.png"
          "ticking-1.png"
          "ticking-2.png"
          "ticking-3.png"
          "ticking-4.png"
		  "ticking-5.png"
		  "ticking-6.png"
		  "ticking-7.png"
		  "ticking-8.png"
		  "ticking-9.png"))

 
 (action
  (name "ticking-right")
  (fps 15.0)
  (hitbox 14 19 32 32)
  (ticking-glow-action "ticking")
  (mirror-action "ticking-left"))

  (linked-sprites
    (ticking-glow "images/creatures/dive_mine/ticking_glow/ticking_glow.sprite")
  )
)
```


The way the tag would work is that it will check for action names within the linked sprite file (in this case `dive_mine/ticking_glow/ticking_glow.sprite`), i.e. "idle" and "ticking".

Vankata453:
@Rusty-Box Action-specific `linked-sprites` are already supported, so what I believe only needs to be done to support this is to be able to link sprites with custom default actions.

I modified your example, so by default it globally (for all actions) links the `ticking-glow` sprite with the action `idle` (since most actions seem to use it). Since only the last two actions use the `ticking` action from the sprite, it's declared again in a new `linked-sprites` block inside of them, with the same name (thus it will override the global `ticking-glow` linked sprite, which has the `idle` action).

At the end, the modified example looks like this:

```
 (action
  (name "left")
  (fps 12.0)
  (hitbox 14 19 32 32)
  (images "left-0.png"
          "left-1.png"
          "left-2.png"
          "left-3.png"
		  "left-4.png"
		  "left-5.png"
		  "left-6.png"
		  "left-7.png"
		  "left-8.png"
		  "left-9.png"
		  "left-10.png"
		  "left-11.png"))

 (action
  (name "right")
  (fps 12.0)
  (hitbox 14 19 32 32)
  (mirror-action "left"))
 
 (action
  (name "iced-left")
  (hitbox 5 8 32 32)
  (images "left-0.png"))
 
 (action
  (name "iced-right")
  (hitbox 5 8 32 32)
  (mirror-action "iced-left"))

 (action
  (name "ticking-left")
  (fps 15.0)
  (hitbox 14 19 32 32)
  (linked-sprites
    (ticking-glow "images/creatures/dive_mine/ticking_glow/ticking_glow.sprite" "ticking")
  )
  (images "ticking-0.png"
          "ticking-1.png"
          "ticking-2.png"
          "ticking-3.png"
          "ticking-4.png"
		  "ticking-5.png"
		  "ticking-6.png"
		  "ticking-7.png"
		  "ticking-8.png"
		  "ticking-9.png"))

 
 (action
  (name "ticking-right")
  (fps 15.0)
  (hitbox 14 19 32 32)
  (linked-sprites
    (ticking-glow "images/creatures/dive_mine/ticking_glow/ticking_glow.sprite" "ticking")
  )
  (mirror-action "ticking-left"))

 (linked-sprites
   (ticking-glow "images/creatures/dive_mine/ticking_glow/ticking_glow.sprite" "idle")
 )
```

Vankata453:
> Wait but Dive Mine's glow never changes. Rather is disappears when it should begins playing the ticking action, as of now. Is there just no way to define that right now? I'm so confused

As I mentioned in the beginning, it still has to be done, I just offered a way it could work.

Rusty-Box:
Okay. Good to know. I thought I was going crazy here 😅
In that case I guess my only note would be that I feel like my proposal looks a bit better in terms of readbilty and having to type less but I ain't a coder so I don't wanna pretend I know better. Atleast I know now that I wasn't crazy, so that's nice :]

Rusty-Box:
Is it just me or is it not working? I updated Dive Mine's sprite file to look like you showed in this example [here](https://github.com/SuperTux/supertux/pull/2753#issuecomment-1995652343) but it doesn't seem to work. The light keeps disapearing when you approach Dive Mine instead of changing the action of the linked sprite. Am I doing something wrong here?

Rusty-Box:
Also the candle light doesn't seem to be rendered as a lightmap since you get a black box when you place it

Rusty-Box:
Another thing: the lit object crashes the game if you place it

Rusty-Box:
@MatusGuy Why does Dart Traps light have numbers in its linked sprite and why (when I change it) it either crashes my game or doesn't show any glowing. Is it hardcoded? It should care what I put there, should it? Since this is were you set it up to how you please when it comes to naming. It works with Dive Mine seemingly and I tried a similar naming convention.

Rusty-Box:
So there's missing sprites for the sparkle effect on keys and gold bomb uses the wrong glow path. It uses the same as Mr.Bomb but the image path looks for one in the gold bomb folder (obviously there's nothing there).

I'll test adding more linked sprites to some enemies myself and see if that works as well.

weluvgoatz (changes requested):
The strawbox also is missing its lightsprite when enflamed, as it produces a black missing texture image.

src/object/bonus_block.cpp
```diff
@@ -748,7 +759,7 @@ BonusBlock::preload_contents(int d)
     case 6: // Light.
     case 15: // Light (On).
       SoundManager::current()->preload("sounds/switch.ogg");
-      m_lightsprite=Surface::from_file("/images/objects/lightmap_light/bonusblock_light.png");
+      m_lightsprite = m_sprite->create_linked_sprite("light");
```
MatusGuy, line 762: ...What? So `on-light` and `light` are different? And isn't `light` reserved for the light linked sprite? And don't you think the `m_lightsprite` name gets confusing?

src/object/explosion.cpp
```diff
@@ -171,7 +164,8 @@ Explosion::update(float )
       break;
 
     case E_STATE_EXPLODING:
-      m_color.alpha = std::min(m_fading_timer.get_progress(), 1.f);
+      if (m_light_sprite)
+        m_light_sprite->get_color().alpha = std::min(m_fading_timer.get_progress(), 1.f);
```
MatusGuy, line 168: So if sprites can have their own color then why is `Candle` for example using its own `m_lightcolor`?

src/object/moving_sprite.cpp
```diff
@@ -56,13 +58,12 @@ MovingSprite::MovingSprite(const ReaderMapping& reader, const std::string& sprit
...
   m_default_sprite_name(sprite_name_),
   m_sprite(),
+  m_light_sprite(),
   m_layer(layer_),
   m_flip(NO_FLIP),
   m_sprite_found(false),
   m_custom_layer(reader.get("z-pos", m_layer))
 {
-  reader.get("x", m_col.m_bbox.get_left());
-  reader.get("y", m_col.m_bbox.get_top());
```
MatusGuy, line 65: What happened to these fellas?

src/object/powerup.cpp
```diff
@@ -96,7 +93,7 @@ PowerUp::get_default_sprite_name() const
     case COFFEE:
       return "images/powerups/retro/coffee.png";
     case HERRING:
-      return "images/powerups/retro/golden_herring.png";
+      return "images/powerups/retro/golden_herring.sprite";
```
MatusGuy, line 96: Why was the herring changed to a sprite but not the coffee? Does the herring glow?

src/object/bullet.cpp
```diff
@@ -53,6 +51,10 @@ Bullet::Bullet(const Vector& pos, const Vector& xm, Direction dir, BonusType typ
       break;
   }
 
+  lightsprite = sprite->create_linked_light_sprite();
+  if (lightsprite)
+    lightsprite->set_blend(Blend::ADD);
```
MatusGuy, line 56: Hey, isn't this already done by the `create_linked_light_sprite` function?

src/sprite/sprite_data.cpp
```diff
@@ -34,10 +35,64 @@
...
+      if (arr.size() >= 3) // Default action has been specified
+      {
+        linked_light_sprite->action = arr[2].as_string();
+
+        if (arr.size() >= 6) // Color has been specified
+        {
+          linked_light_sprite->color = Color(arr[3].as_float(), arr[4].as_float(),
+                                             arr[5].as_float());
+        }
+      }
```
MatusGuy, line 71: You don't need to nest these
```suggestion
      if (arr.size() >= 3) // Default action has been specified
      {
        linked_light_sprite->action = arr[2].as_string();
      }
      
      if (arr.size() >= 6) // Color has been specified
      {
        linked_light_sprite->color = Color(arr[3].as_float(), arr[4].as_float(),
                                           arr[5].as_float());
      }
```

src/sprite/sprite_data.cpp
```diff
@@ -34,10 +35,64 @@
...
+    {
+      LinkedSprite linked_sprite = LinkedSprite(filepath);
+
+      if (arr.size() >= 3) // Default action has been specified
+      {
+        linked_sprite.action = arr[2].as_string();
+
+        if (arr.size() >= 4) // Default action loops have been specified
+          linked_sprite.loops = arr[3].as_int();
+      }
```
MatusGuy, line 83:
```suggestion
      if (arr.size() >= 3) // Default action has been specified
      {
        linked_sprite.action = arr[2].as_string();
      }
      
      if (arr.size() >= 4) // Default action loops have been specified
        linked_sprite.loops = arr[3].as_int();
```

src/supertux/menu/async_dialog.hpp
MatusGuy, line 1: Why?

Vankata453:
> When a sprite is loaded, and then the same sprite is loaded, does it parse again?

No, it's created through `SpriteManager::create()`, which caches `SpriteData`.

src/object/powerup.cpp
Vankata453: Yes, the herring glows.

src/sprite/sprite_data.cpp
Vankata453: I do, if it's not higher than or equal to 3, it can't be higher than or equal to 6.

src/supertux/menu/async_dialog.hpp
Vankata453: How did I push this??

src/object/moving_sprite.cpp
Vankata453: Already parsed by `MovingObject`.

weluvgoatz:
image
The key linked sprite is fixed!  But the strawbox one isn't.  Also, now the gold bomb just doesn't have a glow at all.

weluvgoatz:
I will test once the conflicts are resolved : )

Rusty-Box (approved):
Looking good now!

weluvgoatz (approved):
Works perfectly now!

external/sexp-cpp
MatusGuy, line 1: Wow really? What did you change here?

src/object/bonus_block.cpp
```diff
@@ -139,7 +138,7 @@ BonusBlock::BonusBlock(const ReaderMapping& mapping) :
   if (m_contents == Content::LIGHT || m_contents == Content::LIGHT_ON)
   {
     SoundManager::current()->preload("sounds/switch.ogg");
-    m_lightsprite = Surface::from_file("/images/objects/lightmap_light/bonusblock_light.png");
+    m_lightsprite = m_sprite->create_linked_sprite("on-light");
```
MatusGuy, line 141: This `m_lightsprite` name still hasn't been changed. In my opinion, if you are required to load & allocate another sprite, just to control when to draw it, then this system is heavily flawed. At least I think that's what's going on...

src/object/flower.cpp
```diff
@@ -26,34 +26,32 @@ Flower::Flower(BonusType _type, const std::string& custom_sprite) :
...
   }
   else if (type == BONUS_AIR) {
     sprite = SpriteManager::current()->create(custom_sprite.empty() ? "images/powerups/airflower/airflower.sprite" : custom_sprite);
     SoundManager::current()->preload("sounds/fire-flower.wav");
-    lightsprite->set_color(Color(0.15f, 0.0f, 0.15f));
   }
   else if (type == BONUS_EARTH) {
     sprite = SpriteManager::current()->create(custom_sprite.empty() ? "images/powerups/earthflower/earthflower.sprite" : custom_sprite);
     SoundManager::current()->preload("sounds/fire-flower.wav");
-    lightsprite->set_color(Color(0.0f, 0.3f, 0.0f));
```
MatusGuy, line 52: Where did all these lightsprite colors go?

MatusGuy (changes requested):
Still not sure about the efficiency on this. I think we should build some sort of system to store light textures as surfaces instead of sprite files. Most of these "light sprite files" have only one light with one action anyway. But we can leave this suggestion for another PR, because this one is already heavily demanded, even by me.

Almost there!

src/object/flower.cpp
Vankata453: The file.

src/object/bonus_block.cpp
Vankata453: What this is doing is only creating the "on-light" sprite when it's needed, which is only if the bonus block contains light.

Hypernoot:
What are the remaining tasks?

Hypernoot (approved):
I tested all flammable badguys, light sources, glowing powerups and weak block.
all work

tobbi:
Can you mark all `get_linked_sprites` methods as const?

Hypernoot:
Unfortunately, this cannot work, as a const function cannot return non-const references. Even if you change the definition of `LinkedSprites` to `std::unordered_map<std::string, const SpritePtr&>`, it won't work because the function `MovingSprite::on_sprite_update()` cannot work with `const SpritePtr&`.

provigz:
Since #3418 was reverted, I'm reopening this PR. It only has to have master merged in to fix conflicts.

Hypernoot:
Bugs known so far:
* [ ] Decals don't render (#3427)
* [ ] Linked sprites have no default values. If the definition is missing, it renders invalid texture placeholder instead. (#3434)
* [ ] Herring displays as missing texture.
* [ ] Invisible candles don't have their linked sprite defined.
* [ ] Falling blocks aren't rendered.

tobbi:
@provigz You didn't reopen. I'm going to.

PR #2756, opened 2024-01-30
Improve snow particle customizability and code style

bruhmoent:
Exposes more variables to the reader to allow for more customizable snow particles.

### DIFF
--- a/src/object/snow_particle_system.cpp
+++ b/src/object/snow_particle_system.cpp
@@ -22,15 +22,14 @@
 
 #include "math/random.hpp"
 #include "supertux/sector.hpp"
-#include "util/reader.hpp"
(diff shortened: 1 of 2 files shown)


src/object/snow_particle_system.cpp
```diff
@@ -21,140 +22,157 @@
 
 #include "math/random.hpp"
 #include "supertux/sector.hpp"
+#include "util/reader.hpp"
```
Vankata453, line 25: You don't need to include this.

Vankata453 (approved):

mrkubax10:
It looks like Clang GNU/Linux builds are failing.

mrkubax10 (approved):

PR #2757, opened 2024-01-30
Remove `CameraConfig`

### LINKED ISSUE #2602: `CameraConfig` provides an obscure way to configure the camera
`Camera` has an instance of the `CameraConfig` class, which sets various properties for the camera. The file, used for configuration, is `camera.cfg`, in the root directory.

This is very obscure. Ideally, it would either be made available via the config, configurable through the `Camera` object itself, or removed as a configuration feature altogether.

Vankata453:
I'm proposing the removal of this feature, because of its extremely obscure design and not being something regarded as needed.

Closes #2602.


mrkubax10:
I agree. Camera is an integral part of our game engine and it doesn't make sense to allow configuring it. It should be consistent.

MatusGuy (approved):
I also agree. This thing is just weird to deal with

mrkubax10 (approved):

PR #2769, opened 2024-02-04
Update HTML template page

Semphriss:
This will display a cleaner title to search engines, as well as a better description.


mrkubax10 (approved):

tobbi:
Speaking of which, I wish we could have it hosted on supertux.org

MatusGuy (approved):

Semphriss:
IIRC, we did discuss this a while ago, and it was said that the host used for the website couldn't handle the load of serving the files. If that's the case, I can offer to host a subdomain like `play.supertux.org` to serve the files.

PR #2622, opened 2023-09-04
MacOS workflow: Install harfbuzz normally again

tobbi:
See
https://github.com/orgs/Homebrew/discussions/4735#discussioncomment-6902146

PR #2779, opened 2024-02-13
Fix macOS CI

ZhongRuoyu:
- Run macOS CI on `macos-12`. macOS 11 is no longer supported by Homebrew (and Apple); many packages now need to be built from source.
- Fix `libjxl_cms`'s install name in `libjxl`. This caused CI to fail, as reported in https://github.com/orgs/Homebrew/discussions/4735#discussioncomment-8197975. CMake's `fixup_bundle` does not recognise install names with `@rpath`, so it fails while bundling `libjxl_cms` as `libjxl`'s dependency. Make CMake happy by changing its install name to a full path instead.

See also: https://github.com/SuperTux/supertux/pull/2622#issuecomment-1706816598

CC @tobbi

### DIFF
--- a/.ci_scripts/package.sh
+++ b/.ci_scripts/package.sh
@@ -2,7 +2,7 @@
 
 shopt -s nullglob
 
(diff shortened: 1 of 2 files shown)


tobbi:
Thanks again, @ZhongRuoyu ! 👍

PR #2781, opened 2024-02-13
Save VSync setting

### LINKED ISSUE #2780: VSync setting is not saved and resets upon restart
Your vsync setting is not saved and as far as I can tell it actually gets reset when you close and open the game again. It always seems to default to "Adaptive", but when I change it to "Yes" or "No" it will only apply until I restart my game.

tobbi:
Fixes #2780

I'm not sure if this is the right approach, as we seem to be checking for the best possible vsync value earlier and picking one.

What's saved is: boolean yes / no: Check for vsync and not the exact vsync value.

### DIFF
--- a/src/supertux/gameconfig.cpp
+++ b/src/supertux/gameconfig.cpp
@@ -51,7 +51,7 @@ Config::Config() :
   use_fullscreen(false),
 #endif
   video(VideoSystem::VIDEO_AUTO),
-  try_vsync(true),
+  vsync(),
   show_fps(false),
   show_player_pos(false),
   show_controller(false),
@@ -260,7 +260,7 @@ Config::load()
     std::string video_string;
     config_video_mapping->get("video", video_string);
     video = VideoSystem::get_video_system(video_string);
-    config_video_mapping->get("vsync", try_vsync);
+    config_video_mapping->get("vsync", vsync);
 
     config_video_mapping->get("fullscreen_width",  fullscreen_size.width);
     config_video_mapping->get("fullscreen_height", fullscreen_size.height);
@@ -416,7 +416,7 @@ Config::save()
   } else {
     writer.write("video", VideoSystem::get_video_string(video));
   }
-  writer.write("vsync", try_vsync);
+  writer.write("vsync", VideoSystem::current()->get_vsync());
 
   writer.write("fullscreen_width",  fullscreen_size.width);
   writer.write("fullscreen_height", fullscreen_size.height);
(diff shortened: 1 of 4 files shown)


MatusGuy:
Thanks tobbi

src/supertux/gameconfig.cpp
```diff
@@ -51,7 +51,7 @@ Config::Config() :
   use_fullscreen(false),
 #endif
   video(VideoSystem::VIDEO_AUTO),
-  try_vsync(true),
+  vsync(),
```
MatusGuy, line 54: Initialize this to something like 1

MatusGuy (changes requested):

mrkubax10 (approved):

MatusGuy (changes requested):
Something is horribly wrong about this. The program crashes upon close with this call stack:
```
Error: signal 11:
supertux2(_ZN12ErrorHandler17print_stack_traceEv+0x2f)[0x562d557935df]
supertux2(_ZN12ErrorHandler12handle_errorEi+0x3b)[0x562d5579366b]
/lib/x86_64-linux-gnu/libc.so.6(+0x42520)[0x7f5bbee42520]
supertux2(_ZN6Config4saveEv+0x1109)[0x562d557ae319]
supertux2(_ZN15ConfigSubsystemD1Ev+0x16)[0x562d555a3c76]
supertux2(_ZN4MainD1Ev+0x524)[0x562d555076a4]
supertux2(+0x254919)[0x562d55505919]
supertux2(main+0x6d)[0x562d554d17bd]
/lib/x86_64-linux-gnu/libc.so.6(+0x29d90)[0x7f5bbee29d90]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0x80)[0x7f5bbee29e40]
supertux2(_start+0x2e)[0x562d5550584e]
```

Could this be because you changed a boolean to an integer? If that's the case then we could make a new setting and deprecate the old one.

This pr also started doing weird things like saving to `~/.supertux2/config`, but I'm going to ignore that because i think that wasan issue on my end.

Take a closer look.

Vankata453:
At a first glance, I'd assume the issue is that `VideoSystem` has already been destroyed by the time the config is destroyed.

tobbi:
Okay, please try again.

MatusGuy (changes requested):
It looks like it doesn't crash on exit anymore. But it still doesn't save. It does look like it saves to the config file, but doesn't choose the correct option in the settings menu and instead always chooses on.

mrkubax10 (approved):

MatusGuy (approved):

PR #2786, opened 2024-02-18
Fix some enemies not going down slopes

MatusGuy:
Some enemies with big hitboxes didn't go down slopes. This was because the max drop height was too small. The max drop height for smart enemies is now `width / 2`.

I also added `LedgeBehavior`, which can be used with `WalkingBadguy::set_ledge_behavior`, instead of setting numbers like -1, 16 or 600

### DIFF
--- a/src/badguy/goldbomb.cpp
+++ b/src/badguy/goldbomb.cpp
@@ -168,7 +168,7 @@ GoldBomb::active_update(float dt_sec)
     return;
   }
 
-  if ((tstate == STATE_FLEEING || tstate == STATE_CORNERED) && on_ground() && might_fall(600+1))
+  if ((tstate == STATE_FLEEING || tstate == STATE_CORNERED) && on_ground() && might_fall(s_normal_max_drop_height+1))
   {
     // also check for STATE_CORNERED just so
     // the bomb doesnt automatically turn around
--- a/src/badguy/goldbomb.cpp
+++ b/src/badguy/goldbomb.cpp
@@ -40,8 +40,6 @@ static const float SAFE_DIST = 32.f * 10.f;
 
 static const float NORMAL_WALK_SPEED = 80.0f;
 static const float FLEEING_WALK_SPEED = 180.0f;
-static const int NORMAL_MAX_DROP_HEIGHT = 16;
-static const int FLEEING_MAX_DROP_HEIGHT = 600;
 
 GoldBomb::GoldBomb(const ReaderMapping& reader) :
   WalkingBadguy(reader, "images/creatures/gold_bomb/gold_bomb.sprite", "left", "right"),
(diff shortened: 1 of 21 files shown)


MatusGuy:
AWit
what

Rusty-Box:
Since you set it to "draft" you still want some reviews or... we wait with that?

MatusGuy:
I set this to draft because something went horribly wrong with the Git side. I'll fix this shortly

MatusGuy:
Should be good now

tobbi:
Very good PR, a-ledge-edly.

src/badguy/walking_badguy.hpp
```diff
@@ -24,6 +24,33 @@ class Timer;
...
+     * offscreen.
+     */
+    NORMAL,
+
+    /**
+     * Fall off any ledge.
+     */
+    FALL
+  };
+
```
Vankata453, line 53: You could probably fit these comments on one line, and by removing the empty lines between the enumerators, this would take up way less lines.

src/badguy/goldbomb.cpp
```diff
@@ -170,7 +168,7 @@ GoldBomb::active_update(float dt_sec)
     return;
   }
 
-  if ((tstate == STATE_FLEEING || tstate == STATE_CORNERED) && on_ground() && might_fall(FLEEING_MAX_DROP_HEIGHT+1))
+  if ((tstate == STATE_FLEEING || tstate == STATE_CORNERED) && on_ground() && might_fall(600+1))
```
Vankata453, line 171: What about having the non-hitbox relative ledge behaviour modes as private static constant variables to use in situations like these?

tobbi: Can the 600 be moved to a constant again?

MatusGuy: What do you mean by this?

Vankata453: In `WalkingBadguy` class, on the top:

```
static const float s_fleeing_max_drop_height;
```

for example. Same for the other 2. Behaviour would still be determined by `set_ledge_behavior()`, but objects will also have access to these constants.

src/badguy/walking_badguy.hpp
MatusGuy: I think the names are fine. Kind of the same level of verbosity as HitResponse

Rusty-Box (commented):
Does not fix 45°+ slopes but since that is not affected by the max drop height and seems to be an issue on its own, I'll approve! We can look into 45°+ slopes at another time

Rusty-Box (approved):

src/badguy/walking_badguy.hpp
```diff
@@ -21,9 +21,21 @@
 
 class Timer;
 
+static const int s_normal_max_drop_height = 600;
+
```
Vankata453, line 25: Put this under `WalkingBadguy` as `protected`.

MatusGuy: Ok but why?

src/badguy/walking_badguy.hpp
```diff
@@ -21,9 +21,21 @@
...
 /** Base class for Badguys that walk on the floor. */
 class WalkingBadguy : public BadGuy
 {
+public:
+  enum class LedgeBehavior
+  {
+
```
Vankata453, line 32: Empty line

src/badguy/walking_badguy.hpp
Vankata453: Scoping.

Vankata453 (approved):

PR #2787, opened 2024-02-18
Fix windows builds

MatusGuy:


tobbi:
Why was this merged when literally every OTHER CI failed?

MatusGuy:
tobbi i only realised afterwards

PR #2788, opened 2024-02-18
Fix crusher not changing sprites and only choosing small ice crusher

MatusGuy:


Rusty-Box (approved):
Minus some builds not working, approve!

mrkubax10 (approved):

PR #2789, opened 2024-02-19
Show editor mode below mouse cursor

tobbi:
I find it more intuitive to have the editor mode show below the cursor.

Even better would it be if the cursor itself could be replaced with the actual mode (rubber for "delete") etc.


Hume2:
I see just some code quality improvements. Did you commit the correct code?

Hypernoot:
Implemented by the editor overhaul

Issue #2592, opened 2023-08-20
Improving the `Wind` object


Rusty-Box:
While functional, wind could be giving a couple tweaks to make it more robust.

If you make wind too strong you get push with no way of fighting against it. Make it too low and you won't get blown at all and only get slowed down when walking against it. At lower speeds wind should actually push Tux slightly when standing

It feels to me that wind, instead of pushing Tux, it gives him a set amount of speed in a specific direction, resulting in it feeling very unnatural!

https://github.com/SuperTux/supertux/assets/20615048/fadd5c72-e612-419f-96c2-8c3fda26a459

https://github.com/SuperTux/supertux/assets/20615048/e19b3a8e-c0de-456b-8890-bec8a5ea30c9

---

Also, another idea I had regarding wind:

An underwater object type for the wind object, `Current`, that instead of wind particles uses bubbles.
It will function nearly identical to wind, however I assume it may require some slight adjustments for the underwater setting.

Some of these things maybe be:
- Dive Mine is unaffected by it (this allows for fun dodging underwater)
- You have better control within the current for better navigation and correcting

tylerandari13:
Sometimes the wind feels more natural when the acceleration is set to `10`, but this is only sometimes. I do agree though a lot of the time wind does feel unnatural, especially at slower speeds.

Also while I'm here can there also be a way to hide the particles altogether? Sometimes turning fancy particles off just isn't cutting it.

Rusty-Box:
> Sometimes the wind feels more natural when the acceleration is set to `10`, but this is only sometimes. I do agree though a lot of the time wind does feel unnatural, especially at slower speeds.

Eh... I tried that and uh, yeah no, unless in very, very, VERY specific circumstances I suppose.

> Also while I'm here can there also be a way to hide the particles altogether? Sometimes turning fancy particles off just isn't cutting it.

I do agree with that!

biggeryetbetter:
I would like to work on this; I agree that wind is quite buggy at the moment and could use some fixing. Issues #2570, #2490, and #2189 are closely related and could be fixed at the same time.

tobbi:
@biggeryetbetter I've assigned the issue to you.

tobbi:
If you need anything, just ask here.

Hypernoot:
Okay, so graphical tweaks may definitely stay in 0.7.0 milestone.

[...]

PR #2793, opened 2024-02-20
Wind fixes

biggeryetbetter:
Done:
- Player is not slowed down when they work with wind.
- Most flying enemies act as expected with wind.

Todo:
- "Current" variant of wind.

### DIFF
--- a/src/object/player.cpp
+++ b/src/object/player.cpp
@@ -2880,4 +2880,20 @@ Player::remove_collected_key(Key* key)
                          m_collected_keys.end());
 }
 
+void
+Player::add_wind_velocity(const Vector& velocity, const Vector& end_speed)
+{
+  // Only add velocity in the same direction as the wind.
+  if (end_speed.x > 0 && m_physic.get_velocity_x() < end_speed.x)
+    m_physic.set_velocity_x(std::min(m_physic.get_velocity_x() + velocity.x, end_speed.x));
+  if (end_speed.x < 0 && m_physic.get_velocity_x() > end_speed.x)
+    m_physic.set_velocity_x(std::max(m_physic.get_velocity_x() + velocity.x, end_speed.x));
+  if (end_speed.y > 0 && m_physic.get_velocity_y() < end_speed.y)
+    m_physic.set_velocity_y(std::min(m_physic.get_velocity_y() + velocity.y, end_speed.y));
+  if (end_speed.y < 0 && m_physic.get_velocity_y() > end_speed.y)
+    m_physic.set_velocity_y(std::max(m_physic.get_velocity_y() + velocity.y, end_speed.y));
+}
+
+
+
 /* EOF */
--- a/src/badguy/dart.cpp
+++ b/src/badguy/dart.cpp
@@ -96,6 +96,10 @@ void
 Dart::active_update(float dt_sec)
 {
   BadGuy::active_update(dt_sec);
+
+  m_physic.set_velocity_y(m_physic.get_velocity_y() * 0.9f);
+  m_col.set_movement(m_physic.get_movement(dt_sec));
+
   sound_source->set_position(get_pos());
 }
 
--- a/src/badguy/flyingsnowball.cpp
+++ b/src/badguy/flyingsnowball.cpp
@@ -32,6 +32,7 @@ const float GLOBAL_SPEED_MULT = 0.8f; /**< The overall movement speed/rate. */
 FlyingSnowBall::FlyingSnowBall(const ReaderMapping& reader) :
   BadGuy(reader, "images/creatures/flying_snowball/flying_snowball.sprite"),
   total_time_elapsed(),
+  prev_height(0.f),
   puff_timer()
 {
   m_physic.enable_gravity(false);
@@ -77,14 +78,21 @@ FlyingSnowBall::active_update(float dt_sec)
 
   // Put that function in a graphing calculator :
   // sin(x)^3 + sin(3(x - pi/3))/3
-  float targetHgt = std::pow(std::sin(delta), 3.f) +
-                    std::sin(3.f *
-                             ((delta - math::PI) / 3.f)
-                            ) / 3.f;
-  targetHgt = targetHgt * 100.f + m_start_position.y;
-  m_physic.set_velocity_y(targetHgt - get_pos().y);
+  float targetHgt = (
+                        std::pow(std::sin(delta), 3.f) +
+                        std::sin(3.f *
+                            ((delta - math::PI) / 3.f)
+                        ) / 3.f
+                    ) * 100.f;
 
-  m_col.set_movement(m_physic.get_movement(1.f));
+  m_physic.set_velocity_y(m_physic.get_velocity_y() + (targetHgt - prev_height) / dt_sec);
+  prev_height = targetHgt;
+
+  m_physic.set_velocity_x(m_physic.get_velocity_x() * pow(0.5f, dt_sec));
+
+  m_col.set_movement(m_physic.get_movement(dt_sec));
+
+  m_physic.set_velocity_y(0.f);
 
   auto player = get_nearest_player();
   if (player) {
--- a/src/badguy/flyingsnowball.hpp
+++ b/src/badguy/flyingsnowball.hpp
@@ -40,6 +40,7 @@ protected:
 
 private:
   float total_time_elapsed;
+  float prev_height;
   Timer puff_timer; /**< time until the next smoke puff is spawned */
 
 private:
--- a/src/badguy/kamikazesnowball.cpp
+++ b/src/badguy/kamikazesnowball.cpp
@@ -17,6 +17,7 @@
 #include "badguy/kamikazesnowball.hpp"
 
 #include "audio/sound_manager.hpp"
+#include "supertux/direction.hpp"
 
 namespace{
   static const float KAMIKAZE_SPEED = 200;
@@ -37,6 +38,14 @@ KamikazeSnowball::initialize()
   m_physic.set_velocity_x(m_dir == Direction::LEFT ? -KAMIKAZE_SPEED : KAMIKAZE_SPEED);
   m_physic.enable_gravity(false);
   set_action(m_dir);
+}
+
+void
+KamikazeSnowball::active_update(float dt_sec) {
+  m_physic.set_velocity_y(m_physic.get_velocity_y() * 0.9f);
+  m_col.set_movement(m_physic.get_movement(dt_sec));
+
+  BadGuy::active_update(dt_sec);
 }
 
 bool
--- a/src/badguy/kamikazesnowball.hpp
+++ b/src/badguy/kamikazesnowball.hpp
@@ -29,6 +29,7 @@ public:
 
   virtual void initialize() override;
   virtual void collision_solid(const CollisionHit& hit) override;
+  virtual void active_update(float dt_sec) override;
   static std::string class_name() { return "kamikazesnowball"; }
   virtual std::string get_class_name() const override { return class_name(); }
   static std::string display_name() { return _("Snowshot"); }
--- a/src/badguy/owl.cpp
+++ b/src/badguy/owl.cpp
@@ -102,6 +102,8 @@ Owl::is_above_player() const
 void
 Owl::active_update (float dt_sec)
 {
+  m_physic.set_velocity_y(m_physic.get_velocity_y() * 0.9f);
+
   BadGuy::active_update (dt_sec);
 
   if (m_frozen)
--- a/src/badguy/zeekling.cpp
+++ b/src/badguy/zeekling.cpp
@@ -184,6 +184,12 @@ Zeekling::active_update(float dt_sec) {
     BadGuy::active_update(dt_sec);
     return;
   } else if (state == DIVING) {
+    // Something took away our diving velocity so we should go back up.
+    if (m_physic.get_velocity_y() <= 0.0f) {
+      state = CLIMBING;
+      m_physic.set_velocity_y(-2*fabsf(m_physic.get_velocity_x()));
+      set_action(m_dir);
+    }
     BadGuy::active_update(dt_sec);
     return;
   } else if (state == CLIMBING) {
--- a/src/object/player.hpp
+++ b/src/object/player.hpp
@@ -164,6 +164,9 @@ public:
   /** Adds velocity to the player until given end speed is reached */
   void add_velocity(const Vector& velocity, const Vector& end_speed);
 
+  /** Version of `add_velocity` with modifications for wind physics */
+  void add_wind_velocity(const Vector& velocity, const Vector& end_speed);
+
   /** Returns the current velocity of the player */
   Vector get_velocity() const;
 
--- a/src/object/wind.cpp
+++ b/src/object/wind.cpp
@@ -136,18 +136,18 @@ Wind::collision(GameObject& other, const CollisionHit& )
     player->override_velocity();
     if (!player->on_ground())
 	  {
-      player->add_velocity(speed * acceleration * dt_sec, speed);
+      player->add_wind_velocity(speed * acceleration * dt_sec, speed);
     }
     else
     {
       if (player->get_controller().hold(Control::RIGHT) || player->get_controller().hold(Control::LEFT))
 	    {
-	      player->add_velocity(Vector(speed.x, 0) * acceleration * dt_sec, speed);
+	      player->add_wind_velocity(Vector(speed.x, 0) * acceleration * dt_sec, speed);
 	    }
 	    else
       {
 	      //When on ground, get blown slightly differently, but the max speed is less than it would be otherwise seen as we take "friction" into account
-	      player->add_velocity((Vector(speed.x, 0) * 0.1f) * (acceleration+1), (Vector(speed.x, speed.y) * 0.5f));
+	      player->add_wind_velocity((Vector(speed.x, 0) * 0.1f) * (acceleration+1), (Vector(speed.x, speed.y) * 0.5f));
 	    }
     }
   }


Rusty-Box:
A issue that does still persist is that when the wind is set to values (I believe) below Tux max. walk/run speed, Tux will not be moved by the wind when standing still, stuck in his run animation.

Also, I feel that wind should always affect things when they are in their zone. Right now if you don't jump it will not affect you at all, even though you are clearly inside the wind zone.


https://github.com/SuperTux/supertux/assets/20615048/a5823825-2fb1-44d5-8e3b-91ed76a7da46

biggeryetbetter:
I have made several changes to the wind physics process that seem to have fixed these behaviors. I've also added a prototype "current" that does not affect dive mines. I would like some clarification about what you mean in #2592:

> You have better control within the current for better navigation and correcting

By this, do you mean faster acceleration/deceleration?

Rusty-Box:
So the thought behind the current was that (since it is underwater) it can be used to suck/push Tux into an area. However, Tux should be able to swim against the current if it is not too strong. Now that should probably apply to both the current and wind. Main difference is that the current type would have different particles (i.e. bubbles) to better fit the aquatic theme.

Rusty-Box:
Somethig I noticed with the "Current". It is very hard to fight it. Especially with an acceleration of 100. You would expect setting the speed of wind lower than Tux's swim/boost speed, he'd still be able to slowly go against the current. However, I seemingly get force-halted by the current even thought techincally Tux should still be able to swim through (obviously slower). It lets me swim for a short moment before halting all my progress, even when oosting which is moves very fast! Even with an acceleration of 10.

https://github.com/SuperTux/supertux/assets/20615048/9c560152-edef-4080-bc6b-9b4ed24145bd

Also, make sure to have the "Current" particles fade out rather than suddenly disappearing in an instant.

Rusty-Box:
I would not be suprised if something similar applies to the normal wind as well

biggeryetbetter:
I'm admittedly a bit stuck on this issue. While the normal wind seems to be working correctly, it seems that in water, the acceleration from current is still vastly overpowering that from swimming. Unfortunately I'm not really sure how to fix that without also altering how swimming works- any ideas would be appreciated.

Rusty-Box:
> Unfortunately I'm not really sure how to fix that without also altering how swimming works- any ideas would be appreciated.

Altering swimming in what way? Swimming is still a fairly "unpolished" mechanic so alterations are still legitimate as long as they improve it. So depending on how would have to alter it, you might be able to go through with it.

biggeryetbetter:
Alright, I've tried tweaking swimming to actually work with wind. Let me know what you think. Boosting in particular seems to sometimes create large radii on turns which might need to be fixed.

Also, how can I make the water bubbles fade out?

Rusty-Box:
Found a couple bugs:
- Walking through wind and standing, Tux gets pushed much weaker as when he jumps afterwards

https://github.com/SuperTux/supertux/assets/20615048/db496486-2581-419d-a4ed-f804570aa505


- When underwater Tux will attempt to always move left when letting go of the directional keys.

https://github.com/SuperTux/supertux/assets/20615048/7908f659-6aa1-400a-9ed1-d474b8dc039d


- At higher acceleration, around 100 (default value) but also noticeable aroud 50, Tux begins kinda jittering in a current and also somewhat stop an go when at higher speed wind

https://github.com/SuperTux/supertux/assets/20615048/1b5d0fbf-a4d6-4497-b205-362b3c1251a9


---

It also seems that the wind affect Tux statically, i.e. instead of his speed accelerating or decelerating "smoothly" it changes instantly. Atleast that's how it felt to me. It should ease into the new faster/slower speed to make the transition more smooth. This seems to apply to both wind and current, if that is truly what is happening, maybe I'm mistaken here.

biggeryetbetter:
I can confirm the first and third bugs but I'm not able to recreate the second one. Can you upload the setup that is shown in the video?

Another bug: Pressing the down arrow rapidly while moving into fast wind somehow allows you to go through easier. This is caused by some stuff around line 1586 of `player.cpp` and I don't know what they are supposed to do so I won't touch them.

Rusty-Box:
> Can you upload the setup that is shown in the video?

This is the level I tested and discovered that bug:

[wind_test_level.zip](https://github.com/SuperTux/supertux/files/14735655/wind_test_level.zip)

Rusty-Box:
More bugs I found:
- Buttjumping in downwards moving wind breaks the animation, i.e. Tux gets stuck in his buttjump an/or stomp animation after touching the ground.
- Tux cannot duck when in wind that goes downwards

Rusty-Box:
I also noticed that when you swim in the Current and you move in the direction it goes, you kinda seem to break the speed at which it is pushing you, like you get a bit slower than when you just get pushed without moving yourself. This might also apply to normal wind. Not sure on that. I usually notice it best while swimming.

Rusty-Box:
Okay I keep finding many more bugs - sry 'bout that :D

So apparently, Dive Mine is still affected by currents but also when they leave the current/wind area they keep on going, never stopping. Also getting pushed underwater is still somewhat jittery. Feathering help a little bit but overall needs a much higher value than normal wind to be smoothed out fully, for the most part.

Speaking of feathering, so idk if I fully get what it does specifically but the best values I've seen with it seems to be around 350? Can shed some more light on what it does exactly because with default value of 16 it looks horribly jittery

biggeryetbetter:
Don't worry; that's what open-source is all about. I haven't done a particularly good job testing each change, to be honest.

Also, the way feathering works is that it calculates how far the center of your hitbox is from the edge of the wind area (in pixels), then divides it by the feathering distance and clamps it between 0 and 1. It wasn't introduced to fix the jittering (that is caused by friction) but rather the fact that tux would keep "bouncing" off the edge of strong wind instead of pushing up against it like you might expect.

My main problem with how it works right now is that if you put wind in a tunnel, like in the test level you showed previously, the top and bottom edges of the wind will have feathering, even though this makes no sense inside a tunnel.

Also, I found out that I was calculating `m_wind_velocity` (should actually be `m_wind_acceleration`) completely wrong which seems to have been what was going on when you said
> It is very hard to fight it. Especially with an acceleration of 100. You would expect setting the speed of wind lower than Tux's swim/boost speed, he'd still be able to slowly go against the current.

I will keep attacking these issues and hopefully we can get this to work soon!

Rusty-Box:
Any further progress on this? Just checking :)

biggeryetbetter:
Not really as I have again been busy with other things. It seems like the cause of the weird jittering bug is that wind only collides with tux every few game ticks and adds velocity, in between which, friction causes tux to suddenly slow down. That is what I have been stuck on currently.

biggeryetbetter:
Alright I think I got the jittering issue fixed finally... Sorry that took awhile. Let me know what you think and of any bugs you find.

Rusty-Box:
- Jittering appears to be fixed indeed. The only thing that happens is that Tux's walk animation start,stops,start,stops the whole time if you don't hold the direction key you're going.




https://github.com/SuperTux/supertux/assets/20615048/1da20ea0-b2c9-4507-b5dd-d073e9441d27




- Knowing that Tux walk and run speed can go up to iirc. 230 and 320 repsectivly, I'm suprised that if you walk against wind with a 100 value Tux barely manages to walk against it. As if the wind was suddenly way stronger. Also when Tux jumps still gets pushed stronger than when he is on the ground, which is odd.


https://github.com/SuperTux/supertux/assets/20615048/a869c117-5fa0-4745-9526-eff5089649ce


- Reaching the top of a wind area occassionally push Tux much higher + objects that walk or stand in up facing wind don't get pushed up. They're not even affectd by wind at all even the they should



https://github.com/SuperTux/supertux/assets/20615048/7e7680c7-27ff-4240-9057-21f8cf3e6edc




https://github.com/SuperTux/supertux/assets/20615048/496f1026-a648-4f7e-a13c-daae5fef3e27

biggeryetbetter:
I've been procrastinating on this a lot but I have finally gotten around to letting the vast majority of badguys interact with wind. I have not done this for ice crushers and ghouls because I'm not sure if that is desired.
Unfortunately, due to changes in how strong acceleration is, levels like "A Leaf In the Wind" will be broken with wind in its current state; the "acceleration" option will need to be cranked up quite a bit. Wind is however much more consistent with other sources of acceleration such as gravity, which is 1000 px/s^2 by default and can be mostly negated by using a wind object pointed upwards with an acceleration of 1000. For this reason I would personally recommend that levels simply be updated. Let me know what you think about this and the icecrushers/ghouls.

Rusty-Box:
> I have not done this for ice crushers and ghouls because I'm not sure if that is desired.

Sounds fine. Ghoul is a ghost so makes sense they are not affected really and Crushers don't really need it. At best have them fall slower wind etc but its not too essential.

> Unfortunately, due to changes in how strong acceleration is, levels like "A Leaf In the Wind" will be broken with wind in its current state; the "acceleration" option will need to be cranked up quite a bit.

Totally fine. Levelshave been fully remade from scratch for the next update thus will take this changes into account.

I will do further and more proper testing as soon as possible! Thank you for all this effort btw. :)

weluvgoatz (changes requested):
1) While I do like the swimming changes, please clarify them more verbosely next time (such as in the title or description of PR) because this changes Tux's swimming in a pretty big way.
2) The wind/current seems to not really work.  Okay, well, to be exact, it *barely* works.  I literally have to set the acceleration to 100,000 in order for it to replicate what a previous value of 100 would do.  I understand why it was made this way but it breaks a bunch of current levels, so it would be appreciated if something could be done to more closely align with the current design so that fewer levels break.

Rusty-Box:
Swimming feels now quite heavy. Not the worst thing in the world, I definitely don't hate the idea. Anyway, back to wind:

There still persists the issue that when getting pushed by wind and Tux standing on the ground he is getting pushed much weaker than when you are mid-air which feels rather off. I also agree with Daniel that the acceleration is quite weird as well. To get a descend upwards wind I had to set it to 2000, for example. Downward facing wind meanwhile doesn't seem to have any effect on you at all, even at 2000 acceleration.

biggeryetbetter:
I've made a change that was suggested by @MatusGuy in Discord a month or so back which was to put wind logic with the `Physic` class. Team members should let me know what they think.

> There still persists the issue that when getting pushed by wind and Tux standing on the ground he is getting pushed much weaker than when you are mid-air which feels rather off.

To be honest, I'm not really sure how to fix this. I would say just keep it as a sidenote for now.

> I also agree with Daniel that the acceleration is quite weird as well. To get a descend upwards wind I had to set it to 2000, for example. Downward facing wind meanwhile doesn't seem to have any effect on you at all, even at 2000 acceleration.

I can understand this, however it aligns much better with how physical values like speed and acceleration are handled within the code. Nonetheless we do already multiply sector gravity by 100 in the `Physic` class, so I decided to apply the same logic to wind. Levels like A Leaf In the Wind sort of work like they used to.

Rusty-Box:
Could you please resolve the merge conflicts on the PR so we can get builds for testing?

Rusty-Box:
Swimming has gotten rather worse here, or rather swim boosting. It is very sluggish now, you can't build up speed anymore properly for swim jumping. I like how it eases into faster speeds but the execution isn't so great.

The max speed should be reached much faster but still ease into it, turning also needs to be made just a tiny bit tinder. I'm also unable to push against currents of weaker strength when swim boosting even though technically Tux should be able to swim through. This might be linked to how slow you build up speed when boosting right now.

MatusGuy:
> To be honest, I'm not really sure how to fix this. 

Have you taken a look at the ground movement manager? It could possibly be related

Rusty-Box:
Swimming feels much better again. Still the issue of being unable to swim against a weaker currents is present but the rest seems great 👍

Rusty-Box:
Hold up, is Tux swim boost faster? I'm testing out some levels and he zoomin'. Maybe a bit too fast. It's like he keeps gaining speed underwater. I keep ramming into stuff :D

swagtoy:
I haven't tested this PR, nor do i like it's description, because I have a question regarding the swim updates. Is it directional? Meaning, does the turn still happen instantly or is there a bit of slowdown now when turning in the water?

I think an issue i've noticed with the old swimming is that turning is too easy, it's much like the frog suit in the mario games. I always wished there was some turnaround 'lag', where the player's turning is clamped to where they want to go (up, down, left, right). This creates an immersion of realistic swimming and also makes swimming feel more challenging (not just flying in midair sort-of)

Rusty-Box:
> I haven't tested this PR, nor do i like it's description, because I have a question regarding the swim updates. Is it directional? Meaning, does the turn still happen instantly or is there a bit of slowdown now when turning in the water?

With this PR turning has gotten a bit looser now, though Tux is rather fast which causes some issues. Especially while boosting.

> Not to mention, I would really like if swimming wasn't as fast as it is now.

I wouldn't be against fine tuning a couple values here (as long as swim boosting remains on the faster side for swim jumping mostly) though not sure if required for this PR specically as the main focus is wind.

swagtoy:
> With this PR turning has gotten a bit looser now, though Tux is rather fast which causes some issues. Especially while boosting.

Can you show a video of you turning?

biggeryetbetter:
Here you go:

https://github.com/user-attachments/assets/39c23414-40a1-442b-9a34-506db30495d4

Frostwithasideofsalt:
One thing i wish would be added back to this pr's swimming is the ability to jump out of the water with a lot of speed after dropping into the water with a lot of speed. 

https://github.com/user-attachments/assets/1ab5ec65-2eef-4622-8551-0d6f68dc875b

it let to a lot of cool and fun speed-running opportunities. currently when you drop into water from a high place you just loose all your momentum. I dont feel like something like this is required for this PR, but it would be really nice if this got added to this Pr, because it would be a shame to see this go.

Rusty-Box:
Definitely getting closer. While normal swimming is better now and boosting technically feels much better to control. An issue that arises is the jump height you gain from jumping out of the water. It is very low. Previously you used to be able to jump about 4-ish tiles. Now it's barely 2. While adding a little bit more speed for boosting, it may run the risk of making boosting uncontrollable again. You could try increasing the boost speed to 500? Not sure how much that will fix it.

Alasdairbugs (changes requested):
- The Wind is definitely an improvement
- The currents feature is very welcomed 

- There is still more to be desired when it comes to diving:

There is too high an emphasis placed on vertical height when holding up+right after diving into a water source, when coming out of it, it doesn't look or feel like a 45-degree angle, and the speed you have UP is higher than the speed you have RIGHT. It looks more like a 30-degree angle from falling. A consequence of this is that now you dont cover much horizontal ground from these kinds of dives.

Secondly, The speed conversion from falling doesnt seem to convert as fully into speed while in the water, this has been completely numbed, it seems like you "speed-cap" in the water too harshly. this looks especially odd when you leave it because it looks like you just gain a ton of speed from tux's flippers.

Thirdly the speed ratio from falling to the ratio when coming out of the water seems to be more than halved than what tux' momentum when diving.

The last two points seemed to be sourced from the fact that just diving out of water looks to give the exact same momentum as diving in to water from a height. it looks like they grant the same momentum, but only if you're falling ~30+ tiles or more does marginal speed gains start to occur, but then tux' terminal velocity hits and there isnt really much to be gained.

Fourthly, Tux doesnt gain speed as he reaches a liquid edge, I think in terms of Biology and physics this would make more sense because A: you'd assume the penguin knows they're surfacing, so would put some extra kick into trying to get out, and B: its easier to push around water on a surface than a meter down, the pressure is less there.

Fifth: Currently if diving in and holding UP+Direction, Tux falls into the water about 4-5 tiles deep, i'd rather this be 2-tiles deep at most. Not only would it make the Bubble tileset actually viable, but i think it also would look more natural especially if we're going to have a 45-degree (or close to) trajectory.

Sixth: previously when diving into water, if you had a small amount of horizontal momentum (by tapping right, or by being slightly pushed by wind) and then held UP only, you could get different angles of trajectories. I'd like to see this kind of feature maintained, it gives designers a lot of options for using diving mechanics.

Overall The wind changes are really good but the swimming changes are unfortunately really undesireable! It heavily reduces the amount of fun you get from gaining lots of horizontal speed from them, it reduces options for getting different trajectories based on what input you press during a dive entry. It improves swimming while deep in water in terms of controlling tux, but unfortunately is undesireable elsewhere. If this PR was split into two, where wind was in one and swimming in another, I would approve the Wind ones immediately. Please reach out if you want clarification on any points :)

Zwatotem:
Can we merge this into local branch, separate out the wind and then merge into master? @Alasdairbugs

Alasdairbugs:
Doing that would be best i think for now and let swimming changes be another PR. Swimming changes will require a lot more R&D and feedback from level designers and it wont make sense to halt changes to wind for that. If @Zwatotem or @biggeryetbetter could do that that would be nice :)

Personally i think any swimming changes are kind of lower priority? Swimming seems fine to me rn. the only issue is underwater currents not being a thing, thats the only *issue* i can really think of.

biggeryetbetter:
Sorry guys, I've been a bit busy lately but I can isolate the wind updates soon so you can merge them.

biggeryetbetter:
Wind changes are now isolated, I'm going to turn this into a normal pull request.

src/math/vector.hpp
```diff
@@ -42,6 +42,33 @@ inline Vector at_angle(Vector const& v, float angle)
...
+  if (to.x > 0 && from.x < to.x)
+    result.x = std::min(from.x + diff.x, to.x);
+  if (to.x < 0 && from.x > to.x)
+    result.x = std::max(from.x + diff.x, to.x);
+  if (to.y > 0 && from.y < to.y)
+    result.y = std::min(from.y + diff.y, to.y);
+  if (to.y < 0 && from.y > to.y)
+    result.y = std::max(from.y + diff.y, to.y);
+
+  return result;
```
Hypernoot, line 69: This looks pretty much like:
```
if (glm::dot(to - from, from) > 0) {
  return move_towards(from, to, d);
} else {
  return from;
}
```

Hypernoot (changes requested):

src/object/player.cpp
```diff
@@ -1173,6 +1184,11 @@ Player::apply_friction()
     friction *= (ICE_FRICTION_MULTIPLIER*(m_sliding ? 4.f : m_stone ? 5.f : 1.f));
   else
     friction *= (NORMAL_FRICTION_MULTIPLIER*(m_sliding ? 0.8f : m_stone ? 0.4f : 1.f));
+
+  // Air friction does not make sense when the air is moving with you!
```
Hypernoot, line 1188: Air drag depends on the object speed with respect to the air. If the velocity of the air is equal to the velocity of Tux, then indeed, there is no drag. However, when they have both different velocities, there is drag. That is actually the reason why wind blows object away.

However, it looks like the term friction stands for the friction between Tux and the ground. In such a case, the friction is one force and the air drag is another force.

Hypernoot:
@biggeryetbetter Are you going to work on the issues? If not, I will close this PR. You may reopen it any time later.

biggeryetbetter:
Please close this PR for now. I am currently working on a lot of other stuff at the moment and while I'd like to come back to this at some point I think it's going to be a bit tricky to get much further without some changes to the underlying physics engine, such as support for multiple sources of force/acceleration and some changes to how Tux moves. 
I think for 7.0 the existing wind should be fine.

Hypernoot:
Indeed, I concluded that the physics overhaul would be too much a breaking change. The underwater current subtype was already implemented, so I removed this task from the 0.7.0 milestone.

PR #2801, opened 2024-02-24
Fix fish dissapear when going to the bottom of the sector in water

### LINKED ISSUE #2800: Chasing fish dissapears after killing Tux
So while editing a new level i found out that when a chassing fish kills Tux it dissapears? Note that this is when Tux is swimming and it does not happen 100% of the time. I dont know what exactly caused this to happen nor was this bug already discovered. Im assuming because i used a slightly older nightly build but i really dont know. I even have a video of it.

https://github.com/SuperTux/supertux/assets/161161755/d84455c1-f362-4f90-b5be-c5463d6ca2b9

MatusGuy:
fixes #2800

### DIFF
--- a/src/badguy/fish_swimming.cpp
+++ b/src/badguy/fish_swimming.cpp
@@ -127,9 +127,20 @@ FishSwimming::collision_badguy(BadGuy& badguy, const CollisionHit& hit)
   return CONTINUE;
 }
 
+void FishSwimming::update(float dt_sec)
+{
+  // Don't allow dying by going below the sector.
+  if (m_in_water && get_bbox().get_bottom() >= Sector::get().get_height())
+  {
+    set_pos(Vector(get_bbox().get_left(),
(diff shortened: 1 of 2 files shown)


rapic45:
Thanks marty :)

Vankata453:
Perhaps a better way to get around this issue would be to modify `BadGuy`'s out-of-sector check (`badguy/badguy.cpp:190`)?

MatusGuy:
No

Vankata453:
You mean it's not a better solution?

MatusGuy:
It's not a better solution because this behaviour is specific to the fish and we need to check if its on water

Vankata453 (approved):

tobbi:
You guys both can't spell "disappear".

MatusGuy:
ok bro

mrkubax10 (approved):

PR #2802, opened 2024-02-25
ST OST: Improve the midday theme

Curly68:
Basically i made midday.ogg have a more funky and catchy bassline instead of the more muted and aimless one it had before. I also changed the melody near the end of the loop too.
License is CC-BY-SA 4.0 as always


mrkubax10:
Next time please use separate branch to open a pull request because if you reuse the same branch from previous PR (master in this case) previous commits will appear in new pull request.

Curly68:
Okay, noted. I need to figure out how Github works

Curly68:
Closed so i can make a new separate branch

PR #2804, opened 2024-02-25
Do not increment powerup counter on restart_level()

### LINKED ISSUE #2295: Powerup stacks are not reset when the level is reset
**SuperTux version:** Latest @ master (`v0.6.3-583-g04856232f`)

Entering a level with a flower stack, increasing that stack during the level, then resetting the level does not also reset the powerup stack to what is was when the level was initially started.

- Enter any level with a certain powerup stack.
- While playing, increase that stack by collecting flowers of the same type.
- While still having the (increased) stack, pause the game, and select "Restart level".
- Notice the stack is not reset to what is was at the first step.

Brockengespenst:
This prevents the powerup counter to be incremented unintentionally when the level is restarted for the first time.

Fixes #2295

### DIFF
--- a/src/object/player.cpp
+++ b/src/object/player.cpp
@@ -1878,7 +1878,7 @@ Player::add_bonus(BonusType type, bool animate)
 }
 
 bool
-Player::set_bonus(BonusType type, bool animate)
+Player::set_bonus(BonusType type, bool animate, bool increment_powerup_counter)
 {
   if (m_dying) {
     return false;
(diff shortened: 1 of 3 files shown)


Vankata453:
I've had this issue in the back of my mind for a long time (ever since it started being used as a speedrun strat), thanks for looking into it!

src/object/player.cpp
```diff
@@ -1863,10 +1863,10 @@ Player::set_bonus(BonusType type, bool animate)
...
     m_player_status.max_earth_time[get_id()] = 0;
   }
-  if (type == FIRE_BONUS) m_player_status.max_fire_bullets[get_id()]++;
-  if (type == ICE_BONUS) m_player_status.max_ice_bullets[get_id()]++;
-  if (type == AIR_BONUS) m_player_status.max_air_time[get_id()]++;
-  if (type == EARTH_BONUS) m_player_status.max_earth_time[get_id()]++;
+  if (type == FIRE_BONUS && increment_powerup_counter) m_player_status.max_fire_bullets[get_id()]++;
+  if (type == ICE_BONUS && increment_powerup_counter) m_player_status.max_ice_bullets[get_id()]++;
+  if (type == AIR_BONUS && increment_powerup_counter) m_player_status.max_air_time[get_id()]++;
+  if (type == EARTH_BONUS && increment_powerup_counter) m_player_status.max_earth_time[get_id()]++;
```
mrkubax10, line 1869: `increment_powerup_counter` condition could be moved to outer `if` containing other `if`s.

Brockengespenst: Good point, this is indeed much more readable. Thanks.
I changed it. At least git tells me - github might be a bit laggy at the moment.

mrkubax10 (approved):

Vankata453 (approved):

MatusGuy (approved):

PR #2807, opened 2024-02-26
Fix camera following tux when dying 

### LINKED ISSUE #2799: Camera keeps following Tux when dying
As of recently when Tux dies, the Camera keeps following him rather than remaing stationary.
Most noticeable in tall levels and dying high up.

Miner34dev:
fixes #2799

### DIFF
--- a/src/object/camera.cpp
+++ b/src/object/camera.cpp
@@ -315,13 +315,16 @@ Camera::update(float dt_sec)
 
   switch (m_mode) {
     case Mode::NORMAL:
-      if (Sector::current() && Sector::current()->get_object_count<Player>() > 1)
+      if (Sector::current())
       {
-        update_scroll_normal_multiplayer(dt_sec);
-      }
-      else
-      {
-        update_scroll_normal(dt_sec);
+        if (Sector::current()->get_object_count<Player>() > 1)
+        {
+          update_scroll_normal_multiplayer(dt_sec);
+        }
+        else if (!(Sector::current()->get_players()[0]->is_dying()))
+        {
+          update_scroll_normal(dt_sec);
+        }
       }
       break;
     case Mode::MANUAL:


src/object/camera.cpp
```diff
@@ -310,31 +310,34 @@ Camera::draw(DrawingContext& context)
...
-    default:
-      break;
+  if (Sector::current() && !(Sector::current()->get_players()[0]->is_dying()))
+  {
+    // Minimum scale should be set during the update sequence; else, reset it.
+    m_enfore_minimum_scale = false;
+
+    switch (m_mode) {
+      case Mode::NORMAL:
+        if (Sector::current() && Sector::current()->get_object_count<Player>() > 1)
```
tobbi, line 320: Redunant check for `Sector::current()`

Vankata453:
Perhaps this check should only be done right before `update_scroll_normal()`? This will allow for it to not include multiplayer (even so, it currently only checks for the first player anyway), as well as any other custom modes.

Miner34dev:
Anything else to change?  I'm ready to squash the commits

Vankata453:
> Perhaps this check should only be done right before `update_scroll_normal()`? This will allow for it to not include multiplayer (even so, it currently only checks for the first player anyway), as well as any other custom modes.

What about this?

src/object/camera.cpp
```diff
@@ -319,7 +319,7 @@ Camera::update(float dt_sec)
       {
         update_scroll_normal_multiplayer(dt_sec);
       }
-      else
+      else if (Sector::current() && !(Sector::current()->get_players()[0]->is_dying()))
```
Vankata453, line 322: Seeing the `if` above this also checks for `Sector::current()`, both statements could be nested in an `if (Sector::current())` to not check twice.

Vankata453 (approved):

mrkubax10 (approved):

tobbi (approved):

PR #2812, opened 2024-03-02
Fix max drop height on slopes

MatusGuy:
for example mr tree now falls off 1 tile ledges and walking badguys cant walk up some slopes


Rusty-Box:
Few bugs I found:
- Captain Snowball has trouble jumping of slopes that end in pits. They just drop to their doom
- Haywire constantly attempts to jump over up-facing slopes
- Mr. Tree, Stumpy and Big Granito turn around after the end of a slope, going left, and turn back half way going down to the right

src/badguy/badguy.cpp
```diff
@@ -139,6 +142,22 @@ BadGuy::BadGuy(const ReaderMapping& reader, const std::string& sprite_name,
...
+  Vector eye(0, get_bbox().get_bottom() + 1.f);
+  eye.x = (m_dir == Direction::LEFT ? get_bbox().get_left() : get_bbox().get_right());
+
+  Vector end(eye.x, eye.y + 256.f);
+  context.color().draw_line(eye, end, Color::GREEN, LAYER_GUI);
+
+  float dirmult = (m_dir == Direction::LEFT ? 1.f : -1.f);
+  float rearx = (m_dir == Direction::LEFT ? get_bbox().get_right() : get_bbox().get_left());
+
+  float soff = (get_width() / 5.f) * dirmult;
```
tobbi, line 154: What's `soff` stand for?

src/badguy/badguy.cpp
```diff
@@ -139,6 +142,22 @@ BadGuy::BadGuy(const ReaderMapping& reader, const std::string& sprite_name,
...
+  Vector seye(rearx - soff, eye.y);
+
+  float eoff = soff - (2.f * dirmult);
```
tobbi, line 157: Same for `eoff`. I think a better variable name is needed here.

src/badguy/badguy.cpp
```diff
@@ -833,25 +852,104 @@ BadGuy::try_activate()
...
+    RaycastResult result = Sector::get().get_first_line_intersection(eye, end, false, nullptr);
+    //bool slope = false;
+
+    auto tile_p = std::get_if<const Tile*>(&result.hit);
+    if (result.is_valid && result.box.get_top() - eye.y < static_cast<float>(height))
+    {
+      // The ground is within max drop height. Continue.
+      return false;
+    }
+    else if (tile_p && (*tile_p) && (*tile_p)->is_slope())
```
tobbi, line 882: Since we return from the previous statement, you can actually omit the `else`.

src/badguy/badguy.cpp
```diff
@@ -833,25 +852,104 @@ BadGuy::try_activate()
...
+      // Check if we are about to go down a slope.
+      AATriangle tri((*tile_p)->get_data());
+      if (tri.is_south() && (m_dir == Direction::LEFT ? tri.is_east() : !tri.is_east()))
+      {
+        // Switch to slope mode.
+        m_detected_slope = tri.dir;
+      }
+
+      // Otherwise, climb the slope like normal.
+      return false;
```
tobbi, line 893: You can remove the else block and put te `return false` at the end of the function, I think.

src/badguy/badguy.cpp
```diff
@@ -833,25 +852,104 @@ BadGuy::try_activate()
...
+  if (m_detected_slope != 0)
+  {
+    //AATriangle tri(m_detected_slope);
+    std::cout << "SLOPE " << m_detected_slope << std::endl;
+    //slope = true;
+
+    float dirmult = (m_dir == Direction::LEFT ? 1.f : -1.f);
+    float rearx = (m_dir == Direction::LEFT ? get_bbox().get_right() : get_bbox().get_left());
+
+    float soff = (get_width() / 5.f) * dirmult;
```
tobbi, line 911: see above.

src/badguy/badguy.cpp
```diff
@@ -833,25 +852,104 @@ BadGuy::try_activate()
...
+    Vector eye(rearx - soff, oy);
+
+    float eoff = soff - (2.f * dirmult);
```
tobbi, line 914: see above.

src/badguy/badguy.cpp
```diff
@@ -833,25 +852,104 @@ BadGuy::try_activate()
...
+      if (tile_p && (*tile_p) && (*tile_p)->is_slope())
+      {
+        // Still going down a slope. Continue.
+        //m_detected_slope = (*tile_p)->get_data();
+        return false;
+      }
+
+      // No longer going down a slope. Switch off slope mode.
+      m_detected_slope = 0;
+      std::cout << "aggh" << std::endl;
```
tobbi, line 947: aggh.

src/badguy/badguy.cpp
```diff
@@ -833,25 +852,104 @@ BadGuy::try_activate()
...
+
+    std::cout << result.is_valid << std::endl;
+
+    if (!result.is_valid)
+    {
+      // Turn around and climb the slope.
+      m_detected_slope = 0;
+      return true;
+    }
+    else
```
tobbi, line 927: same as above: Since we return, no need for an `else` block.

src/badguy/badguy.hpp
```diff
@@ -230,7 +230,7 @@ class BadGuy : public MovingSprite,
 
   /** Returns true if we might soon fall at least @c height
       pixels. Minimum value for height is 1 pixel */
-  bool might_fall(int height = 1) const;
+  bool might_fall(int height = 1);
```
tobbi, line 233: Can you tell me why this method is no longer const? Do we modify the object state with the changes? And if so: Could we prevent that?

src/badguy/walking_badguy.cpp
```diff
@@ -99,7 +99,7 @@ void WalkingBadguy::set_ledge_behavior(LedgeBehavior behavior)
       break;
 
     case LedgeBehavior::SMART:
-      max_drop_height = static_cast<int>(get_bbox().get_width()) / 2;
+      max_drop_height = 16.f;
```
tobbi, line 102: Does this number have any significance? Seems like a magic number. Just curious.

MatusGuy: It's half a tile

src/badguy/badguy.hpp
MatusGuy: It changes m_detected_slope. I do this because I want to prevent having two raycasts at a time. So to fix that I alternate between them by checking if m_detected_slope is 0 or not.

src/badguy/badguy.cpp
MatusGuy: This is debug printing :sweat_smile:

src/badguy/badguy.cpp
```diff
@@ -833,25 +852,104 @@ BadGuy::try_activate()
...
+  float oy = get_bbox().get_bottom() + 1.f;
+
+  if (m_detected_slope == 0)
+  {
+    Vector eye(0, oy);
+    eye.x = (m_dir == Direction::LEFT ? get_bbox().get_left() : get_bbox().get_right());
+
+    // TODO: change to max possible drop height
+    Vector end(eye.x, eye.y + 256.f);
+    //std::cout << eye << " " << end << std::endl;
```
tobbi, line 871: Just a reminder to remove this before merging.

Frostwithasideofsalt:
Left - how it acts with this PR
Right - how it acts without the PR/ how it should act
https://github.com/user-attachments/assets/e96b294d-5016-459f-aaa4-d09e60ab4ee2

Issue #2421, opened 2023-04-05
Suggestions for Supertux level editor improvments


Frostwithasideofsalt:
-Adding Zooming in and out in the editor

-Adding option to disable undo and backups (would be useful for reducing lag among other things) 

-Copy and paste for colors in editor 

-Adding tilemap groups

-Adding option to add a tint to the tilemap icons

PR #2813, opened 2024-03-02
Editor zooming

Vankata453:
By holding CTRL and scrolling with the mouse, the camera in the editor can now be zoomed in and out. There is a set zoom limit: `0.5` - `3.0`.

When zooming in, the camera position will progressively focus on the position of the mouse, relative to the screen center.

Additionally, "src/editor/util.*" has been removed, since it only provided an unneeded `draw_tile` wrapper function.

Mentioned in #2421.


mrkubax10:
It doesn't seem right :/

https://github.com/SuperTux/supertux/assets/63350514/da4a428a-b0db-4ebb-8e88-34dff68040d9

Vankata453:
> It doesn't seem right :/

You mean where it goes outside the sector border?

mrkubax10:
Yeah and also weird snapping while zooming but it's not visible on the video

Vankata453:
> Yeah and also weird snapping while zooming but it's not visible on the video

Not sure how to do smooth zooming. Maybe a solution would be to decrease the zoom sensitivity further.

Rusty-Box:
Can we have a way to tell how far in/out you zoomed in? Like a "x1.0", "x2.0" etc somewhere
I'll give this a closer look soon

Rusty-Box:
> Can we have a way to tell how far in/out you zoomed in? Like a "x1.0", "x2.0" etc somewhere I'll give this a closer look soon

This should still be added, also, would be nice to have a keyboard shortcut to instantly go back to a x1.0 zoom.

Rusty-Box (approved):
Looks good!

src/editor/overlay_widget.cpp
```diff
@@ -1487,7 +1491,20 @@ EditorOverlayWidget::draw(DrawingContext& context)
...
+  if (scale_percentage != 100)
+    context.color().draw_text(Resources::big_font, std::to_string(scale_percentage) + '%',
+                              Vector(context.get_width() - 140.f, 15.f),
+                              ALIGN_RIGHT, LAYER_OBJECTS + 1, Color::WHITE);
+
+  context.push_transform();
+  context.set_translation(m_editor.get_sector()->get_camera().get_translation());
+  context.transform().scale = scale;
+
+  draw_path(context);
```
MatusGuy, line 1507: what

MatusGuy (approved):

src/editor/overlay_widget.cpp
Vankata453: Please be more verbose with comments.

tobbi (approved):
I have no idea what I'm doing with my life.

Frostwithasideofsalt:
time 2 test

Frostwithasideofsalt (approved):

PR #2815, opened 2024-03-03
Move default spawnpoint and sector name ("main") into  static variables

tobbi:
Instead of hardcoding "main" for both, I moved it to a variable.

I don't know if this is the right place, though.

### DIFF
--- a/src/object/spawnpoint.hpp
+++ b/src/object/spawnpoint.hpp
@@ -25,6 +25,13 @@
 class ReaderMapping;
 class DrawingContext;
 
+namespace {
+ /**
+  * The spawnpoint that gets activated at the start of a game session)
+  */
+  const std::string DEFAULT_SPAWNPOINT_NAME = "main";
+}
+
 class SpawnPointMarker final : public MovingObject
 {
 public:
--- a/src/supertux/sector.hpp
+++ b/src/supertux/sector.hpp
@@ -20,6 +20,7 @@
 #include "supertux/sector_base.hpp"
 
 #include <vector>
+#include <string>
 #include <stdint.h>
 
 #include "collision/collision_system.hpp"
@@ -47,6 +48,13 @@ class Size;
 class TextObject;
 class TileMap;
 class Writer;
+
+namespace {
+ /**
+  * The sector that gets activated by default when a level is started
+  */
+  const std::string DEFAULT_SECTOR_NAME = "main";
+}
 
 /** Represents one of (potentially) multiple, separate parts of a Level.
     Sectors contain GameObjects, e.g. Badguys and Players. */
--- a/src/badguy/willowisp.cpp
+++ b/src/badguy/willowisp.cpp
@@ -22,6 +22,7 @@
 #include "editor/editor.hpp"
 #include "object/lantern.hpp"
 #include "object/player.hpp"
+#include "object/spawnpoint.hpp"
 #include "sprite/sprite.hpp"
 #include "supertux/game_session.hpp"
 #include "supertux/sector.hpp"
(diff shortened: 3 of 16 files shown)


MatusGuy (approved):
ok code. But someone please test this in game. You never know when something breaks

tobbi:
> ok code. But someone please test this in game. You never know when something breaks

My code never breaks.

MatusGuy:
"my code never breaks" Remember the "prevent drawing offscreen" pr?

tobbi:
> "my code never breaks" Remember the "prevent drawing offscreen" pr?

The rest of the game wasn't written well enough to support my incredible ideas.

MatusGuy:
But it broke the game!! Therefore it broke!!

tobbi:
I thought the game would adapt to my code style and just "not break" because it was convinced by the incredible quality of my code.

MatusGuy:
You stupid

src/object/spawnpoint.hpp
```diff
@@ -25,6 +25,13 @@
 class ReaderMapping;
 class DrawingContext;
 
+namespace {
+ /**
+  * The spawnpoint that gets activated at the start of a game session)
+  */
+  const std::string DEFAULT_SPAWNPOINT_NAME = "main";
+}
+
```
mrkubax10, line 34: Wouldn't it be better to declare `DEFAULT_SPAWNPOINT_NAME` as `extern` in header file and define it in source file? Currently this will cause that identifier getting duplicated for each translation unit.

src/supertux/sector.hpp
```diff
@@ -48,6 +49,13 @@ class TextObject;
 class TileMap;
 class Writer;
 
+namespace {
+ /**
+  * The sector that gets activated by default when a level is started
+  */
+  const std::string DEFAULT_SECTOR_NAME = "main";
+}
+
```
mrkubax10, line 58: Same as earlier.

MatusGuy:
Have you playtested these changes?

tobbi:
It is literally just a code change.

src/supertux/sector.hpp
```diff
@@ -49,6 +50,14 @@ class TextObject;
 class TileMap;
 class Writer;
 
+namespace {
```
MatusGuy, line 53: What is this empty namespace doing here

MatusGuy:
Ok but have you at least ran the game and the title screen showed up correctly?

src/supertux/sector.hpp
tobbi: In case you don't have enough names already and need space for more. 

Like for example: Albus Percival Wulfric Brian [... enter more names here] Dumbledore.

src/object/spawnpoint.hpp
```diff
@@ -25,6 +25,11 @@
 class ReaderMapping;
 class DrawingContext;
 
+/**
+* The spawnpoint that gets activated at the start of a game session)
+*/
+extern std::string DEFAULT_SPAWNPOINT_NAME;
```
Vankata453, line 31: Maybe these constants would be more fit in `supertux/constants.hpp`? You wouldn't need to `#include` the whole `SpawnPoint` or `Sector` object to use any of them.

Also, declare them as `const`, since they're not meant to be modified.

MatusGuy (approved):

src/object/spawnpoint.hpp
tobbi: I don't think `const` applies to this, since it's not bound to a class.

Vankata453: Not sure what you mean. I mean making the variable `const`.

tobbi: Yes of course. I'm stupid.

mrkubax10 (approved):

PR #2819, opened 2024-03-04
Custom layer option for `MovingSprite`

Vankata453:
`MovingSprite`s can now have their layer ("Z-pos") changed via an option.

The layer is compatible with object types. Any object, which must have a different layer, based on type, should override `MovingSprite::get_layer()` (the same way it's done with `MovingSprite::get_default_sprite_name()`).


tobbi (approved):
Seems to be alright.

mrkubax10 (approved):

Issue #2811, opened 2024-03-01
cannon Sprite Error


Chiquinho7788:
**SuperTux V0.6.3:** (2024-02-27)

In supertux, the latest development version has a problem with the automatic cannon sprite error.

#####EXPECTED BEHAIVOR

Should have both sides when you put the auto steering

![image](https://github.com/SuperTux/supertux/assets/139497254/0e00daf7-bbf0-449a-98e3-144df246a444)

##### ACTUAL BEHAVIOR;

There is only the left side where enemies exit when you turn on automatic steering.

![Captura de ecrã 2024-03-01 210626](https://github.com/SuperTux/supertux/assets/139497254/4c3d6be8-9ff1-496b-bb10-32a3ecd4966e)

PR #2821, opened 2024-03-06
Fix cannon sprite bug when direction set to auto (#2811)

springrain5916:
Fix cannon sprite bug when direction set to auto (#2811)

The sprite is now set to "center" in the draw method (in case the dispenser is a cannon and direction is set to auto). I am not sure if this is the best solution (the draw method is probably called very often, and will slow things down). The code changes were tested both in the level editor and in the game itself, seems working without errors: 
Screenshot from 2024-03-06 12-17-42


mrkubax10:
Superseded by #2824

PR #2822, opened 2024-03-08
Fix building on MSYS2 (CLANG64)

iahung2:
Trivial changes to `.cmake` files. Please have a look. Thank you.


tobbi:
I have no idea what re-percussions this has (I'm not a drumset, after all), but seems reasonable. Just not sure what would happen if you're cross-compiling.

iahung2:
> I have no idea what re-percussions this has (I'm not a drumset, after all), but seems reasonable. Just not sure what would happen if you're cross-compiling.

`MSVC` implied `WIN32`. Only `MSVC` and `clang-cl` define `MSVC`, and both are only available on `WIN32`.

https://cmake.org/cmake/help/latest/variable/MSVC.html

iahung2:
> I have no idea what re-percussions this has (I'm not a drumset, after all), but seems reasonable. Just not sure what would happen if you're cross-compiling.

Updated. Thank you for reminding me about cross-compiling.

tobbi (approved):
I'm approving this, but keep in mind I don't have any deeper insight into compiler issues. So I'd recommend at least one other review.

CMakeLists.txt
```diff
@@ -247,12 +247,14 @@ if(EMSCRIPTEN)
 endif()
 
 if(WIN32 AND NOT VCPKG_BUILD)
+if(NOT MINGW)
```
MatusGuy, line 250: Why no `AND NOT MINGW`?

iahung2: If you cross-compile from Linux, you will always use some form of MinGW.

MatusGuy: Ok... I actually didn't know that WIN32 is true when cross compiling, but still, why not merge both into a single if statement?

iahung2: `WIN32` is true. `NOT VCPKG_BUILD` is true. `NOT MINGW` is false. The final result is false. I afraid that it will break cross-compiling. I don't know anything about cross-compiling. I only use MSYS2 on Windows. But @tobbi raised the issue. I tried to do my best not to break it.

iahung2: Tested with `if(WIN32 AND NOT VCPKG_BUILD AND NOT MINGW)` on MSYS2 CLANG64, the command is still being executed. I'm not an expert in this so I can't debug to know why.

MatusGuy: Oh well. It's not a big deal anyway

MatusGuy (approved):

PR #2823, opened 2024-03-10
ST OST - Yet more songs for the Icy Island

Curly68:
I think these are the remaining songs for the Icy Island world campaign, along with an extra bonus level theme included. I also fixed a mistake i did when it came to putting the new version of the midday theme in the wrong folder.

Added Songs:
- glacier.ogg (For the Between Two Glaciers secret level)
- snowfort.ogg (For the snowfort and outpost levels)
- icecastle.ogg (for the Icy Island castle level)
- yetiboss.ogg (For the Icy Island boss)

icecastle and yetiboss are based on the old fortress and boss battle theme made by Wantsti. Like prior contributions these songs are released under CC-BY-SA 4.0

Extra stuff
- Fixed the new midday theme version being in the wrong folder.


Rusty-Box:
I recommand giving the Between Two Glacier track a more simple file name. Something like "glacier.ogg" and "glacier.music"

serano01:
What will happen to the old songs? Are they going to be deleted?

Frostwithasideofsalt:
they will be removed post 0.7, i believe

Vankata453:
> they will be removed post 0.7, i believe 

Why not directly for 0.7.0?

Frostwithasideofsalt:
it just needs to be decided which ones should stay, as i think some of the old songs should really just not be removed. like salcon, and the airship theme(s), and jewels, and some others, as their all fairly good.

Rusty-Box:
We could relocate them to a special folder though

Vankata453:
> We could relocate them to a special folder though

If they stay in the OST, why should they be moved? Also would serve for unnecessary problems with compatibility.

Rusty-Box:
Who says they are just cuz they are in the game files. At best they are unused tracks. They will not appear in any official level going forward

Vankata453:
Can't we have OST tracks even if they aren't used in Story Mode?

Rusty-Box:
Why keep that junk around if we have fantastic replacements instead? They only bring down the quality of consistency by just existing for the sake of it.

Vankata453:
But we were talking about tracks that would be kept and not replaced either... that's what I gathered from Frost's comment.

Rusty-Box:
Sure but still. Me personally would not consider them part of the OST still. The only reason they are still there is because of nostalgia and old favorites. Outside of that they have no role in the game, so no role in an actual OST.

Songs like Chipdisko for example would fit better as a "retro" song now, so moving it over to the retro folder makes more sense.  So the other keepers should go into their own folder too. Something like "old", "unused" to clearly show they are no longer representative of the games style, vision. They are but bonuses for those that can't let go.

I for one would even go so far and recommend removing them all together and add them to a music pack you can instal if need be rather than keep them in the core files in the first place.

Rusty-Box (changes requested):
`snowfort.music` and `icecastle.music` use the wrong file, playing midday and sunset instead :D

Also, the ice castle music should go into the `music/castle/` folder instead!

Curly68:
Okay, i fixed the issues you described.

Rusty-Box (approved):

VincentAraneo:
I have an idea for the old songs that were there before. Those songs like, for example, bossattack.ogg, let's move it to a new folder called *olds* and put the old songs in. So the songs don't get lost

PR #2826, opened 2024-03-13
Ignore reserved indentifier warnings

MatusGuy:
Because the __ function is used everywhere, and apparently this is a reserved identifier, we must add this warning flag.


mrkubax10:
Wouldn't it be better to rename the function in question?

MatusGuy:
No because then i dont think the plural translation would get recognized by the .pot generator program. Also, it's really not that big of a deal

mrkubax10 (approved):

PR #2827, opened 2024-03-13
Camera: Customizable peek multiplier

Vankata453:
Allows for customizing the fractional distance towards the camera peek position to move each frame, from "Options" -> "Video".

Additionally removes a peek cap, which was the reason peeking caused a slightly offset position on return, the offset being bigger the smaller the peek multiplier is.


mrkubax10 (approved):

weluvgoatz (approved):

MatusGuy (approved):
Untested

PR #2828, opened 2024-03-16
Fix water not being electrocuted by thunderstorms and deprecate most tiles in the `others` menu.

tylerandari13:
I added back the thunderstorm... thing... at the end of `tiles.strf` as it was removed some time ago.

I also deprecated a lot of tiles in the `others` tab because they're duplicates. They shouldn't be used and just having them in the others tab makes converted levels look fine even though they aren't.

For the tiles i didn't remove my reasons are:

- 1, 359: They're such classics. Make for nice test tiles. (we should add them into the misc tab or something)
- 2075, 2076, 2079,  2164, 2165: Special cases for tiles that need multiple tilemaps but used to use a single tilemap. Someone told me a reason they're not deprecated but I forgot. Would probably need discussion before deprecating.
- 3873-3877: Lightning water tiles. These are used to electrocute the water with thunder. Pretty simple. (The rain drop tiles aren't actually needed for rain particles, hence why i deprecated them. If someone can confirm otherwise i will undeprecate them.)
- 5658: This tile is missing from the `Forest Background` tab. Should probably ask Rusty where to put this as i couldn't find a good place
- 6539, 6540, 6545 Missing from the `Corrupted Forest` tab and missing from the autotile definitions leaving some corners looking a bit wonky. Would also need to ask Rusty where to put them. (Id need help adding them to the autotiles as i don't know what to do there.)

Any other tile was either a duplicate or shouldnt be used (like the end sequence tiles) except for an empty tile which was... well... emtpy.


MatusGuy:
finally... thanks a crap ton orange

mrkubax10 (approved):

data/images/tiles.strf
```diff
@@ -5117,7 +5145,7 @@
     (width 8)
     (height 3)
     (ids
-      3398 3399 3400 3401 3402 3403 3404 3405
+         0    0 3401    0    0    0    0    0
```
Vankata453, line 5148: Is 3401 intended to be 1 to the left here?

Those tiles should also be marked as 0 in `attributes`.

data/images/tiles.strf
```diff
@@ -6893,6 +6941,17 @@
     )
     (image "tiles/ghostwood/ghostwood-12.png")
   )
+  (tiles
+    (width 4)(height 1)
+    (deprecated #t)
+    (ids
+      6295 0    0    6298
```
Vankata453, line 6948: The `0`s should be to the right (consistency).

data/images/tiles.strf
```diff
@@ -7169,6 +7228,17 @@
     )
     (image "tiles/ghostwood/underground/ghostwood-12_under.png")
   )
+  (tiles
+    (width 4)(height 1)
+    (deprecated #t)
+    (ids
+      6753 0    0    6756
```
Vankata453, line 7235: Same here.

Vankata453 (approved):

PR #2831, opened 2024-03-17
Particles for badguys

weluvgoatz:
Part of what turns a meh game into a great game is having visual feedback whenever you engage in a special event.  One way in which this can be incorporated into SuperTux is by adding some particles upon killing or squishing enemies.  I added exactly that in this PR, alongside some particles for when significant impacts are made (like when an enemy hits a wall at a fast speed.)
     Please test this PR, it is my first contribution in almost a year and I made it rather quickly, and surely there will be some design decisions to be made, and I will definitely be willing to work with that.  But I think this is a good start.
     The graphics added in this PR are created by RustyBox.


tobbi:
Video would be nice.

weluvgoatz:
Here is a video:
https://www.youtube.com/watch?v=y0qKR7CU4Rg

tobbi:
IDK if I like this. The way the particles spawn is too generic. Could we have different effects per badguy?

mrkubax10:
I like the idea but I would leave out collision particles because it may get laggy very quickly.

Rusty-Box:
Some suggestions after testing:
- I think particles should slowly begin fading out after a couple seconds. They stay too long, especially noticable with leaf particles.
- The "dust-type" / neutral paritcle should also have lower gravity. I wanna update their sprite to actually look more dust like. Right now they look too much like a rock
- Snowballs etc currently have two instances of particle effect play at once. You can at least hear the melt sound twice!

Side note, I'm gonna make some more particles sprites soon that you can use. I was planning for sprite used for: Granito, Flying Snowball, Stalactite. Just to let you know

tobbi:
I think particles should spawn from the center of the squished sprite and not from the center of the normal sprite.

Alasdairbugs:
Revisiting this, I think its a good idea to add particles but it'll be needed for every new enemy in the future. is this something we're willing to add onto the already large workload of making an enemy: drawing, animating, designing, and coding them in? I think it'll be better to worry about this closer to a 1.0 release when most, if not all, enemies are in the game.

weluvgoatz:
I was also thinking about this yesterday, I think I might limit the scope for now - just particles for Tux (buttjump), explosions, and crushers (all using the generic particles already here). Then we can focus on more robust particles for later.

PR #2833, opened 2024-03-17
Check for all things with cppcheck

tobbi:
<matusguy> I said do it on a PR, idiot.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -52,5 +52,5 @@ jobs:
         run: |
           ./danmar-cppcheck-*/cppcheck \
             --quiet --inline-suppr --language=c++ --error-exitcode=10 \
-            -j"$(nproc)" --std=c++17 --enable=warning,style \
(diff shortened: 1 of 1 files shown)


tobbi:
I don't know how to add the src directory as an include base for the cppcheck thing.

tobbi:
Well, this works, but...

tobbi:
I don't know if we want all the reported errors in there.

tobbi:
https://pastebin.com/2a3SHYE9

tobbi:
I feel like this adds way too many warnings. Maybe we should cut down the number of things it checks?

HybridDog:
> I don't know how to add the src directory as an include base for the cppcheck thing.

It is possible to run cmake with -DCMAKE_EXPORT_COMPILE_COMMANDS=1 and pass the compile commands with --project=compile_commands.json to cppcheck. I don't know if using cppcheck with this is still fast.

PR #2834, opened 2024-03-17
Small cosmetic wind adjustments

weluvgoatz:
This PR makes the wind particles point in the proper direction now, as well as putting wind on a layer behind others so that you can actually grab other objects in the editor when there's wind in the level.

### DIFF
--- a/src/object/sprite_particle.hpp
+++ b/src/object/sprite_particle.hpp
@@ -39,6 +39,8 @@ public:
                  int drawing_layer = LAYER_OBJECTS-1, bool notimeout = false, Color color = Color::WHITE);
   ~SpriteParticle() override;
 
+  SpritePtr& get_sprite() { return sprite; }
+
 protected:
   virtual void update(float dt_sec) override;
   virtual void draw(DrawingContext& context) override;
--- a/src/object/sprite_particle.hpp
+++ b/src/object/sprite_particle.hpp
@@ -39,7 +39,7 @@ public:
                  int drawing_layer = LAYER_OBJECTS-1, bool notimeout = false, Color color = Color::WHITE);
   ~SpriteParticle() override;
 
-  SpritePtr& get_sprite() { return sprite; }
+  Sprite& get_sprite() { return *sprite; }
 
 protected:
   virtual void update(float dt_sec) override;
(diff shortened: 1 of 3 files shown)


src/object/sprite_particle.hpp
```diff
@@ -39,6 +39,8 @@ class SpriteParticle final : public GameObject
                  int drawing_layer = LAYER_OBJECTS-1, bool notimeout = false, Color color = Color::WHITE);
   ~SpriteParticle() override;
 
+  SpritePtr& get_sprite() { return sprite; }
```
MatusGuy, line 42: Why is this a SpritePtr reference?

weluvgoatz: Why not?

tobbi: You can make this const if you return `SpritePtr` (which is `std::unique_ptr<Sprite>`) instead of `SpritePtr&`. It doesn't change anything.

mrkubax10: Maybe `Sprite& get_sprite() { return *sprite; }` would be better?

weluvgoatz: > You can make this const if you return `SpritePtr` (which is `std::unique_ptr<Sprite>`) instead of `SpritePtr&`. It doesn't change anything.

Doing it this way creates many errors.

tobbi: "creates many errors"...WHAT errors?

Vankata453: It's a `unique_ptr`, so copy issues I'd assume.

weluvgoatz: Many copies of this:
"error C2280: 'std::unique_ptr<Sprite,std::default_delete<Sprite>>::unique_ptr(const std::unique_ptr<Sprite,std::default_delete<Sprite>> &)': attempting to reference a deleted function"

tobbi: Or maybe return a `SpritePtr`instance by using std::unique_ptr::get()

mrkubax10 (approved):

tobbi:
This will fail if sprite ever is NULL (null dereference).

tobbi:
@mrkubax10 Were you so certain that sprite could never be null that you deref'ed it without checking?

mrkubax10:
If you return a pointer instead it will fail in other place.

tobbi:
Not when there's suitable null checks.

weluvgoatz:
Closing as there doesn't seem to be any desire to have this.

PR #2795, opened 2024-02-21
Improve Tux' sliding behavior

Brockengespenst:
* Do smooth transition on slope angle change
* Continue to use slidejump animation when releasing jump button during slidejump
* Fix animation glitches in transition from crawling to sliding
* Add possibility to pause/resume animations and apply this for crawling and climbing when not moving

Fixes #2785
Fixes #2723

PR #2810, opened 2024-02-29
Fix some Tux animation glitches

Brockengespenst:
* Wait until the growing animation is finished before changing to stone animation when smallTux gains an earth flower.
* Fix animation glitch from sliding to slidegrow
* Remain in crawling action when down is held and Tux stops sliding due to friction

Fixes #2533

PR #2838, opened 2024-03-22
Fix sliding player clipping through floor

weluvgoatz:
For whatever reason that is beyond me, the PR #2795 appears to have intentionally introduced a bug where Tux would clip through the floor if he stopped sliding but was still moving (meaning, the player stopped pressing the down key).  In areas where Tux ended sliding in a 1-block hole, he would clip into the floor.  This clipping was also noticeable under normally-tall terrain gaps, but it resolves itself immediately.  However, this PR fixes it.  Let me know if there are any issues or side effects with it.

### DIFF
--- a/src/object/player.cpp
+++ b/src/object/player.cpp
@@ -793,13 +793,8 @@ Player::update(float dt_sec)
 
     //if you stop holding down when sliding, then it stops.
     //or, stop sliding if you come to a stop and are not on a slope.
-    if (!m_controller->hold(Control::DOWN))
-    {
-      m_sliding = false;
-      m_was_crawling_before_slide = false;
-      m_slidejumping = false;
-    }
-    else if (m_floor_normal.y == 0.f && std::abs(m_physic.get_velocity_x()) <= 1.f)
+
+    if (!m_controller->hold(Control::DOWN) || (m_floor_normal.y == 0.f && std::abs(m_physic.get_velocity_x()) <= 1.f))
     {
       if (is_big())
       {
--- a/src/object/player.cpp
+++ b/src/object/player.cpp
@@ -793,8 +793,8 @@ Player::update(float dt_sec)
 
     //if you stop holding down when sliding, then it stops.
     //or, stop sliding if you come to a stop and are not on a slope.
-
-    if (!m_controller->hold(Control::DOWN) || (m_floor_normal.y == 0.f && std::abs(m_physic.get_velocity_x()) <= 1.f))
+    if (!m_controller->hold(Control::DOWN) ||
+      (m_floor_normal.y == 0.f && std::abs(m_physic.get_velocity_x()) <= 1.f))
     {
       if (is_big())
       {


src/object/player.cpp
```diff
@@ -793,13 +793,8 @@ Player::update(float dt_sec)
...
     //or, stop sliding if you come to a stop and are not on a slope.
-    if (!m_controller->hold(Control::DOWN))
-    {
-      m_sliding = false;
-      m_was_crawling_before_slide = false;
-      m_slidejumping = false;
-    }
-    else if (m_floor_normal.y == 0.f && std::abs(m_physic.get_velocity_x()) <= 1.f)
+
+    if (!m_controller->hold(Control::DOWN) || (m_floor_normal.y == 0.f && std::abs(m_physic.get_velocity_x()) <= 1.f))
```
tobbi, line 797: Can you maybe add line breaks so the max length of 80 characters is followed?

Brockengespenst: It was definitely not my intention to introduce bugs, if that happened, I am really sorry.
The code was introduced for the following scenario:
* Tux is big
* Player presses and holds down key for the rest of the described sequence -> Tux is crawling
* Player crawls towards a slope -> Tux starts to slide on the slope
* Slope ends and Tux comes to a stop -> Tux remains in crawling animation
The original behavior - that your PR now restores - is that Tux sets  the "duck" action. If you then press left or right, the action will switch from duck to crawling animation which looks a bit weird. In my opinion remaining in the crawl action looks much better than switching to duck action.
So m_was_crawling_before_slide should in my opinion not be set to false in case player stopped sliding due to friction.

Brockengespenst:
It was definitely not my intention to introduce bugs, if that happened, I am really sorry.
The code was introduced for the following scenario:
* Tux is big
* Player presses and holds down key for the rest of the described sequence -> Tux is crawling
* Player crawls towards a slope -> Tux starts to slide on the slope
* Slope ends and Tux comes to a stop -> Tux remains in crawling animation

The original behavior - that your PR now restores - is that Tux sets  the "duck" action. If you then press left or right, the action will switch from duck to crawling animation which looks a bit weird. In my opinion remaining in the crawl action looks much better than switching to duck action.
So m_was_crawling_before_slide should in my opinion not be set to false in case player stopped sliding due to friction.

Brockengespenst:
At least that was my intention that Tux does not go to duck when it stops after sliding, but it seems that it is not working like that (I was pretty sure I tested this several times), so at this part your PR does not seem to change the behavior.
Anyway: The flag is used to skip the "duck" part of the sliding animation when Tux was crawling before (otherwise Tux would crawl, standup and then transit to slide again in the above described scenario [transition from crawl to slide]).

Brockengespenst:
I remember now, why I was so sure that the described behavior is working: Please check out also my changes in #2810 . With these changes, the described behavior should work. Looks like you're in a flow right now, so maybe you could take all these information into account and do the ultimate fix. :-)

weluvgoatz:
> I remember now, why I was so sure that the described behavior is working: Please check out also my changes in #2810 . With these changes, the described behavior should work. Looks like you're in a flow right now, so maybe you could take all these information into account and do the ultimate fix. :-)

Can you be a little more verbose regarding this message, please? I'm afraid I don't understand what you mean.  And also sorry for coming across as mean in the original PR message.

Brockengespenst:
Sure, sorry for the confusion.

I thought that Tux would already remain in the crawl action when he was sliding before and comes to a stop while the down key is still pressed (but without left or right being pressed). I forgot that I added/fixed this in #2810 (among the other things that are mentionened in the PR description), so I thought your PR here would break this behavior. Maybe you want to have a look at #2810 and check if it either make things better or maybe you have a better idea how to improve it otherwise.

Also I mixed this discussion here a bit up with the behaviour regrading m_was_crawling_before_slide, which is supposed to skip the first frames of the slide action when Tux is crawling and directly goes into the slide action. The first frames of the slide actions show Tux changing from standing to sliding, but this leads to animation glitches if Tux is already crawling.

I hope this makes it a bit clearer. Please let me know if it is still too confusing. :-)

weluvgoatz:
I just tested both PRs together and they seem to work pretty independently of each other and both of them together seem to fix the bugs they set out to, so I think both are fine right now.

Brockengespenst:
Perfect, thanks!

Frostwithasideofsalt (approved):

mrkubax10 (approved):

PR #2839, opened 2024-03-23
Fix branch name display on download.supertux.org

tobbi:
Let's see if this works.

### DIFF
--- a/.ci_scripts/deploy.sh
+++ b/.ci_scripts/deploy.sh
@@ -13,6 +13,7 @@ for file in upload/SuperTux*; do
         shasum=$(shasum -a 256 "$file" | cut -d " " -f 1)
     fi
     echo "Checksum: $shasum";
+    echo "Branch: $BRANCH_NAME";
     curl --data "apikey=$DOWNLOAD_APIKEY" \
          --data "url=$url" \
          --data "size=$size" \
--- a/.ci_scripts/deploy.sh
+++ b/.ci_scripts/deploy.sh
@@ -16,7 +16,7 @@ for file in upload/SuperTux*; do
     curl --data "apikey=$DOWNLOAD_APIKEY" \
          --data "url=$url" \
          --data "size=$size" \
-         --data "branch=$(git branch --show-current)" \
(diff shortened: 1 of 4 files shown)


tobbi:
Okay. This worked! 👍

tobbi:
The echo was only because I suspected the code wasn't run due to whatever thing. But now that it works, let's just keep the echo there. I mean, it might make the CI builds run 0.0000000000000001% slower, but that's a risk I'm willing to take:
https://stackoverflow.com/questions/72564680/am-i-correct-in-assuming-that-echo-and-cat-are-too-fast-for-the-time-comma

Zwatotem (approved):
I have no idea, what this fixes. Is it just to debug something else?
But whatever it is – sure, go ahead!

mrkubax10 (approved):

Vankata453:
In my opinion it would be better if we don't show non-`master` builds at all - I would imagine the average player wouldn't care what branch they are downloading a Nightly from.

tobbi:
> I have no idea, what this fixes. Is it just to debug something else? But whatever it is – sure, go ahead!

Fixes the empty branch cells in https://download.supertux.org/

tobbi:
I feel like having the ability to download feature branches to test them quickly could be beneficial for testers.

MatusGuy:
They can do this through github actions pages. Not through download.supertux.org

tobbi:
download.supertux.org contains untested changes anyway. It's not like commits on master could be considered "stable" enough to be tested by end users.

tobbi:
What would be great is if we could have a branch selector on download.supertux.org that defaults to "master".

MatusGuy:
You're missing the point. Commits on pull requests are much much less stable than master commits. Most master commits consist of merging pull requests that have been reviewed and tested

MatusGuy:
Or we can have a branch selector, that also works, do you want to implement that?

tobbi:
I don't have access to the server hosting download.supertux.org, only @maths22 could do that.

weluvgoatz (approved):

PR #2844, opened 2024-03-24
Door and teleport improvements

### LINKED ISSUE #2409: Proper Teleport Transition
What if instead of simply fading to black (using doors) or no transition at all (using `Level.spawn();`) we could use transition effect closer to entering a level, the circular one.

Now for doors it should become the new default and only play if the door is actually being entered (currently the transition effect happens even if I run away from the door in the process), same for scripted level teleports. However, for scripting the transition-less teleport should remain an option.

### LINKED ISSUE #2649: Invincibility frames after using a door/teleport
It is always quite annoying and unfair if you enter a door but an enemy is right in front of the exit door, getting Tux hurt. You could build a level to prevent that from ever happening, however, i see this as more trouble some and believe an even better solution would be to grant Tux a short moment of "invincibility" so that players have time to move away from an enemy should they be right in front of the door.

Brockengespenst:
Provides the following changes to door and teleport transitions:
- ShrinkFade is now used as fade effect instead of FadeToBlack
- Player controls are disabled while fading out to prevent Tux from walking away the door while it opens
- Tux walks automatically in front of door if up key is pressed and Tux is not directly in front of the door
- Tux cannot be hurt by enemies while controls are disabled
- Added some time of invincibility when spawning at the target door
- Added scripting support for Level.spawn_transition

Default fade effect when entering a GameSession is now the ShrinkFade effect.

Fixes #2409
Fixes #2649

### DIFF
--- a/src/object/player.cpp
+++ b/src/object/player.cpp
@@ -203,6 +203,7 @@ Player::Player(PlayerStatus& player_status, const std::string& name_, int player
   m_invincible_timer(),
   m_skidding_timer(),
   m_safe_timer(),
+  m_is_intentionally_safe(false),
   m_kick_timer(),
   m_buttjump_timer(),
   m_dying_timer(),
@@ -2205,7 +2206,7 @@ Player::draw(DrawingContext& context)
   */
 
   /* Draw Tux */
-  if (!m_visible || (m_safe_timer.started() && m_safe_due_to_hurt && size_t(g_game_time * 40) % 2))
+  if (!m_visible || (m_safe_timer.started() && !m_is_intentionally_safe && size_t(g_game_time * 40) % 2))
   {
   }  // don't draw Tux
 
@@ -2378,7 +2379,7 @@ void
 Player::make_temporarily_safe(const float safe_time)
 {
   m_safe_timer.start(safe_time);
-  m_safe_due_to_hurt = false;
+  m_is_intentionally_safe = true;
 }
 
 void
@@ -2408,11 +2409,11 @@ Player::kill(bool completely)
       || get_bonus() == AIR_BONUS
       || get_bonus() == EARTH_BONUS) {
       m_safe_timer.start(TUX_SAFE_TIME);
-      m_safe_due_to_hurt = true;
+      m_is_intentionally_safe = false;
       set_bonus(GROWUP_BONUS, true);
     } else if (get_bonus() == GROWUP_BONUS) {
       m_safe_timer.start(TUX_SAFE_TIME /* + GROWING_TIME */);
-      m_safe_due_to_hurt = true;
+      m_is_intentionally_safe = false;
       m_duck = false;
       stop_backflipping();
       set_bonus(NO_BONUS, true);
@@ -2511,7 +2512,9 @@ Player::check_bounds()
   }
 
   /* fallen out of the level? */
-  if ((get_pos().y > Sector::get().get_height()) && (!m_ghost_mode)) {
+  if ((get_pos().y > Sector::get().get_height())
+      && (!m_ghost_mode)
+      && !(m_is_intentionally_safe && m_safe_timer.started())) {
     kill(true);
     return;
   }
--- a/src/object/player.cpp
+++ b/src/object/player.cpp
@@ -2375,9 +2375,9 @@ Player::make_invincible()
 }
 
 void
-Player::make_temporarily_safe()
+Player::make_temporarily_safe(const float safe_time)
 {
-  m_safe_timer.start(TUX_SAFE_TIME);
+  m_safe_timer.start(safe_time);
   m_safe_due_to_hurt = false;
 }
 
--- a/src/object/player.cpp
+++ b/src/object/player.cpp
@@ -2205,7 +2205,7 @@ Player::draw(DrawingContext& context)
   */
 
   /* Draw Tux */
-  if (!m_visible || (m_safe_timer.started() && size_t(g_game_time * 40) % 2))
+  if (!m_visible || (m_safe_timer.started() && m_safe_due_to_hurt && size_t(g_game_time * 40) % 2))
   {
   }  // don't draw Tux
 
@@ -2375,6 +2375,13 @@ Player::make_invincible()
 }
 
 void
+Player::make_temporarily_safe()
+{
+  m_safe_timer.start(TUX_SAFE_TIME);
+  m_safe_due_to_hurt = false;
+}
+
+void
 Player::kill(bool completely)
 {
   if (m_dying || m_deactivated || is_winning() )
@@ -2401,9 +2408,11 @@ Player::kill(bool completely)
       || get_bonus() == AIR_BONUS
       || get_bonus() == EARTH_BONUS) {
       m_safe_timer.start(TUX_SAFE_TIME);
+      m_safe_due_to_hurt = true;
       set_bonus(GROWUP_BONUS, true);
     } else if (get_bonus() == GROWUP_BONUS) {
       m_safe_timer.start(TUX_SAFE_TIME /* + GROWING_TIME */);
+      m_safe_due_to_hurt = true;
       m_duck = false;
       stop_backflipping();
       set_bonus(NO_BONUS, true);
--- a/src/object/player.cpp
+++ b/src/object/player.cpp
@@ -2376,7 +2376,7 @@ Player::make_invincible()
 }
 
 void
-Player::make_temporarily_safe(const float safe_time)
+Player::make_temporarily_safe(float safe_time)
 {
   m_safe_timer.start(safe_time);
   m_is_intentionally_safe = true;
@@ -2513,7 +2513,7 @@ Player::check_bounds()
 
   /* fallen out of the level? */
   if ((get_pos().y > Sector::get().get_height())
-      && (!m_ghost_mode)
+      && !m_ghost_mode
       && !(m_is_intentionally_safe && m_safe_timer.started())) {
     kill(true);
     return;
--- a/src/scripting/level.cpp
+++ b/src/scripting/level.cpp
@@ -57,7 +57,7 @@ Level_spawn_transition(const std::string& sector, const std::string& spawnpoint,
   else if (transition == "circle")
     fade_type = ScreenFade::FadeType::CIRCLE;
 
-  game_session.respawn_with_fade(sector, spawnpoint, fade_type, {0.0f, 0.0f});
+  game_session.respawn_with_fade(sector, spawnpoint, fade_type, {0.0f, 0.0f}, true);
 }
 
 void
--- a/src/scripting/level.cpp
+++ b/src/scripting/level.cpp
@@ -16,12 +16,10 @@
 //  along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 #include "scripting/level.hpp"
-
 #include "supertux/d_scope.hpp"
 #include "supertux/flip_level_transformer.hpp"
 #include "supertux/game_session.hpp"
 #include "supertux/sector.hpp"
-
 #include "util/log.hpp"
 
 namespace scripting {
--- a/src/scripting/level.cpp
+++ b/src/scripting/level.cpp
@@ -22,6 +22,8 @@
 #include "supertux/game_session.hpp"
 #include "supertux/sector.hpp"
 
+#include "util/log.hpp"
+
 namespace scripting {
 
 void
@@ -56,6 +58,8 @@ Level_spawn_transition(const std::string& sector, const std::string& spawnpoint,
     fade_type = ScreenFade::FadeType::FADE;
   else if (transition == "circle")
     fade_type = ScreenFade::FadeType::CIRCLE;
+  else
+    log_warning << "Invalid transition type '" << transition << "'." << std::endl;
 
   game_session.respawn_with_fade(sector, spawnpoint, fade_type, {0.0f, 0.0f}, true);
 }
--- a/src/scripting/level.cpp
+++ b/src/scripting/level.cpp
@@ -46,6 +46,21 @@ Level_spawn(const std::string& sector, const std::string& spawnpoint)
 }
 
 void
+Level_spawn_transition(const std::string& sector, const std::string& spawnpoint, const std::string& transition)
+{
+  SCRIPT_GUARD_GAMESESSION();
+
+  GameSession::FadeType fade_type = GameSession::FadeType::NONE;
+
+  if (transition == "fade")
+    fade_type = GameSession::FadeType::FADE;
+  else if (transition == "circle")
+    fade_type = GameSession::FadeType::CIRCLE;
+
+  game_session.respawn_with_fade(sector, spawnpoint, fade_type, {0.0f, 0.0f});
+}
+
+void
 Level_set_start_point(const std::string& sector, const std::string& spawnpoint)
 {
   SCRIPT_GUARD_GAMESESSION();
--- a/src/scripting/level.cpp
+++ b/src/scripting/level.cpp
@@ -16,6 +16,7 @@
 //  along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 #include "scripting/level.hpp"
+
 #include "supertux/d_scope.hpp"
 #include "supertux/flip_level_transformer.hpp"
 #include "supertux/game_session.hpp"
--- a/src/scripting/level.cpp
+++ b/src/scripting/level.cpp
@@ -50,12 +50,12 @@ Level_spawn_transition(const std::string& sector, const std::string& spawnpoint,
 {
   SCRIPT_GUARD_GAMESESSION();
 
-  GameSession::FadeType fade_type = GameSession::FadeType::NONE;
+  ScreenFade::FadeType fade_type = ScreenFade::FadeType::NONE;
 
   if (transition == "fade")
-    fade_type = GameSession::FadeType::FADE;
+    fade_type = ScreenFade::FadeType::FADE;
   else if (transition == "circle")
-    fade_type = GameSession::FadeType::CIRCLE;
+    fade_type = ScreenFade::FadeType::CIRCLE;
 
   game_session.respawn_with_fade(sector, spawnpoint, fade_type, {0.0f, 0.0f});
 }
--- a/src/scripting/level.hpp
+++ b/src/scripting/level.hpp
@@ -66,6 +66,19 @@ bool Level_has_active_sequence();
  */
 void Level_spawn(const std::string& sector, const std::string& spawnpoint);
 
+
+/**
+ * Respawns Tux in sector named ""sector"" at spawnpoint named ""spawnpoint"" with the given transition ""transition"".${SRG_TABLENEWPARAGRAPH}
+   Exceptions: If ""sector"" or ""spawnpoint"" are empty, or the specified sector does not exist, the function will bail out the first chance it gets.
+   If the specified spawnpoint doesn't exist, Tux will be spawned at the spawnpoint named “main”.
+   If that spawnpoint doesn't exist either, Tux will simply end up at the origin (top-left 0, 0).
+ * @param string $sector
+ * @param string $spawnpoint
+ * @param string $transition
+ */
+void Level_spawn_transition(const std::string& sector, const std::string& spawnpoint, const std::string& transition);
+
+
 /**
  * Sets the default start spawnpoint of the level.
  * @param string $sector
--- a/src/scripting/level.hpp
+++ b/src/scripting/level.hpp
@@ -74,7 +74,7 @@ void Level_spawn(const std::string& sector, const std::string& spawnpoint);
    If that spawnpoint doesn't exist either, Tux will simply end up at the origin (top-left 0, 0).
  * @param string $sector
  * @param string $spawnpoint
- * @param string $transition
+ * @param string $transition Valid transitions are ""circle"" and ""fade"". If any other value is specified, no transition effect is drawn.
  */
 void Level_spawn_transition(const std::string& sector, const std::string& spawnpoint, const std::string& transition);
 
--- a/src/supertux/game_session.cpp
+++ b/src/supertux/game_session.cpp
@@ -49,6 +49,7 @@
 #include "worldmap/worldmap.hpp"
 
 static const float FADE_TIME = 1.0f;
+static const float SAFE_TIME = 1.0f;
 
 GameSession::GameSession(const std::string& levelfile_, Savegame& savegame, Statistics* statistics,
                          bool preserve_music) :
@@ -539,7 +540,7 @@ GameSession::update(float dt_sec, const Controller& controller)
       for (auto* p : m_currentsector->get_players())
       {
         // Make all players temporarily safe after spawning
-        p->make_temporarily_safe();
+        p->make_temporarily_safe(SAFE_TIME);
       }
     }
 
--- a/src/supertux/game_session.cpp
+++ b/src/supertux/game_session.cpp
@@ -50,6 +50,8 @@
 
 static const float SAFE_TIME = 1.0f;
 static const int SHRINKFADE_LAYER = LAYER_LIGHTMAP - 1;
+static const float TELEPORT_FADE_TIME = 1.0f;
+
 
 GameSession::GameSession(const std::string& levelfile_, Savegame& savegame, Statistics* statistics,
                          bool preserve_music) :
--- a/src/supertux/game_session.cpp
+++ b/src/supertux/game_session.cpp
@@ -67,7 +67,7 @@ GameSession::GameSession(const std::string& levelfile_, Savegame& savegame, Stat
   m_activated_checkpoint(),
   m_newsector(),
   m_newspawnpoint(),
-  m_spawn_fade_type(FadeType::NONE),
+  m_spawn_fade_type(ScreenFade::FadeType::NONE),
   m_spawn_fade_point(0.0f, 0.0f),
   m_spawn_fade_timer(),
   m_spawn_with_invincibilty(false),
@@ -496,7 +496,7 @@ GameSession::update(float dt_sec, const Controller& controller)
   check_end_conditions();
 
   // Respawning in new sector?
-  if (!m_newsector.empty() && !m_newspawnpoint.empty() && (m_spawn_fade_timer.check() || m_spawn_fade_type == FadeType::NONE)) {
+  if (!m_newsector.empty() && !m_newspawnpoint.empty() && (m_spawn_fade_timer.check() || m_spawn_fade_type == ScreenFade::FadeType::NONE)) {
     auto sector = m_level->get_sector(m_newsector);
     std::string current_music = m_currentsector->get_singleton_by_type<MusicObject>().get_music();
     if (sector == nullptr) {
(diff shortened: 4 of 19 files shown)


Brockengespenst:
Right now the same approach as the safe timer after getting hurt is used when doing a teleport using doors (except for Tux doing the "blinking"). The same amount of time (1.8 sec) is used - which might be a bit long.
The scripting part is preliminary, just for testing. It adds a command `Level.spawn_transition("[sector]", "[spawnpoint]", "[transition_type]");` with transition type  `"circle"` (for shrink fade), `"fade"` (fade to black) or any other string will be interpreted as none, which means no fade effect at all.

Brockengespenst:
For local multiplayer a proper strategy might be needed to determine the focus for the shrink fade when the script is used. Right now the first alive player will be chosen. This should work for single player, but for local multiplayer another approach might be better. Suggestions are welcome.

src/supertux/game_session.hpp
```diff
@@ -88,6 +95,11 @@ class GameSession final : public Screen,
   /** ends the current level */
   void finish(bool win = true);
   void respawn(const std::string& sectorname, const std::string& spawnpointname);
+  void respawn_with_fade(const std::string& sectorname,
+                         const std::string& spawnpointname,
+                         const FadeType fade_type,
+                         const Vector fade_point,
```
tobbi, line 101: I wonder if passing by reference here is preferable.

src/supertux/game_session.hpp
```diff
@@ -88,6 +95,11 @@ class GameSession final : public Screen,
...
+                         const bool make_invincible = false);
```
tobbi, line 102: Does `const bool` mean that make_invincible is always false?

tobbi:
Sorry, I thought this was already ready for review, thus adding my comments. You can ignore them if you like. No pressure.

src/supertux/game_session.hpp
Vankata453: I think so.

src/supertux/game_session.hpp
Vankata453: I don't think those two relate. It's just that the parameter can't be modified.

src/supertux/game_session.hpp
```diff
@@ -76,6 +76,13 @@ class GameSession final : public Screen,
   };
 
 public:
+  enum class FadeType
+  {
+    NONE,
+    FADE,
+    CIRCLE
+  };
```
Vankata453, line 84: Maybe these enumerators would be better suited in the `ScreenFade` class ("src/supertux/screen_fade.hpp)?

src/supertux/game_session.cpp
```diff
@@ -475,6 +515,35 @@ GameSession::update(float dt_sec, const Controller& controller)
...
+    case FadeType::CIRCLE:
+    {
+      const Vector spawn_point_position = sector->get_spawn_point_position(m_newspawnpoint);
+      const Vector shrinkpos = spawn_point_position - sector->get_camera().get_translation();
+
+      ScreenManager::current()->set_screen_fade(std::make_unique<ShrinkFade>(shrinkpos, FADE_TIME, ShrinkFade::FADEIN));
+      break;
+    }
+    case FadeType::NONE:
+    default:
```
Vankata453, line 534: `default:` would take care of `NONE` either way.

src/supertux/game_session.cpp
```diff
@@ -475,6 +515,35 @@ GameSession::update(float dt_sec, const Controller& controller)
...
+      break;
```
Vankata453, line 535: Indent the `switch` statement cases two more spaces in.

src/supertux/game_session.cpp
```diff
@@ -282,6 +298,23 @@ GameSession::on_escape_press(bool force_quick_respawn)
...
+  {
+    if (!player->is_dead() && !player->is_dying())
+    {
+      return player->get_bbox().get_middle();
+    }
+  }
+
+  return Vector(0.0f, 0.f);
+}
+
```
Vankata453, line 317: Perhaps the fade could be done where the door is, instead of where the first alive player is?

For the script, a parameter as to where to do the fade at could work.

src/supertux/game_session.hpp
Brockengespenst: Good idea.

src/supertux/game_session.hpp
Brockengespenst: Indeed, it just says that the parameter cannot be altered inside the function body.

src/supertux/game_session.hpp
Brockengespenst: Yes, I will move it there.

src/supertux/game_session.cpp
Brockengespenst: This function is only used when the point cannot be determined automatically. If there is a door, that triggers the respawn, its coordinates are used (see door passing its center to the respawn_with_fade function).
This functions is mainly for the scripting part where the trigger cannot be determined (or at least I did not figure out how to get it in case of the "Level.*" commands or if invoked from console.

Vankata453: For the scripting function, screen-relative coordinates could be required.

Rusty-Box:
This is lovely! Though I do think Tux should not be able to receive any further inputs during the teleport transition + door transition (kinda like when using the scrippted controller script). Imo. it does feel kinda weird being able to keep on going during it.

src/supertux/shrinkfade.cpp
```diff
@@ -41,7 +42,7 @@ void
 ShrinkFade::draw(DrawingContext& context)
 {
   float progress = m_accum_time / m_fade_time;
-  float diameter = 2 * m_initial_size * (1.0f - progress);
+  float diameter = 2 * m_initial_size * (m_direction == FADEOUT ? (1.0f - progress) : progress);
   context.color().draw_inverse_ellipse(m_dest, Vector(1.1f * diameter, diameter),
                                          Color(0, 0, 0), LAYER_GUI + 1);
```
MatusGuy, line 47: Don't forget to change the layer to something lower than LAYER_GUI or else the effect goes above the pause menu gui. I recommend LAYER_GUI-5

Rusty-Box:
Can we also have teleportation transition make you immune to death, like falling offscreen? For example if you want Tux to jump in a hole in the ground at the bottom of the screen, teleport, without having him die.

src/supertux/shrinkfade.cpp
Brockengespenst: Thanks for the hint!

Brockengespenst:
> Though I do think Tux should not be able to receive any further inputs during the teleport transition + door transition (kinda like when using the scrippted controller script). Imo. it does feel kinda weird being able to keep on going during it.

I totally agree. First, I was playing around with setting the gamespeed to zero, but than then fade animation was stopped as well. I can check if blocking the controls works better. I think in that case it would also make sense to make Tux invincible already at the point when the controls are blocked, before the actual teleport is done.
Btw. right now Tux is invincible only when going through a door. Should this be applied to Level.spawn_transition scripting as well?

> Can we also have teleportation transition make you immune to death, like falling offscreen? For example if you want Tux to jump in a hole in the ground at the bottom of the screen, teleport, without having him die.

You mean before the actual teleport is done? I was not aware that this might be a use case, but I will check.

Rusty-Box:
> Btw. right now Tux is invincible only when going through a door. Should this be applied to Level.spawn_transition scripting as well?

I think so yeah. Also, the time Tux remains invincible after the door is indeed too long. maybe shorten it to like 1 is probably more appropriate.

> You mean before the actual teleport is done? I was not aware that this might be a use case, but I will check.

To be fair, it happens rather rarely to be used in this way but can be quite useful, I'd say. Basically once the script thus the transition is triggered Tux should not be affected by "death barriers" etc.

Brockengespenst:
I shortened the safe time to 1 sec now and disabled the controls for the player that uses the door (if in mulitplayer, the other players are not affected) until the fadout part of the effect is done. After actual teleport they are reactivated. Tux is also safe while controls are disabled.

Regarding the "falling off screen": When calling scripts like Level.spawn*, the information which player triggered the script is lost quite early in the chain, so it is not that easy to find out which player this should affect. If making all players invincible, I think it might lead to unwanted behavior in case of multiplayer.

I am wondering if it would make sense to introduce an own trigger type (similar to the door, but without being visible and all the animation stuff), because this would solve some problems: The player that should be invincible could be determined, as well as the point for the fade effect.
It would also be easier to make this configurable, e.g. 
* Target sector / spawnpoint
* Chosing the transition type from a selection menu
* Either checkbox if invincibility after respawn should be applied or provide option to specify the time of invincibility 
* Checkbox if Tux should be invincible until actual teleport (to allow him to fall off from screen while fade out effect is in progress)

What do you think?

Rusty-Box:
While the transition effect no longer blocks the menu text it still blocks the textbox of it.

Also, one question I have: is it possible to make the circle of the "circle transition" in the center of the screen when the Camera mode is set to manual? I noticed that during cutscenes where the player object is often offscreen, it looks weird having it anchored to the player. All these you a "manual" camera.

Rusty-Box:
> I am wondering if it would make sense to introduce an own trigger type (similar to the door, but without being visible and all the animation stuff), because this would solve some problems: The player that should be invincible could be determined, as well as the point for the fade effect. It would also be easier to make this configurable, e.g.

Honestly, I think I wouldn't mind that.

Brockengespenst:
> While the transition effect no longer blocks the menu text it still blocks the textbox of it.

Thanks, should be fixed now.

> Also, one question I have: is it possible to make the circle of the "circle transition" in the center of the screen when the Camera mode is set to manual? I noticed that during cutscenes where the player object is often offscreen, it looks weird having it anchored to the player. All these you a "manual" camera.

While in cutscenes the center of screen is now chosen. Looked indeed a bit weird.
For the multiplayer I changed the fade point to the middle of all alive players, this feels a bit more fair. But I am open to better ideas.

Open point would be the invinvibility for "falling" off screen. So when you say you do not mind to introduce a separate trigger to determine the player that actually started the script, shall all players be invincible in case of multiplayer?

Brockengespenst:
I agree that it would make sense to only make the player that triggered the script invincible, but that would require too much rework in the scripting part. Maybe this will become easier once the SimpleSquirrel PR from @Vankata453 is merged.
So I added it for all players right now, so you can fall off from screen right now without being killed when Level.spawn_transition is invoked.

I will have a look at the Squirrel Enumeration for the fade type parameter now.

tobbi:
Hey, @Brockengespenst We were wondering whether you'd be willing to join our SuperTux server (external/internal) one so we can coordinate / communicate better?

tobbi:
> You should mention that it is a discord server specifically

You're right. I changed it. Thought I'd written that.

Rusty-Box:
So it appears that the center transition doesn't seem to work quite with levels using the `Camera.set_sclae(x);` script. It look quite offset. Also, I still think the centered effect should apply when Camera Mode is "Manual".

This is a cutscene that uses a zoomed in Camera. As you you can see the effect is not fully centered:



---



This is one of the text scroller "levels". It uses only a manual Camera as I'm not sure how to really use the cutscene scripts here, I was hoping we could have that settled through the manual Camera, having the effect centered. Right now it still does focus Tux who's meant to be hidden in this scene.

Brockengespenst:
I will have a look a the scaling and the manual camera (seems I mixed this up with cut scenes, sorry for that).

Rusty-Box:
> [...] (seems I mixed this up with cut scenes, sorry for that).

It's okay. They usually go hand in hand. But not for those text scrollers. :D

Brockengespenst:
Camera scale should now be taken into account. Also for camera in manual mode, the middle of the screen should be used now.

Rusty-Box:
Very nice! One last thing regarding the door. I find it a bit annoying that I have to wait until the door fully opens before Tux "freezes" and the transition begins to play. In my opinion, once you press up, Tux should freeze and the transition play while the door is opening

Brockengespenst:
> In my opinion, once you press up, Tux should freeze and the transition play while the door is opening

Makes sense, I will change the it.

Rusty-Box:
Okay I said "last one" but, uh, the coin, enemy and secret counter during the pause screen, the timer, coin and powerup counter, as well as "Show Controller, Show Framerate, Show Player Position" overlay/text get covered by the transition effect. Sorry about that 😅

Brockengespenst:
I adjusted the layers so that the fade effect will not cover the player hud anymore. I noticed that the InfoBlock used a GUI layer which caused it not to be covered by the fade effect (see rooted forest, first level, just the one before the door). I would consider InfoBlocks as part of the game and not of the GUI, so I moved it into a lower layer. It seems that it does not have any side effects, but please check.

Brockengespenst:
I would have another small feature for this: If Tux is not standing close to the center of the door when up is pressed, I could make him automatically move closer to the door using the scripted controller. Code would be ready to push, if you appreciate this feature.

Rusty-Box:
> I would have another small feature for this: If Tux is not standing close to the center of the door when up is pressed, I could make him automatically move closer to the door using the scripted controller. Code would be ready to push, if you appreciate this feature.

Depending on how it turns out I would not be against this, honestly.

Brockengespenst:
Give it a try. If you do not like it, I will revert it.

Rusty-Box:
I really like that! This is cool. For once Tux looks like he actually is going through that door :D


> I am wondering if it would make sense to introduce an own trigger type (similar to the door, but without being visible and all the animation stuff), because this would solve some problems: The player that should be invincible could be determined, as well as the point for the fade effect. It would also be easier to make this configurable, e.g.


I assume, if this would be something that gets "greenlit" this would not be part of this PR? If not I would go ahead and approve this PR

Brockengespenst:
> I assume, if this would be something that gets "greenlit" this would not be part of this PR?

It could be done independently of this PR, if someone needs this feature.

Right now the last scripting parameter for Level.spawn_transition (determining the fade effect type) is still a string and not yet an enumeration. To be honest, I have not had a look how to teach Squirrel a new Enumeration, yet.

Rusty-Box:
> It could be done independently of this PR, if someone needs this feature.

Either way I probably should. That way we can merge this sooner rather than later. So yeah, best to just make that its own PR.

> Right now the last scripting parameter for Level.spawn_transition (determining the fade effect type) is still a string and not yet an enumeration. To be honest, I have not had a look how to teach Squirrel a new Enumeration, yet.

Does it need to be though? I think there's a couple of scripts that deal with strings for certain things, like the easing on the `Camera.ease_scale(x, t, "EaseMode");` script. Personally I think the way it is now is totally fine, but perhaps I am missing something?

Brockengespenst:
@Vankata453 suggested this. If he is fine with the current solution, I would be fine, too.

Vankata453:
> Does it need to be though?

Looking at it, an enumerator is not necessarily needed, the possible strings would be documented either way.

Rusty-Box:
I found one major bug. Since you use scripted controller, Tux drops whatever he is carrying while going into a door!

Brockengespenst:
Uh, thanks for the hint! Sorry, I will have a look.

Brockengespenst:
I removed the usage of the scripting controller and used other mechanism instead. Tested with lantern and it did not drop anymore when going through the door.

src/supertux/game_session.cpp
```diff
@@ -282,6 +298,58 @@ GameSession::on_escape_press(bool force_quick_respawn)
...
+      else
+      {
+        fade_point = m_currentsector->get_camera().get_center();
+      }
+    }
+  }
+
+  fade_point = (fade_point - m_currentsector->get_camera().get_translation()) * m_currentsector->get_camera().get_current_scale();
+
+  return fade_point;
```
Vankata453, line 350: The result could be `return`ed directly.

src/supertux/game_session.hpp
```diff
@@ -34,6 +34,7 @@
 #include "supertux/sequence.hpp"
 #include "supertux/timer.hpp"
 #include "video/surface_ptr.hpp"
+#include "supertux/screen_fade.hpp"
```
Vankata453, line 37: Should go upwards for alphabetical sorting.

src/supertux/game_session.hpp
```diff
@@ -135,12 +141,17 @@ class GameSession final : public Screen,
...
+  Vector get_fade_point(const Vector& position) const;
+
 public:
   bool reset_button;
   bool reset_checkpoint_button;
 
   bool m_prevent_death; /**< true if players should enter ghost mode instead of dying */
 
+  static constexpr float TELEPORT_FADE_TIME = 1.0f; /**< Duration of teleport fade animation */
+
```
Vankata453, line 154: Instead, use a static variable, located at the start of the class declaration, prefixed with `s_`:

```
public:
  static const float s_teleport_fade_time;
```

The static variable should be defined at the top of the .cpp file:

```
const float GameSession::s_teleport_fade_time = 1.f;
```

src/supertux/sector.hpp
```diff
@@ -47,6 +47,7 @@ class Size;
 class TextObject;
 class TileMap;
 class Writer;
+class SpawnPointMarker;
```
Vankata453, line 50: Should be moved upwards (alphabetical).

src/supertux/game_session.cpp
```diff
@@ -282,6 +298,58 @@ GameSession::on_escape_press(bool force_quick_respawn)
...
+        fade_point = average_position / alive_players;
+      }
+      else
+      {
+        fade_point = m_currentsector->get_camera().get_center();
+      }
+    }
+  }
+
+  fade_point = (fade_point - m_currentsector->get_camera().get_translation()) * m_currentsector->get_camera().get_current_scale();
```
tobbi, line 348: Maybe move camera into a variable before this?

src/supertux/shrinkfade.cpp
```diff
@@ -41,9 +42,9 @@ void
 ShrinkFade::draw(DrawingContext& context)
 {
   float progress = m_accum_time / m_fade_time;
-  float diameter = 2 * m_initial_size * (1.0f - progress);
+  float diameter = 2 * m_initial_size * (m_direction == FADEOUT ? (1.0f - progress) : progress);
   context.color().draw_inverse_ellipse(m_dest, Vector(1.1f * diameter, diameter),
-                                         Color(0, 0, 0), LAYER_GUI + 1);
+                                         Color(0, 0, 0), LAYER_LIGHTMAP - 1);
```
Vankata453, line 47: These transitions are not only used in levels, but also in the GUI as well. I wonder if we could use a variable here to know which is the case?

src/supertux/sector.cpp
```diff
@@ -179,6 +179,29 @@ Sector::activate(const std::string& spawnpoint)
     }
   }
 
+  return sp;
+}
+
+
```
tobbi, line 185: Extraneous line. (I'm nitpicking here, I know)

src/supertux/shrinkfade.hpp
```diff
@@ -24,7 +24,8 @@
 class ShrinkFade final : public ScreenFade
 {
 public:
-  ShrinkFade(const Vector& point, float fade_time);
+  enum Direction { FADEOUT, FADEIN };
+  ShrinkFade(const Vector& point, float fade_time, Direction = FADEOUT);
```
Vankata453, line 28: Separating enumerators from constructor:

```suggestion
  enum Direction { FADEOUT, FADEIN };

public:
  ShrinkFade(const Vector& point, float fade_time, Direction = FADEOUT);
```

tobbi:
Looks good otherwise. Just those few changes.

src/trigger/door.cpp
```diff
@@ -29,34 +29,43 @@
 #include "supertux/flip_level_transformer.hpp"
 #include "util/reader_mapping.hpp"
 
+
+static const float STAY_OPEN_TIME = 1.0f;
+static const float LOCK_WARN_TIME = 0.5f;
+static const float CENTER_EPSILON = 5.0f;
+static const float WALK_SPEED = 100.0f;
+
+
```
Vankata453, line 38: You could remove the excess newlines, if you want to.

src/scripting/level.cpp
```diff
@@ -45,6 +45,21 @@ Level_spawn(const std::string& sector, const std::string& spawnpoint)
...
+Level_spawn_transition(const std::string& sector, const std::string& spawnpoint, const std::string& transition)
+{
+  SCRIPT_GUARD_GAMESESSION();
+
+  ScreenFade::FadeType fade_type = ScreenFade::FadeType::NONE;
+
+  if (transition == "fade")
+    fade_type = ScreenFade::FadeType::FADE;
+  else if (transition == "circle")
+    fade_type = ScreenFade::FadeType::CIRCLE;
```
Vankata453, line 58: An additional `else`, which prints out a warning (using `log_warning`) that the transition provided is not valid, would be great and would allow for consistency with other similar methods (for example, ones that require easing).

```suggestion
  if (transition == "fade")
    fade_type = ScreenFade::FadeType::FADE;
  else if (transition == "circle")
    fade_type = ScreenFade::FadeType::CIRCLE;
  else
    log_warning << "Invalid transition type '" << transition << "'." << std::endl;
```

src/object/player.cpp
```diff
@@ -2374,6 +2375,13 @@ Player::make_invincible()
   Sector::get().get_singleton_by_type<MusicObject>().play_music(HERRING_MUSIC);
 }
 
+void
+Player::make_temporarily_safe(float safe_time)
+{
+  m_safe_timer.start(safe_time);
+  m_is_intentionally_safe = true;
```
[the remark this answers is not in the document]
Brockengespenst: Otherwise Tux would start to blink once he wants to enter the door. This flag is used to avoid the blinking effect.

Brockengespenst:
> Makes sense as Tux walks towards the center of the door anyway.

I just noticed that I use the door's center position already for the fade effect.

Rusty-Box:
Have requested changes been adressed or... what the status?

Vankata453:
@Rusty-Box I have to re-review and test - will try to soon.

jonathanpoatato:
w

src/supertux/shrinkfade.hpp
```diff
@@ -24,18 +24,23 @@
 class ShrinkFade final : public ScreenFade
 {
 public:
-  ShrinkFade(const Vector& point, float fade_time);
+  enum Direction { FADEOUT, FADEIN };
+
+public:
+  ShrinkFade(const Vector& point, float fade_time, int draw_layer, Direction = FADEOUT);
```
Vankata453, line 30: I suggest having the previous value, used in `draw()`, which is `LAYER_GUI + 1`, as a default.

src/scripting/level.cpp
```diff
@@ -22,6 +22,8 @@
 #include "supertux/game_session.hpp"
 #include "supertux/sector.hpp"
 
+#include "util/log.hpp"
```
Vankata453, line 25: There shouldn't be a separation of includes here:

```suggestion
#include "util/log.hpp"
```

src/scripting/level.hpp
```diff
@@ -66,6 +66,19 @@ bool Level_has_active_sequence();
...
 
+
+/**
+ * Respawns Tux in sector named ""sector"" at spawnpoint named ""spawnpoint"" with the given transition ""transition"".${SRG_TABLENEWPARAGRAPH}
+   Exceptions: If ""sector"" or ""spawnpoint"" are empty, or the specified sector does not exist, the function will bail out the first chance it gets.
+   If the specified spawnpoint doesn't exist, Tux will be spawned at the spawnpoint named “main”.
+   If that spawnpoint doesn't exist either, Tux will simply end up at the origin (top-left 0, 0).
+ * @param string $sector
+ * @param string $spawnpoint
+ * @param string $transition
```
Vankata453, line 77: Valid transition types should be specified under the description of this parameter.

Vankata453 (commented):
Tested this and it seems to work well!

However, I noticed a few other things on the code side, so I have commented my suggestion for those. There is also one comment from my previous review left unresolved.

src/supertux/shrinkfade.hpp
Brockengespenst: Done.

src/scripting/level.cpp
Brockengespenst: Done.

src/scripting/level.hpp
Brockengespenst: Done.

Brockengespenst:
>     * Currently, entering a `GameSession` always plays the `ShrinkFade` transition. It seems to be intended, so I wonder why it was added.

Sorry, I do not really get the point?

Vankata453:
> Sorry, I do not really get the point?

I was wondering why a circle fade is drawn when entering a level (because I didn't find comments where it has been discussed, may have missed them).

Brockengespenst:
> I was wondering why a circle fade is drawn when entering a level (because I didn't find comments where it has been discussed, may have missed them).

From #2409:

> What if instead of simply fading to black (using doors) or no transition at all (using Level.spawn();) we could use transition effect closer to entering a level, the circular one.

I thought that it was intended by @Rusty-Box, that's why I added it.

Vankata453:
> but until now there were no complaints about it

Personally I'm not necessarily against it, but it was a surprise. :D

Vankata453:
In terms of my reviews, one more [comment](https://github.com/SuperTux/supertux/pull/2844#discussion_r1548724420) is left unresolved.

Rusty-Box:
> I thought that it was intended by @Rusty-Box, that's why I added it. Maybe I mixed this up with the world map, but until now there were no complaints about it. :-)

It was never mentioned anywhere, yes, but I really like that effect for level start so I'd personally love to keep it around.

src/supertux/game_session.hpp
Brockengespenst: Sorry, somehow I missed this. It could be moved completely to the implementation file as the GameSession will make Tux safe during the fadeout animation anyway. I hope it is Ok that I did it similar to the other static const's in the other code instead of using your suggestion. Otherwise I can change it again.

Vankata453: It's okay, my suggestion was keeping in mind usage from other files.

Vankata453 (approved):

MatusGuy (approved):
Untested.

src/trigger/door.cpp
```diff
@@ -85,79 +92,98 @@ Door::after_editor_set()
...
+    if (fabs(diff_to_center.x) >= CENTER_EPSILON)
+    {
+      const bool move_right = diff_to_center.x > 0.0f;
+      m_triggering_player->set_dir(move_right);
+      m_triggering_player->walk(move_right ? WALK_SPEED : -WALK_SPEED);
+    }
+    else
+    {
+      m_triggering_player->walk(0.0);
```
tobbi, line 157: Maybe: `0.0f`? Shouldn't make a difference.

tobbi (changes requested):
Just the small thing that I marked. Otherwise looking good. Untested, though. I'll test it in a few.

src/trigger/door.cpp
Brockengespenst: Done.

Rusty-Box (approved):
Apparently I didn't post a proper review here so... yeah, approved!

tobbi:
Okay, so a few things that I noticed: Is it necessary to have the teleport transition circle when starting a level? IDK, to me it kinda seems out of place.

Also, I'd like the speed of the transition to pick up a bit. Seems a little bit too slow for me.

Rusty-Box:
> Okay, so a few things that I noticed: Is it necessary to have the teleport transition circle when starting a level? IDK, to me it kinda seems out of place.

I think it looks nicer. More cartoony. Fits the game's style more. It also fits since entering a level from the worldmap uses the exact same transition

> Also, I'd like the speed of the transition to pick up a bit. Seems a little bit too slow for me.

Entering a door is quite quicker than it used to be, idk if it needs to be faster. A tad maybe. Personally I think it's fine enough

Vankata453:
> Entering a door is quite quicker than it used to be, idk if it needs to be faster. A tad maybe. Personally I think it's fine enough

I agree, faster than this would be very odd.

tobbi:
Okay, in that case: Approval from me.

tobbi (approved):

PR #2847, opened 2024-03-26
Remove unnecessary randomness of mole

weluvgoatz:
Having some enemies with random behavior is so horrible for stuff like speedruns, or situations in which something should not be random.  There is an instance of randomness in badguys that literally does not need to exist, and I have removed it here.  That instance is this:
- Random throwing of rocks by moles has been changed to a certain pattern, which is still very similar to the old one.

### DIFF
--- a/src/badguy/mole.cpp
+++ b/src/badguy/mole.cpp
@@ -76,7 +76,7 @@ Mole::collision_squished(GameObject& )
 void
 Mole::throw_rock()
 {
-  float angle = math::radians(135.f - (float(cycle_num) * 30.f));
+  float angle = math::radians(135.f - (static_cast<float>(cycle_num) * 30.f));
 
   SoundManager::current()->play("sounds/dartfire.wav", get_pos());
   Sector::get().add<MoleRock>(m_col.m_bbox.get_middle(),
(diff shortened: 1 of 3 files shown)


Rusty-Box:
Zeekling's random speed is already addressed in #2624

Alasdairbugs:
i believe skullyhops also have some random function which affect their delay before hopping

Frostwithasideofsalt (approved):

src/badguy/mole.cpp
```diff
@@ -75,14 +76,13 @@ Mole::collision_squished(GameObject& )
 void
 Mole::throw_rock()
 {
-  float angle;
-  float base_angle = (m_flip == NO_FLIP ? 90.0f : 270.0f);
-  angle = math::radians(gameRandom.randf(base_angle - 15.0f, base_angle + 15.0f));
-  float vx = cosf(angle) * THROW_VELOCITY;
-  float vy = -sinf(angle) * THROW_VELOCITY;
+  float angle = math::radians(135.f - (float(cycle_num) * 30.f));
```
Vankata453, line 79: Use `static_cast` instead:

```suggestion
  float angle = math::radians(135.f - (static_cast<float>(cycle_num) * 30.f));
```

Vankata453 (approved):
Works well! The consistent pattern also allows making strategies on how to go around a mole in certain cases possible.

Code-wise, one little thing:

src/badguy/mole.cpp
MatusGuy: Look! It's a giant commit button

tobbi: I want to be able to squash unrelated commit suggestions.

PR #2846, opened 2024-03-25
Incorporate certain object dynamics

weluvgoatz:
For my 100th PR, I decided to fix some old issues I had involving the dynamics between some objects.  Included in this PR are the following:
- Rocks no longer hurt Tux and bounce off of him, making rock puzzles more fair.
- Rocks now interact as expected with slick ice, water, bumpers, and trampolines.
- Enemies now interact as expected with water, bumpers, and trampolines. 
- Objects oriented to the side (switches, darttraps, bumpers, stalactites, yeti stalactites, pushbuttons, ispys, and shards) now properly stick to moving objects which look like they can be stuck to (moving tilemaps, platforms, crushers and fallblocks). [Fixes #1782]

Note: I was hoping to incorporate some more general method for point 4, where I didn't have to specify every single interaction and I could just label some objects as "sticky" and "sticked" objects, but this proved to be difficult as I don't think the game is currently capable of describing objects in such a way that I could do that, and I did not want to spend this PR refactoring some part of the game just to include that little thing.  Each of the different possible interactions tends to have to some slight variation from one another, and there aren't very many of these possible interactions, anyway.  

Fixes #2909 Fixes #2912

PR #2849, opened 2024-03-26
Fix some fallblock issues

weluvgoatz:
There existed an issue where the fallblock would not properly convey that it was going to fall, and I fixed that in this PR.  Also, enemies are properly squished by fallblocks.  
NOTE: This PR removes the feature where bumpers would cling to fallblocks.  This is because the feature is re-added in a more consistent way in PR #2846.  
NOTE 2: I think the fallblock needs a new, unique graphic.  Maybe when it is drawn it can be added to this PR.

### DIFF
--- a/src/object/fallblock.cpp
+++ b/src/object/fallblock.cpp
@@ -16,6 +16,7 @@
 #include "object/fallblock.hpp"
 
 #include "audio/sound_manager.hpp"
+#include "badguy/badguy.hpp"
 #include "object/bumper.hpp"
 #include "object/player.hpp"
 #include "object/camera.hpp"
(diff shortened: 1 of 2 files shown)


Rusty-Box:
If you have any ideas/suggestions of what to use for a fallblock graphic I can work on one. I know that I am using this object for ruin themed levels where some bricks of the ceiling may fall down, so one graphic I will be doing is a cracked/broken version of the ruin brick

weluvgoatz:
I decided that the graphics for fallblock can be decided later.  This is solely for bug fixing.

weluvgoatz:
Closing because all this stuff has been added into the game already.

PR #2853, opened 2024-03-27
Fix dispenser dying

MatusGuy:
- [x] make them unaffectable by rocks
- [x] fix cannon left & right sprites
- [x] unfreeze after a while


src/badguy/dispenser.cpp
```diff
@@ -163,6 +163,12 @@ Dispenser::active_update(float dt_sec)
   }
 }
 
+void
+Dispenser::kill_fall()
+{
+  return;
```
tobbi, line 169: Isn't the return kinda pointless? You can just leave the method empty.

MatusGuy: No. It overrides the kill_fall method to do absolutely nothing, making the dispenser truly unkillable

MatusGuy: wait now i get what you mean. Sure i can remove the return

tobbi (approved):
IDK, just looking at the code, this looks fine. Didn't test it.

Frostwithasideofsalt (approved):
Tested it, seems to work

PR #2854, opened 2024-03-28
Fix crusher issues

### LINKED ISSUE #1880: Icecrusher Bug List
Here I am going to keep track of icecrusher bugs that I have discovered so that someone or me can fix them in the future:

- Rocks placed atop icecrushers cause it to bounce oddly on the floor.
- Icecrusher doesn't return to exact original position and gets a little offset each time.

Feel free to append any of your own, if you have found any.

### LINKED ISSUE #2038: Icecrushers don't propagate movement on some objects
A bug I noticed is that the icecrusher will not propagate movement on frozen badguys, rocks, or scripted objects.  It should be noted that icecrushers and these objects are all movingstatics (in COLGROUP_MOVING_STATIC), and that other movingstatics that I *assume* would be loaded before the icecrushers due to their alphabetic placement (falling blocks, dispensers) or recency in creation (frozen enemies are created and loaded after icecrushers since they can't be instantly created at the start of the level) allow the icecrusher to propagate their movement perfectly fine.  And objects of other colgroups (specifically COLGROUP_MOVING), such as players, powerups, etc. also allow icecrusher propagation, and other COLGROUP_STATIC objects that propagate objects propagate them all.  So maybe this was an oversight with movingstatics when movement propagation was added?

weluvgoatz:
This PR aims to fix some icecrusher issues.  Please test it, it might have some issues.  

- [x] Fix, or add a workaround, for the weird offset experienced whenever a crusher returns to its position.
- [x] Fix the part of the game where crushers would be weighed down by things that are obviously too small for them.

KNOWN ISSUES:
- [x] Rocks can be shoved into the floor by Crushers.

Fixes #1880.
Fixes #2038.

### DIFF
--- a/src/badguy/crusher.cpp
+++ b/src/badguy/crusher.cpp
@@ -137,12 +137,16 @@ Crusher::collision(GameObject& other, const CollisionHit& hit)
   }
 
   auto* badguy = dynamic_cast<BadGuy*>(&other);
-  if (badguy && m_state == CRUSHING) {
+  if (badguy && m_state == CRUSHING && ((!m_sideways && hit.bottom) ||
+    (m_sideways && ((hit.left && m_physic.get_velocity_x() < 0.f) || (hit.right && m_physic.get_velocity_x() > 0.f)))))
+  {
     badguy->kill_fall();
   }
 
   auto* rock = dynamic_cast<Rock*>(&other);
-  if (rock && !rock->is_grabbed() && ((hit.bottom && !m_sideways) || (m_sideways && (hit.left || hit.right))) && m_state == CRUSHING) {
+  if (rock && !rock->is_grabbed() && m_state == CRUSHING && ((!m_sideways && hit.bottom) ||
+    (m_sideways && ((hit.left && m_physic.get_velocity_x() < 0.f) || (hit.right && m_physic.get_velocity_x() > 0.f)))))
+  {
     SoundManager::current()->play("sounds/brick.wav", get_pos());
     m_physic.reset();
     set_state(RECOVERING);
--- a/src/badguy/crusher.cpp
+++ b/src/badguy/crusher.cpp
@@ -138,14 +138,16 @@ Crusher::collision(GameObject& other, const CollisionHit& hit)
 
   auto* badguy = dynamic_cast<BadGuy*>(&other);
   if (badguy && m_state == CRUSHING && ((!m_sideways && hit.bottom) ||
-    (m_sideways && ((hit.left && m_physic.get_velocity_x() < 0.f) || (hit.right && m_physic.get_velocity_x() > 0.f)))))
+      (m_sideways && ((hit.left && m_physic.get_velocity_x() < 0.f) ||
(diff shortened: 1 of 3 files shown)


src/badguy/crusher.cpp
```diff
@@ -136,10 +137,22 @@ Crusher::collision(GameObject& other, const CollisionHit& hit)
...
   }
 
+  auto* rock = dynamic_cast<Rock*>(&other);
+  if (rock && !rock->is_grabbed() && m_state == CRUSHING && ((!m_sideways && hit.bottom) ||
+    (m_sideways && ((hit.left && m_physic.get_velocity_x() < 0.f) || (hit.right && m_physic.get_velocity_x() > 0.f)))))
+  {
+    SoundManager::current()->play("sounds/brick.wav", get_pos());
+    m_physic.reset();
+    set_state(RECOVERING);
+    return ABORT_MOVE;
```
MatusGuy, line 153: This means the rock interrupts the crusher even if it's airborne. Is this what you want?

MatusGuy (commented):
Tested

src/object/rock.cpp
```diff
@@ -151,11 +151,7 @@ Rock::collision(GameObject& other, const CollisionHit& hit)
 
   auto crusher = dynamic_cast<Crusher*> (&other);
   if (crusher) {
-    auto state = crusher->get_state();
-    if(state == Crusher::CrusherState::RECOVERING ||
-       state == Crusher::CrusherState::IDLE) {
-        return ABORT_MOVE;
-       }
+    return FORCE_MOVE;
```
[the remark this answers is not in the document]
MatusGuy: wait what thats the wrong line

MatusGuy, line 154: This if statement probably becomes useless. Idk. You'd have to test

weluvgoatz: It doesn't.  It helps with making sure the crusher doesn't push the rock into the ground.  (I think.)

src/badguy/crusher.cpp
weluvgoatz: I might test around with this, we'll see.

src/collision/collision_system.cpp
```diff
@@ -419,6 +419,14 @@ CollisionSystem::collision_static(collision::Constraints* constraints,
   // Collision with other (static) objects.
   for (auto* static_object : m_objects)
   {
+    float static_size = static_object->get_bbox().get_width() * static_object->get_bbox().get_height();
+    float object_size = object.get_bbox().get_width() * object.get_bbox().get_height();
+    float forgiveness = 16.f * 16.f;
```
mrkubax10, line 424: `const float` for first two, and also `static const float FORGIVENESS` since it's an unchanging constant.

MatusGuy (approved):

Frostwithasideofsalt:
gonna try to test this.

Frostwithasideofsalt (approved):

PR #2855, opened 2024-03-28
Update CMakeLists.txt

iahung2:
Fix building on MSYS2 (UCRT64).


tobbi:
This is odd. The lines are equivalent, they were just moved around.

Brockengespenst:
Probably the link order is important here. I guess LibSDL2_ttf needs things from LibSDL2 and unless you work with mechanics like -Wl,--start-group and -Wl,--end-group, this may result in linker errors.

tobbi:
Oh, I don't know much about link order. Interesting.

MatusGuy (approved):

PR #2858, opened 2024-03-28
Update snail behavior to be more consistent

### LINKED ISSUE #2845: Snail throw + up-throw is swapped + general idea
The way Tux is throwing Snail from his hands appears to be swapped:

- If you hold up while throwing Snail, they will just get pushed forward rather than flying up in an arch as you would expect
- Letting go of Snail normally will throw it upwards in an arch instead

---

I also had a thought that Snail should also not get flung upwards in an arch if you kick it by jumping on it again and instead just get simply pushed forward. I feel the way it works now makes it kinda not worth it to use them at all as they often hit you instead by accident.

weluvgoatz:
This PR makes snail behavior more consistent.  It now works like this:
- Throwing the snail now throws it just like every other object in the game.
- Kicking the snail by any means makes it bounce up and spin (may need changing).
- A snail hitting a wall while spinning makes it also bounce up (may need changing).
- Throwing a snail while swimming makes it spin in the direction it was thrown .
Known issues:
- [x] Throwing a snail in water straight up or down makes it return to normal immediately.  This is because the game makes the snail return to normal if its x speed is less than its walking speed, but may just need to be changed to the absolute value of its speed overall.

Fixes #2845.

### DIFF
--- a/src/badguy/snail.cpp
+++ b/src/badguy/snail.cpp
@@ -212,7 +212,7 @@ Snail::active_update(float dt_sec)
 
     case STATE_KICKED:
       m_physic.set_velocity_x(m_physic.get_velocity_x() * powf(0.99f, dt_sec/0.02f));
-      if (fabsf(m_physic.get_velocity_x()) < walk_speed) be_normal();
+      if (on_ground() && (fabsf(m_physic.get_velocity_x()) < walk_speed)) be_normal();
       break;
 
     case STATE_GRABBED:
--- a/src/badguy/snail.cpp
+++ b/src/badguy/snail.cpp
@@ -327,7 +327,7 @@ Snail::collision_player(Player& player, const CollisionHit& hit)
       m_dir = Direction::LEFT;
     }
     player.kick();
-    be_kicked(true);
+    be_kicked(false);
     return FORCE_MOVE;
   }
 
@@ -371,7 +371,7 @@ Snail::collision_squished(GameObject& object)
           m_dir = Direction::LEFT;
         }
       }
-      be_kicked(true);
+      be_kicked(false);
       break;
 
     default:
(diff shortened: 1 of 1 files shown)


Rusty-Box:
Looks good, but imo snail should no bounce upwards, at least not that high when jump upon and all that. Very annoying to use offensively without killing yourself

Rusty-Box (approved):

src/badguy/snail.cpp
```diff
@@ -404,19 +404,49 @@ Snail::ungrab(MovingObject& object, Direction dir_)
...
+      else
+      {
+        float swimangle = player->get_swimming_angle();
+        m_col.m_bbox.move(Vector(std::cos(swimangle) * 48.f, std::sin(swimangle) * 48.f));
+        be_kicked(false);
+        m_physic.set_velocity(SNAIL_KICK_SPEED * 1.5f * Vector(std::cos(swimangle), std::sin(swimangle)));
+        m_dir = m_physic.get_velocity_x() > 0.f ? Direction::RIGHT : Direction::LEFT;
+      }
+    }
+    else if (owl) {
```
MatusGuy, line 448: I believe this brace goes on the next line but I'm not sure.

```suggestion
    else if (owl)
    {
```

MatusGuy (commented):
Haven't tested this yet. Will approve once I do

MatusGuy (commented):
I don't hear anything when I kick/throw the snail. Intended?

src/badguy/snail.cpp
weluvgoatz: It's only 1 line, so it doesn't.

MatusGuy (approved):

PR #2859, opened 2024-03-29
Fix lightning strikes

### LINKED ISSUE #2842: Scripted Lightning Flash brightens level permanently
The lightning flash works fine normally, however, if you disable to lighning object in your level and have it perform a lighnting flash through scripting, using `[Thunderstorm].lightning();` the level begins to slowly brighten up with every strike until the whole screen is just white.

Another thing, it appears that when the background tints it also reduces the alpha value, which should not happen!

weluvgoatz:
Fixes #2842.

### DIFF
--- a/src/object/thunderstorm.hpp
+++ b/src/object/thunderstorm.hpp
@@ -61,7 +61,7 @@ public:
   void thunder();
 
   /** Play lightning, i.e. call flash() and electrify() */
-  void lightning();
+  void lightning(bool is_scripted = false);
 
   /** Display a nice flash */
   void flash();
@@ -72,7 +72,7 @@ public:
   /** @} */
 
 private:
-  void change_background_colors(bool is_lightning);
+  void change_background_colors(bool is_lightning, bool is_scripted = false);
   void restore_background_colors();
 
 private:
--- a/src/object/thunderstorm.hpp
+++ b/src/object/thunderstorm.hpp
@@ -60,12 +60,9 @@ public:
   /** Play thunder */
   void thunder();
 
-  /** Play lightning, i.e. call flash() and electrify() */
+  /** Methods for doing lightning by different methods. Necessary for the future implementation of SimpleSquirrel. */
+  void lightning_general();
   void lightning();
-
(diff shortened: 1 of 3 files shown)


Rusty-Box:
It's not quite fixed. If `running = false`, in the thunderstorm object's properties, and you run the script, the flash is now completely turned off.

Also regarding the background darkening, while the transparency issue seems fixed, now instead of remaining darkened until lightning strikes it gets darker for but a short moment.

weluvgoatz:
> It's not quite fixed. If `running = false`, in the thunderstorm object's properties, and you run the script, the flash is now completely turned off.

Well that's how it is currently in the nightly.  I don't recall you specifying this issue in the original report.  Did you want some other behavior here instead?

> Also regarding the background darkening, while the transparency issue seems fixed, now instead of remaining darkened until lightning strikes it gets darker for but a short moment.

Is the intended behavior to stay dark? If so I can change that.  I didn't know.

Rusty-Box:
> Well that's how it is currently in the nightly. I don't recall you specifying this issue in the original report. Did you want some other behavior here instead?

I wasn't aware this made a difference. If you ask me it should work no matter what as the lightning script causes lightning to happen no matter what. So the flash should work in all cases!

> Is the intended behavior to stay dark? If so I can change that. I didn't know.

It is indeed intended. People wanted a visual indicator for when lightning is about to strike in case they play with no audio.

Rusty-Box (approved):
Seems to be working!

Frostwithasideofsalt (approved):
Tested it, seems to work

src/object/thunderstorm.hpp
```diff
@@ -60,7 +61,7 @@ class Thunderstorm final : public GameObject,
   void thunder();
 
   /** Play lightning, i.e. call flash() and electrify() */
-  void lightning();
+  void lightning(bool is_scripted = false);
```
Vankata453, line 64: When we migrate to `simplesquirrel`, the `bool is_scripted` parameter wouldn't be supported - the scripting function to be exposed should rather be a separate function in the code, preferably with the same name as the one it's exposed with (easier for documentation).

PR #2862, opened 2024-03-29
Prevent corrupted big granito from dying twice

MatusGuy:


Vankata453:
What was the original issue and how can I attempt to reproduce it?

Vankata453 (approved):
Fixes the issue.

tobbi:
Seems trivial. But since this is a PR by Marty, I'm not gonna give my final review until the code went through government-level security analysis first. Don't want to have anything broken.

PR #2865, opened 2024-03-30
Jump to path node immediately

tobbi:
This currently only works for the platform thing. I didn't add the same thing into the Tilemap class yet.

I don't know if this has further repercussions.

### DIFF
--- a/src/object/path_walker.cpp
+++ b/src/object/path_walker.cpp
@@ -40,7 +40,7 @@ PathWalker::PathWalker(UID path_uid, bool running_) :
   m_running(running_),
   m_current_node_nr(0),
   m_next_node_nr(),
-  m_stop_at_node_nr(m_running?-1:0),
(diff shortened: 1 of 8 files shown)


tobbi:
@Rusty-Box Try this out, please.

Rusty-Box (changes requested):
The `set_node(x);` script works on its own pretty well. However, after using the `set_node(x);` script and then running the `start_moving(bool);` script, the object/tilemap will not start moving anymore.

Rusty-Box (approved):
Now the `start_moving();` script works fine after the `set_node(x);` was run. Neat!

MatusGuy (approved):

tobbi (commented):
If I could approve, I would. 

Here's my "Approve" review:

A very exceptional pull request, Tobbi. Only a person as intelligent and superb as you could produce such superior code. Spread the word.

PR #2866, opened 2024-03-30
Update PhysFS to 3.2.0

tobbi:
We have been on some rather old-ish version of PhysFS since quite a while. Even though we support building against locally installed versions of PhysFS; I feel like updating to the latest release version can have benefits.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -9,7 +9,7 @@
 	url = https://github.com/SuperTux/sexp-cpp.git
 [submodule "external/physfs"]
 	path = external/physfs
-	url = https://github.com/SuperTux/physfs.git
+	url = https://github.com/icculus/physfs.git
 [submodule "external/SDL_ttf"]
 	path = external/SDL_ttf
 	url = https://github.com/SuperTux/SDL_ttf


tobbi:
I think we should update the submodule reference to point to upstream PhysFS. All the changes that were on the supertux-changes branch on our local fork of the submodule have already been incorporated upstream.

We should keep the repository archived, so existing builds don't start failing. 

Unfortunately, the history of our "fork" is not compatible with the upstream here: github.com/icculus/physfs because one was created using some hg-to-git

tobbi:
Maybe I'll rename the branch name on our fork so that we can have the same branches as upstream does.

MatusGuy (approved):
Tested. Works on ubuntu jammy

mrkubax10 (approved):

Vankata453 (approved):
> I think we should update the submodule reference to point to upstream PhysFS.

Yeah, would be good, since with this PR, we are using an upstream release either way.

Otherwise, this builds fine on Arch Linux and runs properly.

tobbi:
Can I make our physfs clone read-only?

Vankata453:
> Can I make our physfs clone read-only?

Sure.

tobbi:
I've archived our clone now and added a note stating the reasons for the deprecation.

tobbi:
I'm rerunning this one mac build and will merge tomorrow.

tobbi:
Since I can't sleep, I might as well merge now.

PR #2884, opened 2024-04-02
Editor: Adding layers from layers widget

### LINKED ISSUE #638: Rework adding layers to sector
Having layers and backgrounds / particles in the object menu might not be a good idea. I think they should stay close to the original layer selector, e.g. by having a plus button next to it.

Upon clicking the plus button the user can choose particles, tilemaps, backgrounds and everything else that doesn't technically belong to the "object" side of levels.

Vankata453:
Adding layers to a sector now makes more sense, since it's done through a "+" button, found after the last layer on the layers widget. Pressing the button shows an object selection, just like in the toolbox widget, only containing the objects, formerly found in the "Sector" category for non-worldmap levels, and "Tilemap" and "Background" for worldmap levels. Pressing an object directly adds it to the sector.

image

To achieve this, most of the `EditorToolboxWidget` code, which managed tile/object selection, was moved to a different class. The resulting `EditorTilebox` class can also be used for features like tile backpacks in the future.

Additionally, this PR allows objects to be moved around right after the editor is started (without an objectgroup having to be active).

Closes #638.


tobbi:
Difficult to review. From my preliminary look, it seems like the code looks good. But let's avoid disasters and test this thoroughly. I like how this reduces complexity in the editor classes.

tobbi:
Maybe add a border around that thing so it isn't too "out of place"?

data/images/engine/editor/add.png
MatusGuy, line 1: Because this icon is black, it will blend in with people who have dark menu customization. I'll make an icon

MatusGuy (commented):
Well done. I like this pr. I may now test in game

tobbi:
Can you rebase this, please?

MatusGuy (approved):

tobbi (approved):

PR #2886, opened 2024-04-02
Fix swimming on slopes

MatusGuy:


MatusGuy:
i'm figuring a few things out. when im done i will reopen this

MatusGuy:
the fix is now present in #2812

Brockengespenst:
I think it would be good to apply these changes, as it should at least solve several issues with slopes. Additional fixes may be done in subsequent PRs.

PR #2891, opened 2024-04-05
Add ability to drag and delete objects

MatusGuy:
This creates an empty object group.


tobbi:
Does this mean that both object mode and tile mode will be allowed when starting? Or only object mode at first?

MatusGuy:
The editor starts in object mode, with an empty object group

PR #2898, opened 2024-04-07
Bycicle platform improvements

MatusGuy:
- [x] easier to place
- [x] custom sprites
- [x] unisolid
- [ ] follow paths

### DIFF
--- a/src/object/bicycle_platform.cpp
+++ b/src/object/bicycle_platform.cpp
@@ -19,6 +19,7 @@
 #include <algorithm>
 #include <math.h>
 
+#include "editor/editor.hpp"
 #include "math/util.hpp"
 #include "object/player.hpp"
 #include "object/portable.hpp"
@@ -35,6 +36,7 @@ BicyclePlatformChild::BicyclePlatformChild(const ReaderMapping& reader, float an
   m_momentum(),
   m_contacts()
 {
+  m_col.set_unisolid(true);
 }
 
 void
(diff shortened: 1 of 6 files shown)


src/object/bicycle_platform.cpp
```diff
@@ -55,8 +61,7 @@ BicyclePlatformChild::collision(GameObject& other, const CollisionHit& )
   const float gravity = Sector::get().get_gravity();
 
   // Somehow the hit parameter does not get filled in, so to determine (hit.top == true) we do this:
```
mrkubax10, line 63: Is this still true?

src/object/bicycle_platform.cpp
```diff
@@ -156,58 +174,43 @@ BicyclePlatform::update(float dt_sec)
...
-
-void
-BicyclePlatform::editor_delete()
-{
-  // Remove children.
-  for (auto& child : m_children)
-  {
-    child->remove_me();
-  }
+  std::cout << m_angular_speed << std::endl;
```
mrkubax10, line 182: Remove debug code (in other places too).

MatusGuy: This pull request is not ready for review

mrkubax10: You requested review from me.

MatusGuy: Oh, you're right, I did. It was an accident, sorry

PR #2903, opened 2024-04-13
Use external/glm fallback when system GLM not found

tobbi:
Basically, what I wrote in the title above.

This will use the GLM fallback in external/glm if system GLM could not be found.

If anyone has any ideas on how to improve the code even further go right ahead.


MatusGuy (approved):
ah yes. my favourite library and file

tobbi:
> ah yes. my favourite library and file

It's everyone's favourite library.

Zwatotem (approved):
If it fixes the pipeline, I'm fine with this.

tobbi:
It fixes no pipeline currently, but it's the sane thing to do.

PR #2907, opened 2024-04-15
Prevent infinite recursion in rock's grab and ungrab scripts

tobbi:
This prevents #2906 from happening.


weluvgoatz:
However this PR does fix the crash. Thank you!!!!

weluvgoatz (approved):

tobbi:
I'd really like to fix the flickering issue as well so that it only rotates once.

tobbi:
I think the solution might be easier than I initially thought.

Vankata453 (approved):

weluvgoatz (approved):

PR #2908, opened 2024-04-15
Handle errors with a stacktrace dialog

MatusGuy:


tobbi (commented):
I don't know anything about this.

src/supertux/error_handler.cpp
```diff
@@ -60,7 +71,58 @@ ErrorHandler::handle_error(int sig)
...
+
+    free(symbol);
+    SymCleanup(GetCurrentProcess());
+
+    stacktrace = stacktrace_stream.str();
+  }
+
+stacktrace_error:
+
+  if (stacktrace.size() > 0)
```
Vankata453, line 111: `if (!stacktrace.empty())` would work, no need for size.

src/supertux/error_handler.cpp
```diff
@@ -60,7 +71,58 @@ ErrorHandler::handle_error(int sig)
...
+  std::string msg = "SuperTux has encountered an unrecoverable error!\n";
+
+  {
+    std::stringstream stacktrace_stream;
+
+    // Initialize symbols
+    SymSetOptions(SYMOPT_UNDNAME | SYMOPT_DEFERRED_LOADS);
+    if (!SymInitialize(GetCurrentProcess(), NULL, TRUE))
+    {
+      goto stacktrace_error;
```
Vankata453, line 85: Instead of using `goto`, perhaps getting the stacktrace from symbols could be moved to a different function (only compiled for `WIN32`) instead?

tobbi:
Can something like this also exist for Linux and MacOS?

MatusGuy:
like, a dialog? Not sure about MacOS, but on Linux there is no universal way to create a dialog, and I wouldn't want to rely on xmessage

tobbi:
Couldn't we have an error handler application that posts stacktraces to some API endpoint?

MatusGuy:
you mean, post stacktraces to a server??? just make a github issue...

tobbi:
I meant collect stacktraces on our server the way mozilla does.

tobbi:
We should switch to `RelWithDebInfo` builds for both nightly and release versions. That way we can get line numbers from stacktraces.

mrkubax10:
> Can something like this also exist for Linux and MacOS?

We already print stacktrace in terminal when running on Linux with glibc.

mrkubax10 (approved):

tobbi:
Do we include the main supertux config file contents and / or graphics driver here?

tobbi:
A couple more things: For me, SuperTux hangs indefinitely, no matter which button I click on. Can we force-quit?

tobbi:
We're having a problem:

The call to exit() in the error handler tries to call a bunch of stuff (in "atexit") and therefore SuperTux lingers indefinitely. Any idea what to d?

tobbi:
Okay, so this fixes my issues with the program not exiting.

Vankata453 (approved):

src/supertux/error_handler.cpp
```diff
@@ -49,33 +193,173 @@ ErrorHandler::handle_error(int sig)
...
+
+  // cppcheck-suppress unknownMacro
+  url << "https://github.com/supertux/supertux/issues/new"
+         "?template=crash.yml"
+         "&labels=type:crash,status:needs-confirmation"
+         "&supertux-version=" << FileSystem::escape_url(PACKAGE_VERSION) <<
+         "&system-info=" << FileSystem::escape_url(get_system_info()) <<
+         "&debug-stacktrace=" << FileSystem::escape_url(details);
+
+  FileSystem::open_url(url.str());
```
MatusGuy, line 356: You should escape the entire string instead of the individual strings

Vankata453: I guess that's also a way.

tobbi: Wouldn't that lead to the ampersands getting escaped as well?

MatusGuy: true. Nvm

src/supertux/error_handler.cpp
```diff
@@ -32,9 +54,131 @@ bool ErrorHandler::m_handing_error = false;
...
+    info << "Windows 7";
+  else if (IsWindowsVistaOrGreater())
+    info << "Windows Vista";
+  else if (IsWindowsXPOrGreater())
+    info << "Windows XP";
+  else
+    info << "Windows";
+
+  info << " ";
+*/
```
mrkubax10, line 141: Is this comment needed?

src/util/file_system.cpp
```diff
@@ -230,6 +232,27 @@ void open_path(const std::string& path)
 #endif
 }
 
+std::string escape_url(const std::string& url)
+{
+#ifndef __EMSCRIPTEN__
+  std::string result = url;
+  CURL *curl = curl_easy_init();
```
mrkubax10, line 239: Maybe it would be better to not initialize curl context every time `escape_url` is called?

src/supertux/error_handler.cpp
MatusGuy: not really

Vankata453: I think it's good to have, so this can be enabled if someone knows how to fix it.

mrkubax10: Fair enough.

src/util/file_system.cpp
MatusGuy: How else would you do this?

tobbi (approved):

PR #2916, opened 2024-04-20
Remove useless warnings

MatusGuy:


tylerandari13 (approved):

src/supertux/sector.cpp
```diff
@@ -233,7 +233,6 @@ Sector::activate(const Vector& player_pos)
     // spawning tux in the ground would kill him
     if (!is_free_of_tiles(player.get_bbox())) {
       std::string current_level = "[" + Sector::get().get_level().m_filename + "] ";
```
Vankata453, line 235: This variable seems to now be unused.

tobbi:
Well, IDK, I guess it makes sense to show this warning for people who don't know what they're doing.

But the detection code needs to get fixed as there are a lot of false positives.

bruhmoent:
"Changing a Sector's gravitational constant might have unforeseen side-effects: "
"Tried spawning Tux in solid matter. Compensating."

These warning are literally worthless.

tylerandari13:
> "Changing a Sector's gravitational constant might have unforeseen side-effects: " "Tried spawning Tux in solid matter. Compensating."
> 
> These warnings are literally worthless.



> "Changing a Sector's gravitational constant might have unforeseen side-effects: " "Tried spawning Tux in solid matter. Compensating."
> 
> These warnings are literally worthless.

The solid matter warning is nothing more than a nuisance, and the gravity warning has probably discouraged at least one person (me) from using the gravity script.

Vankata453 (approved):

pazkero:
My opinion has no weight but I'll comment anyway just so this PR seems more lively. I sent this on IRC earlier anyway.

> Surprisingly, I agree with Tobbi on that one...
Also because design perspective ─ if you try to spawn Tux in solid matter, that's illegal. The game tries to compensate instead of outright crashing on your face or doing it anyway and rendering you stuck, but this also means actual spawn point won't be the specified one. (The other "beware side-effects" are the kind of notes which should be in some sort of documentation for level makers, but I'm pretty sure there wasn't anything like that)
(so personally I think these could be moved to outside the source code, but not entirely removed unless their reason no longer exists e.g. gravity constant changes are always predictable and do not affect the behavior besides what's obvious)
In other hand, if you had some sort of level validator, the validator could issue these warnings instead
I think that's how FreedroidRPG does it, actually. You run the validator whenever you want, and it informs of any potential problems with the maps which you should be aware of before publishing them.

Well, here, this PR is lively again, review/merge when 🥺🤣

PR #2919, opened 2024-04-21
Use code font for console and script editor

MatusGuy:


tobbi:
Can you specify the license somewhere so we don't forget to keep track of that?

Also, that font sounds familiar. Was that in the tree already at an earlier point?

MatusGuy:
not sure. I just got a nice code font by google and put it in. the only monospace font we've had is a bitmap font.

Thanks for warning me about the license though

PR #2924, opened 2024-04-24
Set camera to manual in credits screen

Brockengespenst:
Since the shrinkfade effect is used when a game session is created, the fade point in the credits screen is somewhere in the lower left corner of the screen.
Setting the camera to manual mode will center the shrinkfade effect in the credits screen.


MatusGuy:
Note to reviewers: You don't need a CI build. Just change the credits.stl file.

Rusty-Box (approved):

PR #2925, opened 2024-04-27
Fix badguys not climbing certain slopes

### LINKED ISSUE #761: Badguys turn away from slopes >22.5°
**SuperTux version:** Latest revision

##### Expected behavior
Badguys walk up steeper slopes.

##### Actual behavior
Badguys turn away from steeper slopes, as if they were a ledge (including badguys that usually fall off ledges, e.g. snowballs)
Breaks a bunch of designs to make badguys from a dispenser come upwards out of pipes, e.g. in level 2 of the addon 'Mattie's World'. Instead they get caught between the slope and dispenser and die.

Can be seen in any level with badguys and slopes.

tobbi:
Fixes #761

I don't know if this introduces other bugs, so please test this thoroughly before merging.

### DIFF
--- a/src/badguy/walking_badguy.cpp
+++ b/src/badguy/walking_badguy.cpp
@@ -195,8 +195,9 @@ WalkingBadguy::collision_solid(const CollisionHit& hit)
     if (m_physic.get_velocity_y() > 0) m_physic.set_velocity_y(0);
   }
 
-  if ((hit.left && (m_dir == Direction::LEFT)) || (hit.right && (m_dir == Direction::RIGHT))) {
-    if(hit.slope_normal.x == 0.0f)
+  if ( hit.slope_normal.x == 0.0f &&
+      (hit.left && m_dir == Direction::LEFT) ||
+      (hit.right && m_dir == Direction::RIGHT) ) {
       turn_around();
   }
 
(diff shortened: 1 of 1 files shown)


Rusty-Box:
I does kinda work. Only problem is that smart enemies go up all slopes but refuseto go down >=45° slopes.

MatusGuy:
> I does kinda work. Only problem is that smart enemies go up all slopes but refuseto go down >=45° slopes.

This is unrelated and planned to be fixed in #2812

MatusGuy (approved):

src/badguy/walking_badguy.cpp
```diff
@@ -196,7 +196,8 @@ WalkingBadguy::collision_solid(const CollisionHit& hit)
   }
 
   if ((hit.left && (m_dir == Direction::LEFT)) || (hit.right && (m_dir == Direction::RIGHT))) {
-    turn_around();
+    if(hit.slope_normal.x == 0.0f)
```
Vankata453, line 199: Can you append this check to the `if` above (preferably at the beginning, so the `hit.left` and `hit.right` checks are done afterwards)?

Vankata453 (commented):
Tested this and it works great! Not sure how it could introduce other bugs really - only affects the check of turning around, which is what is being fixed.

PR #2931, opened 2024-04-29
`Canvas::get_pixel`: Use a virtual colorspace

### LINKED ISSUE #1939: Magic Blocks Rublight/Ambient light inconsistency
v0.6.3-beta.1.2-gdb0689f97
win 10

rublights dont function to respond to magic tiles upon light level change. So a rublight can activate a magic tile when its bright, but when the level goes dark it wont activate it.

i think this feature is actually working "as intended" but practically speaking, it makes NO sense. how on earth can light not activate a tile if its dark in the level. surely that gives MORE of a reason for the magic tile to activate??

### LINKED ISSUE #2661: Magic Tiles affected by light intensity
Something I keep noticing is that depending on how dark a sector/ambient light is the smaller the range of affect light has over magic tiles becomes. In my opinion this makes magic tiles quite annoying to deal with and "unfair" in a way as even though the light of a lantern clearly covers a magic tile, it will not appear active.

A magic tile should just ignore the intensity of the light and rather check for just the range of the actual light graphic.

As you can see in the screenshots below, in a darker sector, the light barely covers 3 magic tiles whereas in a brighter sector in covers much more at once!

**Screenshots:**
Sector with 0 light
Sector with 0.5 light

Vankata453:
Virtual colorspaces, consisting of colored shapes, can now be used for the `get_pixel` API, instead of relying on the GPU.

**TODO:**

[ ] Detect, or allow to specify lightsprites as circles with a radius and border radius, making the light check more visually correct. [ ] Make sure the shape of every lightsprite, needed to be detected by the `get_pixel` API, is being added into the `LIGHTSPRITES` colorspace. [ ] More extensive testing.

Fixes #2661. Fixes #1939 (theoretically, should also be tested).

### DIFF
--- a/src/badguy/badguy.cpp
+++ b/src/badguy/badguy.cpp
@@ -170,7 +170,7 @@ BadGuy::draw(DrawingContext& context)
 
       if (m_glowing)
       {
-        m_lightsprite->draw(context.light(), m_col.m_bbox.get_middle(), 0);
+        m_lightsprite->draw(context.light(), m_col.m_bbox.get_middle(), 0, NO_FLIP, ColorSpace::LIGHTSPRITES);
       }
     }
   }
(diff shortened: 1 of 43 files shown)


v1993:
Perhaps call it something other than a color space, given that the term already has a well-established [definition](https://en.wikipedia.org/wiki/Color_space)? Something like a virtual canvas layer, maybe, since it's basically an additional simple CPU-side render target.

PR #2933, opened 2024-05-06
Fix issues with unisolid tiles and frozen enemies

### LINKED ISSUE #2929: Strange interactions with Unisolid Tiles and Frozen Enemies
**SuperTux version:** SuperTux-v0.6.3-1198-g5edb95ede from 2024-04-28 06:12
**System information:** Windows 10, 64bit

##### Expected behavior

+ Behavior 1: Tux can duck and stand up if he wouldn't hit his head on the unisolid tile.

+ Behavior 2: Bouncy Snowballs keep moving normally when inside an unisolid tile hitbox.

+ Behavior 3: Tux stands normally on top of a frozen enemy.

+ Behavior 4: Frozen enemies don't do anything to buttons

##### Actual behavior

+ Behavior 1: Tux can't stand up after ducking/crawling when inside or below an unisolid tile.

+ Behavior 2: Bouncy snowballs change their moving direction when they touch the hitbox of an unisolid tile, and get horizontally stuck in the tile when going inside it from below.

+ Behavior 3: Tux uses the wrong animation when standing on a frozen enemy that is over a solid object that isn't a tile, except for conveyors and invisible walls.

+ Behavior 4: Buttons start flickering when a frozen enemy is placed on it

+ Behavior 1: Duck when Tux is inside the hitbox of an up-facing unisolid tile, then try to stand up again.

+ Behavior 2: Place some bouncy snowballs below some unisolid platforms in a way so that they can reach the unisolid tiles after jumping.

+ Behavior 3: Freeze an enemy, put it over a bonus block, a rock, a brick or a similar object, then stand on the frozen enemy that is on top of this object.

+ Behavior 4: Place a frozen enemy on top of a button.

**[Video with all behaviors being reproduced](https://youtu.be/LKMIWcTcC-k)**
Make sure to pay attention to the controllers on the screen.

Brockengespenst:
Addresses the following topics:
- Tux can duck and stand up if he wouldn't hit his head on the unisolid tile.
- Bouncy Snowballs keep moving normally when inside an unisolid tile hitbox.
- Tux stands normally on top of a frozen enemy.
- Frozen enemies don't do anything to buttons

Fixes #2929


Brockengespenst:
These changes should fix all mentioned issues in #2929.
I changed the collision group of the PushButton to static, I guess this makes more sense. Also it prevents the PushButton to be pushed away a bit by throwing frozen enemies at it. If this behavior was intended, please let me know.

Rusty-Box:
> I changed the collision group of the PushButton to static, I guess this makes more sense. Also it prevents the PushButton to be pushed away a bit by throwing frozen enemies at it. If this behavior was intended, please let me know.

Finally, no more pushing the push(hihi)button around! 🎉

Brockengespenst:
Just one thought: If standing on a frozen enemy - should it be treated like ice such that it is a bit slippery?

Brockengespenst:
Does anyone else want to share an opinion?

Rusty-Box:
> Speaking of this, I have an idea for a little change on the physics of frozen enemies, they could be thrown just like a Mr. ice cube or a snail, but break on impact with a wall.

I do like that. Gives the ice flower even more offensive options.

Rusty-Box (approved):
Looks good to me 👍

Eauix (approved):
Probably a bit too late to comment, but as I opened the issue that this PR fixes, I approve the changes, it works as expected

MatusGuy:
I actually wanted to redesign these is_free_of_x functions to support more edge cases but these bugs are indeed very annoying so fixing them makes sense.

I do remember that ivies also switch to their walking state when touching a unisolid from any angle so you should take a look at that

Brockengespenst:
> I actually wanted to redesign these is_free_of_x functions to support more edge cases but these bugs are indeed very annoying so fixing them makes sense.
> 
> I do remember that ivies also switch to their walking state when touching a unisolid from any angle so you should take a look at that

Maybe as a first step we could merge this PR until your rework/improvements are ready to go.
I can check the Ivies, but earliest mid of next week, I only have access to my mobile right now.

MatusGuy:
> Maybe as a first step we could merge this PR until your rework/improvements are ready to go.

Not worth it, in my opinion. We'd probably be waiting for this pr several months.

Brockengespenst:
> I do remember that ivies also switch to their walking state when touching a unisolid from any angle so you should take a look at that

I have added a check that the floating animation will not be changed to walking, if Ivy is passing a unisolid from below (means moving upward). I guess this is only a rare case (e.g. floating while gravity is negative).
I guess the "normal" cases look good, I may need to check some more sophisticated usecases (I'd be grateful for any provided example).

Brockengespenst:
I think there is one thing about Ivy and unisolids, that is a bit strange: When they hit a unisolid from the side while floating, it might end up in a slightly weird movement until they got below the unisolid. We could overcome this by changing their direction in that case. What do you think about this?

PR #2953, opened 2024-05-22
Fix Tarantula on level flip

MatusGuy:


Rusty-Box (commented):
Well, it flips - however, considering Tarantula is being usually placed on ceilings, it would make more sense for them to move upwards rather than still downwards after the level flips.

I would approach it more like how it is done with Crushers. Their sprite does not flip and they shoot upwards to the ground (now ceiling)

weluvgoatz (approved):

tobbi (approved):
From a code standpoint this looks good. Didn't test it, though.

Hume2:
What are the remaining tasks here?

Hypernoot:
The PR got named after the first commit, so it was confusing. There are more things to be fixed.

MatusGuy:
I will be opening a new PR to fix git issues

PR #2954, opened 2024-05-25
Scale parallax layers perspectively-correctly

Hume2:
I implemented it for both tilemaps and backgrounds. DEMO: https://www.youtube.com/watch?v=GnFArTYBL4E

Scaling now works so that it imitates a perspective camera. The layers scale differently based on their scrolling speed.

Also note that the z-position based on which the scaling is calculated, does not depend on the "z-pos" value at all, so it is nicely compatible with the old levels.

It also fixes the bug which allows players to see around paralax layers when the viewport is scaled down. If the layer is large enough to cover the level in the default scale, it now works in all scales.

### DIFF
--- a/src/object/background.cpp
+++ b/src/object/background.cpp
@@ -334,7 +334,6 @@ Background::draw_image(DrawingContext& context, const Vector& pos_)
   const int end_y   = static_cast<int>(ceilf((cliprect.get_bottom() - (pos_.y + img_h/2.0f)) / img_h)) + 1;
 
   Canvas& canvas = context.get_canvas(m_target);
-  context.set_flip(context.get_flip() ^ m_flip);
 
   if (m_fill)
   {
@@ -407,7 +406,6 @@ Background::draw_image(DrawingContext& context, const Vector& pos_)
         break;
     }
   }
-  context.set_flip(context.get_flip() ^ m_flip);
 }
 
 void
@@ -418,6 +416,14 @@ Background::draw(DrawingContext& context)
 
   if (!m_image)
     return;
+    
+  context.push_transform();
+  if (!context.perspective_scale(m_parallax_speed.x, m_parallax_speed.y)) {
+    //The background is placed behind the camera.
+    context.pop_transform();
+    return;
+  }
+  context.set_flip(context.get_flip() ^ m_flip);
 
   Sizef level_size(d_gameobject_manager->get_width(),
                    d_gameobject_manager->get_height());
@@ -431,6 +437,7 @@ Background::draw(DrawingContext& context)
              level_size.height / 2);
   draw_image(context, pos + m_scroll_offset + Vector(center_offset.x * (1.0f - m_parallax_speed.x),
                                                      center_offset.y * (1.0f - m_parallax_speed.y)));
+  context.pop_transform();
 }
 
 namespace {
--- a/src/object/background.cpp
+++ b/src/object/background.cpp
@@ -316,8 +316,7 @@ void
 Background::draw_image(DrawingContext& context, const Vector& pos_)
 {
   const Sizef level(d_gameobject_manager->get_width(), d_gameobject_manager->get_height());
-  const Sizef screen(context.get_width(),
-                     context.get_height());
+  const Sizef screen = context.get_viewport().get_size();
   const Sizef parallax_image_size((1.0f - m_parallax_speed.x) * screen.width + level.width * m_parallax_speed.x,
                                   (1.0f - m_parallax_speed.y) * screen.height + level.height * m_parallax_speed.y);
 
@@ -427,8 +426,7 @@ Background::draw(DrawingContext& context)
 
   Sizef level_size(d_gameobject_manager->get_width(),
                    d_gameobject_manager->get_height());
-  Sizef screen(context.get_width(),
-               context.get_height());
+  Sizef screen = context.get_viewport().get_size();
   Sizef translation_range = level_size - screen;
   Vector center_offset(context.get_translation().x - translation_range.width  / 2.0f,
                        context.get_translation().y - translation_range.height / 2.0f);
--- a/src/object/background.cpp
+++ b/src/object/background.cpp
@@ -424,15 +424,15 @@ Background::draw(DrawingContext& context)
   }
   context.set_flip(context.get_flip() ^ m_flip);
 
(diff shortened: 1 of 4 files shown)


src/video/drawing_context.hpp
```diff
@@ -77,6 +77,9 @@ class DrawingContext final
 
   float get_scale() const { return transform().scale; }
   void scale(float scale) { transform().scale *= scale; }
+  
+  /** Recalculates the scaling factor for paralax layers.*/
```
tobbi, line 81: Typo: parallax

Hume2: corrected, thanks ^.^

src/object/tilemap.cpp
```diff
@@ -438,9 +438,18 @@ TileMap::draw(DrawingContext& context)
 {
   // skip draw if current opacity is 0.0
   if (m_current_alpha == 0.0f) return;
-
+
```
mrkubax10, line 441: Remove these spaces.

src/object/tilemap.cpp
```diff
@@ -438,9 +438,18 @@ TileMap::draw(DrawingContext& context)
 {
   // skip draw if current opacity is 0.0
   if (m_current_alpha == 0.0f) return;
-
+  
   context.push_transform();
 
+  const bool normal_speed = m_editor_active && Editor::is_active();
+  float speed_x = normal_speed ? 1.0f : m_speed_x;
+  float speed_y = normal_speed ? 1.0f : m_speed_y;
```
mrkubax10, line 446: These probably could be marked as `const`.

Hume2: I fixed both, thanks ^.^

src/video/drawing_context.cpp
```diff
@@ -132,4 +132,27 @@ DrawingContext::get_size() const
...
+    //Trivial or unreal situation: Do not apply perspective.
+    return true;
+  }
+  const float speed = sqrt(speed_x*speed_y);
+  if (speed == 0) {
+    //Special case: The object appears to be infinitely far.
+    tfm.scale = 1.0;
+    return true;
+  }
+  const float t = tfm.scale*(1/speed - 1) + 1;
```
MatusGuy, line 149: Adding spaces in between operators makes the expression more readable
```suggestion
  const float t = tfm.scale * (1 / speed - 1) + 1;
```

src/video/drawing_context.cpp
```diff
@@ -132,4 +132,27 @@ DrawingContext::get_size() const
...
 
+bool
+DrawingContext::perspective_scale(float speed_x, float speed_y)
+{
+  DrawingTransform& tfm = transform();
+  if (tfm.scale == 1 || speed_x < 0 || speed_y < 0) {
+    //Trivial or unreal situation: Do not apply perspective.
+    return true;
+  }
+  const float speed = sqrt(speed_x*speed_y);
```
MatusGuy, line 143:
```suggestion
  const float speed = sqrt(speed_x * speed_y);
```

src/object/tilemap.cpp
```diff
@@ -453,9 +462,7 @@ TileMap::draw(DrawingContext& context)
 
   const float trans_x = context.get_translation().x;
   const float trans_y = context.get_translation().y;
-  const bool normal_speed = m_editor_active && Editor::is_active();
-  context.set_translation(Vector(trans_x * (normal_speed ? 1.0f : m_speed_x),
-                                 trans_y * (normal_speed ? 1.0f : m_speed_y)));
+  context.set_translation(Vector(trans_x*speed_x, trans_y*speed_y));
```
MatusGuy, line 465:
```suggestion
  context.set_translation(Vector(trans_x * speed_x, trans_y * speed_y));
```

src/video/drawing_context.cpp
Hume2: Well, when the spaces are between all operators, it is actually less readable IMO. But I can fix that if that is the code style.

MatusGuy (approved):
Untested

Rusty-Box:
Could you sync your branch so I can make use of the portable zip for windows builds please? Makes it more convenient for me to test

Rusty-Box (approved):
Love this

Frostwithasideofsalt (approved):

src/object/background.cpp
```diff
@@ -418,11 +415,18 @@ Background::draw(DrawingContext& context)
...
+    context.pop_transform();
+    return;
+  }
+  context.set_flip(context.get_flip() ^ m_flip);
 
   Sizef level_size(d_gameobject_manager->get_width(),
                    d_gameobject_manager->get_height());
-  Sizef screen(context.get_width(),
-               context.get_height());
+  Sizef screen = context.get_viewport().get_size();
```
tobbi, line 429: Can this be a reference?

src/object/background.cpp
```diff
@@ -316,8 +316,7 @@ void
 Background::draw_image(DrawingContext& context, const Vector& pos_)
 {
   const Sizef level(d_gameobject_manager->get_width(), d_gameobject_manager->get_height());
-  const Sizef screen(context.get_width(),
-                     context.get_height());
+  const Sizef screen = context.get_viewport().get_size();
```
tobbi, line 319: Can this be a reference as well?

src/object/background.cpp
Hume2: `DrawingContext::get_viewport` returns an instance of `Rect&`.
`Rect::get_size` returns an instance of `Size`, which is not a reference type.
`Rect::get_size` cannot be modified to return an instance of `Size&`, because the result is calculated on demand and is not stored within the class.

src/object/background.cpp
Hume2: same as above

src/object/background.cpp
mrkubax10: `screen` can be `const` though.

tobbi: const ref, right?

mrkubax10: Hume explained why it cannot be a reference above.

tobbi: true, I apparently can't keep focus.

Hume2: I made it const, together with a few more variables.

PR #2958, opened 2024-05-30
Cache nfo filename locations

tobbi:
Well, I thought this would save some startup time but the effect is less than I thought it would be.

Anyway, instead of iterating over all files in each add-on to find the necessary .nfo file, I decided to cache the name of the file for each add-on in the config and then directly ask whether an nfo file can be found at that specific location.

Maybe if you have 1k add-ons installed this will make a noticeable difference.

### DIFF
--- a/src/addon/addon_manager.cpp
+++ b/src/addon/addon_manager.cpp
@@ -141,13 +141,15 @@ static PHYSFS_EnumerateCallbackResult remove_from_dictionary_path(void *data, co
 } // namespace
 
 AddonManager::AddonManager(const std::string& addon_directory,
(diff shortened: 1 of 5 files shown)


tobbi:
Well, there's a lot of add-ons that don't follow that scheme so, thinking of our friend backwards compatibility, I think we should account for that.

Vankata453:
Then what about if, for example, an `info.nfo` file is not found, it goes looking for other `.nfo` files? With time, the support for `.nfo` files with other names can be removed, but deprecated for now. The in-game add-on creator in the editor can also be adapted to use the set name.

Issue #2318, opened 2022-10-16
Tower Of Ghosts: sometimes strange timing enemy timing


erichstuder:
**SuperTux version:** 0.6.2
**System information:** Debian 11, 64 bit

In the level Toser Of Ghosts the two enemies seen in the picture sometimes have strange jumping timing making it impossible to pass them by without shooting them. The timing changes from run to run.
I would expect the timing to be always the same an in a way to pass them without shooting them.

![Screenshot from 2022-10-16 18-49-57](https://user-images.githubusercontent.com/28953680/196047867-58e311b3-5959-4773-9c70-1ac26edc8da5.png)

By the way: Great game! Can't wait for the next version!

Regards
Erich

bruhmoent:
This is version 0.6.2. The latest non nightly release is 0.6.3. This issue is irrelevant anyways as the levels are getting revamped.

mrkubax10:
> The timing changes from run to run.
I would expect the timing to be always the same an in a way to pass them without shooting them.

Inconsistency probably comes from the fact that objects are loaded at different times depending on how and where player moves since there is something like "active area" inside of which objects are updated and outside not.

Semphriss:
@erichstuder In short, this is an engine problem. Basically, everything that is too far from the player is "frozen in time" to avoid calculating their movements, which saves some CPU time and allows making big levels without lag. That causes a desynchronization of the enemies depending on where you are in the level over time.

There's no easy way to avoid this, unfortunately, except restarting the level and hoping that the enemies are properly synchronized.

The good news is, this level is known to be excessively difficult, and will be replaced in the future.

In the meantime, perhaps the engine should be adapted to account for this? (@ devs)

Alasdairbugs:
This is due to loading the objects at certain times. Level design issue. Fixed in upcoming update, which isnt merged in atm.

PR #2964, opened 2024-06-02
Add badguy property to activate/deactivate badguys in groups

mstoeckl:
Putting two jumpy badguys in the same group will ensure they are activated/deactivated at the same time; as a result they will maintain synchronization, no matter how quickly Tux approaches them.

Related level design issue: https://github.com/SuperTux/supertux/issues/2318 .

I'm not certain this is the best approach to solve the synchronization problem, but it is the simplest thing I could think of.

I've only edited one level ("Above the Arctic Skies") to use this feature, and can edit more if people wish. (Although where exactly to use this feature is a matter of taste.) I've only tested this with jumpy enemies and floating snowballs so far; while I don't expect other badguy types to break with this, I don't know the code well enough to be certain. Names are hard to change after the fact, so suggestions for improvements are welcome.

### DIFF
--- a/data/levels/world1/above_arctic_skies.stl
+++ b/data/levels/world1/above_arctic_skies.stl
@@ -116,11 +116,13 @@
     )
     (flyingsnowball
       (direction "left")
+      (load-group "stair")
       (x 6880)
       (y 992)
     )
     (flyingsnowball
       (direction "left")
+      (load-group "stair")
       (x 6816)
       (y 896)
     )
@@ -141,16 +143,19 @@
     )
     (flyingsnowball
       (direction "left")
+      (load-group "final")
       (x 13984)
       (y 912)
     )
     (flyingsnowball
       (direction "left")
+      (load-group "final")
       (x 14144)
       (y 960)
     )
     (flyingsnowball
       (direction "left")
+      (load-group "final")
       (x 13824)
       (y 864)
     )
@@ -192,21 +197,25 @@
     )
     (jumpy
       (direction "left")
+      (load-group "jpair1")
       (x 11616)
       (y 768)
     )
     (jumpy
       (direction "left")
+      (load-group "jpair1")
       (x 11680)
       (y 768)
     )
     (jumpy
       (direction "left")
+      (load-group "jpair2")
       (x 11920)
       (y 768)
     )
     (jumpy
       (direction "left")
+      (load-group "jpair2")
       (x 11984)
       (y 768)
     )


src/badguy/badguy.hpp
```diff
@@ -234,7 +240,9 @@ class BadGuy : public MovingSprite,
   void set_colgroup_active(CollisionGroup group);
 
 private:
-  void try_activate();
+  bool may_deactivate();
+  bool may_activate();
```
tobbi, line 244: This can also be marked as `const`.

src/supertux/game_object_manager.cpp
```diff
@@ -499,6 +517,21 @@ GameObjectManager::this_before_object_remove(GameObject& object)
     assert(it != vec.end());
     vec.erase(it);
   }
+
+  {
+    // By badguy load group
+    BadGuy *bg = dynamic_cast<BadGuy*>(&object);
```
tobbi, line 523: same here about the `auto` thing.

src/badguy/badguy.hpp
```diff
@@ -234,7 +240,9 @@ class BadGuy : public MovingSprite,
   void set_colgroup_active(CollisionGroup group);
 
 private:
-  void try_activate();
+  bool may_deactivate();
```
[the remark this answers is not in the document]
mstoeckl: Yes, good point.

src/badguy/badguy.hpp
mstoeckl: Done.

src/supertux/game_object_manager.cpp
```diff
@@ -473,6 +480,17 @@ GameObjectManager::this_before_object_add(GameObject& object)
     m_objects_by_type_index[std::type_index(typeid(object))].push_back(&object);
   }
 
+  {
+    // By badguy load group
+    BadGuy *bg = dynamic_cast<BadGuy*>(&object);
```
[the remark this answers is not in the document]
mstoeckl: Updated.

src/supertux/game_object_manager.cpp
mstoeckl: Updated.

mstoeckl:
1. I've rebased this to fix a merge conflict. This PR is not very important and is somewhat experimental, so I may forget about future conflicts; please reply if you ever want me to rebase it.

3. Looking at most of the existing levels, I only see a few places where load-group would be helpful, probably because existing levels have been designed to avoid sections whose difficulty depends on loading patterns. (Also, features that might need load-group (dense groups of enemies, bundles of flames, steep flying snowball step ascents, and neighboring jumpies) generally reduce the number of safe spots and force the player to keep moving, which would make levels unusually hard for world1/world2.)

PR #2969, opened 2024-06-11
Update yeti behavior

### LINKED ISSUE #979: Restore Yeti's "throwing bouncing snowball" attack
In an older version of SuperTux (I believe it was v0.3.0) the Yeti had a second attack besides stomping. Throwing a few bouncing snowballs when he's on either side of the boss arena. This attack should be restored to make the first boss a bit more challenging. As an additional idea would be that this attack only activates once the yeti has only 2 or 3 lives left making it some sort of "special attack".

### LINKED ISSUE #1083: Pinch Mode for boss fights
An idea for making boss fights in SuperTux more interesting and more challeniging during the final hits would be adding pinch modes for boss enemies (e.g. yeti, ghost tree and future bosses).

In their pinch mode bosses could get more agressive by either getting faster, using a new attack or having existing attacks get more difficult than before, similar to how it is done with the stalactite attack pattern of the yeti.
The pinch mode could be initiated after a boss only has 40% or less of their total health points left (e.g. A yeti with 5 health points initiates his pinch mode after having only 2 or less health points left).

weluvgoatz:
This PR contains probably one of the final big features for 0.7.0, which is the revamp of the yeti, to fit the standards set in place by the other aspects of the game.  Here is a todo list:

- [x] Add pinch mode
- [x] Put the lives, display HUD and pinch mode activation behavior into a general "Boss" class that the Yeti and Ghost Tree inherit from
- [x] Revamp Yeti for code quality
- [x] Can't spam hit Yeti, it grabs and throws you away
- [x] Re-add yeti snowball throwing

Discussion for this PR can happen here, on Discord, or IRC, since the new actions of the bosses are up to a lot of interpretation right now. 

Fixes #1083 
Fixes #979


Frostwithasideofsalt (approved):

bruhmoent (approved):
Works well, but remember to put opening curly brackets in a new line.

src/badguy/yeti.cpp
```diff
@@ -334,19 +448,27 @@ Yeti::collision_solid(const CollisionHit& hit)
           // Go to the other side after 3 jumps.
           if (m_stomp_count == 3)
           {
-            jump_down();
+            m_just_hit = false;
+            if (m_pinch_mode) {
```
tobbi, line 452: Maybe opening brace in a new line.

src/badguy/yeti.cpp
```diff
@@ -334,19 +448,27 @@ Yeti::collision_solid(const CollisionHit& hit)
...
+              throw_big_snowballs();
+            }
+            else {
```
tobbi, line 455: ditto.

tobbi (approved):
Just looked through this, I think looking good.

PR #2970, opened 2024-06-12
Only call glGetUniformLocation/glGetAttribLocation once

mstoeckl:
Quoting the commit message:

> These values never change after a program is linked, so it is safe to cache the uniform and attribute locations. In addition to the obvious benefit (fewer function calls), calling glGet-type functions inhibits OpenGL driver optimizations: if there is a mixture of setting/getting commands, drivers cannot do batch optimizations where they accumulate a bunch of commands and process them all of them at once, since it is possible one command could affect the output of a glGet-type function, whose results are needed immediately.

The last glGet-type function that SuperTux calls a lot is `glGetError()`; however, I have not removed it, because it is useful when developing and debugging code. I believe there are more efficient ways to debug OpenGL (see https://www.khronos.org/opengl/wiki/Debug_Output ) than glGetError, but a) haven't looked into them in detail b) think this is already a strict improvement.

This change reduces the number of instructions that I see executed when running the title screen for ten seconds. Running `perf stat -- timeout 10 supertux2` a few times to estimate the amount of CPU work done, I measured:
* without this patch: ~9.4+/-0.1 billion cycles, and 14.4+/-0.1 billion instructions
* with this patch: ~8.6+/-0.1 billion cycles, and 13.3+/-0.1 billion instructions

### DIFF
--- a/src/video/gl/gl_program.hpp
+++ b/src/video/gl/gl_program.hpp
@@ -34,19 +34,43 @@ public:
 
   GLuint get_handle() const { return m_program; }
 
-  GLint get_attrib_location(const char* name) const;
-  GLint get_uniform_location(const char* name) const;
+  GLint get_backbuffer_location() const { return check_valid(m_backbuffer_location, "backbuffer"); }
+  GLint get_fragcoord2uv_location() const { return check_valid(m_fragcoord2uv_location, "fragcoord2uv"); }
+  GLint get_diffuse_texture_location() const { return check_valid(m_diffuse_texture_location, "diffuse_texture"); }
+  GLint get_displacement_texture_location() const { return check_valid(m_displacement_texture_location, "displacement_texture"); }
+  GLint get_framebuffer_texture_location() const { return check_valid(m_framebuffer_texture_location, "framebuffer_texture"); }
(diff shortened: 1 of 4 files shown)


src/video/gl/gl_program.hpp
```diff
@@ -34,20 +34,44 @@ class GLProgram final
 
   GLuint get_handle() const { return m_program; }
 
-  GLint get_attrib_location(const char* name) const;
-  GLint get_uniform_location(const char* name) const;
+  GLint get_backbuffer_location() const { return check_valid(m_backbuffer_location, "backbuffer"); }
+  GLint get_fragcoord2uv_location() const { return check_valid(m_fragcoord2uv_location, "fragcoord2uv"); }
+  GLint get_diffuse_texture_location() const { return check_valid(m_diffuse_texture_location, "diffuse_texture_"); }
```
tobbi, line 39: Is the underscore after `diffuse_texture` a typo or not?

mstoeckl: It is a typo; fixed.

tobbi:
Can you rebase against latest master? Should fix the windows builds.

tobbi (approved):

mrkubax10 (approved):

PR #2978, opened 2024-06-16
Support pre-compiled headers and interprocedural/link-time optimisation

Semphriss:
This adds CMake code to enable [pre-compiled headers](https://en.wikipedia.org/wiki/Precompiled_header) (PCH) and [interprocedural optimisation](https://en.wikipedia.org/wiki/Interprocedural_optimization) (IPO, or LTO for link-time optimisation).

## PCH

PCH makes builds 50% to 66% faster. A simple `time make -j$(nproc)` on my 12-core CPU shows:
```
With PCH:
real    0m58.048s
user    5m33.425s
sys    1m56.666s

Without PCH:
real    1m50.533s
user    15m20.049s
sys    2m56.815s
```

`real` is the total time spent, from start to finish, as perceived by the developer. `user` is the sum of the time spent by each thread, which gives a better idea of what would happen on a single-threaded system. [Details](https://stackoverflow.com/questions/556405/what-do-real-user-and-sys-mean-in-the-output-of-time1)

## IPO

IPO makes Release builds 15% smaller. I have not tested the runtime performance.

---

### Notes

- As instructed, I've opened a single PR for both of these.
- For PCH to work properly, I had to add some missing include guards and include headers in some files.
- `cmake_policy(SET CMP0069 NEW)` requires CMake >=3.9, but I did not update `cmake_minimum_required` to discuss whether or not to keep the policy.


tobbi:
I tested this a couple days ago. Seems to be working fine.

tobbi (approved):

CMakeLists.txt
```diff
@@ -63,6 +64,23 @@ include(ExternalProject)
 include(CheckCXXCompilerFlag)
 include(CheckSymbolExists)
 
+if(EMSCRIPTEN)
+  # FIXME: As of writing, Emscripten fails with "wasm-ld: error: /path/to/emsdk/upstream/emscripten/cache/sysroot/lib/wasm32-emscripten/thinlto/libc.a(fileno.o): attempt to add bitcode file after LTO (fileno)"
+  # Seems to be a known issue: https://github.com/emscripten-core/emscripten/issues/20275
+  option(SUPERTUX_IPO "Use interprocedural optimisation (Takes more RAM at compilation, gives smaller and faster builds)" OFF)
+else()
+  option(SUPERTUX_IPO "Use interprocedural optimisation (Takes more RAM at compilation, gives smaller and faster builds)" ON)
```
HybridDog, line 72: Does "faster builds" mean the compiled program is faster or the compilation is faster?

src/editor/scroller_widget.cpp
```diff
@@ -24,7 +24,7 @@
 #include "video/video_system.hpp"
 #include "video/viewport.hpp"
 
-namespace {
+namespace CONSTS {
 
 const float TOPLEFT = 16;
 const float MIDDLE = 48;
```
HybridDog, line 30: Why is it necessary to name the namespace?
And could the `const`s be changed to `constexpr`s?

HybridDog:
This sounds like enabling pre-compiled headers has only advantages and no disadvantage. 
If a developer edits a single header file and recompiles the program, is the recompilation faster or slower with pre-compiled headers enabled?

CMakeLists.txt
Semphriss: Compilation is faster, 50-66% less time on my system. I've hesitated to put an explanation at all, but I thought it might help less familiar people understand the reasons. I'm open to changing the wording if necessary.

src/editor/scroller_widget.cpp
Semphriss: Ah, I should've documented that.

MSVC gave [build errors](https://productionresultssa0.blob.core.windows.net/actions-results/2a8be01b-b1a0-431c-988a-83f65d3719bf/workflow-job-run-58628982-a7ad-5965-3a32-4c1b2d88b7fa/logs/job/job-logs.txt?rsct=text%2Fplain&se=2024-06-27T20%3A50%3A18Z&sig=Wnfv1buWUDTPyc4LXnuIHuTjadufTadLjsILbXGxeu4%3D&ske=2024-06-28T06%3A51%3A57Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2024-06-27T18%3A51%3A57Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2023-11-03&sp=r&spr=https&sr=b&st=2024-06-27T20%3A40%3A13Z&sv=2023-11-03) on [390137b](https://github.com/SuperTux/supertux/pull/2978/commits/390137b739a37edcdf101357a9a77a7b859fd960) about `SIZE` being an ambiguous symbol, probably because a similarly-named variable is defined in some header. Adding the namespace level prevents the symbol clash and makes MSVC happy.

As for `constexpr`, it's most likely possible, but I haven't checked in detail.

Semphriss:
@HybridDog It's something I've been thinking about, whether or not to enable them by default. They're not really relevant during development, only when making releases, which is IMO the preferred place for lesser-known options to live.

I haven't tried actively developing with those options enabled, but I would presume, especially with LTO, that it substantially slows down development compiling. I'm not sure how large the impact is, and active developers would be better placed to report their experience with that.

CMakeLists.txt
HybridDog: "gives smaller and faster builds" sounds ambiguous to me because "builds" can refer to the output of the build process or the build process itself, as far as I know.
Maybe "reduces program size and compilation time" can be used instead if the statement is true.

src/editor/scroller_widget.cpp
HybridDog: const values still have internal linkage even if they are not in an anonymous namespace, so constexpr and const have the same effect here I think.
https://en.cppreference.com/w/cpp/language/cv#Notes
https://stackoverflow.com/a/33402258

HybridDog:
If they are only relevant for release builds, maybe they can be enabled if and only if the build type is Release or RelWithDebInfo (-DRELEASE).

CMakeLists.txt
Semphriss: My apologies, I confused PCH and IPO. PCH is what makes compilation much faster, LTO takes more time to compile and gives better binaries, both in terms of runtime speed and executable size.

Semphriss: I've clarified the wording with the latest commit.

Semphriss:
PCH and IPO are now only enabled on Release|RelWithDebInfo builds.

HybridDog (approved):

src/port/emscripten.hpp
```diff
@@ -38,60 +41,8 @@ void onDownloadError(int id);
...
 } // extern "C"
 
-void
-init_emscripten()
-{
-  EM_ASM({
-    if (window.supertux_onready)
-      window.supertux_onready();
-  }, 0); // EM_ASM is a variadic macro and Clang requires at least 1 value for the variadic argument
-}
```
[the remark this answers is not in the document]
Semphriss: Yes, I've moved them. Thanks for the catch!

MatusGuy:
We will cherry-pick this into #3093

PR #2980, opened 2024-06-20
Make "[DISABLED]" translatable

Narre:


src/supertux/menu/addon_menu.cpp
```diff
@@ -121,13 +121,13 @@ AddonMenu::rebuild_menu()
...
     const std::string text = addon_string_util::generate_menu_item_text(addon);
-    add_entry(MAKE_UPDATE_MENU_ID(index), fmt::format(fmt::runtime(_("{} {}*UPDATE*")), text, !addon.is_enabled() ? "[DISABLED] " : ""));
+    add_entry(MAKE_UPDATE_MENU_ID(index), fmt::format(fmt::runtime(_("{} {}*UPDATE*")), text, !addon.is_enabled() ? (_("[DISABLED]") + " ") : ""));
   }
   for (const auto& index : addons_to_list)
   {
     const Addon& addon = m_addon_manager.get_installed_addon(m_installed_addons[index]);
     const std::string text = addon_string_util::generate_menu_item_text(addon);
-    add_entry(MAKE_INSTALLED_MENU_ID(index), fmt::format(fmt::runtime(_("{}{}")), text, !addon.is_enabled() ? " [DISABLED]" : ""));
+    add_entry(MAKE_INSTALLED_MENU_ID(index), fmt::format(fmt::runtime(_("{}{}")), text, !addon.is_enabled() ? (" " + _("[DISABLED]")) : ""));
```
HybridDog, line 130: The `_("{}{}")`, which was already here before the changes, looks wrong to me. Does it translate the string `{}{}` before substitutions are applied?

tobbi: It doesn't make sense for that to have translation markers. It doesn't have any translatable content.

Narre: I'll fix it tomorrow.

HybridDog: Nitpick question: Is the `fmt::runtime` still needed if the string is known at compile time?

HybridDog (approved):

src/supertux/menu/addon_menu.cpp
tobbi: Honestly, it should be "{} [disabled]" as translated string, because some rtl locales need the add-on name on the right.

tobbi: Honestly, it should be "{} [disabled]" as translated string, because some rtl locales need the add-on name on the right.

mrkubax10 (approved):

tobbi (approved):
My OCD kicks in with so much near-code-duplication.

But looking good.

PR #2985, opened 2024-06-24
ST OST - Two forest songs 

Curly68:
Here are two new forest songs for Supertux 0.7 that I've made lately.

- forest_new.ogg - A new version of the old forest theme.
- forest_new_alternative.ogg - An alternative version, meant for more fast paced levels.

These two songs are based on forest.ogg and forest3.ogg by Wantsi, 
However they also have some original songwritting in between.

I will continue making new songs for the forest world as soon as i can.
Sorry for my 3 month absence


Rusty-Box (commented):
For consistency, I'd recommend naming these two files midday and midday_alt. Other than that everything looks good 👍

Curly68:
The problem is that there are already two existing "midday" and "midday_alt" songs located in the antartic music folder, which could lead to some confussion in for example the AUTHORS file
Maybe "forest_midday" and "forest_midday_alt" would be better?

Rusty-Box:
> The problem is that there are already two existing "midday" and "midday_alt" songs located in the antartic music folder, which could lead to some confussion in for example the AUTHORS file Maybe "forest_midday" and "forest_midday_alt" would be better?

Good catch ☝️  Yeah, let's call them forest_midday and forest_midday_alt instead.

Curly68:
Ok, i think that it should be all good by now.

Rusty-Box (approved):

weluvgoatz (approved):

mrkubax10 (approved):
This is very good.

PR #2987, opened 2024-06-26
Fix windows crash viewer

MatusGuy:


NickerAban:
finally!

MatusGuy:
wtfreaky

MatusGuy:
@tobbi How can I ignore the cppcheck errors?

tobbi:
Can you not simply fix them? Add a check for the path being empty?

MatusGuy:
I still need to pass `&path[0]` which shouldn't cause issues anyway because I already reserve MAX_PATH bytes

tobbi:
I think it can still be empty

MatusGuy:
The contents of the string are empty but the size isn't. path[0] would probably just give a null character

tobbi:
14:20:39]  Tobbi: If I have an std::string and call reserve on it, can &path[0] still be out of bounds if the string itself is empty?
[14:21:00]  * tsimpson has disconnected (Quit: https://quassel-irc.org - Chat comfortably. Anywhere.)
[14:24:37]  * tsimpson has joined
[14:32:43]  PJBoy: yeah
[14:32:58]  PJBoy: well
[14:33:05]  PJBoy: you can take the address just fine
[14:33:30]  PJBoy: and if you make the string non-empty without exceeding the capacity you reserved, that address will become dereferencable

MatusGuy:
Alright then how do I check if the string is invalid before passing it into GetModuleFileName (the function that fills the string)?

tobbi:
Just check the length: https://cplusplus.com/reference/string/string/length/

MatusGuy:
If GetModuleFileName fills the string then won't checking for the length of the string always reach the same result?

tobbi:
Just check whether it's greater than 0, to silence the warning.

PR #2990, opened 2024-06-28
Always keep dispensers active

### LINKED ISSUE #2837: Dispensers scripting is broken!
Example script: `Dispenser.deactivate();`

When you run the deactivate script for dispensers and they are not loaded in the level (i.e. being close to them) when that script runs they will ignore the script and remain activated. However, if they are loaded in during the time the script runs it will work but unload them and load them in again (i.e. going far enough away and come back), the dispenser is activated again.

I have the theory that it acts this way because Dispenser might be coded as badguys, as moving sprites such as Platform objects never unload so never run into that issue with scripting

weluvgoatz:
Fixes #2837 

This PR makes dispensers active in terms of being an active/inactive badguy always.  This is beneficial for many reasons:
1) Dispensers don't really function as badguys anyway, moreso as other objects, which are always active.
2) It fixes the bug where dispensers scripting would break if you went away from the dispenser and then came back.
3) I believe it fixes an unreported issue where gravity-obeying dispensers would fall into the void if they were placed on moving platforms because the dispenser would load below/inside/off the moving platform because the platform was moving and the dispenser wasn't.

However, I know this could be considered kind of a cheap/non-optimal solution to the bug.  So if anyone has any better suggestions feel free to take a look at it.

### DIFF
--- a/src/badguy/badguy.cpp
+++ b/src/badguy/badguy.cpp
@@ -220,7 +220,7 @@ BadGuy::update(float dt_sec)
   }
 
   // Deactivate badguy, if off-screen and not falling down.
-  if (m_is_active_flag && is_offscreen() && m_physic.get_velocity_y() <= 0.f)
+  if (m_is_active_flag && is_offscreen() && m_physic.get_velocity_y() <= 0.f && !always_active())
   {
     deactivate();
     set_state(STATE_INACTIVE);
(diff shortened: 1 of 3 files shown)


tobbi (approved):
Trivial shit.

MatusGuy (approved):

Issue #1682, opened 2021-03-03
The game is very slow when there are many angrystones


HybridDog:
**SuperTux version:** latest (40ab22f63)
**System information:** Lubuntu 18.04

##### Expected behavior

Reasonable frame rate while there are many enemies 

##### Actual behavior

In a level in Oliver W2, many angrystones spawn and SuperTux becomes unplayable (less than 4 FPS).
![many_emies](https://user-images.githubusercontent.com/3192173/109843275-2def9700-7c4b-11eb-9711-288de6daf7eb.jpg)
Here's the (modified) level: [level26.stl.zip](https://github.com/SuperTux/supertux/files/6077768/level26.stl.zip)

##### Steps to reproduce actual behavior

Download and play the level

Semphriss:
There are 43 cannons in this level (23 on the right and another 20 divided in sub-sections on the left), each of them firing at a rate of one enemy each 0.3 or 0.4 seconds.

After a quick calculation and 10 seconds of gameplay, you have around 1300 enemies in the level - counting only those badguys, excluding the tilemaps and other already existing objects. All of these need to be updated, animated, and drawn 60 times per second.

Adding to this an optimization behavior of the game: Enemies that are too far away from the camera automatically stop being updated and drawn. If you stay far away of the cannons for a bit and suddenly come back in, you have all the badguys that accumulated at the edge of the updater limit that resume being updated all at once. Considering the wall on the left where badguys are supposed to crash is beyond that limit when you jump towards the right, they will accumulate with each second of gameplay, resulting quickly in tens of thousands of badguys after mere minutes of gameplay.

I don't think it is reasonable to expect the game to deliver such capacity at a framerate of 60 FPS. I'd label this as a level design issue and recommend not putting that many enemies at once in a level, especially in such a small zone.

HybridDog:
I'm not the creator of this level; I'm wondering if it worked well at the time when the author created it.
The frame rate drops below 30 FPS already before 6 s, i.e. when there are less than 1300 enemies. In my opinion the calculations for 1300 enemies should take less than 16 ms. Perhaps each object is compared to all other objects somewhere in the code (which leads to quadratic run time).

Semphriss:
> Perhaps each object is compared to all other objects somewhere in the code (which leads to quadratic run time).

You reminded me of something:
src/video/canvas.cpp, lines 51 to 61:
```cpp

void
Canvas::render(Renderer& renderer, Filter filter)
{
[...]

PR #2991, opened 2024-06-30
Index by base classes in GameObjectManager::m_objects_by_type_index

mstoeckl:
Previously, the `m_objects_by_type_index` map only listed the objects by their (main) class. Finding all objects inheriting a given base class (like BadGuy, Portable, or PathObject) required scanning through the entire list of objects and trying a `dynamic_cast` on each. This PR changes the map to, for a given GameObject, record it in `m_objects_by_type_index` for all its notable intermediate base classes. Thus, for example, a Lantern is registered in the `m_objects_by_type_index` map for the lists for type indexes corresponding to Lantern, Rock, Portable, MovingSprite, and MovingObject. Now, one can efficiently look up all e.g. Portable objects by querying `m_objects_by_type_index[std::type_index(typeid(Portable))]`, and this will be fast even if the level has 10000 unrelated objects.

In practice, this PR significantly speeds up processing for the Player and Crusher classes, which no longer need to search through many unrelated objects when looking for e.g. other players, Bullets, Portables, Bricks, etc. (A test level that I made with many crushers and about 1000 coins, which used to lag, now runs quickly.)

The awkward part of this approach is identifying the various base classes of an object when it is time to register them in `GameObjectManager::this_before_object_add()`. This is done by adding a new virtual function to `GameObject`, which when called provides all the classes under which to record an object. Every class inheriting GameObject must then implement the method to add itself to its own list. (This is admittedly tedious, but the method body is only a very formulaic line.)

This supersedes but does not conflict with https://github.com/SuperTux/supertux/pull/2986 .

### DIFF
--- a/src/supertux/game_object.hpp
+++ b/src/supertux/game_object.hpp
@@ -22,6 +22,7 @@
 #include <algorithm>
 #include <string>
 #include <vector>
+#include <typeindex>
 
 #include "editor/object_settings.hpp"
 #include "supertux/game_object_component.hpp"
@@ -46,6 +47,23 @@ struct GameObjectType
   const std::string name;
 };
 typedef std::vector<GameObjectType> GameObjectTypes;
+
+/**
+ A helper structure to list all the type_indexes of the classes in the
+ type hierarchy of a given class. This makes it easier to register e.g.
+ a MrIceblock in lists for MrIceBlock, WalkingBadguy, Badguy, Portable,
+ MovingSprite, MovingObject, and GameObject.
+ */
+struct GameObjectClasses
+{
+  std::vector<std::type_index> types;
+
+  GameObjectClasses& add(const std::type_info &info) {
+    std::type_index idx(info);
+    types.push_back(idx);
+    return *this;
+  }
+};
 
 /**
    This class is responsible for:
@@ -102,6 +120,9 @@ public:
    * @description Returns the display name of the object, translated to the user's locale.
    */
   virtual std::string get_display_name() const { return _("Unknown object"); }
+  /** List notable classes in inheritance hierarchy of class. This makes it possible
+      to efficiently look up all objects deriving from a particular intermediate class */
+  virtual GameObjectClasses get_class_types() const;
 
   /** Version checking/updating, patch information */
   virtual std::vector<std::string> get_patches() const;
--- a/src/badguy/angrystone.hpp
+++ b/src/badguy/angrystone.hpp
@@ -38,6 +38,7 @@ public:
   virtual std::string get_class_name() const override { return class_name(); }
   static std::string display_name() { return _("Angry Stone"); }
   virtual std::string get_display_name() const override { return display_name(); }
+  virtual GameObjectClasses get_class_types() const override { return BadGuy::get_class_types().add(typeid(AngryStone)); }
 
 protected:
   virtual std::vector<Direction> get_allowed_directions() const override;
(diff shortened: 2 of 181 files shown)


src/supertux/game_object.hpp
```diff
@@ -41,6 +42,23 @@ struct GameObjectType
...
 
+/**
+ A helper structure to list all the type_indexes of the classes in the
+ type hierarchy of a given class. This makes it easier to register e.g.
+ a MrIceblock in lists for MrIceBlock, WalkingBadguy, Badguy, Portable,
+ MovingSprite, MovingObject, and GameObject.
+ */
+struct GameObjectClasses
+{
+  std::vector<std::type_index> types;
```
HybridDog, line 53: The compiler warns that `types` should be initialised.
```
supertux/src/supertux/game_object.hpp: In constructor 'GameObjectClasses::GameObjectClasses()':
supertux/src/supertux/game_object.hpp:51:8: warning: 'GameObjectClasses::types' should be initialized in the member initialization list [-Weffc++]
   51 | struct GameObjectClasses
      |        ^~~~~~~~~~~~~~~~~
supertux/src/supertux/game_object.cpp: In member function 'virtual GameObjectClasses GameObject::get_class_types() const':
supertux/src/supertux/game_object.cpp:111:23: note: synthesized method 'GameObjectClasses::GameObjectClasses()' first required here
  111 |     GameObjectClasses g;
      |                       ^
```

HybridDog:
I have tried the change on the level from #1682. In flame graphs generated with [hotspot](https://github.com/KDAB/hotspot), `GameObjectManager::update` looks a lot faster now.

#### `GameObjectManager::update` flame graphs

master:
gameobjectmanager_update_master

this PR:
gameobjectmanager_update_pr

#### Main game loop flame graphs 

master:
screenmanager_run_master

this PR:
screenmanager_run_pr

mstoeckl:
> The return value of get_class_types depends only on the type of the object it is called on dynamically. Is it possible to somehow change get_class_types, e.g. with constexpr, such that it does not need to call the parent class get_class_types method during runtime?

It should be possible, instead of a std::vector, to return some type of list constructable at compile time using `constexpr`. However, I could not find any simple solutions. As `get_class_types` is called at most twice per created object, I do not think the cost is significant.

> Is there a way to avoid adding get_class_types manually to all GameObjects with compile-time programming?

Yes, but I wouldn't recommend them.
* It may be possible to define a macro to handle class definitions which both defines the class and automatically adds get_class_types.
* Code generation always works, but can complicate the build.
* If one has a (template parameter) list of _all_ types that could be tracked, `std::is_base_of` might be usable to filter that list at compile time. However, 1) constructing that list requires including headers for all types 2) one ultimately trades a single line defining `get_class_types` for two (an `#include` and an entry in a list of template parameters), 3) I haven't actually tried this, so it may not even work -- there could be recursion depth issues.

src/supertux/game_object.hpp
mstoeckl: This is one of the `-Weffc++` style advice warnings, which recommends initializing _all_ member variables of a struct, since some types do not have default initializers. `std::vector` does. I've avoided the warning by aggregate-initializing `g`.

HybridDog:
Thanks for the explanation. The changes look fine for me.

HybridDog (approved):

tobbi (approved):

mstoeckl:
I've rebased this to fix merge conflicts.

tobbi:
Thank you 👍

weluvgoatz:
Sorry, but please re-open this, as this breaks every object that has a path attached to it by making them disappear.

mstoeckl:
> Sorry, but please re-open this, as this breaks every object that has a path attached to it by making them disappear.

Thank you for letting me know. Github doesn't seem to have a way to reopen merged&reverted PRs, so I've made a new PR that should fix the issue.

PR #2993, opened 2024-07-03
Replace asserts with clamping in color constructor

MatusGuy:


HybridDog:
In my opinion both the assertion and the clamping are bad because values outside $[0,1]$ are physically plausible. Values larger than 1 correspond to brighter light and values smaller than 0 correspond to wide-gamut colours, i.e. colours outside the sRGB gamut.

What is the purpose of the replacement?

MatusGuy:
Some people kept getting mad about the game aborting when a color exceeds 1 or goes under 0

bruhmoent (approved):

PR #2997, opened 2024-07-04
Add an option to hide wind particles + set fancy as default

bruhmoent:


tobbi (approved):
lgtm

MatusGuy (commented):
I think this would be more intuitive if you used an enum instead of two booleans. Up for discussion.

Other than that, seems to work fine.

bruhmoent:
I don't think that's necessary, I think it's already very clear.

MatusGuy (approved):

PR #3000, opened 2024-07-05
Add an option for the thunderstorm flash

### LINKED ISSUE #2914: Have option to make lightning flash way less intense
As of now, the lightning flash is very bright and sort of hurts my eyes. Additonally, i can see this bothering some other people a ton, so i propose a option to make the flash a lot more subtle. While not required, it would also probably be nice for this option to be enabled by default, as i think most people would probably prefer the more subtle flash, and it would mean there's no chance of anyone that would be very effected by this miss turning the setting on.

bruhmoent:
This pr adds a brightness setting for the thunderstorm flash as a percentage value.

Fixes #2914


Rusty-Box (approved):

MatusGuy (approved):
Untested

PR #3002, opened 2024-07-07
Merge Bomb with MrBomb

MatusGuy:
This PR internally rewrites MrBomb to be more like GoldBomb. It doesn't add any considerable features.

It also removes lots of code duplication. The Bomb class is now obsolete.

As a result, the death script stays consistent between MrBomb and GoldBomb.

This PR also fixes a bug with gold bomb, in which getting out of the safe distance and going back in would freeze gold bomb in the same sprite action until another state change.


tylerandari13 (approved):
Just what I needed. Works great.

Frostwithasideofsalt (approved):
yeah this really does work yup mhm

weluvgoatz (approved):

src/badguy/mrbomb.cpp
```diff
@@ -1,5 +1,7 @@
-//  SuperTux
+//  SuperTux BadGuy MrBomb
```
Vankata453, line 1: I noticed we don't modify this part of the copyright header anymore. Should we modify it here?

src/badguy/goldbomb.hpp
```diff
@@ -19,66 +19,42 @@
...
-  virtual void freeze() override;
-  virtual bool is_freezable() const override;
-
-  virtual void kill_fall() override;
-  virtual void ignite() override;
   static std::string class_name() { return "goldbomb"; }
   virtual std::string get_class_name() const override { return class_name(); }
   static std::string display_name() { return _("Gold Bomb"); }
   virtual std::string get_display_name() const override { return display_name(); }
-  virtual GameObjectClasses get_class_types() const override { return WalkingBadguy::get_class_types().add(typeid(GoldBomb)); }
```
Vankata453, line 52: Why was this removed?

src/badguy/goldbomb.hpp
```diff
@@ -19,66 +19,42 @@
...
+  void recover();
+  void normalize();
 
 private:
-  enum Ticking_State {
-    STATE_NORMAL,
-    STATE_TICKING,
-    STATE_REALIZING,
+  enum State : uint8_t {
+    STATE_REALIZING = 2,
```
Vankata453, line 50: In the `MrBomb::State` enumerators, add one at the end called `STATE_LAST`. Use it here to set `STATE_REALIZING`, instead of hardcoding the value in this file (may prevent future bugs when updating `MrBomb::State`).

Vankata453 (commented):
Mr. Bomb and Gold Bomb seem to work fine. I'm not sure what the mentioned issue was, so I trust you that it's been fixed.

Left some code-related comments.

src/badguy/goldbomb.hpp
MatusGuy: Must've been the automatic merge

MatusGuy:
@weluvgoatz I accidentally requested your review.

src/badguy/goldbomb.hpp
```diff
@@ -19,66 +19,43 @@
...
 
 private:
-  enum Ticking_State {
-    STATE_NORMAL,
-    STATE_TICKING,
-    STATE_REALIZING,
-    STATE_FLEEING,
-    STATE_CORNERED
+  enum State : uint8_t {
+    GB_STATE_REALIZING = 2,
```
Vankata453, line 51: This should be `MB_STATE_COUNT`:

```suggestion
    GB_STATE_REALIZING = MB_STATE_COUNT,
```

PR #3005, opened 2024-07-08
Configurable rise speed for special riser

tobbi:
I have no fucking clue what it is I'm doing or why I'm doing it. 

Actually, I'm working on my own stuff that you won't like and probably shred to pieces once you realize I've done something, just like you always do.

Whatever, this is just a small code change.

### DIFF
--- a/src/object/specialriser.cpp
+++ b/src/object/specialriser.cpp
@@ -19,10 +19,11 @@
 #include "supertux/sector.hpp"
 #include "video/drawing_context.hpp"
 
-SpecialRiser::SpecialRiser(const Vector& pos, std::unique_ptr<MovingObject> child, bool is_solid) :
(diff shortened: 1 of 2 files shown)


Rusty-Box:
> I have no fucking clue what it is I'm doing or why I'm doing it.

How reassuring... So what is this even supposed to do?

tobbi:
It allows configuring the speed with which the fireflower raises. I need it for something I'm working on that you guys will criticise to no avail anyway

Rusty-Box:
What is it then? If you need it, how about you actually share with us what it is? Because with no context I don't even know why we need it, go figure

Hume2 (approved):
This seems harmless. I see no way to actually test it. I don't know why this should be added but it seems quite natural. You have my approval. I don't know if you actually need my approval but you have it.

Issue #3008, opened 2024-07-09
[Bug]: Moving platforms/coins don't appear in level editor


epicguy2222:
### SuperTux Version

v0.6.3-1304-g00a307005

### System Information

Windows 10 64-bit

### Expected Behavior

The moving platforms/coins appear and move normally like the previous versions

### Actual Behavior

The moving platforms/coins disappears

### Steps To Reproduce Actual Behavior

Step 1: Make a platform moves in editor
Step 2: Test the level and the moving platform will disappear

### Additional Information

Error messages I think they might help:
<img width="271" alt="Screenshot 2024-07-07 212048" src="https://github.com/SuperTux/supertux/assets/168249975/e9476c4f-93e6-4646-a11f-c066ce748ec2">

<img width="960" alt="Screenshot 2024-07-07 212857" src="https://github.com/SuperTux/supertux/assets/168249975/38f21312-5040-46b0-9038-fad2e76284d9">

### Guidelines For Reporting Issues

- [X] I have read <https://github.com/SuperTux/supertux/blob/master/CONTRIBUTING.md#bug-reports>.
- [X] I have verified this isn't an issue that's already been reported.
- [X] I have verified this isn't a [discussion](https://github.com/SuperTux/supertux/discussions), or an issue about a crash or a feature request, but rather an actual bug ─ that is, the game did something not intended.
- [X] I have verified this issue is **not** about wrong translations (use Transifex for those), or anything unsupported (e.g. third-party add-ons).
- [X] In this report, I have only included details about **one** (1) bug.
- [X] If I make a mistake while submitting this report, I agree to use the "Edit" feature to correct it, instead of closing this issue and opening a new one.

weluvgoatz:
This was due to PR #2991 and has since been reverted.

PR #3009, opened 2024-07-09
Index intermediate classes of GameObjects, try two

mstoeckl:
See #2991 for details; in short, this significantly speeds up `GameObjectManager::get_objects_by_type()` and `GameObjectManager::get_object_count()`, by no longer scanning over unrelated object types. 

However, the first version had an issue with disappearing objects with paths when a level was being tested from the level editor (#3008). I believe it was caused by the PathObject base class not being recorded in `GameObject::get_class_types` implementations. I've updated the code to do this (and also record the other intermediate classes, TriggerBase and CollisionListener, even though those are not currently searched for.) I was able to reproduce #3008 with the old version and no longer see it with the new.

### DIFF
--- a/src/badguy/angrystone.hpp
+++ b/src/badguy/angrystone.hpp
@@ -38,6 +38,7 @@ public:
   virtual std::string get_class_name() const override { return class_name(); }
   static std::string display_name() { return _("Angry Stone"); }
   virtual std::string get_display_name() const override { return display_name(); }
+  virtual GameObjectClasses get_class_types() const override { return BadGuy::get_class_types().add(typeid(AngryStone)); }
 
 protected:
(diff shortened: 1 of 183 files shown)


MatusGuy (approved):

tobbi (approved):
I didn't look through the list of all classes to see if all are accounted for, but code-wise this looks good.

mstoeckl:
I've updated this PR to account for the new Boss and BigSnowball classes.

weluvgoatz (approved):
Tested, looks good!

tobbi (approved):
Haven't tested this, but looks good, code-wise.

tobbi:
@Vankata453 Can you merge this, please?

Vankata453 (approved):

PR #3011, opened 2024-07-11
Run init-script only once after a sector has loaded

### LINKED ISSUE #698: InitScript bug when teleporting
SuperTux version: current git version (28.04.2017)
System information: Ubuntu 17.04 64-bit

Expected behaviour: InitScript should be either stopped and triggered again when Tux is teleported, or it should be neither stopped nor triggered again.

Actual behaviour: InitScript is triggered again when Tux is teleported, but it isn't stopped when it's already running.

Steps to reproduce actual behaviour: Try it with a script that runs for a long time or contains an infinite loop.

Video (might help if you don't know what I mean): https://www.youtube.com/watch?v=ZEkpM3VxFWU

Note: I haven't tried this with multiple sectors.
Note: I have included swimming before compiling, so, even though it is very unlikely, this might be an issue that only occurs with swimming included.

tobbi:
Fixes #698

### DIFF
--- a/src/supertux/sector.cpp
+++ b/src/supertux/sector.cpp
@@ -287,7 +287,7 @@ Sector::activate(const Vector& player_pos)
 
   // Run init script
   if (!m_init_script.empty() && !Editor::is_active() &&
-      ((m_init_script_run_once && !m_init_script_run) || !m_init_script_run_once)) {
+      (!m_init_script_run || !m_init_script_run_once)) {
     run_script(m_init_script, "init-script");
     m_init_script_run = true;
   }
--- a/src/supertux/sector.cpp
+++ b/src/supertux/sector.cpp
@@ -69,7 +69,8 @@ Sector::Sector(Level& parent) :
   m_foremost_opaque_layer(),
   m_gravity(10.0f),
   m_collision_system(new CollisionSystem(*this)),
-  m_text_object(add<TextObject>("Text"))
+  m_text_object(add<TextObject>("Text")),
+  m_init_script_run()
 {
   add<DisplayEffect>("Effect");
   add<TextArrayObject>("TextArray");
@@ -284,8 +285,9 @@ Sector::activate(const Vector& player_pos)
   }
 
   // Run init script
-  if (!m_init_script.empty() && !Editor::is_active()) {
+  if (!m_init_script.empty() && !Editor::is_active() && !m_init_script_run) {
     run_script(m_init_script, "init-script");
+    m_init_script_run = true;
   }
 
   // Do not interpolate camera after it has been warped
--- a/src/supertux/sector.cpp
+++ b/src/supertux/sector.cpp
@@ -286,10 +286,10 @@ Sector::activate(const Vector& player_pos)
   }
 
   // Run init script
-  if (!m_init_script.empty() && !Editor::is_active() &&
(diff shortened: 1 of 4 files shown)


Narre:
This is a great option to have, especially for infinitely looping scripts that are only meant to trigger once, such as a loop that randomises the time between lightning strikes, however, I think there remains the need for scripts that are triggered every time on sector entry (such as my STAGE project, that needs to check for changes made in other sectors when a sector is entered). I have a few possible solutions in mind.

- init script is triggered every time the sector is entered, much as it works now, but its execution is terminated when the sector is unloaded (so it gets triggered again upon re-entry but the old instance does not continue running)
- there is an option in the sector settings that determines whether the init script should only trigger once upon the sector being loaded, or every time
- there are two separate scripts, the init script that's only triggered on first initialisation, and an enter script that's triggered every time (in this case, I think it would be better for retrocompatibility if existing init scripts in old levels are interpreted as enter scripts, so they work how they did in the version they were written in)

tobbi:
Basically you want an option that you can check to specify whether it's run every time. I think I need to also reset the flag when the sector is reset.

tobbi:
Discussion question:
Should I keep the default of "run every time" for backward compatibility reasons, but make an option of "run only once".

Can I get some input on this?

@Narre @tylerandari13 @Vankata453

Vankata453:
Yeah, I believe a script to run only on sector initialization should be a separate option.

Narre:
I agree, with both options available, the default should be the old behaviour to avoid needlessly breaking older levels.

tobbi:
This is ready for review. I haven't tested the actual behaviour but to quote "I, Robot": "My logic is undeniable".
https://www.youtube.com/watch?v=LH-G8c3TUac

tylerandari13:
Every time you load a sector it replaces all the instances with new instances. Is this also gone with this PR or is that untouched?

tobbi:
> Every time you load a sector it replaces all the instances with new instances. Is this also gone with this PR or is that untouched?

What do you mean by "instances" in this case?

tylerandari13:
> > Every time you load a sector it replaces all the instances with new instances. Is this also gone with this PR or is that untouched?
> 
> What do you mean by "instances" in this case?

Basically the exposed objects. (theyre referred to as instances by Squirrel)

src/supertux/sector.cpp
```diff
@@ -287,8 +289,10 @@ Sector::activate(const Vector& player_pos)
   }
 
   // Run init script
-  if (!m_init_script.empty() && !Editor::is_active()) {
+  if (!m_init_script.empty() && !Editor::is_active() &&
+      ((m_init_script_run_once && !m_init_script_run) || !m_init_script_run_once)) {
```
mrkubax10, line 293: I suppose this could be replaced with `(!m_init_script_run_once || !m_init_run_script)`.

Frostwithasideofsalt:
What's the status on this PR?

tobbi:
Nothing was done since the last comments. It's another PR I need to tackle.

Alasdairbugs:
Assigned 0.7.0 as it seems like a crucial thing for level makers

swagtoy:
Tested with a willow wisp, works fine.

PR #3013, opened 2024-07-13
Fix egg not showing shade sprite when placed on its own.

tylerandari13:
To put it simply heres what it looked like before...
image

...and heres what it looks like now.
image

I noticed a Bonus Island level used a powerup on its own so I decided to take a whirl at fixing it for 0.7.

### DIFF
--- a/src/object/powerup.cpp
+++ b/src/object/powerup.cpp
@@ -110,6 +110,11 @@ PowerUp::initialize()
   SoundManager::current()->preload("sounds/fire-flower.wav");
   SoundManager::current()->preload("sounds/gulp.wav");
 
+  if(m_type == EGG) {
+    shadesprite = SpriteManager::current()->create("images/powerups/egg/egg.sprite");
+    shadesprite->set_action("shadow");
+  }
+
   // Older levels utilize hardcoded behaviour from the chosen sprite
   if (get_version() == 1)
   {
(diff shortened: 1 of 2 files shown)


src/object/powerup.cpp
```diff
@@ -110,6 +110,11 @@ PowerUp::initialize()
   SoundManager::current()->preload("sounds/fire-flower.wav");
   SoundManager::current()->preload("sounds/gulp.wav");
 
+  if(m_type == EGG) {
```
tobbi, line 113: Please place the opening brace in its own line.

tobbi (approved):
Once the code style issue is fixed, this looks good to go.

src/object/powerup.cpp
```diff
@@ -110,6 +110,11 @@ PowerUp::initialize()
...
+    m_shadesprite = SpriteManager::current()->create("images/powerups/egg/egg.sprite");
+    m_shadesprite->set_action("shadow");
+  }
```
MatusGuy, line 116: Fix this by placing this code snippet at the end of the after_editor_set function.

MatusGuy (changes requested):
Game crashes when changing from non egg to egg.

juanmagdev (approved):

PR #3014, opened 2024-07-13
Tux emits bubbles while being underwater

bruhmoent:
This pr adds a visual effect - when Tux is underwater, he will emit bubble particles that will follow a sine pattern until they reach the water surface.

Make sure to test.


Brockengespenst:
I think this is a pretty nice feature, but is it possible that the air_bubble-*.png files are missing?
Maybe some idea: What about using std::array for m_bubble_particles? This would make it easier to iterate over the array or do a check for bounds (e.g. when picking a random bubble).

bruhmoent:
Oh you're right, I forgot to push the bubbles.

Brockengespenst:
Looks pretty nice so far. Some comments from my side:
- Sometimes it seems the bubbles are not emitted at the correct place after Tux has turned around. That looks a bit weird when the bubbles are emitted at the wrong side of Tux ;-)
- I am not sure if the bubbles should really leave the water and vanish in the air. Probably they should start to vanish a bit earlier

bruhmoent:
I like the effect of them "popping" in the air. And for the bubbles not being emitted at the right place, I'm aware - but I don't know what to do about it - sometimes the sprite rotation gets registered improperly - At-least that's what I've observed.

Brockengespenst:
Some suggestions:
- Use m_swimming_angle as it is already in radians
- I do not fully understand your direction check. If comparing the angle in radians, I thinks something like m_swimming_angle > M_PI_2 && m_swimming_angle < M_PI_2 * 3.0 should check if Tux is facing left, all other cases Tux is facing right
- If the beak was at Tux.height / 2, there would even be no need to check for direction, just rotate it. But if doing so in this case, this might lead to a small offset, maybe you can check if this looks still OK
- I would not use std::vector for m_active_bubbles as it might lead to lots of copy operations when deleting entries as vector must guarantee that all entries are placed in a linear order on the heap. When deleting items in the middle, they have to be copied to a new place in linear order. As you do not need to access the bubles by index, you could simply use a std::list or similar

bruhmoent:
Thanks for the suggestions!

Brockengespenst:
No problem. I really like this feature, but I think there are two small tweaks to be done:
- I guess small and big Tux need to be handled differently. Small Tux swimming left looks good, but when small Tux swims right, there is a bit of space between the beak and the point where bubbles are emitted
- Also, kind of related probably, left and right direction of Tux seem to need different offsets. Big Tux swimming right emits bubbles direcly at his beak, but big Tux swimming left emits the bubbles somwhere "behind his head"
It would be great if you could have a look at these two things.

src/object/player.cpp
```diff
@@ -514,6 +536,64 @@ Player::update(float dt_sec)
...
+        glm::vec2 beak_local_offset(30.f, 0.0f);
+        float big_offset_x = is_big() ? 4.0f : 0.0f;
+
+        // Calculate the offsets based on the sprite angle
+        float offset_x = std::cos(m_swimming_angle) * 10.0f;
+        float offset_y = std::sin(m_swimming_angle) * 10.0f;
+
+        // Rotate the beak offset based on the sprite's angle
+        float rotated_beak_offset_x = beak_local_offset.x * std::cos(m_swimming_angle) - beak_local_offset.y * std::sin(m_swimming_angle);
+        float rotated_beak_offset_y = beak_local_offset.x * std::sin(m_swimming_angle) + beak_local_offset.y * std::cos(m_swimming_angle);
```
tobbi, line 551: Is the first `beak_local_offset.x` correct or is it a typo and you meant to use the y coordinate?

tobbi (approved):
Approve on account of code style. Haven't done any actual testing in-game, though.

src/object/player.cpp
```diff
@@ -514,6 +536,64 @@ Player::update(float dt_sec)
...
+
+        glm::vec2 player_center = m_col.m_bbox.get_middle();
+        glm::vec2 beak_position;
+
+        // Determine direction based on the radians
+        if (m_swimming_angle > static_cast<float>(M_PI_2) && m_swimming_angle < 3.0f * static_cast<float>(M_PI_2)) // Facing left
```
MatusGuy, line 557: We use `math::PI_2`
```suggestion
        if (m_swimming_angle > static_cast<float>(math::PI_2)) && m_swimming_angle < 3.0f * static_cast<float>(math::PI_2)) // Facing left
```

Rusty-Box:
@bruhmoent Why did you not use the .sprite file I send you?

src/object/player.cpp
```diff
@@ -254,6 +254,12 @@ Player::Player(PlayerStatus& player_status, const std::string& name_, int player
   SoundManager::current()->preload("sounds/invincible_start.ogg");
   SoundManager::current()->preload("sounds/splash.wav");
   SoundManager::current()->preload("sounds/grow.wav");
+  m_bubble_particles[0] = Surface::from_file("images/particles/air_bubble-1.png");
+  m_bubble_particles[1] = Surface::from_file("images/particles/air_bubble-2.png");
+  m_bubble_particles[2] = Surface::from_file("images/particles/air_bubble-3.png");
+  m_bubble_particles[3] = Surface::from_file("images/particles/air_bubble-4.png");
+  m_bubble_timer.start(3.0f + (rand() % 2));
```
MatusGuy, line 261: For usage of rng in graphics effects, use `graphicsRandom`
```suggestion
  m_bubble_timer.start(3.0f + graphicsRandom.randf(2));
```

Rusty-Box:
Also, I think the bubbles vertical speed could be increased by quite a bit

src/object/player.hpp
```diff
@@ -552,6 +555,10 @@ class Player final : public MovingObject
 
   SurfacePtr m_airarrow; /**< arrow indicating Tux' position when he's above the camera */
 
+  std::array<SurfacePtr, 4> m_bubble_particles; /**< bubble particles for swimming */
+  Timer m_bubble_timer; /**< timer for spawning bubble particles */
+  std::list<std::pair<SurfacePtr, glm::vec2>> m_active_bubbles; /**< active bubble particles */
```
MatusGuy, line 560: Here you should use a vector.
```suggestion
  std::vector<std::pair<SurfacePtr, glm::vec2>> m_active_bubbles; /**< active bubble particles */
```
Don't forget to include `<vector>`!

bruhmoent:
> @bruhmoent Why did you not use the .sprite file I send you?

I just used the images - I found the .sprite file to be unnecessary? Idk how does it matter really

MatusGuy (changes requested):
Replace all `glm::vec2`s with `Vector`s

Despite what I mentioned. It looks great! Thanks for your contribution!

src/object/player.hpp
Brockengespenst: I suggested to not use std::vector here, because when bubbles are removed from m_active_bubbles, this may lead to unnecessary coping of the remaining elements. In my opinion a std::list is fine as it does not need the containing elements to be placed linearly on the heap.

MatusGuy: fair. Thanks for pointing that out.

Brockengespenst:
Sorry, but I have to slightly update my suggestion regarding how to determine the direction. From the swim sprite logic, I saw that the following code is used:
```std::abs(m_swimming_angle) <= math::PI_2 ? Direction::RIGHT : Direction::LEFT)``` 
This seems to work more reliable, because it takes also negative angles into account.

File: src/object/player.hpp [hunk 5]
```diff
@@ -28,6 +28,9 @@
 #include "video/layer.hpp"
 #include "video/surface_ptr.hpp"
 
+#include <array>
```
Brockengespenst, line 31: This include is not needed anymore.

Brockengespenst (changes requested):
There is just this one include, that is obsolete again. Rest looks fine, thanks for adapting the left/right check.

Brockengespenst (approved):
Looks fine to me.

mstoeckl:
While this looks nice and cartoony, I (random passerby) should warn that cartoons are often wrong, and actual penguins probably don't breathe out underwater. (Nor should any species that can't breathe _in_ underwater, except if very close to the surface :-) 

Instead of emitting periodic bubbles from their beaks, penguins might produce them through:
1. splashing near water surface
2. possible: some bubbles when opening beak underwater to grab things for the first time
3. releasing trapped air after diving in (looks like many small trailing bubbles from the back/wings of penguin; could take some time for all bubbles to go.)

Therefore: I would suggest emitting the bubbles more irregularly in space and time; from the main body instead of from the beak; and maybe progressively reduce the bubble rate the longer Tux stays underwater.

bruhmoent:
SuperTux is not aimed to replicate realistic behaviours. This is just a little visual effect. This is out of scope for this PR - and this might be adjusted in the future.

mstoeckl:
> SuperTux is not aimed to replicate realistic behaviours. This is just a little visual effect. This is out of scope for this PR - and this might be adjusted in the future.

Agreed. While I think foundations in reality are nice to have, in the end art and gameplay are what matter.

src/object/player.cpp
```diff
@@ -514,6 +533,72 @@ Player::update(float dt_sec)
...
+
+            if (num_bubbles > 1)
+            {
+              float burst_offset_x = graphicsRandom.randf(-5.0f, 5.0f);
+              float burst_offset_y = graphicsRandom.randf(-5.0f, 5.0f);
+              bubble_pos.x += burst_offset_x;
+              bubble_pos.y += burst_offset_y;
+            }
+
+            m_active_bubbles.push_back({ bubble_surface, bubble_pos });
```
MatusGuy, line 595: The bubble should not try to spawn if it's not submerged, or else it will spawn the bubble for one frame and then pop in the next frame. To reproduce this bug, get in this position.
image

src/object/player.cpp
```diff
@@ -412,6 +415,22 @@ Player::update(float dt_sec)
...
+    for (auto& bubble : m_active_bubbles)
+    {
+      bubble.second.y -= dt_sec * 40.0f;
+      bubble.second.x += std::sin(bubble.second.y * 0.1f) * dt_sec * 5.0f;
+    }
+
+    m_active_bubbles.remove_if([&](const std::pair<SurfacePtr, Vector>& bubble)
+      {
+        Rectf bubble_box(bubble.second.x, bubble.second.y, bubble.second.x + 16.f, bubble.second.y + 16.f);
+        bool is_out_of_water = Sector::get().is_free_of_tiles(bubble_box, true, Tile::WATER);
```
MatusGuy, line 429: Make sure the bubble pops when hitting solid tiles.

MatusGuy (changes requested):
I can't confirm if the item in `m_active_bubbles` gets popped when the bubble goes offscreen, but make sure it does.

Almost there!

Rusty-Box:
The bubbles are still not animated despite the sprite file having 2 action of animation. One for bigger bubbles and one for smaller ones.

swagtoy (commented):
You may be able to get away string_view instead of arrays of strings

Frostwithasideofsalt (approved):

MatusGuy (approved):
Nice work

PR #3018, opened 2024-07-13
Prevent rock from going through ceiling at negative gravity

### LINKED ISSUE #2982: Rock phases through ceiling
If the sector gravity is flipped and a rock hits a ceiling, it phases through it.  Not sure if this has been reported, but it exists.

Brockengespenst:
Fixes #2982


Brockengespenst:
@weluvgoatz: This *should* solve the problem. But as Rock is used as a base class for several other objects, intensive testing would be good. At a first glance it looks fine, but there might be things I do not have in mind right now.

weluvgoatz (approved):

MatusGuy (approved):
code looks good

PR #3019, opened 2024-07-13
Moving converted objects from tiles with tilemap

Vankata453:
When a `TileMap` is moving on a set path, any objects that have been converted from tiles in it, are now moved together with it as well.

### DIFF
--- a/src/object/tilemap.cpp
+++ b/src/object/tilemap.cpp
@@ -22,10 +22,13 @@
 #include <simplesquirrel/vm.hpp>
 
 #include "editor/editor.hpp"
+#include "object/pulsing_light.hpp"
 #include "supertux/autotile.hpp"
 #include "supertux/debug.hpp"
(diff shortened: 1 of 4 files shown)


Frostwithasideofsalt:
the builds are expired, can you re-run them?

Frostwithasideofsalt:
i.. dont really know either. 
https://github.com/user-attachments/assets/dae6a590-d89d-427d-8113-fb49c75da41d

Hume2 (changes requested):
I rebased the branch to Hume2:objects-move-with-tilemap

It works well, as long as the objects aren't moving on their own. When I place a bonus block into a moving tilemap and bump it, it doesn't return to its original position with respect to the tilemap. Also Tux walking on these bonusblocks acts weirdly.
[level2.zip](https://github.com/user-attachments/files/19555805/level2.zip)

I suggest postponing this to 0.7.1+, as fixing it won't be straightforward.

PR #3025, opened 2024-07-19
Squirrel: Default argument support for exposed C++ functions

Vankata453:
This PR updates the simplesquirrel submodule to a branch which supports exposing C++ functions (including class constructors) wtih default arguments.

The `Camera::set_scale()`, `scripting::Level::spawn()`, `scripting::Globals::wait()` and `MovingSprite::set_action()` functions were modified to support default arguments and their respective other functions were marked as deprecated. In case more functions are found suitable for this, this PR would be updated.


Brockengespenst (approved):
Tested Level.spawn() with all transitions and everything works as expected.
I did not test the camera stuff.

Frostwithasideofsalt (approved):
Works.

tylerandari13 (approved):
As Patrick Star once said, "It can't be that bad. Looks good to me."

tylerandari13:
I have a request for a default argument.

Give `wait()` a second argument that if true will force it to wait through cutscenes even if they've been skipped. This will be especially useful for while loops that would hang the game otherwise. Its a relatively simple addition that makes use of the new default arguments so I decided to request it.

PR #3032, opened 2024-08-02
`Screen`: Virtual methods for SDL and window resize events

Vankata453:
The `Screen` abstract class now provides virtual methods for SDL and window resize events. The `Editor` and `PartcleEditor` classes now `override` the `event(const SDL_Event& ev)` method and `Editor` `override`s `on_window_resize()`.

Resize events in `ScreenManager` are now handled by all `Screen`s in the stack. SDL events are only handled by the current (top) screen.

Additionally, `OptionsMenu` now calls `ScreenManager::on_window_resize()` on video setting changes, instead of just calling `MenuManager::on_window_resize()`. This fixes a bug where the game crashes on changing the "Video Resolution" and "Magnification" settings from the in-editor "Options" menu.

### DIFF
--- a/src/editor/button_widget.cpp
+++ b/src/editor/button_widget.cpp
@@ -62,7 +62,7 @@ ButtonWidget::setup()
 }
 
 void
-ButtonWidget::resize()
+ButtonWidget::on_window_resize()
 {
 }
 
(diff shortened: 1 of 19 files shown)


tobbi (approved):
Looking good, code-wise.

MatusGuy (commented):
Code looks great. Will test later

MatusGuy (approved):

PR #3035, opened 2024-08-06
Add clear selection button for filesystem menu

### LINKED ISSUE #3030: Make teleporting not reload everything
Currently teleporting causes everything to reload, which causes unseeded lag spikes (mainly on the worldmap from my experience?), and causes issues with scripting. it would be nice if it could be changed to only set the players position, as that would solve both of those issues

MatusGuy:
This allows the user to clear the selection of a filesystem option, as if it was brand new. Useful for removing the selection of a worldmap file on a teleporter, for example, which fixes #3030.


MatusGuy:
TODO: should probably make this a "reset to default" option

PR #3036, opened 2024-08-07
`ControlTextbox` multiline overhaul, use in script editor

Vankata453:
This PR overhauls `ControlTextbox` so it supports multiple lines.

A new derived from `ControlTextbox` class has been added - `ControlScriptbox`. It features suggestions when typing (just like the console does when "Tab" is pressed). Additional information and parameters for the suggested functions are also displayed. That information is fetched from the "data/scripts/reference.stsr" file, which is of type "supertux-scripting-reference". Other such files can also be imported using the `register_scripting_reference()` and `unregister_scripting_reference()` scripting functions.

image

[Short video showcase from 2023 development summary video](https://youtu.be/Wcqat_b8ft8?t=174)

**NOTE:** This is a WIP, since it needs to be suited to all additions and changes since the implementation of the simplesquirrel library. It also needs to further be tested for bugs.


swagtoy:
For reasons, Im interested in reviewing this when I get a chance :)

src/interface/control_scriptbox.cpp
```diff
@@ -0,0 +1,496 @@
...
+  if (m_suggestions.empty())
+    return;
+
+  const squirrel::Suggestion& suggestion = m_suggestions.at(m_selected_suggestion);
+  if (!suggestion.reference) return;
+
+  const float max_width = (static_cast<float>(SCREEN_WIDTH) - m_suggestions_rect.get_right() -
+                           SUGGESTION_DESCRIPTION_RECT_SPACING);
+
+  using namespace squirrel;
```
swagtoy, line 311: Move this to the top please. Then change `squirrel::Suggestion` case.

Also TBH it would be way more readable if we did `using sq = squirrel;` if you want to save some keystrokes without axing readability

src/interface/control_scriptbox.cpp
```diff
@@ -0,0 +1,496 @@
...
+  std::vector<std::string> desc_lines;
+  switch (suggestion.reference->get_type())
+  {
+    case ScriptingObject::Type::CONSTANT:
+    {
+      const auto& con = static_cast<const ScriptingConstant&>(*suggestion.reference);
+
+      desc_lines.push_back(con.type + " " + con.name);
+      desc_lines.push_back(con.description);
+      desc_lines.push_back(con.detailed_description);
```
swagtoy, line 322: :/

src/interface/control_scriptbox.cpp
```diff
@@ -0,0 +1,496 @@
...
+    }
+    break;
+
+    case ScriptingObject::Type::FUNCTION:
+    {
+      const auto& func = static_cast<const ScriptingFunction&>(*suggestion.reference);
+
+      const Vector pos = get_suggestion_description_rect_pos();
+
+      std::stringstream out;
```
swagtoy, line 332: Maybe you should use `format` for this instead? There is (theoretically, aka unneccesary, yet likely not practically) a small performance loss doing this that we could probably improve.

src/interface/control_scriptbox.cpp
```diff
@@ -0,0 +1,496 @@
...
+  }
+
+  std::string line_content = get_contents(m_caret.line);
+  const size_t begin = line_content.find_first_not_of(" \t"); // Skip beginning spaces
+  if (begin == std::string::npos)
+  {
+    m_suggestions.clear();
+    return;
+  }
+  line_content = line_content.substr(begin);
```
swagtoy, line 439: This might be able to live in its own private, or even static function... just sayin'

src/interface/control_scriptbox.cpp
```diff
@@ -0,0 +1,496 @@
...
+
+  m_suggestion_description_rect = Rectf(get_suggestion_description_rect_pos(),
+                                        size + Sizef(SUGGESTION_DESCRIPTION_RECT_SPACING * 2,
+                                                     SUGGESTION_DESCRIPTION_RECT_SPACING));
+}
+
+Vector
+ControlScriptbox::get_suggestion_description_rect_pos() const
+{
+  return Vector(m_suggestions_rect.get_right(), m_suggestions_rect.get_top() +
```
swagtoy, line 491: Would be nice if you put a newline after `get_right(),` imo

src/interface/control_scriptbox.hpp
```diff
@@ -0,0 +1,76 @@
...
+  std::unique_ptr<ControlScrollbar> m_suggestions_scrollbar;
+  float m_suggestions_offset;
+
+  /** Selected suggestion */
+  size_t m_selected_suggestion;
+  std::vector<std::string> m_suggestion_description;
+  Rectf m_suggestion_description_rect;
+
+  /** Hovering over a parameter in function suggestion description */
+  std::vector<std::pair<Rectf, std::string>> m_function_parameters;
```
swagtoy, line 66: `std::map` ?

src/interface/control_scriptbox.hpp
```diff
@@ -0,0 +1,76 @@
...
+  const std::pair<Rectf, std::string>* m_hovered_parameter;
```
swagtoy, line 67: typedeffing this would also help, like `Parameter_Pair`,in case this must be changed

src/interface/control_scrollbar.cpp
```diff
@@ -50,7 +50,7 @@ ControlScrollbar::draw(DrawingContext& context)
   context.color().draw_filled_rect(get_bar_rect(),
                                    Color(1.f, 1.f, 1.f, (m_hovering || m_scrolling) ? 1.f : 0.7f),
```
swagtoy, line 51: Color options?

src/interface/control_textbox.cpp
```diff
@@ -30,160 +28,233 @@
 // The time for the caret to change visibility (when it flashes).
 static const float CONTROL_CURSOR_TIMER = 0.5f;
 
-ControlTextbox::ControlTextbox() :
+const float ControlTextbox::TEXT_X_OFFSET = 5.f;
+const float ControlTextbox::TEXT_Y_OFFSET = 5.f;
+
+static const size_t UNDO_STACK_SIZE = 10;
```
swagtoy, line 34: This can certainly be a setting

src/interface/control_textbox.cpp
```diff
@@ -30,160 +28,233 @@
...
   m_shift_pressed(false),
   m_ctrl_pressed(false),
   m_mouse_pressed(),
-  m_current_offset(0)
+  m_scrollbar(),
+  m_offset(0.f, 0.f),
+  m_box_offset(0.f, 0.f),
+  m_box_rect(m_rect)
+{
+  m_scrollbar.reset(new ControlScrollbar(1.f, 1.f, m_offset.y, 35.f));
```
swagtoy, line 56: `std::make_unique`

also unrelated but i think you can just use `operator=` over reset?

src/interface/control_textbox.cpp
```diff
@@ -195,75 +266,151 @@ ControlTextbox::on_key_down(const SDL_KeyboardEvent& key)
...
-        m_secondary_caret_pos = m_caret_pos;
+        set_secondary_caret_pos(m_caret.pos);
     }
 
-    recenter_offset();
+    recenter_offset(true);
     return true;
   }
-  else if (key.keysym.sym == SDLK_BACKSPACE)
+  else if (key.keysym.sym == SDLK_UP)
```
swagtoy, line 305: Why don't we even use switch/case in the code for these, just food for thought? A compiler probably optimizes it to such anyway, hopefully...

src/interface/control_textbox.cpp
```diff
@@ -195,75 +266,151 @@ ControlTextbox::on_key_down(const SDL_KeyboardEvent& key)
...
+    set_caret_pos(m_caret.pos + 1); // Newline
+    set_secondary_caret_pos(m_caret.pos);
+
+    recenter_offset(true);
+    parse_value();
+    return true;
+  }
+  else if (key.keysym.sym == SDLK_TAB)
+  {
+    put_text("  "); // Tab key should insert 2 spaces
```
swagtoy, line 374: malloc

src/interface/control_textbox.cpp
```diff
@@ -195,75 +266,151 @@ ControlTextbox::on_key_down(const SDL_KeyboardEvent& key)
...
+      set_secondary_caret_pos(m_caret.pos);
 
-    recenter_offset();
+    recenter_offset(true);
     return true;
   }
   else if (key.keysym.sym == SDLK_END)
   {
-    m_caret_pos = int(m_charlist.size());
+    set_caret_pos(get_line_start(m_caret.line) + static_cast<int>(m_charlist[m_caret.line].size()));
```
swagtoy, line 408: is this cast even necessary?

src/interface/control_textbox.cpp
```diff
@@ -288,70 +435,136 @@ ControlTextbox::on_key_down(const SDL_KeyboardEvent& key)
...
+    const auto& change = m_redo_stack.back();
+    revert_value(change.string);
+    if (parse_value(true, TextAction::REDO))
+    {
+      m_caret = std::move(change.carets.caret);
+      m_secondary_caret = std::move(change.carets.secondary_caret);
+    }
+    m_redo_stack.pop_back();
     return true;
   }
```
swagtoy, line 472: This is like... 18 if else statements... Sorry but, please refactor this somehow. I did not read this with a straight face.

src/interface/control_textbox.cpp
```diff
@@ -288,70 +435,136 @@ ControlTextbox::on_key_down(const SDL_KeyboardEvent& key)
...
     put_text(std::string(ev.text.text));
+    parse_value();
+  }
 
   return false;
 }
 
 bool
-ControlTextbox::parse_value(bool call_on_change /* = true  (see header)*/)
+ControlTextbox::parse_value(bool call_on_change /* = true  (see header)*/, TextAction action /* = TextAction::NONE  (see header)*/)
```
swagtoy, line 492: Just don't `(see header)` the reader like this at all. They'll see the header if they need to, as it's obvious.

src/interface/control_textbox.cpp
```diff
@@ -288,70 +435,136 @@ ControlTextbox::on_key_down(const SDL_KeyboardEvent& key)
...
-  std::string new_str = get_string();
-  if (m_internal_string_backup != new_str) {
-    m_internal_string_backup = new_str;
+  m_internal_string_backup = new_str;
+  m_previous_caret_backup = { m_caret, m_secondary_caret };
 
-    if (m_string)
-      *m_string = new_str;
+  if (m_string)
+    *m_string = new_str;
```
swagtoy, line 527: what if instead of a pointer we just set some int variables or something, then returned a string view of some sort? Food for thought of the soul. It is certainly how most editors handle such logic, if I'm getting this down correctly

src/interface/control_textbox.cpp
```diff
@@ -288,70 +435,136 @@ ControlTextbox::on_key_down(const SDL_KeyboardEvent& key)
...
+    {
+      if (!m_multiline)
+        continue;
+
+      m_charlist.push_back({});
+      line++;
+      continue;
+    }
+    m_charlist[line].push_back(c);
   }
```
swagtoy, line 562: Standard for statement can't possibly hurt in this case. Not a big deal though.

src/interface/control_textbox.cpp
```diff
@@ -361,102 +574,125 @@ ControlTextbox::get_string() const
 }
 
 std::string
-ControlTextbox::get_contents() const
+ControlTextbox::get_contents(int line) const
```
swagtoy, line 577: This does handle negative logic too? I think it can be made clear with a comment.

src/interface/control_textbox.cpp
```diff
@@ -361,102 +574,125 @@ ControlTextbox::get_string() const
...
+  for (char c : m_charlist[line])
+  {
     if (!(amount--)) break;
-    temp += c;
+    out += c;
   }
 
-  return temp;
+  return out;
 }
```
swagtoy, line 613: Could this logic be improved somehow, anyway?

swagtoy (changes requested):
Taking a break at `control_textbox.cpp` line 669, but dumping what I've seen.

Just little things mostly, but overall looks good. I have not even ran this PR yet.

src/interface/control_scriptbox.cpp
```diff
@@ -0,0 +1,496 @@
...
+ControlScriptbox::draw(DrawingContext& context)
+{
+  ControlTextbox::draw(context);
+
+  /** Line numbers */
+  const Rectf line_number_rect(m_rect.p1(),
+                               m_rect.p1() + Vector(LINE_NUMBER_RECT_WIDTH, m_rect.get_height()));
+  context.color().draw_filled_rect(line_number_rect,
+                                   m_has_focus ? Color(0.8f, 0.8f, 0.8f, 1.0f)
+                                               : Color(0.6f, 0.6f, 0.6f, 1.f),
```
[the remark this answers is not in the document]
tobbi: I think colors should be moved to a variable of some sort.

swagtoy:
@provigz Are you still interested in completing this ever since I implemented external editor support? I do think it would be useful, though, at least for viewing. You pitched at some point that we should just remove the built-in editor entirely, so just checking in :)

provigz:
Hm, I think this would still be nice to have in the future, probably, because of the basic auto-completion and simplicity of using in-game. Though I think the external editor can suffice well for now.

swagtoy:
Yeah i'm cool with it. We still need a way to quickly _view_ a script without opening an external editor.

PR #3037, opened 2024-08-07
Add more convenience functions to Player class

tobbi:
Waddup, biatches! I just added a few convenience methods to the player class that will reduce some complexity in if statements elsewhere.

Have fun "reviewing" (aka tearing into my code changes).

### DIFF
--- a/src/object/player.hpp
+++ b/src/object/player.hpp
@@ -142,8 +142,8 @@ public:
   bool is_alive() const { return !is_dying() && !is_dead(); }
   
   /**
-   * Returns true if the player can be controlled
-   * by the player (alive and not currently in a win sequence)
+   * Returns true if the player can be controlled.
+   * (alive and not currently in a win sequence)
    */
   bool is_active() const { return is_alive() && !is_winning(); }
 
--- a/src/object/player.hpp
+++ b/src/object/player.hpp
@@ -135,6 +135,18 @@ public:
   bool is_invincible() const { return m_invincible_timer.started(); }
   bool is_dying() const { return m_dying; }
 
+  /**
+   * Returns true if the player is currently alive 
+   * (not dying or dead)
+   */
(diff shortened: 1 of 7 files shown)


MatusGuy:
no

tobbi:
Thank you, @MatusGuy for your enthusiasm in reviewing the PR.

mrkubax10:
I agree with Tobbi, could you try being a little bit nicer?

tobbi:
It's an ongoing feud at this point. TBH, I don't know how serious either of us are about this at this point. I just know that I'm more on edge / unhinged than usual.

src/object/player.hpp
```diff
@@ -135,6 +135,18 @@ class Player final : public MovingObject
...
+   * Returns true if the player is currently alive 
+   * (not dying or dead)
+   */
+  bool is_alive() const { return !is_dying() && !is_dead(); }
+  
+  /**
+   * Returns true if the player can be controlled
+   * by the player (alive and not currently in a win sequence)
+   */
+  bool is_active() const { return is_alive() && !is_winning(); }
```
MatusGuy, line 148:
```suggestion
  /**
   * Returns true if the player can be controlled.
   * (alive and not currently in a win sequence)
   */
  bool is_active() const { return is_alive() && !is_winning(); }
```

I just think repeating the word "player" while meaning two different things doesn't sound well and could probably cause confusion.

MatusGuy (approved):

tobbi:
> > I agree with Tobbi, could you try being a little bit nicer?
> 
> I don't think his attitude was correct either. By saying "no", I was joking around, like he was in his PR description.
> 
> Anyway, I hope we can put that aside.

It was quite clear from the start that we were both joking, that's why I edited your message to a very polite one, before you added it back. No worries. This is all fine, unlike some other things I could tell you about.

bruhmoent (approved):
Not tested.

PR #3040, opened 2024-08-09
Replace RelWithDebInfo builds with Release builds

Semphriss:
RelWithDebInfo include debug data separately from the executable code, which needs a debugger to be interpreted properly.

The debug data does not affect stack traces generated by the program itself, but increases the executable by more than 100Mb, which is wasted if the user does not intent to run the exectuable inside a debugger.

### DIFF
--- a/.github/workflows/bsd.yml
+++ b/.github/workflows/bsd.yml
@@ -32,7 +32,7 @@ jobs:
       matrix:
         # TODO: Add the OpenBSD, NetBSD and Solaris VMs whenever possible
         vm_os: [freebsd]
-        build_type: [Debug, RelWithDebInfo]
+        build_type: [Debug, Release]
         # FreeBSD does not have glbinding
         # glbiactions/checkout@v4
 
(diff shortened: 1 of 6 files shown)


tobbi:
Okay, I admit I had no idea what I'm doing. Any idea how to get proper symbol names on stack unwinding when Tux has swallowed a bad fish can?

Or is that a no-*can*-do?

MatusGuy:
On windows I just needed to add a few things and package the pdb file for supertux2.exe

tobbi:
> On windows I just needed to add a few things and package the pdb file for supertux2.exe

So for Windows this is already done? How about MacOS or Linux?

MatusGuy:
SuperTuxKart has an implementation for what you are talking about at https://github.com/supertuxkart/stk-code/blob/master/src/utils/crash_reporting.cpp

Semphriss:
On Linux at least, it looks like a copy-paste of `addr2line`, which is already available on most developer systems that use GCC.

I don't think there is any need to ship debug data, since everything can be obtained locally from the mangled data using local tools like `addr2line` or, preferably, GDB. (You can grab a mangled crash report from a user's Release build and use a local RelWithDebInfo build in GDB to get the file and line info.)

MatusGuy (approved):

Issue #3029, opened 2024-07-30
Big number of compiler warnings


HybridDog:
### SuperTux Version

5a5057859

### System Information

Arch Linux

### Expected Behavior

When compiling SuperTux with -DWARNINGS=ON, there should be no warnings except those caused by myself if I change code.
The CI should fail at Pull Requests if a warning appears, which is documented at [INSTALL.md](https://github.com/SuperTux/supertux/blob/master/INSTALL.md#linuxunix-using-cmake). I think this is broken.

### Actual Behavior

If I compile SuperTux with -DWARNINGS=ON, it shows a lot of warning messages. Most of them are related to simplesquirrel. If I change source code, I want to see warnings to figure out if I have made an accidental mistake. The unrelated warnings make it difficult to find the relevant ones.
Here's a small excerpt of the warnings:
```
[3/356] Building CXX object CMakeFiles/supertux2_lib.dir/src/badguy/captainsnowball.cpp.o
In file included from external/simplesquirrel/include/simplesquirrel/vm.hpp:3,
                 from src/squirrel/squirrel_environment.hpp:24,
                 from src/supertux/sector_base.hpp:22,
                 from src/supertux/sector.hpp:20,
                 from src/badguy/captainsnowball.cpp:20:
external/simplesquirrel/include/simplesquirrel/exceptions.hpp:12:11: warning: 'class ssq::Exception' has pointer data members [-Weffc++]
   12 |     class Exception: public std::exception {
      |           ^~~~~~~~~
external/simplesquirrel/include/simplesquirrel/exceptions.hpp:12:11: warning:   but does not declare 'ssq::Exception(const ssq::Exception&)' [-Weffc++]
external/simplesquirrel/include/simplesquirrel/exceptions.hpp:12:11: warning:   or 'operator=(const ssq::Exception&)' [-Weffc++]
external/simplesquirrel/include/simplesquirrel/exceptions.hpp:14:21: note: pointer member 'ssq::Exception::vm' declared here
   14 |         HSQUIRRELVM vm;
      |                     ^~
[4/356] Building CXX object CMakeFiles/supertux2_lib.dir/src/badguy/crystallo.cpp.o
In file included from external/simplesquirrel/include/simplesquirrel/vm.hpp:3,
                 from src/squirrel/squirrel_environment.hpp:24,
                 from src/supertux/sector_base.hpp:22,
                 from src/supertux/sector.hpp:20,
                 from src/badguy/crystallo.cpp:20:
[...]

PR #3042, opened 2024-08-09
Unify WarningFlags.cmake by checking if compiler supports warning flag

tobbi:
Instead of having different warning flags per compiler, check for each flag if the compiler supports it and then add it


tobbi:
#3029 for reference

tobbi:
https://www.foonathan.net/2018/10/cmake-warnings/
```
The trick is to use target_include_directories(my_library SYSTEM PUBLIC include/). The SYSTEM turns the include directory into a system include directory. Compilers will not issue warnings from header files originating from there.
```

PR #3047, opened 2024-08-21
Various crusher fixes

### LINKED ISSUE #3024: Sideways Crusher should not insta-kill players
A normal Crusher when crushing the player will only hurt them. Sideways Crushers on the other hand kill the player instantaneously when they get crushed at a wall. This is rather unfair as well as inconsistent amongst Crushers thus should be changed

TL;DR - When a player gets hit/crushed by a sideways Crusher, instead of dying instantly, they should only get hurt (i.e. loose a powerup and only die when hit as small Tux)

tobbi:
This also includes some code style improvements to the `Crusher::collision` and relieves some of the mental pain I have when looking at the SuperTux source code.

Fixes #3024

### DIFF
--- a/src/badguy/crusher.cpp
+++ b/src/badguy/crusher.cpp
@@ -652,6 +652,7 @@ Crusher::eye_position(bool right) const
     }
     break;
   case RECOVERING:
+  {
     auto amplitude = static_cast<float>(m_sprite->get_width()) / 64.0f * 2.0f;
     // Eyes spin while crusher is recovering, giving a dazed impression.
     return Vector(sinf((right ? 1 : -1) * // X motion of each eye is opposite of the other.
@@ -665,6 +666,7 @@ Crusher::eye_position(bool right) const
         (m_ic_size == NORMAL ? RECOVER_SPEED_NORMAL : RECOVER_SPEED_LARGE) + m_cooldown_timer * 13.0f) * //Phase factor due to cooldown timer.
       amplitude -  // Amplitude dependent on size.
       amplitude); // Offset to keep eyes visible.
+  }
   default:
     log_debug << "Crusher in invalid state" << std::endl;
     break;
--- a/src/badguy/crusher.cpp
+++ b/src/badguy/crusher.cpp
@@ -123,31 +123,40 @@ HitResponse
 Crusher::collision(GameObject& other, const CollisionHit& hit)
 {
   auto* player = dynamic_cast<Player*>(&other);
+  bool crushed_bottom = m_state == CRUSHING && !m_sideways && hit.bottom;
+  bool crushed_sideways = m_state == CRUSHING && m_sideways &&
+    ((hit.left && m_physic.get_velocity_x() < 0.f) ||
+     (hit.right && m_physic.get_velocity_x() > 0.f));
+  bool is_crushing = crushed_bottom || crushed_sideways;
 
   // If the other object is the player, and the collision is at the
   // bottom of the crusher, hurt the player.
-  if (player && hit.bottom && player->on_ground() && m_state == CRUSHING) {
+  if (player && is_crushing && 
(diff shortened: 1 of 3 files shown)


bruhmoent:
Will test tm.

tobbi:
I've added a fix for an issue found by frost, where roots would spawn in the wrong direction.

tobbi:
@bruhmoent Could you test as well?

bruhmoent:
Yes. I will soon.

Rusty-Box (approved):

tobbi:
I moved some code out but I don't know what `phase_factor` or `amplitude` do. So, well...

tobbi:
@Vankata453 @MatusGuy Can you guys test the code changes and merge maybe?

src/object/player.cpp
```diff
@@ -2324,8 +2325,13 @@ Player::collision_solid(const CollisionHit& hit)
...
   if (hit.crush) {
-    if (hit.left || hit.right) {
-      kill(true);
+    if ((hit.left || hit.right)) {
+      if (m_ignore_sideways_crush) {
+        m_ignore_sideways_crush = false;
+      }
+      else {
+        kill(true);
+      }
```
MatusGuy, line 2334: If tux isn't killed immediately when crushed vertically, then why can't it also happen here? This way you remove the ignore_sidways_crush thing. But I might be missing something

src/badguy/crusher.cpp
```diff
@@ -637,18 +652,27 @@ Crusher::eye_position(bool right) const
...
+      amplitude, // Offset to keep eyes visible.
 
       cosf((right ? 3.1415f : 0.0f) + // Eyes spin out of phase of eachother.
-      (!m_sideways ? get_pos().y / 13 : get_pos().x / 13) - // Phase factor due to y position.
-        (m_ic_size == NORMAL ? RECOVER_SPEED_NORMAL : RECOVER_SPEED_LARGE) + m_cooldown_timer * 13.0f) * //Phase factor due to cooldown timer.
-      static_cast<float>(m_sprite->get_width()) / 64.0f * 2.0f -  // Amplitude dependent on size.
-      static_cast<float>(m_sprite->get_width()) / 64.0f * 2.0f); // Offset to keep eyes visible.
+      phase_factor) * amplitude - 
+      amplitude); // Offset to keep eyes visible.
+  }
```
MatusGuy, line 675: Halelujah thanks for fixing this.

src/badguy/crusher.cpp
```diff
@@ -123,31 +123,40 @@ HitResponse
...
 {
   auto* player = dynamic_cast<Player*>(&other);
+  bool crushed_bottom = m_state == CRUSHING && !m_sideways && hit.bottom;
+  bool crushed_sideways = m_state == CRUSHING && m_sideways &&
+    ((hit.left && m_physic.get_velocity_x() < 0.f) ||
+     (hit.right && m_physic.get_velocity_x() > 0.f));
+  bool is_crushing = crushed_bottom || crushed_sideways;
 
   // If the other object is the player, and the collision is at the
   // bottom of the crusher, hurt the player.
```
Vankata453, line 133: This comment doesn't reflect the changes (sideways collision also now hurts the player).

src/object/player.cpp
Vankata453, line 2334: Couldn't we make it so that despite of the side where Tux was hit from, `kill(false)` is performed? Is there a reason for sideways crushing to instantly kill Tux with `kill(true)`?

If we do that, `m_ignore_sideways_crush` wouldn't be needed.

Vankata453 (commented):
Tested the sideways crushing behaviour, sometimes gets Tux stuck in the wall when crushed, but he's invincible so it should be fine. I'm just not sure if that might open the door for some glitchy skips in levels.

PR #3051, opened 2024-08-29
Add scripting function for turning player invincible (without star)

MatusGuy:
This property uses none of the effects from the star bonus while still maintaining the player invincible.


Alasdairbugs (approved):
works as far as i can tell

PR #3055, opened 2024-09-03
Fix filled background properties

MatusGuy:
An issue introduced by #2748 didn't allow filled backgrounds to have color, alpha or blend. Now they can.

This also allows you to fade backgrounds with fill on.


Frostwithasideofsalt (approved):
Tested, it works

PR #3062, opened 2024-09-13
Allow picking up items while crawling

### LINKED ISSUE #2979: Picking up stuff while crawling
Tux should be able to pick up stuff while he's crawling, because he can otherwise get trapped or softlocked.  Originally reported by Sevenberry in #2846.

marbens-arch:
Allows picking up items (such as springs* and stone blocks), while Tux is crawling (triggered by moving while ducking).

This is to avoid Tux getting trapped by any item that can be grabbed.

Closes #2979.

### DIFF
--- a/src/object/player.cpp
+++ b/src/object/player.cpp
@@ -1736,7 +1736,7 @@ Player::position_grabbed_object(bool teleport)
 bool
 Player::try_grab()
 {
-  if (m_controller->hold(Control::ACTION) && !m_grabbed_object && !m_duck && !m_released_object)
+  if (m_controller->hold(Control::ACTION) && !m_grabbed_object && !(m_duck ^ m_crawl) && !m_released_object)
   {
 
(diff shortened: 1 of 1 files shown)


src/object/player.cpp
```diff
@@ -1736,7 +1736,7 @@ Player::position_grabbed_object(bool teleport)
 bool
 Player::try_grab()
 {
-  if (m_controller->hold(Control::ACTION) && !m_grabbed_object && !m_duck && !m_released_object)
+  if (m_controller->hold(Control::ACTION) && !m_grabbed_object && !(m_duck ^ m_crawl) && !m_released_object)
```
MatusGuy, line 1739: haven't seen an xor in a while

MatusGuy (commented):
will test later

Frostwithasideofsalt (approved):

PR #3070, opened 2024-10-05
Item Pocket

MatusGuy:
The Item Pocket allows you to save a powerup for later use. If you collect a flower while having a bonus greater than `GROWUP_BONUS`, that flower gets equipped and the old flower gets stored in the top left corner of the screen. Now, the player can press the new `ITEM` control (usually Select/Back or Left Shift) to use the stored flower by throwing it up and catching it.

This pull request replaces the powerup stacking feature because this new solution is much more balanced. It also limits the amount of concurrent players to 4.


tobbi:
Please fix initialization order:
```
/Users/runner/work/supertux/supertux/src/supertux/player_status.cpp:38:3: error: field 'bonus' will be initialized after field 'm_item_pockets' [-Werror,-Wreorder-ctor]
  bonus(num_players),
  ^~~~~~~~~~~~~~~~~~
  coins(START_COINS)
```

tobbi:
Oh, right, this is a draft, so my comments were maybe a bit premature.

tobbi:
```
/home/runner/work/supertux/supertux/src/badguy/boss.cpp:76:70: error: implicit conversion from 'int' to 'float' may lose precision [-Werror,-Wimplicit-int-float-conversion]
    float startpos = (context.get_width() - (m_hud_head->get_width() * m_max_lives)) / 2;
```

swagtoy:
Has the ability to disable the itembox been discussed?

bruhmoent:
> Has the ability to disable the itembox been discussed?

What do you mean by disabling it...? It's a core gameplay feature

swagtoy:
> What do you mean by disabling it...? It's a core gameplay feature

Pardon? I meant that some level designers may want to disable it for themselves (or per world). I am curious on the consensus. I personally wouldnt enable it for my worlds

Frostwithasideofsalt (approved):
works, doesnt twerks  👩‍🔬

Rusty-Box:
Found tiny bug: after I played through a level and went back to the worldmap, the item pocket visually disappears while on the worldmap. It does come back when entering a level, but goes away again afterwards.

src/control/game_controller_manager.cpp
Vankata453, line 1: You can remove the `sector.hpp` include.

src/control/joystick_manager.cpp
Vankata453, line 1: You can remove the `sector.hpp` include.

src/supertux/player_status.cpp
```diff
@@ -233,32 +257,87 @@ PlayerStatus::read(const ReaderMapping& mapping)
...
+  if (bonustype == NO_BONUS)
+    return;
+
+  m_item_pockets[player->get_id()] = NO_BONUS;
+
+  auto& powerup = Sector::get().add<PocketPowerUp>(bonustype, Vector(0,0));
+  Vector pos = player->get_pos();
+  pos.x = player->get_bbox().get_left();
+  pos.y = player->get_bbox().get_top() - powerup.get_bbox().get_height() - 15;
+  powerup.set_pos(pos);
```
Vankata453, line 276:
```suggestion
  powerup.set_pos(player->get_pos() - Vector(0.f, powerup.get_bbox().get_height() + 15));
```

src/supertux/player_status.cpp
```diff
@@ -233,32 +257,87 @@ PlayerStatus::read(const ReaderMapping& mapping)
...
+      return true;
+    }
+  }
+
+  Level* level = Level::current();
+  int allowed = static_cast<Level::Setting>(level->m_allow_item_pocket);
+
+  if (allowed != Level::INHERIT)
+  {
+    return allowed == Level::ON ? true : false;
```
Vankata453, line 313:
```suggestion
    return allowed == Level::ON;
```

src/supertux/player_status.cpp
```diff
@@ -233,32 +257,87 @@ PlayerStatus::read(const ReaderMapping& mapping)
...
+      return worldmap->is_item_pocket_allowed();
+    }
+    else
+    {
+      // This level is probably in a levelset, pick ON.
+      return true;
+    }
+  }
+
+  Level* level = Level::current();
```
Vankata453, line 308: Why not `GameSession::current()->get_current_level()`? `Level::current()` is funky IMO. Also you checked if `GameSession::current()` is valid beforehand.

src/supertux/player_status.cpp
```diff
@@ -301,17 +377,55 @@ PlayerStatus::remove_player(int player_id)
...
-  max_air_time.resize(m_num_players, 0);
-  max_earth_time.resize(m_num_players, 0);
+  m_item_pockets.resize(m_num_players, NO_BONUS);
+}
+
+PlayerStatus::PocketPowerUp::PocketPowerUp(BonusType bonustype, Vector pos):
+  PowerUp(pos, PowerUp::get_type_from_bonustype(bonustype)),
+  m_cooldown_timer(),
+  m_blink_timer(),
+  m_visible(true)
```
Vankata453, line 391: Simply suggesting another newline here for some separation from the `PlayerStatus` class.

```suggestion


PlayerStatus::PocketPowerUp::PocketPowerUp(BonusType bonustype, Vector pos):
  PowerUp(pos, PowerUp::get_type_from_bonustype(bonustype)),
  m_cooldown_timer(),
  m_blink_timer(),
  m_visible(true)
```

src/supertux/player_status.hpp
```diff
@@ -38,6 +42,7 @@ enum BonusType {
     example when switching maps in the worldmap) */
 class PlayerStatus final
 {
+
```
Vankata453, line 45:
```suggestion
```

src/supertux/player_status.hpp
```diff
@@ -23,9 +23,13 @@
 #include <string>
 #include <vector>
 
+#include "object/powerup.hpp"
+#include "supertux/timer.hpp"
+
 class DrawingContext;
 class ReaderMapping;
 class Writer;
+class Player;
```
Vankata453, line 32:
```suggestion
class DrawingContext;
class Player;
class ReaderMapping;
class Writer;
```

src/supertux/player_status_hud.cpp
```diff
@@ -16,12 +16,19 @@
...
 #include "supertux/player_status.hpp"
 #include "supertux/resources.hpp"
+#include "supertux/title_screen.hpp"
+#include "supertux/screen_manager.hpp"
 #include "video/drawing_context.hpp"
 #include "video/surface.hpp"
 #include "editor/editor.hpp"
+#include "worldmap/worldmap_sector.hpp"
+
+#include <iostream>
```
Vankata453, line 31:
```suggestion

#include <iostream>

#include "sprite/sprite_manager.hpp"
#include "supertux/game_object.hpp"
#include "supertux/level.hpp"
#include "supertux/player_status.hpp"
#include "supertux/resources.hpp"
#include "supertux/title_screen.hpp"
#include "supertux/screen_manager.hpp"
#include "video/drawing_context.hpp"
#include "video/surface.hpp"
#include "editor/editor.hpp"
#include "worldmap/worldmap_sector.hpp"
```

src/supertux/sector.cpp
```diff
@@ -589,6 +589,19 @@ Sector::is_free_of_specifically_movingstatics(float left, float top, float right
   return m_collision_system->is_free_of_specifically_movingstatics(Rectf(Vector(left, top), Vector(right, bottom)), nullptr);
 }
 
+void
+Sector::set_allow_item_pocket(bool allow)
+{
+  //FIXME
+  //m_level.m_allow_item_pocket = allow;
+}
```
Vankata453, line 597: Hasn't been fixed.

src/supertux/sector.hpp
```diff
@@ -176,6 +176,19 @@ class Sector final : public Base::Sector
...
+   * @description Allows/disallows usage of the Item Pocket.
+   * @param bool $allow
+   */
+  void set_allow_item_pocket(bool allow);
+
+  /**
+   * @scripting
+   * @description Checks if the Item Pocket is currently allowed.
+   */
+  bool is_item_pocket_allowed();
```
Vankata453, line 190: Missing `const` qualifier.

Vankata453 (changes requested):
Overall, amazing work on this PR!

**Encountered a bug:** The "Add Player" option in "Multiplayer Settings" does nothing.

Additionally, I believe that launching the powerup from the item pocket to the side (like a Tuxdoll), perhaps to the opposite side of where Tux is facing, would make utilizing the powerup from the item pocket require more concentration and consideration of where to utilize, since the player would also have to catch it, instead of waiting for it in the same spot.

I am also personally not a fan of the blinking effect that launched pocket powerups have. Maybe we could do without it?

MatusGuy:
The blinking effect indicates that you can't pick it up yet. This is a mechanism that prevents immediately grabbing the object upon jumping.

Also, I think the general consensus is that it shouldn't move in x axis whatsoever but I might be wrong so I'll leave this up for discussion.

MatusGuy:
Can you check if power up stacking was truly removed in case I missed something?

src/supertux/player_status.cpp
MatusGuy: Is this value correct when in the editor?

Vankata453: Don't think so.

src/supertux/player_status.hpp
tobbi: Peak change suggestion!

src/squirrel/supertux_api.cpp
```diff
@@ -872,6 +904,9 @@ void register_supertux_scripting_api(ssq::VM& vm)
   level.addFunc("toggle_pause", &scripting::Level::toggle_pause);
   level.addFunc("pause_target_timer", &scripting::Level::pause_target_timer);
   level.addFunc("resume_target_timer", &scripting::Level::resume_target_timer);
+  level.addFunc("override_item_pocket", &scripting::Level::override_allow_item_pocket);
```
Vankata453, line 907: The C++ function name is different. Make sure they both stay the same for accuracy in documentation.

```suggestion
  level.addFunc("override_item_pocket", &scripting::Level::override_item_pocket);
```

src/squirrel/supertux_api.cpp
```diff
@@ -787,6 +787,38 @@ static void resume_target_timer()
   GameSession::current()->set_target_timer_paused(false);
 }
 
+/**
+ * @scripting
+ * @description Override the Item Pocket setting in the Level.
+ * @param string allow Can be "on", "off", or "inherit" (use the setting in the Level)
+ */
+static void override_allow_item_pocket(const std::string& allow)
```
Vankata453, line 795: The C++ function name is different. Make sure they both stay the same for accuracy in documentation.

```suggestion
static void override_item_pocket(const std::string& allow)
```

Vankata453 (commented):
Take a look at the comments below, as well as the following:

* multiplayer_player_menu.cpp:118 uses the same code as `GameSession::on_player_added` so it can be replaced with a call to that function.
* `GameSession::on_player_removed` shouldn't remove the player from `PlayerStatus`, since that will removed all the data for that player, when they may reconnect.
* ~~The "Add Player" bug from "Multiplayer Settings", as mentioned in the previous comment, is still present.~~

src/control/keyboard_config.cpp
```diff
@@ -64,20 +66,6 @@ KeyboardConfig::KeyboardConfig() :
...
-  m_keymap[SDLK_s]         = {1, Control::DOWN};
-  m_keymap[SDLK_e]         = {1, Control::JUMP};
-  m_keymap[SDLK_q]         = {1, Control::ACTION};
-
-  m_keymap[SDLK_j]         = {2, Control::LEFT};
-  m_keymap[SDLK_l]         = {2, Control::RIGHT};
-  m_keymap[SDLK_i]         = {2, Control::UP};
-  m_keymap[SDLK_k]         = {2, Control::DOWN};
-  m_keymap[SDLK_o]         = {2, Control::JUMP};
-  m_keymap[SDLK_u]         = {2, Control::ACTION};
```
Vankata453, line 80: Don't think there's any reason to remove the defaults for players 2 and 3. Perhaps put them back in and set a default control for the Item Pocket?

MatusGuy: > Ok I think I remember why i did this
> Player 1 keybinds are literally incompatible with the rest - just picture it yourself
> You and your friends would be playing a game of twister on your keyboard

bruhmoent (approved):
Nice work - reviewed both the code and the in-game functionality

Frostwithasideofsalt:
image
seems theres a bug with the item pocket that makes it so theres multiple item pockets when theres only one player

Frostwithasideofsalt (approved):
zero twerking found here, bucko! only working.

Frostwithasideofsalt:
being that this seems to be done, could it be merged?

Vankata453 (approved):
Code-wise looks good now!

weluvgoatz (approved):
Works great!

PR #3085, opened 2024-10-21
`GameObject` change undo/redo refactor to track different options

Vankata453:
`GameObject`, instead of comparing all options against each other, now saves the state of and compares individual ones, so the change data is more memory-efficient.

This way of tracking object changes also allows for proper undo/redo tracking with multiple remote users (regarding the [remote level editing concept](https://github.com/SuperTux/supertux/tree/editor-remote-level-networking), which branch these changes originate from).

`TileMap` tile changes are now saved in a way more memory-efficient way. Instead of saving the whole tiles array for each change, only the changed tiles are saved in a list of pairs (tile index, old/new tile ID).

### DIFF
--- a/src/supertux/game_object_change.cpp
+++ b/src/supertux/game_object_change.cpp
@@ -1,0 +1,92 @@
+//  SuperTux
+//  Copyright (C) 2024 Vankata453
+//
+//  This program is free software: you can redistribute it and/or modify
+//  it under the terms of the GNU General Public License as published by
+//  the Free Software Foundation, either version 3 of the License, or
+//  (at your option) any later version.
+//
+//  This program is distributed in the hope that it will be useful,
+//  but WITHOUT ANY WARRANTY; without even the implied warranty of
+//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//  GNU General Public License for more details.
+//
+//  You should have received a copy of the GNU General Public License
+//  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#include "supertux/game_object_change.hpp"
+
+#include "util/log.hpp"
+#include "util/reader_iterator.hpp"
+#include "util/reader_mapping.hpp"
+#include "util/writer.hpp"
+
+GameObjectChange::GameObjectChange(const std::string& name_, const UID& uid_,
+                                   const std::string& data_, const std::string& new_data_,
+                                   Action action_) :
+  name(name_),
+  uid(uid_),
+  data(data_),
+  new_data(new_data_),
+  action(action_)
+{
+}
+
+GameObjectChange::GameObjectChange(const ReaderMapping& reader) :
+  name(),
+  uid(),
+  data(),
+  new_data(),
+  action()
+{
+  reader.get("name", name);
+  reader.get("uid", uid);
+  reader.get("data", data);
+  reader.get("action", reinterpret_cast<int&>(action));
+}
+
+void
+GameObjectChange::save(Writer& writer) const
+{
+  writer.write("name", name);
+  writer.write("uid", uid);
+  writer.write("data", data);
+  writer.write("action", reinterpret_cast<const int&>(action));
+}
+
+
+GameObjectChanges::GameObjectChanges(const UID& uid_, std::vector<GameObjectChange> objects_) :
+  uid(uid_),
+  objects(std::move(objects_))
+{
+}
+
+GameObjectChanges::GameObjectChanges(const ReaderMapping& reader) :
+  uid(),
+  objects()
+{
+  auto iter = reader.get_iter();
+  while (iter.next())
+  {
+    if (iter.get_key() != "object-change")
+    {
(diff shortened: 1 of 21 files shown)


src/supertux/game_object_change.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+#include "util/log.hpp"
+#include "util/reader_iterator.hpp"
+#include "util/reader_mapping.hpp"
+#include "util/writer.hpp"
+
+GameObjectChange::GameObjectChange(const std::string& name_, const UID& uid_,
+                                   const std::string& data_, const std::string& new_data_,
+                                   Action action_) :
+  name(name_),
```
tobbi, line 27: Can you do me a favour and prefix member variables with `m_`? Then you don't need to suffix `_` for the parameters.

Vankata453: The reason I did that is because those are public members, and since the point of these classes is to act more like a structure containing multiple variables, it might look awkward always referring to them as `change.m_...`. If that doesn't seem like an issue to you, I could.

src/supertux/game_object_manager.cpp
```diff
@@ -376,79 +379,233 @@ GameObjectManager::on_editor_save()
...
+
+    default:
+      break;
+  }
+}
+
+void
+GameObjectManager::apply_object_changes(const GameObjectChanges& changes, bool track_undo)
+{
+  for (const auto& change : changes.objects)
```
tobbi, line 429: If you're iterating over the changed objects, shouldn't the variable be called `object`? But then line 433 doesn't make any sense.

Vankata453: The `objects` member of `GameObjectChanges` actually is a `GameObjectChange` vector. So the problem here is more so the naming of that member (maybe I can name it `changes`).

src/supertux/game_object_manager.cpp
```diff
@@ -376,79 +379,233 @@ GameObjectManager::on_editor_save()
...
+  Writer writer(stream);
+
+  writer.start_list("supertux-game-object");
+  if (new_state)
+    settings.save_new_state(writer);
+  else
+    settings.save_old_state(stream);
+  writer.end_list("supertux-game-object");
+
+  return stream.str();
```
tobbi, line 536: I'm quite confused about this. Assuming you call this with new_state = true, nothing gets done with the `stream` parameter. Yet you're returning `stream.str()` from the function. Does that return something usable even though the variable is initialized?

I feel like this is bad code style, but correct me if I'm wrong.

Vankata453: `writer` takes a reference to `stream`. When I pass `writer` to `settings.save_new_state` effectively whatever is written to that writer is written to the stream, as it's the writer's base. So the data always goes there.

tobbi: Can you maybe change both method signatures to take a `Writer` instance? Or is that overkill?

Vankata453: The old state is meant to be pre-saved and not written at the moment of calling, so it wouldn't work there.

tobbi:
MacOS build failure:
```
In file included from /Users/runner/work/supertux/supertux/src/supertux/game_object_change.cpp:17:
/Users/runner/work/supertux/supertux/src/supertux/game_object_change.hpp:65:33: error: implicit instantiation of undefined template 'std::vector<GameObjectChange>'
  std::vector<GameObjectChange> objects;
                                ^
/Applications/Xcode_14.2.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX13.1.sdk/usr/include/c++/v1/iosfwd:259:28: note: template is declared here
class _LIBCPP_TEMPLATE_VIS vector;
```

Windows build failure (maybe if the logs don't deceive me):
https://github.com/SuperTux/supertux/actions/runs/11448138595/job/31851014184?pr=3085#step:6:4092

src/supertux/game_object_manager.cpp
tobbi: How about giving both a reference to `stream`? IDK...

src/supertux/game_object_manager.cpp
tobbi: How about renaming the class to `GameObjectChangeSet`. Then you can rename `objects` to `changes` and keep the UID for identifying that excrement.

src/supertux/game_object_manager.cpp
Vankata453: I just realized that doing that would complicate this more, as the writer already must be created for the "supertux-game-object" header to be added.

tobbi: Okay, even disregard my comments

Frostwithasideofsalt (approved):

src/supertux/game_object_change.cpp
tobbi: Okay, even though every single cell in my body is bothered by this, I'll let it be.

PR #3096, opened 2024-11-09
Update collision.cpp

Iceyshell:
Fix ceiling slope hang so which was fixed by others too.However,Ensure that the code is consistent in determining the validity of the p1 coordinate relative to the area range

### DIFF
--- a/src/collision/collision.cpp
+++ b/src/collision/collision.cpp
@@ -129,7 +129,7 @@ bool rectangle_aatriangle(Constraints* constraints, const Rectf& rect,
 
   Vector outvec = normal * (depth + 0.2f);
 
-  const float RDELTA = 3;
+  const float RDELTA = 1.5f;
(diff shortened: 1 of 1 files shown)


tobbi:
@SuperTux/developers I'm not too familiar with our collision system. Will changing rdelta negatively affect anything else?

MatusGuy (commented):
Seems like this PR does the exact same as #3057 and iirc it didn't work out.

Iceyshell:
I reworked it and was able to implement it locally, so hopefully it will be useful for the main supertux branch!

Frostwithasideofsalt:
image
image

okay, so. this pr works... but not really. it makes it *less common?* but it still happens.

Frostwithasideofsalt:
i'll hold off on approving until it's fully fixed

MatusGuy (commented):
I'm not sure what to do about this. I don't think changing hacky values is the way to go here, especially when I'm currently rewriting this function (#3140)

PR #3099, opened 2024-11-15
Update sexp-cpp submodule to latest upstream version

tobbi:
This updates the sexp-cpp submodule to the revision at upstream/master. Unfortunately, because the upstream is at C++ 20, I had to revert the commit that pushed it that far, because we still compile with 17.


MatusGuy:
Will be fixed in #3093. Can I close this?

tobbi:
> Will be fixed in #3093. Can I close this?

How far along is it?

MatusGuy:
It's already done in the pr but the pr itself isnt done. I still need to finish fixing all the CI and you need to test macos

PR #3100, opened 2024-11-15
Color Clipboard

bruhmoent:
This PR adds three new options to the Color Menu:
- Copy to clipboard (rgb),
- Copy to clipboard (hex),
- Paste from clipboard

These options enable users to copy and paste colors between objects with the color menu via the SDL's clipboard feature.


src/gui/color_clipboard.hpp
```diff
@@ -0,0 +1,45 @@
...
+//  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#ifndef HEADER_SUPERTUX_GUI_COLOR_CLIPBOARD_HPP
+#define HEADER_SUPERTUX_GUI_COLOR_CLIPBOARD_HPP
+
+#include <memory>
+
+#include "video/color.hpp"
+
+class ColorClipboard
```
Vankata453, line 24: Does this need to be a class? The same thing could be done with a static variable under the `Color` class.

src/gui/menu_color.cpp
```diff
@@ -31,13 +34,33 @@ ColorMenu::ColorMenu(Color* color_) :
...
 }
 
 void
 ColorMenu::menu_action(MenuItem& item)
 {
+  if (item.get_id() == 1)
+  {
+    clipboard.set_color(*color);
+    MenuItem& menu_paste_item = get_item_by_id(2);
+    if (&menu_paste_item)
```
Vankata453, line 55: Not needed; references cannot be null.

src/gui/color_clipboard.hpp
bruhmoent: Could be expanded on later.

Vankata453: I'm not sure what other purpose a color clipboard could serve.

src/gui/menu_color.cpp
```diff
@@ -31,13 +33,32 @@ ColorMenu::ColorMenu(Color* color_) :
   add_color_channel_rgba(&(color->alpha), Color::BLACK, -1, true);
   add_color_display(color);
 
+  add_hl();
+  add_item(std::make_unique<MenuItem>(_("Copy"), 1));
+  add_item(std::make_unique<MenuItem>(
```
Vankata453, line 38: Use `add_entry` here for consistency (the color argument from the constructor should be available in `Menu::add_entry` too). Also maybe MNID_ enumerators like other menus?

MatusGuy:
What is this? Please describe this feature in the pr description.

Frostwithasideofsalt (approved):

MatusGuy (commented):
Does this copy to an in-game clipboard or the system clipboard? This feature might be confusing for users who want to copy/paste a color to outside of supertux.

You could parse a hex color value from the system clipboard which in my opinion would provide a better experience.

You can use [SDL's clipboard functionality](https://wiki.libsdl.org/SDL2/SDL_GetClipboardText) to do this.

bruhmoent:
I can do that, sure. I'm still gonna keep the current in game functionality, but I'm gonna also add parsing it from the clipboard upon clicking the paste entry.

MatusGuy (commented):
Nice work

src/gui/menu_color.cpp
```diff
@@ -14,30 +15,108 @@
...
+  add_entry(MNID_COPY_CLIPBOARD_HEX, _("Copy to clipboard (hex)"));
+  if (SDL_HasClipboardText())
+  {
+    const char* clipboard_text = SDL_GetClipboardText();
+    std::optional<Color> clipboard_color;
+
+    if (clipboard_text)
+    {
+      const std::string text(clipboard_text);
+      SDL_free(const_cast<char*>(clipboard_text));
```
Vankata453, line 50: Do you have to make `clipboard_text` `const` when declaring it above?

src/gui/menu_color.cpp
```diff
@@ -14,30 +15,108 @@
...
+  }
+  else
+    add_entry(MNID_PASTE_CLIPBOARD, _("Paste from clipboard"), Color(1.f, 1.f, 1.f));
 
   add_hl();
   add_back(_("OK"));
 }
 
+void
+ColorMenu::set_clipboard_and_update_paste_item(const std::string& color_str)
```
Vankata453, line 67: Could be renamed to something like `copy_color` since the name is too long.

src/video/color.cpp
```diff
@@ -14,9 +14,13 @@
 //  You should have received a copy of the GNU General Public License
 //  along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+#include "../math/util.hpp"
```
Vankata453, line 17:
```suggestion
#include "math/util.hpp"
```

src/video/color.hpp
```diff
@@ -81,6 +82,12 @@ class Color final
                  static_cast<float>(a) / 255.0f);
   }
 
+  static std::optional<Color> deserialize_color_from_rgb(const std::string& rgb_string);
+  static std::optional<Color> deserialize_color_from_hex(const std::string& hex_string);
+
+  static std::string serialize_color_to_rgb(const Color& color);
+  static std::string serialize_color_to_hex(const Color& color);
```
Vankata453, line 89: Just suggestions, change if you feel like it:
1. `_color_` doesn't seem needed in the function names.
2. The serialize functions could be non-`static`, but that doesn't matter much.

Vankata453 (approved):
Looking good code-wise.

src/gui/menu_item.hpp
```diff
@@ -86,6 +88,7 @@ class MenuItem
   std::string m_text;
   std::string m_help;
   FontPtr m_font;
+  std::optional<Color> m_text_color;
```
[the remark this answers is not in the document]
MatusGuy: Hey, why was this marked as resolved?

src/gui/item_action.cpp
```diff
@@ -16,17 +16,19 @@
...
 {
-  if (action == MenuAction::HIT) {
-    if (m_callback) {
+  if (action == MenuAction::HIT)
+  {
+    if (m_callback)
+    {
       m_callback();
     }
   }
```
MatusGuy, line 34:
```suggestion
  if (action == MenuAction::HIT && m_callback)
  {
    m_callback();
  }
```

src/gui/menu_item.hpp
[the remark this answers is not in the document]
bruhmoent: It is a rare circumstance that a menuitem will need a manually given colour

MatusGuy (approved):

src/gui/item_action.hpp
```diff
@@ -22,7 +22,7 @@
 class ItemAction final : public MenuItem
 {
 public:
-  ItemAction(const std::string& text, int id = -1, std::function<void()> callback = {});
+  ItemAction(const std::string& text, int id = -1, std::function<void()> callback = {}, const Color& text_color = Color(1.f, 1.f, 1.f));
```
tobbi, line 25: Don't we have a Color::WHITE constant somewhere?

src/gui/menu_color.cpp
```diff
@@ -14,30 +15,108 @@
...
+
+      clipboard_color = Color::deserialize_from_rgb(text);
+      if (!clipboard_color)
+        clipboard_color = Color::deserialize_from_hex(text);
+    }
+
+    add_entry(MNID_PASTE_CLIPBOARD, _("Paste from clipboard"), clipboard_color.value_or(Color(1.f, 1.f, 1.f)));
+  }
+  else
+    add_entry(MNID_PASTE_CLIPBOARD, _("Paste from clipboard"), Color(1.f, 1.f, 1.f));
```
tobbi, line 60: What's the point to add an item called "Paste from clipboard" here when SDL doesn't support the functionality?

bruhmoent: SDL_GetClipboardText(). What do you mean???

tobbi: Never mind. I misunderstood the meaning of `SDL_HasClipboardText()`.

HybridDog:
> static_cast<int>(color.red * 255.f)

@bruhmoent, this looks slightly wrong to me. Color contains floating point numbers and this code always rounds down. For example, if red is very close to but slightly below 1, it becomes 254 and not 255 (e.g. in `static_cast<int>(254.999f)`).
Since sRGB is non-linear, I think for rounding based on physical brightness, the two choices for rounding up and down need to be tested, e.g. (dummy code):
```
float srgb_to_linear(float channel) {
	if (channel <= 0.04045)
		return channel / 12.92;
	else
		return pow((channel + 0.055) / (1.0 + 0.055), 2.4);
}

u8 color_to_u8(float v) {
	float v_lin = srgb_to_linear(v);
	u8 candidate_1 = static_cast<int>(color.red * 255.f);
	u8 candidate_2 = candidate_1 + 1;
	float error_1 = v_lin - srgb_to_linear(candidate_1 / 255.f);
	float error_2 = srgb_to_linear(candidate_2 / 255.f) - v_lin;
	return error_1 <= error_2 ? candidate_1 : candidate_2;
}

u8 red_int = color_to_u8(color.red);
```
I don't know if rounding based on physical brightness is a good idea or if doing it directly in sRGB is better.

PR #3102, opened 2024-11-16
Fix file path for particle editor

tobbi:
This uses the FileSystem methods for combining the path elements instead of some weird homemade stuff.


swagtoy:
Bahaha, looks good!

On a related note: We should move to std::filesystem soon, unless there are any objections. Feels okay to discuss here real quick

tobbi:
Superseded by #3137

PR #3104, opened 2024-11-17
Change all "normal" actions to "default"

MatusGuy:
This will absolutely remove all sprite file inconsistencies between "normal" and "default". Seriously, what were they thinking!

Now that all files are consistent there should be no more weird cases with objects showing wrong actions or giving cryptic warnings.

Testers: Please test badguys like granito, goldbomb, etc. thoroughly

### DIFF
--- a/data/images/creatures/fish/forest/jumpfish.sprite
+++ b/data/images/creatures/fish/forest/jumpfish.sprite
@@ -1,6 +1,6 @@
 (supertux-sprite
   (action
-    (name "normal")
+    (name "default")
     (fps 18)
     (loops 1)
     (hitbox 12 24 40 46)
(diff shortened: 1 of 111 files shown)


weluvgoatz (approved):
Tested, reviewed code, works good and fine.

MatusGuy:
Fixed that just now

weluvgoatz:
I just noticed that the bumper has the same issue.

MatusGuy:
> I just noticed that the bumper has the same issue.

Can't reproduce this issue

Rusty-Box:
I can confirm the bumper issue. They work upon first time interacting with them but any subsequent time they will be stuck. Also bumpers facing left won't do so unless interacted with once.

swagtoy (approved):
Everything in this change seems consistent. Code looks correct.

PR #3109, opened 2024-11-22
Prevent Tux from wanting to take baths in lava

tobbi:
I propose the following code change to prevent Tux from swimming in lava baths unless he has the invincibility star. That means that once Tux enters lava, he'll fall right through. During invincibility, Tux will swim as usual.


MatusGuy (approved):

Frostwithasideofsalt:
not too big of a fan of this change. i prefer having tux being able to swim in it always, so he has a chance to get out if he falls in.

Vankata453:
IMO letting Tux swim in the lava could allow for some interesting level gimmicks and is also more logical:

* This is inconsistent with how (at least I have perceived) the swimming mechanic logically allows Tux to swim in liquids.
* If Tux shouldn't swim in the lava, the other way it could be done it is to kill him instantly whenever he touches it, but that is then inconsistent with how powerups work.

tobbi:
Closing this as per concensus.

PR #3115, opened 2024-11-29
(cherry pick) Don't call GameSession::restart_level in the constructor

swagtoy:
This stops testing (on the "Test from here")/the title screen (yes, the title screen) from reparsing the level by simply changing the usage pattern of the GameSession constructor.

Cherry picked from commit https://github.com/swagtoy/supertux/commit/0d25b71d0b92d1f2feedcc6735b581fed34dbeaa upon the request of Vankata.

### DIFF
--- a/src/supertux/game_session.cpp
+++ b/src/supertux/game_session.cpp
@@ -249,7 +249,7 @@ GameSession::restart_level(bool after_death, bool preserve_music)
     }
   }
   catch (std::exception& e) {
-    throw std::runtime_error(fmt::format("Couldn't start level: {}", e.what()));
+    throw std::runtime_error(std::string("Couldn't start level: ") + e.what());
     ScreenManager::current()->pop_screen();
   }
 
--- a/src/supertux/game_session.cpp
+++ b/src/supertux/game_session.cpp
@@ -17,6 +17,8 @@
 #include "supertux/game_session.hpp"
 
 #include <cfloat>
+#include <fmt/format.h>
+#include <stdexcept>
 
 #include "audio/sound_manager.hpp"
 #include "control/input_manager.hpp"
@@ -151,7 +153,7 @@ GameSession::on_player_removed(int id)
   return false;
 }
 
-int
+void
 GameSession::restart_level(bool after_death, bool preserve_music)
 {
   const PlayerStatus& currentStatus = m_savegame.get_player_status();
@@ -234,7 +236,7 @@ GameSession::restart_level(bool after_death, bool preserve_music)
     m_currentsector = m_level->get_sector(spawnpoint->sector);
     if (!m_currentsector)
     {
-      throw std::runtime_error("Couldn't find sector '" + spawnpoint->sector + "' to spawn/respawn Tux.");
+      throw std::runtime_error(fmt::format("Couldn't find sector '{}' to spawn/respawn Tux.", spawnpoint->sector));
     }
     // Activate on either the spawnpoint (if set), or the spawn position.
     if (spawnpoint->spawnpoint.empty())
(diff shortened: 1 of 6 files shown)


swagtoy:
wait hold on there's a regression where you can control Tux from the title screen

swagtoy:
Okay, seems to work!

Later i'll actually end up probably creating an `start_game()` function, but this will unfortunately be part of the Level editor PR for now as I just don't have the time to cherry pick stuff and fix it only to stub my foot later.

src/supertux/levelset_screen.cpp
```diff
@@ -27,6 +27,8 @@
 #include "util/file_system.hpp"
 #include "util/log.hpp"
 
+// TODO: wtf?????????
+
```
Vankata453, line 31: You'll fix this anyway, not needed.

src/supertux/game_session.cpp
```diff
@@ -94,9 +94,6 @@ GameSession::GameSession(const std::string& levelfile_, Savegame& savegame, Stat
   m_boni_at_start.resize(InputManager::current()->get_num_users(), BONUS_NONE);
 
   m_data_table.clear();
-
-  if (restart_level(false, preserve_music) != 0)
```
Vankata453, line 98: Since `preserve_music` is no longer used here, it should be removed from the constructor.

src/supertux/title_screen.cpp
```diff
@@ -133,6 +135,7 @@ TitleScreen::refresh_level()
   else if (!m_titlesession || m_titlesession->get_level_file() != DEFAULT_TITLE_LEVEL)
   {
     m_titlesession = std::make_unique<GameSession>(DEFAULT_TITLE_LEVEL, m_savegame, nullptr, true);
+    m_titlesession->restart_level();
```
Vankata453, line 138: Pass `false, preserve_music` as arguments.

src/supertux/levelset_screen.cpp
swagtoy: Ha, this was from my cherry picked commit, my bad :)

src/supertux/game_session.cpp
```diff
@@ -247,9 +249,8 @@ GameSession::restart_level(bool after_death, bool preserve_music)
     }
   }
   catch (std::exception& e) {
-    log_fatal << "Couldn't start level: " << e.what() << std::endl;
+    throw std::runtime_error(fmt::format("Couldn't start level: {}", e.what()));
```
Vankata453, line 252: You don't need FMT for this.

swagtoy:
Seems good now. I did test some things and it doesn't crash and burn. Any final words?

Vankata453:
```
src/supertux/game_session.cpp:253:5: style: Statements following 'throw' will never be executed. [unreachableCode]
    ScreenManager::current()->pop_screen();
    ^
```

src/supertux/levelset_screen.cpp
```diff
@@ -88,8 +88,8 @@ LevelsetScreen::setup()
                                                   m_savegame);
       if (m_start_pos) {
         screen->set_start_pos(m_start_pos->first, m_start_pos->second);
-        screen->restart_level();
       }
+      screen->restart_level();
```
Vankata453, line 92: We can put a try/catch here so the game doesn't crash if level loading fails.

src/worldmap/worldmap_sector.cpp
```diff
@@ -354,10 +354,13 @@ WorldMapSector::update(float dt_sec)
           Vector shrinkpos = Vector(level_->get_pos().x + 16 - m_camera->get_offset().x,
                                     level_->get_pos().y +  8 - m_camera->get_offset().y);
           std::string levelfile = m_parent.m_levels_path + level_->get_level_filename();
+          
+          auto gamesession = std::make_unique<GameSession>(levelfile, m_parent.m_savegame, &level_->get_statistics());
+          gamesession->restart_level();
```
Vankata453, line 359: We can put a try/catch here so the game doesn't crash if level loading fails.

swagtoy:
RE: `src/worldmap/worldmap_sector.cpp`

Your change here would've broken logic as we never pushed anything on the screen to begin with.

swagtoy:
(commenting in a regular comment as Github is broken right now)

Vankata453 (approved):

PR #3116, opened 2024-11-29
[Draft] Refactor Player.cpp

swagtoy:
About damn time.

Highlights:
- Moved Player to MovingSprite, now you can use Squirrel `Tux.set_sprite` calls and have fun with that. Custom sprites are possible now. This was so easy to do it hurts.

Main things on my bucketlist are refactoring the majority of state into either a) their own smaller classes, b) into more functions, and c) maybe bitwise state and/or std::bitset. The input code is kind of disgusting and could be set up some. State machinery stuff related to switching the players sprite likely has room for improvement. Update function is just putrid. Input/shooting/powerup code MIGHT be able to get refactored out of player, i'll see what I can do :)

### DIFF
--- a/src/object/player.cpp
+++ b/src/object/player.cpp
@@ -22,6 +22,7 @@
 
 #include "audio/sound_manager.hpp"
 #include "badguy/badguy.hpp"
+#include "collision/collision_group.hpp"
 #include "control/codecontroller.hpp"
 #include "control/input_manager.hpp"
 #include "editor/editor.hpp"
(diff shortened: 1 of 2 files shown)


MatusGuy:
I think this pr is cool .

swagtoy:
Unrelated note but kind of surprised git doesnt allow for true commit renaming in 2024 because "Oh no" was not a good commit message and force pushing is not a great decision

swagtoy:
State machine code is done, though right now, MovingSprite has broken Tux's logic (still looking into why) but its likely due to disgusting jank code... but anyone is welcome to take a look at this right now and note any thoughts (overall code still sucks ass). You can use squirrel `get_state()` functionality now but I'm not sure if I should keep maintaining our 100 getters when we can just get the state and use an Enum/int now. Internally its `std::vector<bool>` btw instead of bitset

tobbi:
Can you rebase this against latest master, please?

PR #3118, opened 2024-11-30
Fix crash if Block::start_bounce is called with nullptr

Brockengespenst:
Brick may call try_break with nullptr as argument which causes start_bounce to be called with nullptr as argument. This leads to a crash in start_bounce when hitter is accessed.

### DIFF
--- a/src/object/block.cpp
+++ b/src/object/block.cpp
@@ -158,6 +158,8 @@ Block::start_bounce(MovingObject* hitter)
   m_bounce_dir = -BOUNCY_BRICK_SPEED;
   m_bounce_offset = 0;
 
+  if (!hitter) return;
+
   float center_of_hitter = hitter->get_bbox().get_middle().x;
   float offset = (m_col.m_bbox.get_middle().x - center_of_hitter)*2 / m_col.m_bbox.get_width();
 


Brockengespenst:
This seemed to be introcuded by https://github.com/SuperTux/supertux/commit/725a20ad95f378dee41f35063bd39a8db5ca8935#diff-e878a641a312f4d3d83db12c2951aad3ca3923cc6d9b08bb46a32f94f994d74cL162. Happens when Tux causes an iceblock to bounce horizontally between two bricks.

Vankata453:
Good catch! Maybe `if (!hitter) return;` being put before that code would be better because it removes the need to tabulate the other code.

Brockengespenst:
> Good catch! Maybe `if (!hitter) return;` being put before that code would be better because it removes the need to tabulate the other code.

Done.

tobbi:
How was such a thing able to slip through the cracks, considering the enormous amount of unit and functional tests we have available.

Vankata453:
> How was such a thing able to slip through the cracks, considering the enormous amount of unit and functional tests we have available.

Nice one 😂

PR #3119, opened 2024-11-30
Fix nullptr crash in player status hud due to wrong player item pocket access

Brockengespenst:
Fixes a crash if multiple players are in the game and the first player does not have any bonus while other players have a bonus. Due to a wrong access of m_bonus_sprites, a nullptr is returned for the sprite pointer which causes a segfault when the hud tries to draw the sprite.

### DIFF
--- a/src/supertux/player_status_hud.cpp
+++ b/src/supertux/player_status_hud.cpp
@@ -111,7 +111,7 @@ PlayerStatusHUD::draw(DrawingContext& context)
         pos += 20;
 
         Sprite* sprite = m_bonus_sprites[m_player_status.m_item_pockets[i]].get();
-        if (sprite != nullptr)
+        if (sprite)
           sprite->draw(context.color(), pos, LAYER_HUD);
       }
     }
--- a/src/supertux/player_status_hud.cpp
+++ b/src/supertux/player_status_hud.cpp
@@ -109,8 +109,10 @@ PlayerStatusHUD::draw(DrawingContext& context)
       if (m_bonus_sprites.find(m_player_status.m_item_pockets[i]) != m_bonus_sprites.end())
       {
         pos += 20;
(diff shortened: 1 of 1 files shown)


Brockengespenst:
To be honest, I have not checked the whole item pocket logic, but I encountered a crash in local multiplayer mode. First player did not have any bonus while the second player gained a bonus and that caused the game to crash: While iterating the players, always the sprite for the first player was retrieved and drawn. As there is no sprite for NO_BONUS, a nullptr was introduced to the map and the game crashed as it tried to draw a sprite from a nullptr.
I assume the check for ```if (m_bonus_sprites.find(m_player_status.m_item_pockets[i]) != m_bonus_sprites.end())``` should check for the correct sprite, so I used this as well when retrieving the actual sprite. Additionally, I added a nullptr check, just to be sure.
@MatusGuy : Please check if there is anything wrong in my understanding.

src/supertux/player_status_hud.cpp
```diff
@@ -109,8 +109,10 @@ PlayerStatusHUD::draw(DrawingContext& context)
       if (m_bonus_sprites.find(m_player_status.m_item_pockets[i]) != m_bonus_sprites.end())
       {
         pos += 20;
-        Sprite* sprite = m_bonus_sprites[m_player_status.m_item_pockets.front()].get();
-        sprite->draw(context.color(), pos, LAYER_HUD);
+
+        Sprite* sprite = m_bonus_sprites[m_player_status.m_item_pockets[i]].get();
+        if (sprite != nullptr)
```
Vankata453, line 114:
```suggestion
        if (sprite)
```

MatusGuy:
Oh my god wahahaha That slipped tight through me indeed. Nice catch!

MatusGuy (approved):

PR #3120, opened 2024-11-30
Reset tinygettext submodule to a more recent version

tobbi:
I somehow reset the tinygettext submodule a few months ago to an old version. IDK why I did that but I changed it to point to the supertux-modifications branch now.


Vankata453:
Windows can't find `iconv`, MacOS can't link.

PR #3121, opened 2024-11-30
Add new Bonus Island 1, remove old Bonus Islands

weluvgoatz:
This PR, which has been a WIP for over 2 years, includes the revamp of Bonus Island 1, with levels by myself, @Frostwithasideofsalt , @Alasdairbugs , @Rusty-Box , @Sevenberry , and @bruhmoent .  The graphics also included in this PR are by @Frostwithasideofsalt.

This new version of Bonus Island 1 aims to bring the old bonus islands to a higher, more modern standard set in place by the new story mode (which has yet to be publicly available).  Whereas the story mode will be accessible to all players, the Bonus Islands aim to appeal to experienced players.  This revamp also aims to separate the Bonus Islands into the different themes of the story mode, so Bonus Island 1 has the same theme as World 1, aka the Icy Island.  So Bonus Island 1 is snow themed.

NOTE: The levels whose names end in * are WIP! This means that Nolok's Party Pit is WIP.

Further development of Bonus Island 1 should occur on this branch.

Please playtest and let us know what you think.

PR #3124, opened 2024-12-02
Compress `TileMap` tiles in level file

Vankata453:
`TileMap` tiles are now saved with the help of negative integers, the absolute value of each one being a multiplier for the next value (ex. "0 0 0 0 0 5 5 5" -> "-5 0 -3 5").

Newline separation for tiles in file based on `TileMap` width is no longer supported to save off additional file size.

Examples:

* `0 0 0 0 0 0 45 1 1 1 1 3 3 3 3 3 0 0` -> `-6 0 45 -4 1 -5 3 -2 0`
* `0 0 0 0 0 43 12 12 12 12 12 1 1 1 5` -> `-5 0 43 -5 12 -3 1 5`

Results from re-saving some levels from "The Crystal Catacombs":

* **Worldmap**: 1.5MiB -> 74.9KiB
* "The Journey Begins Again": 409.2KiB -> 27.7KiB
* "Fjord of Fortitude": 319.1KiB -> 48.9KiB
* "A Light in the Darkness": 613.3KiB -> 183.9KiB
* "Light and Magic": 456KiB -> 163.0KiB

### DIFF
--- a/src/util/reader_mapping.cpp
+++ b/src/util/reader_mapping.cpp
@@ -201,6 +201,29 @@ ReaderMapping::get(const char* key, std::vector<unsigned int>& value,
   GET_VALUES_MACRO("unsigned int", is_integer, as_int)
 }
 
+bool
+ReaderMapping::get_merge(const char* key, std::vector<unsigned int>& value, unsigned int merge_value) const
+{
+  const auto sx = get_item(key);
+  if (!sx)
+    return false;
+
+  assert_is_array(m_doc, *sx);
+  value.clear();
+  const auto& item = sx->as_array();
+  for (size_t i = 1; i < item.size(); ++i)
+  {
+    assert_is_integer(m_doc, item[i]);
+
+    const int val = item[i].as_int();
+    if (val < 0)
+      value.insert(value.end(), -val, merge_value);
+    else
+      value.emplace_back(val);
+  }
+  return true;
+}
+
 #undef GET_VALUES_MACRO
 
 bool
--- a/src/util/reader_mapping.cpp
+++ b/src/util/reader_mapping.cpp
@@ -201,30 +201,53 @@ ReaderMapping::get(const char* key, std::vector<unsigned int>& value,
   GET_VALUES_MACRO("unsigned int", is_integer, as_int)
 }
 
+#undef GET_VALUES_MACRO
+
 bool
-ReaderMapping::get_merge(const char* key, std::vector<unsigned int>& value, unsigned int merge_value) const
+ReaderMapping::get_compressed(const char* key, std::vector<unsigned int>& value,
+                              const std::optional<std::vector<unsigned int>>& default_value) const
 {
   const auto sx = get_item(key);
   if (!sx)
+  {
+    if (default_value)
+      value = *default_value;
     return false;
+  }
 
   assert_is_array(m_doc, *sx);
   value.clear();
   const auto& item = sx->as_array();
+  int multiplier = 0;
   for (size_t i = 1; i < item.size(); ++i)
   {
     assert_is_integer(m_doc, item[i]);
 
     const int val = item[i].as_int();
-    if (val < 0)
-      value.insert(value.end(), -val, merge_value);
+    if (multiplier)
+    {
+      if (val < 0)
+      {
+        raise_exception(m_doc, item[i], "expected positive integer after multiplier");
+      }
+      value.insert(value.end(), multiplier, val);
+      multiplier = 0;
+    }
+    else if (val < 0)
+    {
+      multiplier = -val;
+    }
     else
-      value.emplace_back(val);
+    {
+      value.push_back(val);
+    }
+  }
+  if (multiplier)
+  {
+    raise_exception(m_doc, item.back(), "expected positive integer after multiplier");
   }
   return true;
 }
-
-#undef GET_VALUES_MACRO
 
 bool
 ReaderMapping::get(const char* key, std::optional<ReaderMapping>& value) const
--- a/src/util/reader_mapping.cpp
+++ b/src/util/reader_mapping.cpp
@@ -218,31 +218,31 @@ ReaderMapping::get_compressed(const char* key, std::vector<unsigned int>& value,
   assert_is_array(m_doc, *sx);
   value.clear();
   const auto& item = sx->as_array();
-  int multiplier = 0;
+  int repeater = 0;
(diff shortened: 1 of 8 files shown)


weluvgoatz (approved):
Tested the first 4 levels of #3121:
all_that_glistens.stl: 437KB -> 194KB
another_cold_day.stl: 918KB -> 167KB
area_42.stl: 838KB -> 454KB
bonus_dias.stl: 337KB -> 107KB
And they all seem to work just fine!

swagtoy:
@Vankata453 Is there a possibility you could implement a unit test for this?

swagtoy:
Can you try breaking your test now? Primarily, we already know it works, but try breaking it, Try reading nothing at the end of the tiles and seeing if it breaks. Assume the level designer would maliciously put a negative at the end to break stuff and see if an exception throws or something.

Vankata453:
I already ensured it throws exception if two negative integers are next to each other and if the array ends with a negative value.

src/util/reader_mapping.cpp
```diff
@@ -203,6 +203,52 @@ ReaderMapping::get(const char* key, std::vector<unsigned int>& value,
...
+  {
+    if (default_value)
+      value = *default_value;
+    return false;
+  }
+
+  assert_is_array(m_doc, *sx);
+  value.clear();
+  const auto& item = sx->as_array();
+  int multiplier = 0;
```
swagtoy, line 221: Call this `repeat` or `repeater`

src/util/reader_mapping.cpp
```diff
@@ -203,6 +203,52 @@ ReaderMapping::get(const char* key, std::vector<unsigned int>& value,
...
+      value.insert(value.end(), multiplier, val);
+      multiplier = 0;
+    }
+    else if (val < 0)
+    {
+      multiplier = -val;
+    }
+    else
+    {
+      value.push_back(val);
```
swagtoy, line 242: Could we possibly use `reserve` on whatever is going on here? Could minimalize some obsessive allocation calls.

src/util/writer.cpp
```diff
@@ -248,6 +248,59 @@ Writer::write(const std::string& name, const sexp::Value& value)
...
+  indent();
+  *out << '(' << name;
+  if (width > 0)
+  {
+    *out << "\n";
+    indent();
+  }
+
+  int count = 0;
+  int multiplier = 0;
```
swagtoy, line 263: repeater

src/util/writer.cpp
```diff
@@ -248,6 +248,59 @@ Writer::write(const std::string& name, const sexp::Value& value)
...
+      if (multiplier > 1)
+        *out << -multiplier << " " << multiplied_value;
+      else if (multiplier == 1)
+        *out << multiplied_value;
+
+      count = 0;
+      multiplier = 0;
+      multiplied_value = 0;
+
+      *out << "\n";
```
swagtoy, line 296: So, is there even a point in putting newlines anymore if we are now compressing? Seems pointless now.

tests/reader_test.cpp
```diff
@@ -148,6 +168,8 @@ TEST(ReaderTest, syntax_error)
     "   (myfloat 1.125 err)\n\r"
     "   (mystring \"Hello World\" err)\n"
     "   (mystringtrans (_ \"Hello World\" err))\n"
+    "   (mycompressedintarray1 -5 0 43 -5 -12 3 -1 5)\n"
```
swagtoy, line 171: Could we also test something like -5 -5 -5?

Also im curious what happens with -0 but im sure sexpr turns it to 0.

swagtoy (changes requested):

swagtoy:
Oh my apologies, i only read the first test function and missed that (i skimmed that as you "adding the test cases"). Moving on then. If the logic works then it looks good :) Sent some suggestions I think

tests/reader_test.cpp
Vankata453: "-5 -5 -5" would break on the 2nd integer.

src/util/reader_mapping.cpp
Vankata453: Why?

tests/reader_test.cpp
swagtoy: oh my goodness im friggin blind. Forgive me

src/util/reader_mapping.cpp
Vankata453: We can use `reserve` on the non-compressed functions for sure, but here, since I don't know the exact size at the end, should I just reserve the size of the uncompressed array and just deal with allocation calls whenever values need to be repeated?

src/util/reader_mapping.cpp
swagtoy: I think it is correcter

src/util/reader_mapping.cpp
swagtoy: Sure, but from what im seeing, these repetitions happen quite often.

std vector and just preallocate it to like 4 once repetition happens. Its not a big deal actually, up to you.

src/util/writer.cpp
Vankata453: I think this should be discussed, if we don't want newlines and tilemap width separation in tiles, it can be removed.

swagtoy: I mean, its just kinda unnecessary now... if we were using the spacing to actually align things.. well it doesnt matter if we are compressing it.

swagtoy:
LGTM. It look great son lets merge it

swagtoy:
Actually, wait no, I do think we could maybe throw a dart and maybe align on a single `Vector::reserve()` size, just so for example all those `32 4 5 5 5 5 5 5 5 5 5 42 5` cases where a tile repeats maybe 10 or 15 times to create a wall won't result in needless reallocations. We could even do step scaling (i.e. for a step of 5, if 6 elements are pushed, then it allocates to 10)... EFL can do this but I see nothing in C++ that can easily let us do a step. We'd have to implement that logic ourselves, though I actually question if that'd help. At this point we would be micro-optimizing to very OS dependent tasks. So at a minimum, lets just reserve and clear for now on a vector so we can very slightly speed up those few cases. Maybe a value of 24 seems generous. Up to you here; once you decide on that you can then merge.

swagtoy:
lgtm

swagtoy (approved):
heh

PR #3127, opened 2024-12-04
Try using macos-13 as Mac OS X CI

tobbi:


tobbi:
```
ld: library 'SDL2_image' not found
```

MatusGuy (commented):
I'm sorry, I can't really review this but I must say, that is a pretty odd workaround

PR #3129, opened 2024-12-07
Remove invalid off-script in leaf_wind.stl

Brockengespenst:
The off script causes the game to crash, because door1-door6 do not exist in the level.

### DIFF
--- a/data/levels/world2/leaf_wind.stl
+++ b/data/levels/world2/leaf_wind.stl
@@ -190,12 +190,6 @@
     )
     (switch
       (script "wind2.start();")
-      (off-script "door1.goto_node(0);
-door2.goto_node(0);
-door3.goto_node(0);
-door4.goto_node(0);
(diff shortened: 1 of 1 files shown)


Brockengespenst:
If last switch in like a leaf in the wind is triggered twice (which starts the off-script), the game crashes with

```
[SCRIPTING] AN ERROR HAS OCCURRED [the index 'door1' does not exist]
[SCRIPTING] CALLSTACK
[SCRIPTING] *FUNCTION [main()] switch[ 8000.000, 1376.000] line [1]
[SCRIPTING] LOCALS
[SCRIPTING] [vargv] ARRAY
[SCRIPTING] [this] TABLE
```


@Rusty-Box : I just removed the off-script from the trigger, the other switches do not have an off script either. If you have different ideas of fixing this, I can delete this PR.

Brockengespenst:
Will become obsolete with #3431

PR #3136, opened 2024-12-08
Fix MD5 digest

Semphriss:
Commit c04b46a58a mistakenly used `snprintf(str, sizeof(str), ...)` with `sizeof` instead of `strlen`.

Since using `strlen` would add a header, and since we know both the buffer length and the length of the printed string, the size can simply be hardcoded (2 characters + 1 terminating NULL character, which will be overwritten for each printf except the last).

I've also removed the now-unneeded manual terminating null byte.

### DIFF
--- a/src/addon/md5.cpp
+++ b/src/addon/md5.cpp
@@ -161,10 +161,10 @@ std::string MD5::hex_digest() {
   for (i=0; i<16; i++)
   {
     char* so = s + i * 2;
-    snprintf(so, sizeof(so), "%02x", digest[i]);
+    snprintf(so, 3 /* 2 chars + terminating NULL */, "%02x", digest[i]);
   }
 
-  s[32]='\0';
+  // Already null-terminated from the last snprintf
 
   // Create string from 's'
   std::string s_str = std::string(s);


swagtoy:
First off, this is the wrong commit, Tobbi simply fixed linting in this one

Second of all, If you're going to out of your way to refactor some weird logic in this library, you might as well just use std::stringstream for this instead of snprintf. Or hell, even just std::fmt might be the preferred solution, but alas:

```
std::stringstream res;
for ... {
    res<< std::hex << digest[i];
}
```

It is slightly slower due to no preallocation but more c++'y and less prone to any issues (i mean... we're already casting it to a string at the end anyway!!!). No need to use these C functions...

And in all honesty we..... shouldn't be relying on a hex digest to begin with :-) We should just rely on the raw md5sum data through the rest of the code. In the usage at `Addon::get_md5()`, it would be preferred if `get_md5()` returned the array of each byte, then we either used `std::equal` or just some std::array comparison to do this. We shall never need to rely directly on the stringified checksum data. It is designed for reading and simpler string comparisons, not for actual code comparisons.

tulpenkiste:
> Second of all, If you're going to out of your way to refactor some weird logic in this library, you might as well just use std::stringstream for this instead of snprintf. Or hell, even just std::fmt might be the preferred solution, but alas

As far as I'm aware, std::fmt is not a viable solution for SuperTux due to needing to support older Ubuntu versions

swagtoy:
We have our own format library in the game. Not a c++20 thing

tobbi:
I still don't understand if I fucked up or not.

PR #3137, opened 2024-12-08
Fix relative paths with the Particle Editor

Semphriss:
Relative paths make the game panic about insecure filenames when they contain '..'; when a custom particle file located outside the particles/ root folder was selected in the editor, and the object was used to 'Open in Particle Editor', the game would crash.

The best fix would've been to dodge the need for relative paths entirely, but that would prevent the settings from opening by default in the particles/ folder. This is, I believe, the second best fix.

### DIFF
--- a/src/editor/editor.cpp
+++ b/src/editor/editor.cpp
@@ -265,7 +265,12 @@ Editor::update(float dt_sec, const Controller& controller)
     m_particle_editor_request = false;
     std::unique_ptr<Screen> screen(new ParticleEditor());
     if (m_particle_editor_filename)
-      static_cast<ParticleEditor*>(screen.get())->open("particles/" + *m_particle_editor_filename);
+    {
+      // realpath() is necessary for particle files outside the particles/ folder
+      std::string path = physfsutil::realpath("particles/" + *m_particle_editor_filename);
+
+      static_cast<ParticleEditor*>(screen.get())->open(path);
+    }
     ScreenManager::current()->push_screen(std::move(screen));
     return;
   }
--- a/src/supertux/menu/particle_editor_open.cpp
+++ b/src/supertux/menu/particle_editor_open.cpp
@@ -20,6 +20,7 @@
 #include "gui/dialog.hpp"
 #include "gui/menu_item.hpp"
 #include "gui/menu_manager.hpp"
+#include "physfs/util.hpp"
 #include "supertux/level.hpp"
 #include "supertux/gameconfig.hpp"
 #include "supertux/menu/menu_storage.hpp"
@@ -55,11 +56,17 @@ ParticleEditorOpen::~ParticleEditorOpen()
 void
 ParticleEditorOpen::menu_action(MenuItem& item)
 {
+  std::string path;
+
   switch (item.get_id())
   {
     case MNID_OPEN:
       std::replace(m_filename.begin(), m_filename.end(), '\\', '/');
-      ParticleEditor::current()->open("/particles/" + m_filename);
+
+      // realpath() is necessary for particle files outside the particles/ folder
+      path = physfsutil::realpath("/particles/" + m_filename);
+
+      ParticleEditor::current()->open(path);
       MenuManager::instance().clear_menu_stack();
       break;
 
--- a/src/object/custom_particle_system_file.cpp
+++ b/src/object/custom_particle_system_file.cpp
@@ -20,6 +20,7 @@
 
 #include "editor/editor.hpp"
 #include "gui/menu_manager.hpp"
+#include "physfs/util.hpp"
 #include "util/reader.hpp"
 #include "util/reader_document.hpp"
 #include "util/reader_mapping.hpp"
@@ -66,7 +67,10 @@ CustomParticleSystemFile::update_data()
 {
   try
   {
-    auto doc = ReaderDocument::from_file("particles/" + ((m_filename == "") ? "default.stcp" : m_filename));
+    // realpath() is necessary for particle files outside the particles/ folder
+    std::string path = physfsutil::realpath("particles/" + ((m_filename == "") ? "default.stcp" : m_filename));
+
+    auto doc = ReaderDocument::from_file(path);
     auto root = doc.get_root();
     auto mapping = root.get_mapping();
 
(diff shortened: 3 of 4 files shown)


tobbi:
Does this supersede #3102  ?

Semphriss:
Ah, I hadn't noticed that PR, my apologies.

I've fixed a few more places that had the same bug though, so I suppose it does supersede the other PR.

tobbi:
My changes apparently didn't work, so this is probably better.

src/editor/editor.cpp
```diff
@@ -265,7 +265,12 @@ Editor::update(float dt_sec, const Controller& controller)
     m_particle_editor_request = false;
     std::unique_ptr<Screen> screen(new ParticleEditor());
     if (m_particle_editor_filename)
-      static_cast<ParticleEditor*>(screen.get())->open("particles/" + *m_particle_editor_filename);
+    {
+      // realpath() is necessary for particle files outside the particles/ folder
+      std::string path = physfsutil::realpath("particles/" + *m_particle_editor_filename);
+
+      static_cast<ParticleEditor*>(screen.get())->open(path);
```
MatusGuy, line 272: You really need to static cast here? You can store the new ParticleEditor in a variable and use that. In my opinion, that would be more readable. And then you can std::make_unique in the push_screen function

src/editor/editor.cpp
```diff
@@ -265,7 +265,12 @@ Editor::update(float dt_sec, const Controller& controller)
     m_particle_editor_request = false;
     std::unique_ptr<Screen> screen(new ParticleEditor());
     if (m_particle_editor_filename)
-      static_cast<ParticleEditor*>(screen.get())->open("particles/" + *m_particle_editor_filename);
+    {
+      // realpath() is necessary for particle files outside the particles/ folder
+      std::string path = physfsutil::realpath("particles/" + *m_particle_editor_filename);
```
MatusGuy, line 270: Hey, whatever happened to `FileSystem::join`? Or `std::filesystem::path::operator/`? Not the only place where this happens so if you're gonna fix this then do it there aswell

src/supertux/menu/particle_editor_open.cpp
```diff
@@ -55,11 +56,17 @@ ParticleEditorOpen::~ParticleEditorOpen()
 void
 ParticleEditorOpen::menu_action(MenuItem& item)
 {
+  std::string path;
```
MatusGuy, line 59: What? Can't you just make the variable inside the switch case?

src/editor/editor.cpp
Semphriss: The static cast was already there. I've copied the code and changed as little as possible so as not to transform the code too much, since I'm not very familiar with the current coding standards and people's preferences. A static cast will be needed either way if `screen`'s type is a pointer to Scene; I can move the static cast on its own line but I can't really avoid it without changing more code. I can't make_unique the screen directly in push_screen because I need to call the `open` function on the scene before pushing it.

src/editor/editor.cpp
Semphriss: Similarly to above, I have opted not to change the code more than necessary to keep the changes minimal and avoid going against eventual preferences. The join function also does not normalize the path, so it would be an extra call rather than a replacement.

src/supertux/menu/particle_editor_open.cpp
Semphriss: https://stackoverflow.com/questions/92396/why-cant-variables-be-declared-in-a-switch-statement

The [coding style documentation](https://github.com/SuperTux/supertux/blob/master/CODINGSTYLE.md) does not specify whether to declare variables outside the switch statement or to create a new scope within the case statement. Since I'm used to the former with SDL and I couldn't find any source as to how to format a scope within a case, I put the variable outside the block. I can move it inside and create a scope, if you could provide me with a preferred formatting template.

Alasdairbugs:
Particles dont seem to load images correctly atm

image
image

Alasdairbugs:
when saving a new particle configuration it says "failed to save particle configuration, retrying" and then is successful

Alasdairbugs:
image
image

these settings, with a tinted texture, alpha being irrelevant, cause the sprite to flicker at its original brightness then return to the tint its at, this weird flickering seems to always happen?

Alasdairbugs:
bi2 level spoiler //

heres kind of another bug(?)

https://youtu.be/Zmwubb0mKe4

basically im trying to spawn in particles using a particle zone set to spawn, and one set to kill the particles.

the particles seem to be spawning anywhere they like, primarily in the spawn region, but still spawning in above it. not too sure whats going on

also as a side note, im not sure how this system interacts with camera.ease_scale, i think zooming out 'dilutes' the particles, because if the amount is set at max 500, then you zoom out, there is more space to be filled by those 500 particles. therefore being diluted. this might be noticeable in some scenarios. in this case, a dynamic max. cap might be required?

Alasdairbugs:
ok so:

- you were correct, i've unticked spawn anywhere.
- now, without normal camera scaling, the particles spawn and behave pretty much as intended
- if i scale the camera out, two things happem:
1. in the particle zone region, particles will spawn in and out, flickering
2. if i peek down to 'load in'(?) the particle zone, then particles will both spawn normally as required, but step 1 will also continue simultaneously.

also, it seems that if im loading in like a smidge of the particle zone, like a tile then only a fraction of particles will spawn.
if i peek down then an array of particles will spawn.

so this behaviour can be 'patchworked' as in, at least in my own unique case, i can place the particle zone under some terrain (spawn) and the flickering is not seen by the player. but still i think that particle zones should perhaps load off-screen? because the alternative rn is to have the particles set for the entire sector, where its always loaded in, but that has flickering issues and i cant discriminate where i want spawning not to occur spatially.

heres the particle file:
[think_fast_or_die_particles.txt](https://github.com/user-attachments/files/18154615/think_fast_or_die_particles.txt)

ive converted it to .txt pls convert to .stcp.

Semphriss:
I've reproduced the flickering bug, I'm working on a fix now.

@tobbi Would you prefer that I put all the fixes in one PR, or that I keep them separate per PR?

tobbi:
> I've reproduced the flickering bug, I'm working on a fix now.
> 
> @tobbi Would you prefer that I put all the fixes in one PR, or that I keep them separate per PR?

Sorry, I forgot about this. I think separate PRs for better readability is better.

PR #3139, opened 2024-12-11
[NO SQUASH] Enforce batching all render requests in SDL2_Renderer

swagtoy:
For all SDL2 Render requests, we (presumably) may not be batching it. This is noticed in the SDL Painter for functions like `SDLPainter::draw_texture` as well as other classes. This means that for our `RenderCopyEx` calls (technically, `SDL_RenderCopyExF`) as well as any shape functions, SDL was immediately passing each RenderCopy call to the GPU. Note: I believe this could be the case, though the documentation is somewhat unreliable in SDL's defense on it. It calls one case "efficient" and the other "compatible"...

With this PR, by settings a single SDL2 Renderer hint, this should result in a possible major performance boost!

They also mention on the docs that it is disabled by default due to some potential issues with usage of rendering, but after some pretty general testing, I did not spot any in particular. I would appreciate some testers that want to boot a build up and see if it at least draws stuff.

SDL3 mentioned in a PR that this is default now, but since we use SDL2 still, let's just pass this.

### DIFF
--- a/src/video/sdl/sdl_video_system.cpp
+++ b/src/video/sdl/sdl_video_system.cpp
@@ -67,6 +67,7 @@ SDLVideoSystem::create_window()
   log_info << "Creating SDLVideoSystem" << std::endl;
 
   SDL_SetHint(SDL_HINT_RENDER_SCALE_QUALITY, "2");
+  SDL_SetHint(SDL_HINT_RENDER_BATCHING, "1");
 
   create_sdl_window(0);
 


tobbi:
You missed out on a huge opportunity: You could've called the branch: `batch-please`

swagtoy:
@tobbi Ha, my new favorite running joke is from our player class, `m_wants_buttjump`, so the branch name is oriented off of that :)

tobbi:
Honestly, I couldn't find any differences between this branch and latest master using SDL.

MatusGuy (commented):
I think an actual practical review would be more helpful so I'm not gonna approve this

swagtoy:
A quick glimpse at the SDL2 branch shows that a `batching` flag is set if the hint is enabled. This calls the inline function `FlushRenderCommandsIfNotBatching`. This flag should be okay to enable. The SDL2 Renderer internally uses `renderer->render_commands_tail` which is okay. I feel confident we should ship this commit now.

SDL3 indeed removes this flag entirely, but we are on SDL2 so until we decide to migrate, enabling this flag cannot hurt for the SDL2 side. It may not be a major speedup actually, but it should be a helpful optimization for the SDL2 Renderer. The android side on SDL2 shows that this doesn't happen however, just letting you know.

PR #3140, opened 2024-12-13
Slope collision algorithm rewrite

MatusGuy:
Ok so I'm not the most math guy in here but I've come up with a simpler and better documented slope collision algorithm. Here, I plan to fix inconsistencies and weird edge cases that would otherwise be difficult to pinpoint with the old system.

Here I also removed all instances of glm::vec2 and replaced it with our own Vector class, which means that glm is ready to get removed.

Testers: Once I fully open this PR, please test all interactions with different types of slopes. And on a seemingly unrelated note, also test big snowball.


swagtoy:
Was GLM not being used in our OpenGL/video code at all?? :O

PR #3143, opened 2024-12-16
Remove `set_gamma`, `exit_screen` and `set_next_worldmap` scripting functions

Frostwithasideofsalt:
**PR description by @Vankata453:**

`set_gamma` changes the brightness of the whole screen, which could be used inaccordingly.

`exit_screen` pops the current screen, meaning it can likely be used to make levels unplayable or the game un-openeable, if used in a title screen level.

`set_next_worldmap` (added in 65da51280c38c867708b0f32eb59748c48a27181) requires the `exit_screen` function to properly work, meaning it was also removed, thus no longer allowing switching from one world to another (when was this needed anyway?)

### DIFF
--- a/src/squirrel/supertux_api.cpp
+++ b/src/squirrel/supertux_api.cpp
@@ -296,18 +296,6 @@ static void load_worldmap(const std::string& filename, const std::string& sector
 }
 /**
  * @scripting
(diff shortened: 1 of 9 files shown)


MatusGuy (approved):

Vankata453 (approved):
Additionally removed the `set_next_worldmap` function, as well as both remnants of it and `set_gamma`.

PR #3145, opened 2024-12-16
[NO SQUASH] Update Yeti sprites to fit new style + new animations

Rusty-Box:
**NOTE: For the best testing experience for this boss behavior change, make sure to download the `yeti_boss.stl` from #3216 - a level/arena designed specifically with this new behavior in mind!**

- [x] Update Yeti sprite set
- [x] Update Yeti health icon
- [x] Update Yeti Stalactite sprites for Level Editor
- [x] Code adjustments (help needed)

Code changes required:
- [x] Use of "walking" action must be renamed to "walk" in the code.
- [x] "dizzy" and "busted" action was added for when the yeti is defeated.
  - upon defeat, the yeti will stumble backwards for a bit before falling to the ground, remaining there instead of falling off-screen.
- [x] "fall" action is added for when the Yeti falls or his jump loses altitude.
- [x] Big Snowball projectile spawn position must be adjusted to fit animation.
  - Same may apply to the Bouncy Snowball spawn position
- [x] Upon entering pinch mode, Yeti shall play his "rage" action, beating his chest, to signify to the player that his pinch mode has started.

Minor behavior adjustments (recommandations):
- [x] Big Snowball the Yeti throws should bounce (feels more dynamic that way).
- [x] Reduce amount of Big Snowballs thrown from 3 to 2 on last health (it's odd how he goes from 1 to 3 between second to last and last health).
~~- In Pinch Mode, instead of walking from one end to the other, Yeti will jump towards the center of the arena, stand still (while the stomp animation plays), allowing for an attack and soon after jump towards the other end of the arena.~~ (postponed)
- [x] Slightly reduce delay between transitions
      - pinch mode rage -> snowball throw
      - landing on platform -> snowball throw
      - delay between individual snowball throws

FIXME:
- [x] Yeti does not play "gna" sound when leaping in to thei air, right before the stomp
- [x] Big Snowballs should hurt Tux and break if "crushed" but not kill Tux


tobbi:
```
/Users/runner/work/supertux/supertux/src/badguy/yeti.cpp:576:13: error: 4 enumeration values not handled in switch: 'ANNOUNCE', 'FALL', 'IDLE'... [-Werror,-Wswitch]
    switch (m_state) {
            ^~~~~~~
/Users/runner/work/supertux/supertux/src/badguy/yeti.cpp:576:13: note: add missing switch cases
    switch (m_state) {
```

Rusty-Box:
> In Pinch Mode, instead of walking from one end to the other, Yeti will jump towards the center of the arena, stand still (while the stomp animation plays), allowing for an attack and soon after jump towards the other end of the arena.

@MatusGuy you still wanna tinker with that or do we skip that for now?

MatusGuy:
Gonna have to skip that, sorry

Frostwithasideofsalt (approved):
it all works, doesn't twerks, but the big snowball sprite is still outdated (awkward!)

weluvgoatz (changes requested):
Okay, to be honest, this leaves a little bit to be desired. For one, I think the fact that the big rolling snowball bounces makes the fight too chaotic and replicates almost the exact same behavior as the bouncing snowball enemies so why even have it in the first place.  Second, I noticed that, at least on my resolution, the yeti lives HUD icon covers some of the stalactites in the yeti boss level so they are not visible - I suggest maybe making the HUD smaller maybe? Third (and this is just a personal wish so I don't think this bit is necessary) but I did like before when you had multiple big snowballs rolling at you near the end, it made them feel a bit more utilized because otherwise they are only used twice in the whole level, which, again, raises the question of why even have it in the first place.  Other than those 3 things, I think this PR is pretty good.

Rusty-Box:
> I think the fact that the big rolling snowball bounces makes the fight too chaotic and replicates almost the exact same behavior as the bouncing snowball enemies so why even have it in the first place.

Well, it is the same challenge as the Bouncing Snowball either way, if you ask me. In both cases the way to dodge the attack is to jump over the Big Snowball. The extra bounce makes the attack look and feel more dynamic, in my opinion compared to the more static flat motion. Also, since it is just one big snowball after a short pause so I don't see how it is chaotic? Having several Bouncing Snowballs and/or Stalactites during the Pinch Phase are even more chaotic since there is much more stuff happening at once.

> Second, I noticed that, at least on my resolution, the yeti lives HUD icon covers some of the stalactites in the yeti boss level so they are not visible - I suggest maybe making the HUD smaller maybe?

What is your resolution? I usually play at Magnification set to 150% and my resolution being 1920x1080 and I see them perfectly fine, despite playing slightly zoomed in myself.

> Third (and this is just a personal wish so I don't think this bit is necessary) but I did like before when you had multiple big snowballs rolling at you near the end, it made them feel a bit more utilized because otherwise they are only used twice in the whole level, which, again, raises the question of why even have it in the first place.

The initial idea was to have him throw 2 on his last HP. Originally, he did throw 3 which was rather odd considering the pattern of adding one extra throw (for Bouncing Snowball) per hit, whereas the Big Snowball goes from 1 to 3. We could do that.

weluvgoatz:
1) I think the chaotic feeling arises from the fact that the bounce makes it hard to predict when you can or can't jump over it.  It feels like I'm fighting a wave in the ocean when I play that part.
2) Okay, I suppose "cover" isn't the right word, but them appearing over the stalactites still looks a tad odd to me.  However it isn't really that big of a deal.
image
3) I think throwing 2 on the last HP would be nice.

Chiquinho7788:
I found an error when the yeti throws the giant snowball:

Captura de ecrã 2025-02-26 114137


When the yeti is in this position in editor mode, and when it's in launch mode and reaches the right-hand side, it can't throw the giant snowball because there's no room for the snowball to fall, so it hits the wall, my suggestion would be to make the level a little wider on the right-hand side.

Captura de ecrã 2025-02-26 221452

Hypernoot:
I tested that and it doesn't work with the old Yeti arena. Testing is possible only with a new arena, which wasn't provided yet.

Rusty-Box:
> I tested that and it doesn't work with the old Yeti arena. Testing is possible only with a new arena, which wasn't provided yet.

You can find the new arena here: #3216

Hypernoot:
screenshot000308

Rusty-Box:
Lol what? Lemme check the level file. Might be an issue on its end perhaps.

Hypernoot:
Actually, my bad. I tested it wrongly. Now it works.

Rusty-Box:
Strange, the sprite path looks fine. Did you play the level with the build from this PR?

Rusty-Box:
> Actually, my bad. I tested it wrongly. Now it works.

Ah okay 👍

Hypernoot:
I found a bug. When I manage to jump up to Yeti's position, he doesn't turn around and then he says goodbye.
[yeti-2025-04-05_13.22.51.zip](https://github.com/user-attachments/files/19615080/yeti-2025-04-05_13.22.51.zip)

PR #3157, opened 2024-12-25
Avoid grabbing an already grabbed portable

Brockengespenst:
Avoid that a player can "grab" an already grabbed portable from another player.


Brockengespenst:
Happens only in multiplayer when another player presses the action button next to a player that has grabbed a portable object. Actually the portable remains visible at the position of the first player that grabbed the protable, but when the other player ungrabs the object, the portable "flies" back to the first player. 
A simple check if the portable is already grabbed solves this problem.

tobbi (approved):

PR #3158, opened 2024-12-27
Support X/Y position periods for autotile "alt-id"s

Vankata453:
Autotile "alt-id"s now support the "period-x" and "period-y" properties, which contain 2 unsigned integer values: first is the divider and second is the desired remainder after dividing the X/Y position of the tile, relative to its tilemap. If the desired remainder is matched, the condition is passed.

NOTE: For an "alt-id" to be chosen, all of its conditions ("period", random "weight") must be met.

This PR takes advantage of these new position periods to achieve proper castle autotiling with line tiles.

### DIFF
--- a/src/supertux/autotile.cpp
+++ b/src/supertux/autotile.cpp
@@ -36,7 +36,7 @@ AutotileMask::matches(uint8_t mask, bool center) const
 
 // Autotile.
 
-Autotile::Autotile(uint32_t tile_id, const std::vector<std::pair<uint32_t, float>>& alt_tiles, const std::vector<AutotileMask>& masks, bool solid) :
+Autotile::Autotile(uint32_t tile_id, const std::vector<std::pair<uint32_t, AltCondition>>& alt_tiles, const std::vector<AutotileMask>& masks, bool solid) :
   m_tile_id(tile_id),
   m_alt_tiles(alt_tiles),
   m_masks(std::move(masks)),
@@ -62,7 +62,7 @@ Autotile::pick_tile(int x, int y) const
 {
   // Needed? Not needed?
   // Could avoid pointless computation.
-  if (m_alt_tiles.size() == 0)
+  if (m_alt_tiles.empty())
     return m_tile_id;
 
   // srand() and rand() are inconsistent across platforms (Windows)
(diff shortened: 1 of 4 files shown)


weluvgoatz (approved):

src/supertux/autotile.cpp
```diff
@@ -77,11 +77,22 @@ Autotile::pick_tile(int x, int y) const
 
   for (const auto& pair : m_alt_tiles)
   {
-    rnd_val -= pair.second;
-    if (rnd_val <= 0)
+    const AltConditions& cond = pair.second;
+    if (cond.weight <= 0.f && !cond.period_x.first && !cond.period_y.first)
+      continue;
+
+    if (cond.period_x.first && x % cond.period_x.first != cond.period_x.second)
```
MatusGuy, line 84: What do you want by checking for only `cond.period_x.first`? Checking if it exists? If it isn't zero?

Vankata453:
Yes, if it isn't 0.

src/supertux/autotile.cpp
Vankata453: Yes, if it isn't 0.

MatusGuy: Can you write that explicitly?
```suggestion
    if (cond.period_x.first != 0 && x % cond.period_x.first != cond.period_x.second)
```

(in the other conditions too)

Vankata453: Why? It cannot be mistaken for anything else, can it?

MatusGuy: it can, and I did, because i'm used to seeing that only in booleans and pointers. i just suggested that to save some time reading that. not a big deal

src/supertux/autotile_parser.cpp
```diff
@@ -196,17 +196,27 @@ AutotileParser::parse_autotile(const ReaderMapping& reader, bool corner)
...
-        alt_ids.push_back(std::pair<uint32_t, float>(alt_id, weight));
+        log_warning << "X/Y period for alt tile must contain exactly 2 values" << std::endl;
+        continue;
       }
+
+      alt_ids.push_back(std::pair<uint32_t, Autotile::AltConditions>(alt_id, {
+          { period_x[0], period_x[1] },
+          { period_y[0], period_y[1] },
+          weight
+        }));
```
MatusGuy, line 219: Does this mean you can put alt-ids with no conditions now?

Vankata453: Yes. If you do, they just won't do anything.

PR #3160, opened 2024-12-29
Add Some Position-Based Autotilesets

Eauix:
As alt-IDs can now be defined by the tile positions in a repeating pattern, this PR adds some extra autotilesets which benefit from this feature.

New Autotilesets:
- Ruin Bricks with vertical, mossy and corrupted variants
- Water and Lava
- Snow and Forest Cave Walls

Tweaks:
- The Forest Castle autotileset has random cracks now
- Fixed the Underground Forest Tileset's 1x2 vines being placed when they shouldn't
- Reduced the amount of roots in the Underground Corrupted Forest Tileset
- Repositioned the 1x2 vines in the tilegroup to make it more evident


Frostwithasideofsalt (approved):
works

Alasdairbugs (changes requested):
Works quite well, but i wanna suggest just one change, and ask one question:

1. Generally, when using these moss-covered castle brick tiles, personally i would expect to see some normal brick still mixed within, because otherwise it looks way too heavy on the green, and quite bad, imo:

image

2. [**mostly speculation, ignore if not possible**] is there infrastructure to potentially support crystal hexagon autotiling given these would probably use alt IDs but not every other tile, just every other 2nd tile, it seems (with the darker hexagoens included as spontaneous variations). just wondering. It might be possible i think with the current infrastructure, so long as people are ok with seeing tile errors briefly as they build their crystal structures(?). If this is possible I think maybe this should be done too, but feel free to chose if this would be best in an additional PR later.

image

Other than that it looks all good, all the variation tiles are in the right places

Eauix:
> 1. Generally, when using these moss-covered castle brick tiles, personally i would expect to see some normal brick still mixed within, because otherwise it looks way too heavy on the green, and quite bad, imo:

I thought about this, however, adding random variants to pattern with tiles that need to another specific tile from this variant adjacent to it would be really complicated with the current system. Currently, there is not a way of making it so an alt-ID appears if there's a specific tile ID near it, so, if we try to add a random moss pattern to this tileset, some of the bricks will get cut in half and the other half will be the mossy variant

image

The only solution I can think about is creating a huge pattern with pseudo-random tiles defined within it, but this would require hundreds of alt-IDs, but not sure if it's worth it. The current pattern is a 4x2 matrix of predefined alt-IDs, to make it feel like a random pattern, probably a 12x12 or larger pattern would be necessary

Eauix:
> 2\. If this is possible I think maybe this should be done too, but feel free to chose if this would be best in an additional PR later.

The same problem happens with the darker crystal variants, for it to be close to random we'd need a huge pattern

And making the crystal background autotile would proably need copies of the same tile IDs for the top and bottom halves, one to be considered solid and one to be considered decoration for the autotiling system. Or it would need a tile ID to have both a solid and a non-solid mask, and I imagine this would break everything

Alasdairbugs (approved):

Alasdairbugs (approved):

PR #3164, opened 2025-01-01
Allow platforms to start moving backwards

tobbi:
Another pull request that will eventually end up abandoned because I have bright ideas but no drive to actually follow through.

### DIFF
--- a/src/object/path_object.cpp
+++ b/src/object/path_object.cpp
@@ -96,10 +96,10 @@ PathObject::set_node(int node_idx)
 }
 
 void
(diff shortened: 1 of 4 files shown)


Alasdairbugs:
unfortunately does not work. start_moving(false); just makes them move i think

Alasdairbugs (changes requested):
unfortunately does not work. start_moving(false); just makes them move i think

tobbi:
I'm gonna close this. I have too many open PRs and this was just a test to see if it could be this easy.

PR #3169, opened 2025-01-08
Use unordered_map instead of ordered_map

tobbi:
unordered_map is preferred over map for accessing particular elements such as we do in those instances. See https://en.cppreference.com/w/cpp/container/unordered_map

### DIFF
--- a/src/video/texture_manager.hpp
+++ b/src/video/texture_manager.hpp
@@ -18,6 +18,7 @@
 #define HEADER_SUPERTUX_VIDEO_TEXTURE_MANAGER_HPP
 
 #include <config.h>
+#include <unordered_map>
 #include <map>
 #include <memory>
 #include <ostream>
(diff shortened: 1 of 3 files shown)


src/video/texture_manager.hpp
```diff
@@ -77,7 +78,7 @@ class TextureManager final : public Currenton<TextureManager>
 
 private:
   std::map<Texture::Key, std::weak_ptr<Texture>> m_image_textures;
```
swagtoy, line 80: What is this btw? Can we hash it for unordered_map as well?

swagtoy (approved):
Lgtm overall

src/video/texture_manager.hpp
tobbi: I tried and it gave me compiler errors because of the std::weak_ptr

swagtoy: Well I just read `texture_manager.cpp` and whoever decided texture locking with a weak pointer like this for a game was necessary has/had serious thinking to do in their life.

We can ignore this for now, but something about this code doesn't smell right to begin with. Im skeptical about this texture code.

MatusGuy (approved):
Code looks good

PR #3170, opened 2025-01-10
Shatter shards on impact

### LINKED ISSUE #3134: Shards should shatter upon impact
The shards dropped by Corrupted Igel, Crystallo and Big Granito tend to be rather annoying and hard to see once on the ground. They are already challenging enough during the initial burst.

So I recommend having those shards be destroyed upon impact with either the player, an object or solid ground to make them a bit less annoying.

Brockengespenst:
Makes shards less annoying. With this change, there is no need for them to be a sticky object, a moving sprite is just enough and makes them less complex.

Fixes #3134


Brockengespenst:
@Rusty-Box : Please give it a try and check if this works better.

weluvgoatz:
Haven't tested yet but in my opinion the crystallo shards should retain their current behavior but every single other usages of darts can have this shattering-upon-impact.

MatusGuy:
Nice. Make sure this works with the corrupted big granito, roof crystallo, and corrupted igel.

MatusGuy:
In my opinion the cooldown should stay but instead be significantly lowered to something like less than a second. This PR might need discussion

Brockengespenst:
> Nice. Make sure this works with the corrupted big granito, roof crystallo, and corrupted igel.

They all use the same class, so it works for all of them.

> In my opinion the cooldown should stay but instead be significantly lowered to something like less than a second. This PR might need discussion

Sure, I absolutely agree about the discusison.
If keeping the old behaviour, I think decreasing the time the shards are still active should definitely decreased. And maybe add some decent fade out effect instead of disappearing instantly.

Brockengespenst:
No feedback yet, so I give you the opportunity to test it out. Shards remain for 0.7 seconds active after impact and fadeout in another 0.7 seconds (not hurting anymore). This feels quite fair for me and looks nicer.
Further idea: Fadout may already start on first impact with Tux or a badguy - kind of "taking the shard out of the game" after first hit of one of them.

MatusGuy (approved):
Code looks good

Alasdairbugs (approved):

bruhmoent (approved):

PR #3172, opened 2025-01-11
Add default arguments to add_object function call

tobbi:
This adds a few default arguments to the add_object function call. Direction will default to `auto` and additional data will default to empty string.

Unfortunately, the name parameter can't be omitted because it's before a non-optional parameter (location).


MatusGuy (commented):
I can't review this because I didn't implement squirrel default arguments.

PR #3184, opened 2025-01-18
WIP: Determine object type from sprite filename if sprite not found

tobbi:
This is another backwards compatibility check: Some of the sprite paths changed in between version 0.6.3 and latest nightly. I've implemented something that checks the old sprite paths against a list of known earlier sprite paths and returns the proper object type in that case.

This only happens if the sprite path in question couldn't be found. To be honest, setting the object type for sprite paths that remained the same in between versions would also make sense, but the code is too flaky for that currently (e.g. we don't check the full path).

Tell me what you think.

### DIFF
--- a/src/badguy/bouncing_snowball.cpp
+++ b/src/badguy/bouncing_snowball.cpp
@@ -99,6 +99,18 @@ BouncingSnowball::get_types() const
   };
 }
 
+bool 
+BouncingSnowball::get_type_from_sprite(const std::string& sprite_name, std::string& type) const
+{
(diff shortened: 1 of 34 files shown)


tobbi:
This is still missing the existing sprite paths to be honest.

tobbi:
> Uh, why a new system when the object versioning system by Vankata already exists? Refer to PowerUp::initialize

Does that work with old levels from 0.6.3 or something? Assuming the sprite was set to "rock_crusher.sprite", will that properly load the rock crusher object type?

PR #3185, opened 2025-01-20
New industrial tiles, New Tux statue, other misc tile changes 

Frostwithasideofsalt:
-New industrial tiles, with more slope variants, a 3x3 plate, and unisolids.
-Add new tuxdoll statue, and reduce color bleeding on nolok statues.
-deprecate "plant1" from doodads tiles folder
-Darken pillars to make them more easily distinguishable as non-solid
-Darken "pedestal", and make it non-solid 
-Update chocolate blocks so that their design reflects the bridge block designs 
-bridge block designs have a 1x1 variant


Alasdairbugs:
Personally i'd leave the area 51 retiling to daniel, I think the level needs just general touch-ups anyway and requires a PR regarding rock squishing.

PR #3187, opened 2025-01-22
Fix Tux being permanently safe after mix of using door and scripting Tux safe

### LINKED ISSUE #3186: [Bug]: Tux immunity after door
### SuperTux Version

v0.6.3-1566-ge43838b1d

### System Information

win 11 64 bit

### Expected Behavior

Tux is not immune to every damage type

### Actual Behavior

After entering a door, possibly in addition to having Tux be set invincible and then not invincible (#3051 related(?)), tux is invincible to all damage types.

go through a door, if thats not enough, possible try make tux invincible and then not invincible, then go through the door.

if that doesnt work, then start by making tux not invincible (he shouldn't be anyway, by default), invincible, then not invincible, then go through a door.

### Additional Information

_No response_

Brockengespenst:
* Use a separate timer for temporary safety instead of reusing m_is_intentionally_safe
* Split timers for post damage safety and temporary safety

Fixes #3186

### DIFF
--- a/src/object/player.cpp
+++ b/src/object/player.cpp
@@ -209,7 +209,8 @@ Player::Player(PlayerStatus& player_status, const std::string& name_, int player
   m_does_buttjump(false),
   m_invincible_timer(),
   m_skidding_timer(),
-  m_safe_timer(),
+  m_post_damage_safety_timer(),
+  m_temp_safety_timer(),
   m_is_intentionally_safe(false),
   m_kick_timer(),
   m_buttjump_timer(),
@@ -2256,7 +2257,7 @@ Player::draw(DrawingContext& context)
   Vector draw_pos = get_pos() + context.get_time_offset() * m_physic.get_velocity();
 
   /* Draw Tux */
-  if (!m_visible || (m_safe_timer.started() && !m_is_intentionally_safe && size_t(g_game_time * 40) % 2))
+  if (!m_visible || (m_post_damage_safety_timer.started() && size_t(g_game_time * 40) % 2))
   {
   }  // don't draw Tux
 
@@ -2387,7 +2388,7 @@ Player::collision(MovingObject& other, const CollisionHit& hit)
 
   auto badguy = dynamic_cast<BadGuy*> (&other);
   if (badguy != nullptr) {
-    if (m_is_intentionally_safe || m_safe_timer.started() || m_invincible_timer.started())
+    if (m_is_intentionally_safe || m_post_damage_safety_timer.started() || m_temp_safety_timer.started() || m_invincible_timer.started())
       return FORCE_MOVE;
     if (m_stone)
       return ABORT_MOVE;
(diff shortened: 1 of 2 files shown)


Frostwithasideofsalt (approved):

src/object/player.hpp
```diff
@@ -575,6 +575,11 @@ class Player final : public MovingObject
    */
   bool m_is_intentionally_safe;
 
+  /**
+   * @description Flag indicating that Tux shall not blink while being safe
+   */
+  bool m_avoid_blinking_while_safe;
+
```
MatusGuy, line 582: *Avoid* blinking while safe?? Man... This file needs a rewrite so bad...

MatusGuy (approved):
Code looks..... ok.... ugh. Don't worry, it's not your fault.

src/object/player.hpp
Brockengespenst: Giving variables a good name can be very challenging. ;-) I‘d be grateful for any suggestion.

src/object/player.cpp
```diff
@@ -2571,7 +2572,7 @@ Player::check_bounds()
   /* fallen out of the level? */
   if ((get_pos().y > Sector::get().get_height())
       && !m_ghost_mode
-      && !(m_is_intentionally_safe && m_safe_timer.started())) {
+      && !(m_avoid_blinking_while_safe && m_safe_timer.started())) {
```
mstoeckl, line 2575: This use of `m_avoid_blinking_while_safe` does not match the variable description.

One strategy to avoid such name/purpose problems is to avoid "derived" names, properties and state as much as possible.  For example, instead of having a `m_avoid_blinking_while_safe` variable, store a variable expressing the _cause_, from which one determine whether the player should blink, or whether it should be allowed to fall out of bounds.
```
enum TemporarySafetyCause {
        Damage, // has blink effect
        Transition, // stay alive even if out of bounds        
};
TemporarySafetyCause m_safe_timer_reason;
```

Another approach is to replace `m_safe_timer` with two distinct timers for the different causes; e.g. use `m_post_damage_timer`, started only in Player::kill, and `m_temp_safety_timer`, started only in `make_temporarily_safe` . Doing this would make it possible to have the player blink for the specified time after being damaged, even when going through a door halfway through.

Brockengespenst: > Another approach is to replace `m_safe_timer` with two distinct timers for the different causes; e.g. use `m_post_damage_timer`, started only in Player::kill, and `m_temp_safety_timer`, started only in `make_temporarily_safe` . Doing this would make it possible to have the player blink for the specified time after being damaged, even when going through a door halfway through.

This totally makes sense in my opinion, so I modified the PR accordingly.

Alasdairbugs (changes requested):
Sorry, wrong PR. will review this one shortly

mstoeckl:
I do not see any problems with the current version of this PR, and have tested it in a variety of scenarios (going through doors, applying damage, sector changes, etc.) without finding any issues.

Alasdairbugs:
Nevermind, I can't review this, because there is no functional windows build. I'm going to assume it works though and i think 2 reviewers is enough.

For context, i made a review in the wrong place. Sorry!

MatusGuy (approved):

Frostwithasideofsalt:
cant re-test: no working windows builds.

Frostwithasideofsalt (approved):

PR #3196, opened 2025-01-30
Fix addon MD5 buffer overflow

tulpenkiste:
The original code was using `sizeof(so)` which, on my machine, is returning `8`. This causes the game to try to write to a point past `so[32]`, causing a buffer overflow.
This PR simply hardcodes the size to 3 (as each char is always 1 byte large, theres 2 chars inserted via the format + 1 NULL char). This makes it not buffer overflow without breaking existing addon functionality.

### DIFF
--- a/src/addon/md5.cpp
+++ b/src/addon/md5.cpp
@@ -161,7 +161,7 @@ std::string MD5::hex_digest() {
   for (i=0; i<16; i++)
   {
     char* so = s + i * 2;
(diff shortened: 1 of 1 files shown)


MatusGuy:
Is this a duplicate of #3136?

tulpenkiste:
Ah, looks like it is, although Semphriss's explanation is wrong

tulpenkiste:
Can something please be done with either this or #3136?

PR #3197, opened 2025-02-02
Allow to ignore multiple objects when checking for collisions

### LINKED ISSUE #3194: [BUG] Enemy Error
SuperTux Version: SuperTux-v0.6.3-1571-g0bf228934

System Information: Windows 64 

Expected Behavior: 

https://github.com/user-attachments/assets/b4013c53-0faf-4568-860d-d1b5cc5bec9b

Actual Behavior:

https://github.com/user-attachments/assets/66d03421-f8bf-4d50-b496-8d4ae1a352e8

tobbi:
Basically, it's in the title.

Currently, only one object can be set as ignore object for the collision system. This is alright when we want to prevent self-collisions. However, sometimes multiple objects need to be ignored. For example in case of ledge behaviour where we want to prevent badguys from assuming players can be walked on.

Therefore, I changed `GameObject* ignore_object` to `const std::vector<GameObject*>&`.

Furthermore, I've changed the code to ignore players when checking whether the badguy may fall because we don't want badguys to consider players as suitable target for continuing to walk when their ledge behaviour is set to stay on the platform.

Fixes #3194

### DIFF
--- a/src/badguy/badguy.cpp
+++ b/src/badguy/badguy.cpp
@@ -850,6 +850,14 @@ BadGuy::might_fall(int height)
   float oy = get_bbox().get_bottom() + 1.f;
 
   float fh = static_cast<float>(height);
+  std::vector<const CollisionObject*> ignore_objects = {};
+
+  for(const auto& player : Sector::current()->get_players())
+  {
+    ignore_objects.push_back(player->get_collision_object());
+  }
+
+  ignore_objects.push_back(&m_col);
 
   if (m_detected_slope == 0)
   {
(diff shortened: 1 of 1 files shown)


MatusGuy (commented):
Why does this ignore only players now? Does this fix a bug?

MatusGuy:
Oops, accidentally reviewed sorry

tobbi:
> Why does this ignore only players now? Does this fix a bug?

>> Fixes https://github.com/SuperTux/supertux/issues/3194 (see title)

tobbi:
Maybe `RayCastIgnore::IGNORE_OBJECTS` would yield the same result? IDK.

MatusGuy (commented):
I believe there's a better way of fixing this. For the BadGuy, you can detect if the object hit by the raycast is static, or, in other words, either COLGROUP_MOVING or COLGROUP_MOVING_STATIC.
If it's really needed to ignore objects (which is not the case in might_fall) we can think about this later. I'm rewriting these functions sooner anyway

PR #3198, opened 2025-02-02
Let squished enemies fade out instead of vanishing instantly

Brockengespenst:
Squished enemies will now fadeout instead of vanishing instantly. This looks a bit smoother and is more aligned with the melting or burning animation.

### DIFF
--- a/src/badguy/badguy.cpp
+++ b/src/badguy/badguy.cpp
@@ -42,6 +42,7 @@
 static const float SQUISH_TIME = 2;
 static const float GEAR_TIME = 2;
 static const float BURN_TIME = 1;
+static const float FADEOUT_TIME = 0.2f;
 
 static const float X_OFFSCREEN_DISTANCE = 1280;
 static const float Y_OFFSCREEN_DISTANCE = 800;
@@ -77,7 +78,8 @@ BadGuy::BadGuy(const Vector& pos, Direction direction, const std::string& sprite
   m_is_active_flag(),
   m_state_timer(),
   m_on_ground_flag(false),
-  m_colgroup_active(COLGROUP_MOVING)
+  m_colgroup_active(COLGROUP_MOVING),
+  m_alpha_before_fadeout(1.0f)
 {
   SoundManager::current()->preload("sounds/squish.wav");
   SoundManager::current()->preload("sounds/fall.wav");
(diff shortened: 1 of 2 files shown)


Brockengespenst:
Just an idea. Looks better to me when squished Badguys do not vanish instanly. If you do not like it, I can delete this PR.

bruhmoent:
Could you please provide video footage?

Brockengespenst:
Sure: 
https://github.com/user-attachments/assets/df038f80-1049-404c-8545-d60e7ddefb70

tobbi:
Looking good to me.

src/badguy/badguy.cpp
```diff
@@ -315,12 +317,18 @@ BadGuy::update(float dt_sec)
     case STATE_SQUISHED:
       m_is_active_flag = false;
       if (m_state_timer.check()) {
-        remove_me();
+        m_alpha_before_fadeout = m_sprite->get_alpha();
```
MatusGuy, line 320: An enemy at 0.5 alpha should fade out faster. In that case, you don't need this variable. You can multiply the fadeout time by the alpha in the set_state function.

src/badguy/badguy.cpp
```diff
@@ -315,12 +317,18 @@ BadGuy::update(float dt_sec)
...
         break;
       }
       m_col.set_movement(m_physic.get_movement(dt_sec));
       break;
-
+    case STATE_SQUISHED_FADING_OUT:
+      if (m_state_timer.check())
+        remove_me();
+      else
+        m_sprite->set_alpha(m_alpha_before_fadeout - (m_alpha_before_fadeout * m_state_timer.get_progress()));
```
MatusGuy, line 330: And here you'd only need to put m_state_timer.get_progress()
```suggestion
        m_sprite->set_alpha(m_state_timer.get_progress()));
```

src/badguy/badguy.cpp
```diff
@@ -760,6 +768,9 @@ BadGuy::set_state(State state_)
     case STATE_SQUISHED:
       m_state_timer.start(SQUISH_TIME);
       break;
+    case STATE_SQUISHED_FADING_OUT:
+      m_state_timer.start(FADEOUT_TIME);
```
MatusGuy, line 772:
```suggestion
      m_state_timer.start(FADEOUT_TIME * m_sprite.get_alpha());
```

src/badguy/badguy.cpp
```diff
@@ -42,6 +42,7 @@
 static const float SQUISH_TIME = 2;
 static const float GEAR_TIME = 2;
 static const float BURN_TIME = 1;
+static const float FADEOUT_TIME = 0.2f;
```
MatusGuy, line 45: I'm not sure because I havent tested this in game but this might be too fast?? Someone else needs to confirm.

src/badguy/badguy.cpp
Brockengespenst: If alpha was already lower than 1, the suggested change would put it up to 1 before fading out again.

src/badguy/badguy.cpp
Brockengespenst: Good point, I changed it.

src/badguy/badguy.cpp
Brockengespenst: This is only valid for the fadeout time, the alpha value at the start of the fadout is still needed.

src/badguy/badguy.cpp
Hume2: Valid point here. Although all badguys currently have alpha=1.0, it doesn't have to hold in future. If they already start at a lower value of alpha, then the suggested change won't work. Just for better readability, I would write it as:
```
m_alpha_before_fadeout * (1.0f - m_state_timer.get_progress())
```

Hume2 (approved):
code OK
test OK

Vankata453 (approved):
The fadeout is something I didn't know I needed, it feels fancy!

PR #3216, opened 2025-03-20
Icy Island Overhaul for v0.7

Rusty-Box:
Update the entirety of Icy Island to make use of all the new graphics, mechanics, music and overall theming for the next versions in hopes this can be considered the "final" major update on world 1 specifically.

There is still a few things left to finalize outside of levels for this to be considered "complete", such as a proper system for handling secret exits and cutscenes. Right now every paths is open so you can acccess all levels on the worldmap.


tobbi:
When writing instructional billboards, do we actually have a key template for <action>?

tylerandari13:
> When writing instructional billboards, do we actually have a key template for "action"?

I doubt it, although I think something like that would be cool. Like if you typed `Press <action> to shoot a fireball!` in an infoblock it would show as `Press Z to shoot a fireball!` or `Press SPACE to shoot a fireball!` ingame.

tobbi:
Oh, yeah, I meant a key graphic for action like we do for up / down / left / right. But yours works as well.

tylerandari13:
That's also a cool idea if we can pull it off.

FILIPOKonGithup:
It would be better if you replaced the **poles** on the billboards, with the 0.7 ones.

komaros1:
why are 23rd airbone and above the arctic skyes the same old levels or is it just me

Rusty-Box:
> why are 23rd airbone and above the arctic skyes the same old levels or is it just me

No? I checked the files again and these are the new ones.

Hypernoot:
resolved by #3338

PR #3218, opened 2025-03-22
Allow referencing actual keys the player uses inside strings

tobbi:
This PR allows for referencing actual player keys inside texts, like info boxes etc.

The syntax is `{keyref up}` or instead of `up` you can name any of the other keys in SuperTux.

Unfortunately, this currently only works for one mention of `{keyref}` per string. This needs to be fixed in order to support multiple mentions.

### DIFF
--- a/src/supertux/menu/keyboard_menu.cpp
+++ b/src/supertux/menu/keyboard_menu.cpp
@@ -74,7 +74,7 @@ KeyboardMenu::KeyboardMenu(InputManager& input_manager, int player_id) :
 }
 
 std::string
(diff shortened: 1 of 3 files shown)


MatusGuy (commented):
This idea seems fundamentally wrong because it doesn't account for players who are playing with touch controls or with joysticks. I can't think of a way to implement this to account for those cases in an intuitive manner. Besides, the player should already know which keys are assigned to which actions by going to the settings menu.

tobbi:
Closing as per the first comment.

PR #3220, opened 2025-03-24
Break Rusty Trampoline on grab instead of ungrab

### LINKED ISSUE #3205: Change Rusty Spring's behavoir to break upon pickup
Change the rusty spring to break upon pickup instead of upon fall, as to keep the same effect of preventing pickup, whilst also allowing owls to carry and drop them without issue.

tobbi:
This fixes #3205

### DIFF
--- a/src/object/rusty_trampoline.cpp
+++ b/src/object/rusty_trampoline.cpp
@@ -129,14 +129,14 @@ RustyTrampoline::collision_solid(const CollisionHit& hit) {
 void
 RustyTrampoline::grab(MovingObject& object, const Vector& pos, Direction dir) {
   Rock::grab(object, pos, dir);
+  set_action("breaking", 1);
+  counter = 0; //remove in update()
 }
 
 void
 RustyTrampoline::ungrab(MovingObject& object, Direction dir)
 {
   Rock::ungrab(object, dir);
-  set_action("breaking", 1);
-  counter = 0; //remove in update()
 }
 
 bool


Hypernoot (approved):
it workz

I just don't know why the rusty trampolines should be grabbable at all, if they decay no matter what. I see no point why the player should grab a rusty trampoline then. Except that a trollface appears saying: you're really dumb trolololol.

Or maybe you can have a passage obstructed by rusty trampolines and you have to break them one by one in order to pass through?

tobbi:
Yeah, I don't know. This question is beyond the scope of this pull request.

Alasdairbugs:
Theres a few problems here!

1. Rusty spring functionality change wasn't agreed by multiple people, so why has it been sent for a PR?
2. Theres a better and more obvious way to give rusty springs a better functionality:

- rusty springs can be picked up, and placed down
- rusty springs disintigerate when they touch water OR touch water when tux is holding them
- rusty springs turn into regular springs when they touch lava OR tux touches lava

and maybe you can keep the "only 5 uses" kind of thing like ice cubes have, but make it so you can pick up and place it as normal.
maybe you can also make it so that it cannot be upthrown or thrown, only placed beside you(?).

these are some way better ideas. this should be discussed more before making changes i think. and we should not go changing features just because one person wrote one sentence on it, thats silly

Hypernoot:
redirecting discussion to #3403.

PR #3225, opened 2025-03-27
Fixes #3208: bounce mriceblock to bonusblock by bumper doesn't go back

### LINKED ISSUE #3208: [Bug]: Stunned iceblock pushed toward bonusblock by bumper doesn't bounce off bonusblock it hit
### SuperTux Version

0.6.3-1599-g6f1078d10

### System Information

win11

### Expected Behavior

i expected the iceblock to bounce off the bonusblock it was pushed toward by the bumper. after it hit the bonusblock, it should've returned to the bumper to be bumped again and so on.

### Actual Behavior

the iceblock did not bounce off the bonusblock and instead got stuck on the bonusblock, unable to move. the bonusblock was also spamming loud sounds due to the iceblock being stuck on it.

1. place any random amount of solid blocks/tiles horizontally.
2. place a bumper on a solid block/tile facing right or left depending on how you set it up.
3. place a bonusblock on top of the end solid block/tile
4. place an iceblock on top of the solid blocks/tiles in between the bumper and the bonusblock.
5. test the level and jump on the iceblock to stun it. kick it toward the bumper or the bonusblock and observe the behavior.

example:
Image

### Additional Information

_No response_

ist1106037:
When a MrIceBlock is kicked into a trampoline, it bounces back facing backwards. The collision_solid() function, when the MrIceBlock's state is KICKED (which is the case), didn't expect to receive a hit from the right while facing left, and vice-versa. I added that condition and made it so it plays the collision sound and goes back, just like normal, but skips the flipping orientation, as it should slide forward and not backwards.
Video demonstration: https://www.youtube.com/watch?v=oAgOuTEf52w

Closes #3208

### DIFF
--- a/src/badguy/mriceblock.cpp
+++ b/src/badguy/mriceblock.cpp
@@ -166,6 +166,9 @@ MrIceBlock::collision_solid(const CollisionHit& hit)
       m_dir = (m_dir == Direction::LEFT) ? Direction::RIGHT : Direction::LEFT;
       SoundManager::current()->play("sounds/iceblock_bump.wav", get_pos());
       m_physic.set_velocity_x(-m_physic.get_velocity_x() * .975f);
+    } else if ((hit.left && m_dir == Direction::RIGHT) || (hit.right && m_dir == Direction::LEFT)) {
+      SoundManager::current()->play("sounds/iceblock_bump.wav", get_pos());
+      m_physic.set_velocity_x(-m_physic.get_velocity_x() * .975f);
     }
     set_action("flat", m_dir, /* loops = */ -1);
     if (fabsf(m_physic.get_velocity_x()) < walk_speed * 1.5f)


ist1106037:
@tobbi could you look at this? Thanks

tobbi:
Technically, this is very similar to the if statement before, could you try to unify those two statements into one?

ist1106037:
Done. Apologies for the delay, had a few technical issues I could only resolve at home. I believe this is the most concise way I can make it.

tobbi:
> Done. Apologies for the delay, had a few technical issues I could only resolve at home. I believe this is the most concise way I can make it.

Looks good from my perspective!

ist1106037:
Should we ask for other reviews?

bruhmoent (approved):
Works!

tobbi:
I think this is trivial enough to merge. Also, it has 2 reviews now.

tobbi:
Thanks for your work! Much appreciated 👍

PR #3226, opened 2025-03-27
Fixes #3215: Ice isn't slippery for rocks

### LINKED ISSUE #3215: [Bug]: Ice isn't really slippery for rocks
### SuperTux Version

v0.6.3, commit fc7965b

### System Information

N/A

### Expected Behavior

Rock slides smoothly on ice and takes a few seconds to slow down.

### Actual Behavior

Tux throws the block from this position with a running start, but the rock stops as shown very quickly after around 5 tiles. It isn't really slippery.
Image

1. Cover a floor in ice
2. Throw a rock on the ice and try to make it slide.

### Additional Information

_No response_

goncalof148:
Since the ice floor wasn't being recognized as ice, specially on the hills on level "Tip of the Ice", the floor didn't have the slippery feature. I changed the tile attributes to address this problem and added a condition on the rock logic to include a slippery property on ice. Now tux can throw a rock to hills and it will slide back down due to the slippery property.

Video: https://youtube.com/shorts/B_z0Fx5yYJc?feature=share

Closes #3215


goncalof148:
@tobbi could you check this, please?

Hypernoot:
It looks like the behaviour is almost the same, both in master and this PR. So what was actually fixed?

0. non-ice solid flat tiles - In both versions, the rock stops instantly. That is correct.
1. solid flat ice - In both versions, the rock slides for approx. 1s and then it stops. Wasn't that supposed to be different?
2. solid ice slope - In master, the rock stops instantly, unless it is given an initial speed. In this PR, the rock slowly slides down and jitters occasionally.
3. unisolid flat ice - In both versions, the rock just stops.
4. unisolid ice slope - In both versions, the rock just stops.
5. non-ice solid tiles with unisolid flat ice overlay - same as case 3
6. non-ice solid tiles with unisolid ice slope overlay - same as case 4

goncalof148:
@Hypernoot could you check this last version?

Solid Flat Ice: I decreased the friction
Solid Ice Slope: Now there is a minimum force to throw the rock
Unisolid flat ice and slope: I updated that part

Hypernoot (changes requested):
I didn't test it in-game but I suspect that anything essential has changed, except that the ice friction was decreased. So that fixes point 1. The rest remains.

If this PR shall be treated as a partial fix, then all you need is decreasing one value, the ice friction. And you can drop all the other changes.

goncalof148:
@Hypernoot I thought I fixed the problems! Could you help me identify the right tiles? On the code I can't identify them by their type.

Hypernoot (approved):
As you need it ASAP for your university, I'm giving you an approval. But normally, I wouldn't accept this code quality. Your code is equivalent to:
```
  if (m_on_ground || (hit.bottom && m_on_ice)) {
    // Full friction!
    m_physic.set_velocity_x(m_physic.get_velocity_x() * (1.f - (GROUND_FRICTION * (m_on_ice ? 0.3f : 10.f))));
  }
```
So you effectively just changed two constants. I don't even understand why you changed 1.f to 10.f, nobody was complaining at rock friction on grass being too low.

The PR cannot be merged yet, as it wasn't tested enough. You need at least one more review. But let's hope that your university is going to give you course credits for that anyway. Fingers crossed.

Hypernoot (changes requested):
Your code is equivalent to:
```
  if (m_on_ground || (hit.bottom && m_on_ice)) {
    // Full friction!
    m_physic.set_velocity_x(m_physic.get_velocity_x() * (1.f - (GROUND_FRICTION * (m_on_ice ? 0.3f : 10.f))));
  }
```
This is almost the original code, except that two values are changed.

Can you tell me why you changed the non-ice friction from 1.0 to 10.0? Nobody complained about that.

goncalof148:
@Hypernoot the non-ice friction value changed by mistake! I will fix that! I had changed the GROUND_FRICTION constant and then also changed that value and forgot to change!

src/object/rock.cpp
```diff
@@ -177,9 +177,9 @@ Rock::collision_solid(const CollisionHit& hit)
     m_at_ceiling = true;
   }
 
-  if (m_on_ground || (hit.bottom && m_on_ice)) {
+    if (m_on_ground || (hit.bottom && m_on_ice)) {
```
Hypernoot: You accidentally messed up indentation.

Hypernoot (approved):

PR #3227, opened 2025-03-28
Fixes #3079 Buttjump bonusblock runs script twice & unports trampoline

### LINKED ISSUE #3079: [Bug]: Hitting a bonus block with buttjump runs its script twice and unportables the trampoline
### SuperTux Version

SuperTux-v0.6.3-1388-g94bf5074d

### System Information

Windows 10, 64-bit

### Expected Behavior

When Tux hits a bonus block that has a script defined in it, the script will get activated normally, no matter how the block was hit or what content it has.

### Actual Behavior

If the bonus block has one of the following contents, hitting it with a buttjump will activate the script twice in a row.
- Coin
- Trampoline
- Light (On or Off)
- Coin Rain

With every other content the block can have, the script will run one single time, as expected.
- Script
- Any Powerup
- Custom Content
- Portable Trampoline
- Coin Explosion
- Rock
- Potion

1. Place a bonus block object in your level
2. Set the bonus block's content to a Coin, a Trampoline, a Light or a Coin Rain
3. Add to the bonus block a script that you can keep track of what it does
4. Hit the bonus block with a buttjump

If you can keep track of what the script does, you will notice that it has done it one extra time. If the script adds a value of 1 to a variable, it will add 2 to the variable. If the script plays a sound, the sound will be twice as loud. If the script spawns an object, it will spawn two objects.

### Additional Information

In this example video, all the bonus blocks have the same script, a function that spawns a rock below Tux with its x position rounded to every 3 tiles. As you can see, when Tux hits the bonus block with a buttjump and the block has one of the described contents, two rocks are spawned at the same time.

https://github.com/user-attachments/assets/1ed69681-5c0c-4fa5-b8d0-d74fc79bd084

Additional note: You have probably noticed as well that when a bonus block containing a portable trampoline is hit with a buttjump, the trampoline that comes out of it is not portable, so this is an extra issue.

https://github.com/user-attachments/assets/3d067e20-12e1-4095-8479-8407bc9a6aa5

- [X] I have read <https://github.com/SuperTux/supertux/blob/master/CONTRIBUTING.md#bug-reports>.
- [X] I have verified this isn't an issue that's already been reported.
- [X] I have verified this isn't a [discussion](https://github.com/SuperTux/supertux/discussions), or an issue about a crash or a feature request, but rather an actual bug ─ that is, the game did something not intended.
- [X] I have verified this issue is **not** about wrong translations (use Transifex for those), or anything unsupported (e.g. third-party add-ons).
- [X] In this report, I have only included details about **one** (1) bug.
- [X] If I make a mistake while submitting this report, I agree to use the "Edit" feature to correct it, instead of closing this issue and opening a new one.

Martim-F:
When someone buttjumps onto a bonus_block it calls the try_drop function in which some objects on that function would call try_open, another function on which it would run a script, then come back to try_drop and run the script again since the condition to run the script didn't include countdown, the bool that checks if try_open is called, so I just added it to the condition.
Then for the trampoline error the add function had the position that the trampoline is created and an enum (1 or 0, or respectivelly true or false) which was incorrect and creating the wrong trampoline, so I just changed from true, which created a stationary trampoline, to a false, a portable one.

Closes #3079

### DIFF
--- a/src/object/bonus_block.cpp
+++ b/src/object/bonus_block.cpp
@@ -616,7 +616,7 @@ BonusBlock::try_drop(Player *player)
     }
     case Content::PORTABLE_TRAMPOLINE:
     {
-      Sector::get().add<Trampoline>(get_pos() + Vector(0, 32), false);
+      Sector::get().add<Trampoline>(get_pos() + Vector(0, 32), Trampoline::PORTABLE);
       countdown = true;
       break;
     }
--- a/src/object/bonus_block.cpp
+++ b/src/object/bonus_block.cpp
@@ -616,7 +616,7 @@ BonusBlock::try_drop(Player *player)
     }
     case Content::PORTABLE_TRAMPOLINE:
     {
-      Sector::get().add<Trampoline>(get_pos() + Vector(0, 32), true);
+      Sector::get().add<Trampoline>(get_pos() + Vector(0, 32), false);
       countdown = true;
       break;
     }
@@ -638,7 +638,7 @@ BonusBlock::try_drop(Player *player)
   if (play_upgrade_sound)
     SoundManager::current()->play("sounds/upgrade.wav", get_pos(), UPGRADE_SOUND_GAIN);
 
-  if (!m_script.empty()) { // Scripts always run if defined.
+  if (!m_script.empty() && countdown) { // Scripts always run if defined.
     Sector::get().run_script(m_script, "powerup-script");
   }
 


tobbi (approved):
I didn't test it but I think we can trust you on this one.

Hume2 (approved):
I tested it and it works.

src/object/bonus_block.cpp
```diff
@@ -616,7 +616,7 @@ BonusBlock::try_drop(Player *player)
     }
     case Content::PORTABLE_TRAMPOLINE:
     {
-      Sector::get().add<Trampoline>(get_pos() + Vector(0, 32), true);
+      Sector::get().add<Trampoline>(get_pos() + Vector(0, 32), false);
```
MatusGuy, line 619:
```suggestion
      Sector::get().add<Trampoline>(get_pos() + Vector(0, 32), Trampoline::PORTABLE);
```

Martim-F:
@MatusGuy the change you requested is done.

tobbi:
Thank you for your contribution, @Martim-F. 👍

I'm gonna merge this.

PR #2848, opened 2024-03-26
Improve ghoul behavior [WIP]

weluvgoatz:
This PR improves the Ghoul's behavior by making it more similar to the concept by @Rusty-Box.  It may have issues.  Please test it.
NOTE: This PR is waiting on new graphics for the Ghoul! Currently it still uses the old sprites as a placeholder.

PR #3230, opened 2025-03-31
Improve ghoul behaviour (rebase+fix)

Hume2:
a fixed and rebased version of #2848

### DIFF
--- a/src/badguy/ghoul.cpp
+++ b/src/badguy/ghoul.cpp
@@ -137,7 +137,7 @@ Ghoul::get_settings()
 }
 
 bool
-Ghoul::collision_squished(GameObject& object)
+Ghoul::collision_squished(MovingObject& object)
 {
   auto player = Sector::get().get_nearest_player(m_col.m_bbox);
   if (player)
--- a/src/badguy/ghoul.cpp
+++ b/src/badguy/ghoul.cpp
@@ -39,8 +39,9 @@ Ghoul::Ghoul(const ReaderMapping& reader) :
   reader.get("speed", m_speed, DEFAULT_SPEED);
   reader.get("track-range", m_track_range, DEFAULT_TRACK_RANGE);
 
-  m_sprite->set_action(m_dir);
+  set_action(m_dir);
   m_physic.set_gravity_modifier(0.2f);
+  set_colgroup_active(COLGROUP_TOUCHABLE);
 
   SoundManager::current()->preload("sounds/fall.wav");
 }
@@ -55,7 +56,7 @@ Ghoul::active_update(float dt_sec)
   if (m_fake_dead && m_respawn_timer.check()) {
     initialize();
     m_physic.enable_gravity(true);
-    set_colgroup_active(COLGROUP_MOVING);
+    set_colgroup_active(COLGROUP_TOUCHABLE);
     m_fake_dead = false;
     m_respawn_timer.stop();
   }
@@ -77,14 +78,14 @@ Ghoul::active_update(float dt_sec)
   {
   case NORMAL:
     m_speed_modifier = std::max(0.f, m_speed_modifier - (dt_sec * 2.f));
-    m_sprite->set_action(player->get_bbox().get_middle().x < get_bbox().get_middle().x ? "normal-left" : "normal-right", 1);
+    set_action(player->get_bbox().get_middle().x < get_bbox().get_middle().x ? "left" : "right", 1);
     if (m_sprite->animation_done()) {
       m_chase_dir = glm::normalize(dist);
       m_sprite_state = SpriteState::FAST;
     }
     break;
   case FAST:
-    m_sprite->set_action(player->get_bbox().get_middle().x < get_bbox().get_middle().x ? "fast-left" : "fast-right", 1);
+    set_action(player->get_bbox().get_middle().x < get_bbox().get_middle().x ? "fast-left" : "fast-right", 1);
     m_speed_modifier = 3.5f;
     if (m_sprite->animation_done()) {
       m_sprite_state = SpriteState::NORMAL;
--- a/src/badguy/ghoul.cpp
+++ b/src/badguy/ghoul.cpp
@@ -15,202 +15,149 @@
 
 #include "badguy/ghoul.hpp"
 
+#include "audio/sound_manager.hpp"
 #include "object/player.hpp"
-#include "editor/editor.hpp"
-#include "sprite/sprite.hpp"
-#include "supertux/game_session.hpp"
 #include "supertux/sector.hpp"
 #include "util/reader_mapping.hpp"
-#include "util/writer.hpp"
 
-static const float FLYSPEED = 80.0f; /**< Speed in px per second. */
-static const float TRACK_RANGE = 2500.0f; /**< At what distance to start tracking the player. */
+static const float DEFAULT_SPEED = 40.0f;
+static const float DEFAULT_TRACK_RANGE = 2500.0f;
+static const float RESPAWN_TIME = 5.f;
 
 Ghoul::Ghoul(const ReaderMapping& reader) :
   BadGuy(reader, "images/creatures/ghoul/ghoul.sprite"),
-  PathObject(),
-  m_mystate(STATE_IDLE),
-  m_flyspeed(),
-  m_track_range()
+  m_speed(),
+  m_track_range(),
+  m_speed_modifier(),
+  m_fake_dead(),
+  m_chase_dir(),
+  m_respawn_timer(),
+  m_sprite_state(SpriteState::NORMAL)
 {
-  reader.get("flyspeed", m_flyspeed, FLYSPEED);
-  reader.get("track-range", m_track_range, TRACK_RANGE);
-  
-  bool running;
-  reader.get("running", running, false);
+  m_countMe = false;
 
-  init_path(reader, running);
+  reader.get("speed", m_speed, DEFAULT_SPEED);
+  reader.get("track-range", m_track_range, DEFAULT_TRACK_RANGE);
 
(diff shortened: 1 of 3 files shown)


Frostwithasideofsalt (changes requested):
there seems to be a bug where the ghoul refuses to move until i pass below it at least once.

Rusty-Box:
Looking pretty good so far. A few suggestions I have:
- Have Ghoul thrust more agressivly towards the player than just up at a set angle.
- If Tux is too far away, I think Ghoul just stop chasing him and simply hover in mid-air waiting.
- Ghouls stun time could be reduced a little bit.

Frostwithasideofsalt:
well it seems my issue with the ghoul not moving is fixed lol

Hypernoot:
Thanks for review!
>     * Have Ghoul thrust more agressivly towards the player than just up at a set angle.
Can you please elaborate on what that means in particular?

>     * If Tux is too far away, I think Ghoul just stop chasing him and simply hover in mid-air waiting.
That is already implemented but the default tracking range is quite large. You can set it to a lower value to actually see it. The default value can be reduced if needed.

>     * Ghouls stun time could be reduced a little bit.
The stunning time is now 5s. I can reduce it to maybe 3s.

Rusty-Box:
> Can you please elaborate on what that means in particular?

Currently, when the player is higher than Ghoul, Ghoul does simply push itself relatively straight upwards whlie slowly moving towards him. What I had in mind was closer to the Blooper from Mario, in that instead of going straight up, Ghoul pushes itself towards Tux, so more diagonal-up. They only move with every thrust, standing still before charging up another thrust to move closer to Tux again. Like a jellyfish.
 
> That is already implemented but the default tracking range is quite large. You can set it to a lower value to actually see it. The default value can be reduced if needed.

Nice, I'll have to try out a few values then to find a good default.

> The stunning time is now 5s. I can reduce it to maybe 3s.

Try 4. Should be enough.

Hypernoot:
As you may have noticed, when the ghoul is bellow its target position, he jumps upwards by a defined step. When Tux is above ghoul but not much higher, then a single jump may suffice for the ghoul to reach Tux. If Tux is higher than that, then the ghoul actually makes multiple consecutive jumps to reach Tux. Shall I make the jumps shorter to make it more apparent?

Rusty-Box:
Yeah, so the horizontal speed is still very low when going up. The idea was that it moves faster horizontally too when it thrusts itself upwards. The speed when it goes down is fine but should have much more speed per thrust horizontally.

Hypernoot:
@Rusty-Box feel free to check

Frostwithasideofsalt (changes requested):
https://github.com/user-attachments/assets/41a62f15-8ab7-43e4-b6ed-0116e0dba26d
there's a bug where a ghoul will sometimes trusts *backwards?* in which it seems to have no speed cap or something, causing it to go flying

Frostwithasideofsalt (approved):

src/badguy/ghoul.cpp
```diff
@@ -15,202 +15,343 @@
...
-Ghoul::deactivate()
+Ghoul::horizontal_thrust()
 {
-  switch (m_mystate) {
-    case STATE_TRACKING:
-      m_mystate = STATE_IDLE;
-      break;
-    default:
-      break;
+  const float vy = (UP_VELOCITY + DOWN_VELOCITY) / 2.0f;
```
MatusGuy, line 212: What is this value I keep seeing? What does it mean? The upwards thrust velocity?

src/badguy/ghoul.cpp
```diff
@@ -15,202 +15,343 @@
...
+    const float a = dist.x > 0.0f ? -HORZ_ACCELERATION : HORZ_ACCELERATION;
+    m_physic.set_acceleration_x(a);
+    const float vx = m_physic.get_velocity_x();
+    const float vx_diff = m_speed * 9.0f;
+    if (t == 0.0f) {
+      m_physic.set_velocity_x(dist.x > 0.0f ? vx - vx_diff : vx + vx_diff);
+    } else {
+      const float vx_needed = dist.x / t - a * t / 2.0f;
+      m_physic.set_velocity_x(std::max(vx - vx_diff, std::min(vx + vx_diff, vx_needed)));
+    }
```
MatusGuy: - Use `std::abs`
- Use a return guard here. Please check if I got the condition right.
- Use `std::clamp`
```suggestion
  if (t * (std::abs(m_physic.get_velocity_x()) + m_speed) / 2.0f <= std::abs(dist.x) && dist.y <= 0)
    // No need for acceleration.
    return;
    
  const float a = dist.x > 0.0f ? -HORZ_ACCELERATION : HORZ_ACCELERATION;
  m_physic.set_acceleration_x(a);
  const float vx = m_physic.get_velocity_x();
  const float vx_diff = m_speed * 9.0f;
  if (t == 0.0f) {
    m_physic.set_velocity_x(dist.x > 0.0f ? vx - vx_diff : vx + vx_diff);
  } else {
    const float vx_needed = dist.x / t - a * t / 2.0f;
    m_physic.set_velocity_x(std::clamp(vx_needed, vx - vx_diff, vx + vx_diff));
  }

```

src/badguy/ghoul.cpp
```diff
@@ -15,202 +15,343 @@
...
+Ghoul::update_speed(const Vector& dist)
 {
-  if (get_walker() && get_walker()->is_running()) {
-    m_mystate = STATE_PATHMOVING_TRACK;
+  const float vx = m_physic.get_velocity_x();
+  if (vx >= -m_speed && vx <= m_speed) {
+  	m_physic.set_acceleration_x(0.0f);
+  	const float vy = dist.y < 0.0f ? (UP_VELOCITY + DOWN_VELOCITY) / 2.0f : DOWN_VELOCITY;
+    const float t = dist.y / vy;
+    if (t * m_speed > fabs(dist.x)) {
```
MatusGuy:
```suggestion
    if (t * m_speed > std::abs(dist.x)) {
```

src/badguy/ghoul.cpp
```diff
@@ -15,202 +15,343 @@
...
-      break;
-
-    default:
-      assert(false);
+void
+Ghoul::roaming_deccel_check()
+{
+  const float vx = m_physic.get_velocity_x();
+  const float ax = m_physic.get_acceleration_x();
+  if (vx * ax > 0) {
```
MatusGuy, line 246: What is `vx * ax`? The resulting x velocity when affected by the acceleration?

Frostwithasideofsalt:
ok testing it further, setting the speed really low causes the ghoul to fly backwards. tbh idk why you can even set the speed i think it might be best to just make you not able to edit it?

src/badguy/ghoul.cpp
Hypernoot: It is the mean velocity of linearly accelerated movement.

src/badguy/ghoul.cpp
Hypernoot: It tests whether `vx` and `ax` have the same sign, similar to the dot product.

Hypernoot:
Good catch! I will remove the speed setting completely, you can't choose the speed of most badguys, so there's no point why this one should be changeable.

src/badguy/ghoul.cpp
Hypernoot: fixed

src/badguy/ghoul.cpp
Hypernoot: fixed

src/badguy/ghoul.cpp
MatusGuy:
```suggestion
  if (std::signbit(vx) == std::signbit(ax)) {
```

Hypernoot: Yes, but I don't want the condition to trigger when either `vx` or `ax` is zero.

MatusGuy: Alright then

MatusGuy (approved):

src/badguy/ghoul.cpp
```diff
@@ -15,202 +15,339 @@
...
+  BadGuy::collision_badguy(badguy, hit);
+  return CONTINUE;
 }
 
 void
-Ghoul::stop_moving()
+Ghoul::collision_solid(const CollisionHit& hit)
 {
-  get_walker()->stop_moving();
+// allows it to continue moving if it hits a wall.
```
bruhmoent:
```suggestion
  // allows it to continue moving if it hits a wall.
```

Hypernoot: fixed

bruhmoent (approved):

Rusty-Box (approved):
Ghoul movement feels much better now. Maybe a few tiny things could be adjusted once used more actively in level design but nothing we cannot address later on along the way separately.

tobbi:
Build failures are unrelated.

src/badguy/ghoul.cpp
```diff
@@ -15,202 +15,339 @@
...
+    m_physic.set_velocity_x(dist.x > 0.0f ? vx - vx_diff : vx + vx_diff);
+  } else {
+    const float vx_needed = dist.x / t - a * t / 2.0f;
+    m_physic.set_velocity_x(std::clamp(vx_needed, vx - vx_diff, vx + vx_diff));
   }
 }
 
 void
-Ghoul::active_update(float dt_sec)
+Ghoul::start_roaming_deccel()
```
tobbi: Shouldn't this be `decel` (typo)?

Hypernoot: Screw your arbitrary gramar rules, I'm an engineer and not a gramar expert :P

Issue #3209, opened 2025-02-23
[Bug]: Tux unable to stand on / jump off frozen badguy that is bouncing on trampoline


celeste5000:
### SuperTux Version

0.6.3-1599-g6f1078d10

### System Information

win11

### Expected Behavior

i expected tux to be able to stand on and jump off the frozen badguy bouncing on the trampoline onto another surface while the frozen badguy is still briefly moving upwards from the trampoline.

### Actual Behavior

tux was not able to stand on or jump off the frozen badguy, instead tux was stuck in the falling animation while on the frozen badguy. backflipping onto the frozen badguy made tux slide off of it while stuck in the backflipping animation rather than land on it normally.

### Steps To Reproduce Actual Behavior

1. place a trampoline in a level
2. place any badguy that can be frozen (ivy, spiky, etc)
3. test the level and freeze the badguy
4. grab the frozen badguy and place it on a trampoline (portable or stationary works)
5. jump onto a solid surface and jump off it to land on the frozen badguy, or you can backflip onto it

example:
![Image](https://github.com/user-attachments/assets/22850eac-8116-45c9-af32-f1b283041b5c)

### Additional Information

![Image](https://github.com/user-attachments/assets/091b7a81-d201-4886-94f8-7e51ca87f3a9)
![Image](https://github.com/user-attachments/assets/3ca34303-dc47-4615-92fc-4d4e91a32af5)
![Image](https://github.com/user-attachments/assets/3347d3ec-2042-4e95-8834-8484a5b19517)

### Guidelines For Reporting Issues

- [x] I have read <https://github.com/SuperTux/supertux/blob/master/CONTRIBUTING.md#bug-reports>.
- [x] I have verified this isn't an issue that's already been reported.
- [x] I have verified this isn't a [discussion](https://github.com/SuperTux/supertux/discussions), or an issue about a crash or a feature request, but rather an actual bug ─ that is, the game did something not intended.
- [x] I have verified this issue is **not** about wrong translations (use Transifex for those), or anything unsupported (e.g. third-party add-ons).
- [x] In this report, I have only included details about **one** (1) bug.
- [x] If I make a mistake while submitting this report, I agree to use the "Edit" feature to correct it, instead of closing this issue and opening a new one.

Brockengespenst:
[...]

PR #3232, opened 2025-04-02
Fixes Bug #3209

Filipe2004:
Tux wasn't able to stand on or jump off a frozen badguy that was bouncing of a trampoline, getting stuck in the falling animation while on the frozen badguy. To solve this I added a verification on Player::collision for when the badguy is frozen and the player hits bottom. Inside the if I adapted the code on collision_solid so that it performs similar to when the player collides with a solid and the hit is on the bottom. Also added a verification for the velocity of the badguy, because for some reason if it is static it enter both the Player:collision and Player:collision_solid.

### DIFF
--- a/src/object/player.cpp
+++ b/src/object/player.cpp
@@ -2392,6 +2392,32 @@ Player::collision(MovingObject& other, const CollisionHit& hit)
       return FORCE_MOVE;
     if (m_stone)
       return ABORT_MOVE;
+    if (badguy->is_frozen() && hit.bottom && badguy->get_physic().get_velocity_y() != 0) {
+      if (m_physic.get_velocity_y() > 0)
+        m_physic.set_velocity_y(0);
+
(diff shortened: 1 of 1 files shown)


tobbi:
Hey, @Filipe2004

thanks for your contribution! I see that you copied a lot of source code from the `Player::collision_solid(CollisionHit&)` method. I'd like to avoid having duplicate source code. I'd appreciate if you could simply call the function above with appropriate parameters instead.

MatusGuy:
> simply call the function above with appropriate parameters instead.

Don't do this. The `collision_solid` function is used by the collision system to run callbacks on static collision.
Instead, write a new function, and call it there and here.

Filipe2004:
Hello, @MatusGuy 
Thanks for the suggestion, that's what I did
Sorry for taking so long to fix it

tobbi (approved):
Looking good from my side.

PR #3231, opened 2025-04-01
Crusher code rewrite

MatusGuy:
- [x] Implement existing functionality 
- [x] Fix Rock-Crusher collision interactions. Rocks may not push crushers. 
- [x] Use Root class for Corrupted Crusher roots
- [x]  Implement upwards facing crusher
- [x]  Add script that runs when crusher hits ground
- [x]  Polish & fix jank

Fixes #2403

PR #3234, opened 2025-04-03
Fix #3105 - Big rocks push small rocks through the floor

### LINKED ISSUE #3105: [Bug]: Big rocks push small rocks through the floor
### SuperTux Version

SuperTux Latest Nightly

### System Information

Windows 10 64-bit

### Expected Behavior

When big rocks are stacked on top of smaller rocks, they should not push the smaller rocks through the floor.

### Actual Behavior

When big rocks are stacked on top of smaller rocks, they push the smaller rocks through the floor.

Take a big rock and place it on top of a smaller rock.

### Additional Information

This is probably a result of the MovingStatic collision changes I made earlier this spring, so it might be worthy to add a check to not let bigger movingstatics push smaller movingstatics through movingstatics that are larger than the small one, or normal statics.

- [X] I have read <https://github.com/SuperTux/supertux/blob/master/CONTRIBUTING.md#bug-reports>.
- [X] I have verified this isn't an issue that's already been reported.
- [X] I have verified this isn't a [discussion](https://github.com/SuperTux/supertux/discussions), or an issue about a crash or a feature request, but rather an actual bug ─ that is, the game did something not intended.
- [X] I have verified this issue is **not** about wrong translations (use Transifex for those), or anything unsupported (e.g. third-party add-ons).
- [X] In this report, I have only included details about **one** (1) bug.
- [X] If I make a mistake while submitting this report, I agree to use the "Edit" feature to correct it, instead of closing this issue and opening a new one.

VaLLeX8:
- Removed an optimization condition in collision_static 
function (collision_system.cpp) that skipped collision 
checks for smaller static objects.
- Previously, when a big rock was processed, it would skip
collision checks for a smaller rock beneath it, causing the
smaller rock to fall through the floor.
- Adjusting the condition to skip larger rocks instead would 
lead to the opposite issue-smaller rocks pushing larger ones
through the floor.

Closes #3105

### DIFF
--- a/src/collision/collision_system.cpp
+++ b/src/collision/collision_system.cpp
@@ -33,7 +33,6 @@
 namespace {
 
 const float MAX_SPEED = 16.0f;
-static const float FORGIVENESS = 256.f; // 16.f * 16.f - half a tile by half a tile.
 
 } // namespace
 
@@ -422,11 +421,7 @@ CollisionSystem::collision_static(collision::Constraints* constraints,
   {
     const float static_size = static_object->get_bbox().get_width() * static_object->get_bbox().get_height();
     const float object_size = object.get_bbox().get_width() * object.get_bbox().get_height();
-    // let's skip this if two colgroup_moving_static's connect and our object is somewhat larger than the static object.
-    if ((object.get_group() == COLGROUP_MOVING_STATIC && static_object->get_group() == COLGROUP_MOVING_STATIC) &&
-      (object_size > static_size + FORGIVENESS)) {
-      continue;
-    }
+
     if ((
           static_object->get_group() == COLGROUP_STATIC ||
           static_object->get_group() == COLGROUP_MOVING_STATIC


Hume2:
You removed a part of code which was introduced in https://github.com/SuperTux/supertux/commit/24087deee6c40d040c5a7cc1c5b07839bbce4ef6 to fix #2854 . Can you ensure that your changes won't re-introduce the bug?

huntekye:
Should probably compare against #3231.

VaLLeX8:
@Hypernoot, you are right - this solution allows crushers to be pushed to the ground by small rocks
as shown here:
https://github.com/user-attachments/assets/231c347f-ee8f-491a-89e7-2c651831008f

In contrast, skipping the check_collisions step allows rocks to pass through the ground:
https://github.com/user-attachments/assets/5510fe71-836d-401d-8ecd-d56c4a7a43cd

Do you think this PR should resolve this bug and reopen #2038, or would you prefer that I address the other bug in this PR as well?

VaLLeX8:
@huntekye , I see that this bug is already being fixed in that PR in the same way I’m addressing it here.

Hypernoot:
The bug was fixed by #3231 , so this PR is no longer relevant.

PR #3236, opened 2025-04-05
Merge Rect and Rectf classes.

tylerandari13:
In honor of Supertux week, I have decided to contribute something of a decent size. This Pull Request merges a lot of int and float duplicates in Supertux into single classes with templates and `using`. These classes include:
- `Rect` and `Rectf`
- `Size` and `Sizef`
- `Vector` and `Vectori` (technically it was added, but still)

The float variants of all of the classes were prioritized because they are more up-to-date than the int versions. This means `Rect` now has some asserts that weren't there before, and it has caught a couple of oversights in a few places... aaand thats why its still a draft PR. I need to finish fixing a bug in the tilemap code. I am releasing this as a draft so that things like codestyle can start being checked.


tylerandari13:
I forgot to mention in the description that I also replaced the `left`, `down`, `up`, and `right` properties with a function instead lol

Hypernoot:
Size_t is not a good name, as it can be confused with std::size_t.

PR #3237, opened 2025-04-05
Implement up and down direction for swimming fish

### LINKED ISSUE #2573: Allow fish badguys to move up/down
Currently Harmless Fish, Swimming Fish and Chasing Fish can only swim from left to right.
Instead, when selecting direction "up" or "down" they should be able to swim from top to bottom.

Hypernoot:
fixes #2573 

I hate the implementation but some compromise has to be done in order to fit it into 0.7.0. To make the code nicer, I need all the following:
* A transpose-action sprite modifier, to make the fish visually facing up or down.
* Make the chasing fish use 4 different rotations instead of just 2.
* A better interoperability between vectors and directions. There should be at least functions to obtain a unit vector from direction and a direction from vector.

As this seems as a bigger task, the question is, whether it makes sense to put it into 0.7.0 or not. There are three options:
1) Merge this PR as it is and then report these tasks as a bug to be solved in 0.7.1.
2) Postpone whole this stuff to 0.7.1
3) Add these tasks to the scope to 0.7.0 and then provide a proper solution.

### DIFF
--- a/src/badguy/fish_chasing.cpp
+++ b/src/badguy/fish_chasing.cpp
@@ -89,7 +89,10 @@ FishChasing::active_update(float dt_sec) {
   const Vector p2 = player->get_bbox().get_middle();
   const Vector dist = (p2 - p1);
   const bool is_player_in_water = player->is_swimming() || player->is_swimboosting() || player->is_water_jumping();
-  const bool is_facing_player = (m_dir == Direction::LEFT && dist.x <= 0.0f) || (m_dir == Direction::RIGHT && dist.x >= 0.0f );
+  const bool is_facing_player = (m_dir == Direction::LEFT && dist.x <= 0.0f)
+                             || (m_dir == Direction::RIGHT && dist.x >= 0.0f )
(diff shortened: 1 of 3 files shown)


Rusty-Box:
I don’t think Swimming fush moving up and down needs to actually look up and down. I‘d even argue that‘d look worse. They simply just need to move up and down on the spot (range can be set in editor). A bit like Dive Mine but obviously a bit further

Hypernoot:
Okay. In that case, the first two points are not needed. The first one may be still handy for other purposes.

There's just one issue with the chasing fish. When the fish stops chasing Tux, shall it continue moving horizontally, vertically, or how? I currently implemented that so it moves horizontally no matter what. But maybe it should move based on how it was moving originally. Or maybe according to the direction in which Tux escaped?

Frostwithasideofsalt (approved):
Visually looks a bit odd, but it works

tobbi:
Can I have a video on how it looks? Don't feel like compiling / downloading a nightly build.

PR #3249, opened 2025-04-17
Fix fullscreen using wrong resolution

MatusGuy:
This happens because when moving SuperTux to the secondary monitor, the Viewport will take the size of the primary monitor.


Vankata453 (approved):
Tested, now it properly adjusts to my second monitor's 4:3 ratio.

Issue #3128, opened 2024-12-04
Add a saturation slider underneath the color picker


Eauix:
Ever since the color sliders were replaced by a 2D color picker, we can no longer adjust the saturation of the colors we choose. If we click anywhere on the picker, we'll get a color with its maximum saturation, and if we want to pick a gray color or a less vibrant tone, we need to directly change the RGB numbers or paste a color copied from an external source.

2 of the 3 sliders' functions were combined into the 2D picker, which is good, but the 3rd slider's function was lost, so what i'm proposing is bringing back this 3rd slider to work alongside the 2D picker

Here's the setup we had before compared to the one we could have after this

![image](https://github.com/user-attachments/assets/491c919f-12c7-4747-9ae0-7904e68e8ed1)

HybridDog:
When I have replaced the Lightness, Hue and Saturation sliders by the 2D color picker, I have put the addition of the 2D color picker and the removal of the sliders into two separate commits in #2895.
I think it should be possible to add back the sliders by reverting that commit: https://github.com/SuperTux/supertux/pull/2895/commits/d7e670366c2dbeec5800ec514adeb6f5a2423704

HybridDog:
>  I think it should be possible to add back the sliders by reverting that commit: d7e6703

Here's a Pull Request doing this: #3764

PR #3251, opened 2025-04-24
Added color saturation slider

nathaniel-jb:
As part of a school project, my friend and I have worked on a solution for a color saturation slider that was proposed and discussed as a listed issue. Even if it is unacceptable, we would greatly appreciate any comments. Thank you.


Eauix:
I really appreciate the work on this PR as it would fix #3128 (which, believe it or not, has bothered me a lot when creating backgrounds)

MatusGuy:
It also seems like these files have been changed to CRLF. Here we use LF line endings, so please configure your editor properly.

nathaniel-jb:
Addressed feedback: removed extra files and (I think) fixed line endings to LF

.clang-tidy
tobbi, line 1: Please restore this file.

src/gui/item_colorchannel_rgba.hpp
```diff
@@ -67,4 +67,4 @@ class ItemColorChannelRGBA final : public MenuItem
 
 #endif
 
-/* EOF */
+/* EOF */
```
tobbi, line 70: Add a new line at the end.

Frostwithasideofsalt:
are there plans on updating this based on tobbi's review?

PR #3252, opened 2025-04-25
Fix WalkingBadguy getting stuck on curved tiles (#3239)

### LINKED ISSUE #3239: [Bug]: Enemies get stuck in some tiles
### SuperTux Version

SuperTux-v0.6.3-1622-g446e9f6aa

### System Information

Windows 10 64-bit

### Expected Behavior

The enemy gets trapped in specific tiles, such as these: 

Image

The enemy is not supposed to remain stuck in motion.

### Actual Behavior

The enemy is not supposed to remain stuck in motion.

By using  certain curved snow or cave tiles enemies get stuck-in-motion there.

### Additional Information

_No response_

mederand:
This pull request fixes issue #3239 where WalkingBadguy enemies would get stuck indefinitely on certain curved snow or cave tiles.

Changes:
- When a collision against a shallow slope is detected (0 < slope_normal.x < 0.7), the enemy is manually nudged upward slightly.
- Otherwise, if the slope is steep or it's a wall, the enemy turns around as usual.

Testing:
- Reproduced the original stuck behavior in a sandbox level.
- After applying the fix, enemies now smoothly turn around properly on curved tiles without getting stuck.
- Regression tested normal wall collisions and enemy behavior across different tile types via playing a couple levels.

Closes #3239


src/badguy/walking_badguy.cpp
```diff
@@ -200,12 +200,18 @@ WalkingBadguy::collision_solid(const CollisionHit& hit)
...
-  if ( hit.slope_normal.x == 0.0f &&
-      ((hit.left && m_dir == Direction::LEFT) ||
-      (hit.right && m_dir == Direction::RIGHT)) ) {
+  if ((hit.left && m_dir == Direction::LEFT) ||
+      (hit.right && m_dir == Direction::RIGHT)) {
+    
+    // Enemies get stuck in some tiles #3239 FIX
+    // When a bad guy is about to step onto a slope, have it manually step onto it
+    if (std::abs(hit.slope_normal.x) > 0.0f && std::abs(hit.slope_normal.x) < 0.7f) {
+      m_col.m_bbox.move(Vector(0, -5));
```
MatusGuy, line 209: Instead of moving the box directly, set the movement of the CollisionObject.

MatusGuy (changes requested):

src/badguy/walking_badguy.cpp
mederand: Thanks for the review. I updated the fix using get_movement() and set_movement(). Let me know if there's anything I need to adjust.

src/badguy/walking_badguy.cpp
```diff
@@ -200,12 +200,21 @@ WalkingBadguy::collision_solid(const CollisionHit& hit)
...
 
-  if ( hit.slope_normal.x == 0.0f &&
-      ((hit.left && m_dir == Direction::LEFT) ||
-      (hit.right && m_dir == Direction::RIGHT)) ) {
+  if ((hit.left && m_dir == Direction::LEFT) ||
+      (hit.right && m_dir == Direction::RIGHT)) {
+    
+    // Enemies get stuck in some tiles #3239 FIX
+    // When a bad guy is about to step onto a slope, have it manually step onto it
+    if (std::abs(hit.slope_normal.x) > 0.0f && std::abs(hit.slope_normal.x) < 0.7f) {
```
tobbi, line 208: Can you maybe move the `std::abs(hit.slope_normal.x)` into a variable? That way, this if statement would fit into a line

PR #3253, opened 2025-04-29
Add `StartupWMClass` to desktop file

fiftydinar:
This should resolve the edge-cases where icon is not shown, like in Gnome's dash for example.


tobbi:
Seems trivial enough to merge

PR #3254, opened 2025-05-01
WIP: Separating collision system into multiple groups

tobbi:
Here's something that I've been thinking about: Instead of iterating over all objects, we could group the possible objects together, depending on what type of collision group they have and then only iterate over those matching the collision group we care about. 

I don't honestly know how much faster this is. All I know is: It crashes. It crashes all the time. And I have no idea why.

So, if anyone wants to check this out and submit fixes to this PR, feel free.

Also, terminology-wise: `collision part` and `collision bucket` are the same thing.

### DIFF
--- a/src/badguy/crusher.cpp
+++ b/src/badguy/crusher.cpp
@@ -705,7 +705,7 @@ CrusherRoot::CrusherRoot(Vector position, Crusher::Direction direction, float de
   }
   else
   {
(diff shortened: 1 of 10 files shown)


tobbi:
Sorry, this was just a experiment, and I feel like I can't fix all the bugs with it in the meantime. I don't even know if it shaves off as much time as I think. So I'm closing this, but I'll likely keep the branch around.

PR #3262, opened 2025-05-23
Fixes #517 - Add a button with the ability to hide/single layers

### LINKED ISSUE #517: Ability to hide / single layers
The C# editor has this and I think it's quite a useful feature:

Hiding layers or displaying one layer only helps people designing levels quite a bit, IMO.

Not sure if we want this for 0.5.0 or later, though.

This was the eye symbol in the old editor.

sofi004:
The Supertux level editor did not have a button to
hide individual layers.
Now this button is at the end of each layer menu.
The button does not appear on the selected layer,
as it does not make sense to hide the selected layer.
When someone decides to hide a layer other than
the selected one and then switches to it, that layer is shown
again. In addition, after hiding a layer, if the hidden layer
menu is opened again, instead of the hide layer button,
a show layer button appears.
The default opacity of 0.5 was kept as the default,
so that users do not find it strange or have to press
this implemented button.

Closes #517

Co-authored-by: Filipe Oleacu <filipeoleacu04@tecnico.ulisboa.pt>

### DIFF
--- a/src/editor/object_menu.cpp
+++ b/src/editor/object_menu.cpp
@@ -54,6 +54,20 @@ ObjectMenu::refresh()
     }
   }
 
+  TileMap* selected_tilemap = m_editor.get_layers_widget()->get_selected_tilemap();
+  if (m_object != selected_tilemap) 
+  {
+    float current_alpha = static_cast<TileMap*>(m_object)->get_alpha();
+    if (current_alpha > 0.0f) 
+    {
+      add_entry(MNID_HIDE_LAYER, _("Hide Layer"));
+    } 
+    else 
+    {
+      add_entry(MNID_HIDE_LAYER, _("Show Layer"));
+    }
+  }
+
   if (!m_object->is_up_to_date())
   {
     add_hl();
@@ -85,6 +99,25 @@ ObjectMenu::menu_action(MenuItem& item)
         m_object->update_version();
         refresh();
       });
+      break;
+    
+    case MNID_HIDE_LAYER:
+      if (m_object) 
+      {
+        TileMap* tilemap = static_cast<TileMap*>(m_object);
+        float current_alpha = tilemap->get_alpha();
+        if (current_alpha > 0.0f) 
+        {
+          tilemap->set_alpha(0.0f);
+        } 
+        else 
+        {
+          tilemap->set_alpha(1.0f);
+        }
+        m_editor.m_reactivate_request = true;
+        refresh();
+      }
+      MenuManager::instance().pop_menu();
       break;
 
     case MNID_PATCH_NOTES:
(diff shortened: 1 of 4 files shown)


tobbi:
Looking good, code-wise. Haven't tested it.

src/editor/object_menu.cpp
```diff
@@ -86,6 +100,25 @@ ObjectMenu::menu_action(MenuItem& item)
         refresh();
       });
       break;
+    
+    case MNID_HIDE_LAYER:
+      if (m_object) 
+      {
+        TileMap* tilemap = static_cast<TileMap*>(m_object);
+        float current_alpha = tilemap->get_alpha();
```
bruhmoent, line 108: Why not do the same thing here? Also consider making current_alpha `const` as well.
`float current_alpha = static_cast<TileMap*>(m_object)->get_alpha();`

bruhmoent (changes requested):
It works, but for some reason the background shifts in the Y axis (editor only)

tobbi:
> It works, but for some reason the background shifts in the Y axis (editor only)

Can you clarify this or send a screenshot?

Hypernoot:
The sector size is calculated from the sizes of solid tilemaps. Is it possible that when you hide the only solid tilemap, then something gets goofed there?

sofi004:
https://github.com/user-attachments/assets/f9fa6d9e-5f7e-442e-9abe-b7528124bbba

I think the problem has nothing to do with what I did, but that it was already an old problem considering that this happens without my code, as you can see in the video above.

src/editor/object_menu.cpp
sofi004: In the meantime I will fix this

sofi004: Done

bruhmoent (approved):

Frostwithasideofsalt (changes requested):
You aren't able to toggle the visibility of a layer your currently selecting, which is quite confusing if you don't already know how it works. Also, i think it would be nice if hidden tilemaps had a icon on them showing that their hidden, so you can know what layers are hidden at a glance. here's a mock up of what it could look like. 
image
(also here is the raw icon) 
hidden_layer
Also, this Pr has a _lot_ of bugs.

Frostwithasideofsalt:
image
image
image
firstly, every single object has a "hide layer" option, which doesnt do anything.

Frostwithasideofsalt:
https://github.com/user-attachments/assets/c420217c-10b8-4aad-8901-20c95b26f1b1

third. this pr has a issue regarding backgrounds moving based on a tilemap is solid or not. This issue is unique to this pr as it doesn't happen on latest master

PR #3265, opened 2025-05-29
Add support for paths for the `WillOWisp` in the editor

### LINKED ISSUE #1020: Add support for Willowisp paths to editor
Willowisps can be made to follow paths using scripting. See lines 138-146, 286 in willowisp.cpp. It would be useful if we could set a path for them in the editor.

bruhmoent:
Fixes #1020

Now WillOWisps support path nodes in the editor, meaning they can be manipulated just as any `PathObject`. They can be configured to start moving along their assigned path and will follow the path until they detect the player or are otherwise interrupted.

### DIFF
--- a/src/badguy/willowisp.cpp
+++ b/src/badguy/willowisp.cpp
@@ -53,7 +53,8 @@ WillOWisp::WillOWisp(const ReaderMapping& reader) :
   {
     reader.get("sector", m_target_sector);
     reader.get("spawnpoint", m_target_spawnpoint);
-  } else
+  }
+  else
   {
     reader.get("sector", m_target_sector, DEFAULT_SECTOR_NAME.c_str());
     reader.get("spawnpoint", m_target_spawnpoint, DEFAULT_SPAWNPOINT_NAME.c_str());
--- a/src/badguy/willowisp.cpp
+++ b/src/badguy/willowisp.cpp
@@ -92,28 +92,43 @@ WillOWisp::WillOWisp(const ReaderMapping& reader) :
 }
 
 void
+WillOWisp::synchronize_position_from_path()
+{
+  if (!has_valid_path())
+    return;
+
+  const auto& nodes = get_path()->get_nodes();
+  if (nodes.empty())
+    return;
+
+  const int size = static_cast<int>(nodes.size());
+
+  if (m_starting_node < 0)
+  {
+    m_starting_node = 0;
+  }
+  if (m_starting_node >= size)
+  {
+    m_starting_node = size - 1;
+  }
+
+  set_pos(m_path_handle.get_pos(m_col.m_bbox.get_size(), nodes[m_starting_node].position));
+}
+
+void
 WillOWisp::finish_construction()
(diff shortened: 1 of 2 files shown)


src/badguy/willowisp.cpp
```diff
@@ -122,55 +189,66 @@ WillOWisp::active_update(float dt_sec)
...
       }
       break;
-
     case STATE_TRACKING:
-      if (glm::length(dist) > m_vanish_range) {
+      if (glm::length(dist) > m_vanish_range)
+      {
         vanish();
-      } else if (glm::length(dist) >= 1) {
+      } else if (glm::length(dist) >= 1)
```
tobbi, line 206: Sorry, also check the brace placement in this file and all the other ones. Thanks! :)

src/badguy/willowisp.cpp
```diff
@@ -122,55 +189,68 @@ WillOWisp::active_update(float dt_sec)
...
+      {
         remove_me();
       }
       break;
     }
 
     case STATE_PATHMOVING:
     case STATE_PATHMOVING_TRACK:
-      if (get_walker() == nullptr)
+      if (get_walker() == nullptr || !get_path() || !get_path()->is_valid())
```
tobbi: I think you can use the shortcut method `PathObject::has_valid_path()` here.

src/badguy/willowisp.cpp
```diff
@@ -91,9 +94,55 @@ WillOWisp::WillOWisp(const ReaderMapping& reader) :
...
+  if (has_valid_path())
+  {
+    const auto& nodes = get_path()->get_nodes();
+    if (!nodes.empty())
+    {
+      if (m_starting_node < 0)
+      {
+        m_starting_node = 0;
+      }
+      if (m_starting_node >= static_cast<int>(nodes.size()))
```
tobbi: It's not mandatory, but I feel like moving 

`static_cast<int>(nodes.size())` into a variable and using that could be beneficial here.

tobbi: And elsewhere in the code obviously.

tobbi (approved):
Looking good to me!

src/badguy/willowisp.cpp
```diff
@@ -105,12 +156,31 @@ WillOWisp::after_editor_set()
...
+    {
+      if (m_starting_node < 0)
+      {
+        m_starting_node = 0;
+      }
+      if (m_starting_node >= size)
+      {
+        m_starting_node = size - 1;
+      }
+      set_pos(m_path_handle.get_pos(m_col.m_bbox.get_size(), nodes[m_starting_node].position));
```
tobbi: Isn't this code semantically identical to https://github.com/SuperTux/supertux/pull/3265/files#diff-3f748aafc72f1aa095eed38ee035f379c712ab01c3f0c7352ca355b16f12d279R129-R147 ?

Frostwithasideofsalt (approved):
seems to work so far

PR #3268, opened 2025-05-30
Another attempt to fix wind

### LINKED ISSUE #2592: Improving the `Wind` object

Hypernoot:
fixes #2592

There will be 3 types of wind:
* legacy - original behaviour
* wind - assumes that the object is in air
* current - assumes that the object is in water

Currently, I'm trying to implement the "wind" wind ^.^

### DIFF
--- a/src/object/wind.cpp
+++ b/src/object/wind.cpp
@@ -28,6 +28,7 @@
 #include "object/sprite_particle.hpp"
 #include "sprite/sprite.hpp"
 #include "sprite/sprite_manager.hpp"
+#include "supertux/physic.hpp"
 #include "supertux/sector.hpp"
 #include "util/reader_mapping.hpp"
 #include "video/drawing_context.hpp"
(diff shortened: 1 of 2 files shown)


tobbi:
```
/supertux/src/object/wind.cpp:139:23: error: implicit conversion loses floating-point precision: 'double' to 'const float' [-Werror,-Wimplicit-float-conversion]
  const float coef = 1/28.8;
```

Hypernoot:
closing as stale

PR #3270, opened 2025-05-31
Wind: Add `current` subtype and fix particle orientation

### LINKED ISSUE #2592: Improving the `Wind` object

Hypernoot:
Partially resolves #2592 but does only the graphical part. The physics is unchanged.

### DIFF
--- a/src/object/sprite_particle.hpp
+++ b/src/object/sprite_particle.hpp
@@ -32,11 +32,11 @@ public:
   SpriteParticle(SpritePtr sprite, const std::string& action,
                  const Vector& position, AnchorPoint anchor,
                  const Vector& velocity, const Vector& acceleration,
-                 int drawing_layer = LAYER_OBJECTS-1, bool notimeout = false, Color color = Color::WHITE, float angle = 0);
+                 int drawing_layer = LAYER_OBJECTS - 1, bool notimeout = false, Color color = Color::WHITE, float angle = 0);
   SpriteParticle(const std::string& sprite_name, const std::string& action,
                  const Vector& position, AnchorPoint anchor,
                  const Vector& velocity, const Vector& acceleration,
-                 int drawing_layer = LAYER_OBJECTS-1, bool notimeout = false, Color color = Color::WHITE, float angle = 0);
+                 int drawing_layer = LAYER_OBJECTS - 1, bool notimeout = false, Color color = Color::WHITE, float angle = 0);
   virtual GameObjectClasses get_class_types() const override { return GameObject::get_class_types().add(typeid(SpriteParticle)); }
   ~SpriteParticle() override;
 
--- a/src/object/sprite_particle.hpp
+++ b/src/object/sprite_particle.hpp
@@ -32,11 +32,11 @@ public:
   SpriteParticle(SpritePtr sprite, const std::string& action,
                  const Vector& position, AnchorPoint anchor,
                  const Vector& velocity, const Vector& acceleration,
-                 int drawing_layer = LAYER_OBJECTS-1, bool notimeout = false, Color color = Color::WHITE);
+                 int drawing_layer = LAYER_OBJECTS-1, bool notimeout = false, Color color = Color::WHITE, float angle = 0);
   SpriteParticle(const std::string& sprite_name, const std::string& action,
                  const Vector& position, AnchorPoint anchor,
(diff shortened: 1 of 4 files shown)


src/object/sprite_particle.hpp
```diff
@@ -32,11 +32,11 @@ class SpriteParticle final : public GameObject
   SpriteParticle(SpritePtr sprite, const std::string& action,
                  const Vector& position, AnchorPoint anchor,
                  const Vector& velocity, const Vector& acceleration,
-                 int drawing_layer = LAYER_OBJECTS-1, bool notimeout = false, Color color = Color::WHITE);
+                 int drawing_layer = LAYER_OBJECTS-1, bool notimeout = false, Color color = Color::WHITE, float angle = 0);
```
bruhmoent: `LAYER_OBJECTS-1` -> `LAYER_OBJECTS - 1`

src/object/sprite_particle.hpp
```diff
@@ -32,11 +32,11 @@ class SpriteParticle final : public GameObject
...
   SpriteParticle(const std::string& sprite_name, const std::string& action,
                  const Vector& position, AnchorPoint anchor,
                  const Vector& velocity, const Vector& acceleration,
-                 int drawing_layer = LAYER_OBJECTS-1, bool notimeout = false, Color color = Color::WHITE);
+                 int drawing_layer = LAYER_OBJECTS-1, bool notimeout = false, Color color = Color::WHITE, float angle = 0);
```
bruhmoent: Same here

src/object/wind.cpp
```diff
@@ -96,6 +97,14 @@ Wind::get_settings()
   return result;
 }
 
+GameObjectTypes
+Wind::get_types() const {
```
bruhmoent: Ensure that `{` always start on a `\n`

src/object/wind.cpp
```diff
@@ -107,19 +116,26 @@ Wind::update(float dt_sec_)
...
   if (graphicsRandom.randf(0.f, 100.f) < (m_col.m_bbox.get_width() / 32.f) * (m_col.m_bbox.get_height() / 32.f))
   {
     // Emit a particle
-	  if (fancy_wind)
+    if (fancy_wind)
     {
-	    Sector::get().add<SpriteParticle>("images/particles/wind.sprite", (std::abs(speed.x) > std::abs(speed.y)) ? "default" : "flip", ppos, ANCHOR_MIDDLE, pspeed, Vector(0, 0), m_layer); 
-	  }
-	  else
+      const float angle = std::atan2(speed.y, speed.x) * float(180/M_PI);
```
bruhmoent: `const float angle = std::atan2(speed.y, speed.x) * static_cast<float>(180.0 / M_PI);`

bruhmoent (changes requested):

Frostwithasideofsalt:
remind me to test this again later

src/object/wind.cpp
Hypernoot:
```
src/object/wind.cpp:126:76: style: Floating-point cast causes loss of precision. [suspiciousFloatingPointCast]
      const float angle = std::atan2(speed.y, speed.x) * static_cast<float>(180.0 / M_PI);
```

tobbi:
> remind me to test this again later

You wanted to be reminded, @Frostwithasideofsalt

bruhmoent (approved):

PR #3271, opened 2025-05-31
Fix bugs around rocks and crushers

Hypernoot:
Fix #3234

yes oh yes


Hypernoot:
unnecessary variables...

Hypernoot:
The bug was fixed by #3231 , so this PR is no longer relevant.

PR #3277, opened 2025-06-04
Fix `CorruptedCrusher`'s behaviour

### LINKED ISSUE #3274: Weird behaviour with corrupted crusher roots
https://github.com/user-attachments/assets/4dcf8bd3-7d7f-43e9-8348-9c6cf153990a

crusher roots are spawning where they aren't supposed to and aren't snapping to the ground

bruhmoent:
This pr:
  - Fixes the incorrect offset of root bases,
  - Prevents roots from spawning on air or when there's no space,
  - Prevents roots from spawning when hitting the player,
  - Improves the visuals of the spawning roots,
  - Implements DRY strategy to the `spawn_roots` method and improves code style.

Fixes #3274


Frostwithasideofsalt:
Im thinking... can we just go back to using the old roots? the old ones felt a lot more polished and more fun to play around. maybe the new ones can stay for the big crusher? but idk.

Rusty-Box:
The problem with the old roots is that the hitbox is active even when the first frames are playing where the roots haven't reached their final position since it an animation which makes more unfair hits, where as the newer one has a moving hitbox since it is one single sprite moving up and down. Personally I much prefer the new one as it is more consistent with any other type of root attack we got.

Frostwithasideofsalt (approved):

bruhmoent:
Note: The accidental file I pushed (CMakeSettings.json) will be removed once the pr is deemed ready to merge.

PR #3279, opened 2025-06-08
Reset Tux' velocity after using a door

### LINKED ISSUE #3272: [Bug]: Bouncing between bumpers when entering door causes bounce effect to apply to tux after exiting door into another sector
### SuperTux Version

0.6.3-1648-g229bf8a7c

### System Information

W11

### Expected Behavior

for tux to not be affected by bumpers in the second sector after exiting the door into the first sector.

### Actual Behavior

when tux was bouncing between bumpers and entering the door, tux bounced after exiting that sector, which should not have happened since tux shouldn't be affected by bumpers in the second sector when tux is in the first sector.

1. connect two doors to each other via sectors, and place bumpers next to the spawnpoint of either door.
2. test the level and use the door with the bumpers next to it.
3. tux should bounce after exiting the door with the bumpers into the other sector.

### Additional Information

https://github.com/user-attachments/assets/5ea14467-58ad-4721-9292-12aff709db94

Brockengespenst:
This avoids taking any movement from last sector to the new one.

Fixes #3272

### DIFF
--- a/src/supertux/game_session.cpp
+++ b/src/supertux/game_session.cpp
@@ -603,6 +603,9 @@ GameSession::update(float dt_sec, const Controller& controller)
 
       if (m_spawn_with_invincibility)
       {
+        // Reset velocity to avoid taking any movement from last sector to new one
+        p->set_velocity(0.0f, 0.0f);
+
         // Make all players temporarily safe after spawning
         p->make_temporarily_safe(SAFE_TIME);
       }


Brockengespenst:
There are different ways to solve the issue. Another one might be to expose the "is temporarily safe" state from Tux (which is only used internally in Player class right now for the door transition) and avoid adding the velocity from Explosion::explode to Tux. The result would be, that the explosion will not affect Tux while entering the door in the old sector. I am not sure if Explosion is the only place that might affect Tux in such a way, so the solution in the GameSession is more general. Hopefully not too general, so that it may prevent some scriptings?

tobbi (approved):
Looking good!

Frostwithasideofsalt (commented):
works for me.

Frostwithasideofsalt (approved):
oops heres a proper review

PR #3284, opened 2025-06-13
Add editor shortcuts for Tiles and Objects navigation (SuperTux#2988)

VaLLeX8:
- Add "Q" to switch between Tiles and Objects;
- Add Shift + "Q" to toggle between current and last subgroup inside the same group (Tiles or Objects);
- Add Shift + Scroll to open the current menu and select on release;
- Add "S" to open a menu of sub-subgroups within each Tiles subgroup;
  - This uses the divisions previously commented in tiles.strf.

Co-authored-by: Afonso Mateus <afonso.mateus@tecnico.ulisboa.pt>

### DIFF
--- a/src/editor/overlay_widget.cpp
+++ b/src/editor/overlay_widget.cpp
@@ -1508,13 +1508,13 @@ EditorOverlayWidget::draw(DrawingContext& context)
   switch (m_editor.get_tileselect_input_type())
   {
     case EditorTilebox::InputType::TILE:
-      mode_text = "TILE MODE  (Click S to open tiles subgroups)";
+      mode_text = _("TILE MODE  (Click S to open tiles subgroups)");
       break;
     case EditorTilebox::InputType::OBJECT:
-      mode_text = "OBJECT MODE";
+      mode_text = _("OBJECT MODE");
       break;
     default:
-      mode_text = "NONE";
+      mode_text = _("NONE");
       break;
   }
   std::string zoom_text = (scale_percentage != 100)
--- a/src/supertux/menu/editor_tilesubgroup_menu.cpp
+++ b/src/supertux/menu/editor_tilesubgroup_menu.cpp
@@ -1,5 +1,6 @@
 //  SuperTux
-//  Copyright (C) 2015 Hume2 <teratux.mail@gmail.com>
+//  Copyright (C) 2025 Vasco Rodrigues <vasco.a.a.rodrigues@tecnico.ulisboa.pt>
+//                2025 Afonso Mateus   <afonso.mateus@tecnico.ulisboa.pt> 
 //
 //  This program is free software: you can redistribute it and/or modify
 //  it under the terms of the GNU General Public License as published by
--- a/src/supertux/menu/editor_tilesubgroup_menu.hpp
+++ b/src/supertux/menu/editor_tilesubgroup_menu.hpp
@@ -1,5 +1,6 @@
 //  SuperTux
-//  Copyright (C) 2015 Hume2 <teratux.mail@gmail.com>
+//  Copyright (C) 2025 Vasco Rodrigues <vasco.a.a.rodrigues@tecnico.ulisboa.pt>
+//                2025 Afonso Mateus   <afonso.mateus@tecnico.ulisboa.pt> 
 //
 //  This program is free software: you can redistribute it and/or modify
 //  it under the terms of the GNU General Public License as published by


swagtoy:
Did you attempt to merge all of master and stage it? You'll have to fix that, as it's difficult to review the code

VaLLeX8:
Yes, you are absolutely right. I think it's good now.

src/editor/tilebox.cpp
```diff
@@ -53,6 +60,22 @@ EditorTilebox::EditorTilebox(Editor& editor, const Rectf& rect) :
   m_scrollbar.reset(new ControlScrollbar(1.f, 1.f, m_scroll_progress, 35.f));
 }
 
+
+int
+EditorTilebox::get_parentgroup_size() const
+{
+  int parent_group_size = 0;
+  std::string  parent_group_name = m_active_tilegroup->parent_group;
+  for (auto& tilegroup :m_editor.get_tileset()->get_tilegroups())
```
tobbi: I think this can be `const auto&`

src/editor/tilebox.cpp
```diff
@@ -53,6 +60,22 @@ EditorTilebox::EditorTilebox(Editor& editor, const Rectf& rect) :
   m_scrollbar.reset(new ControlScrollbar(1.f, 1.f, m_scroll_progress, 35.f));
 }
 
+
+int
+EditorTilebox::get_parentgroup_size() const
+{
+  int parent_group_size = 0;
+  std::string  parent_group_name = m_active_tilegroup->parent_group;
```
tobbi: double space between data type and name.

src/editor/tilebox.cpp
```diff
@@ -370,16 +430,32 @@ EditorTilebox::on_select(const std::function<void(EditorTilebox&)>& callback)
...
 {
+  if (tilegroup_selected) {
+    previous_tilegroup_index = current_selected_tilegroup_index;
+  }
+  current_selected_tilegroup_index = id;
+  tilegroup_selected = true;
   m_active_tilegroup.reset(new Tilegroup(m_editor.get_tileset()->get_tilegroups()[id]));
   m_input_type = InputType::TILE;
   reset_scrollbar();
+  if (subgroup) {
```
tobbi: This can be shortened to 
```
subgroup_selected = subgroup;
```

src/supertux/menu/editor_tilegroup_menu.cpp
```diff
@@ -26,9 +26,13 @@ EditorTilegroupMenu::EditorTilegroupMenu()
   add_label(_("Tiles"));
   add_hl();
 
+  std::string previous_parent_group = "";
   int id = 0;
   for (auto& tg : Editor::current()->get_tilegroups()) {
-    add_entry(id, _(tg.name));
+    if (previous_parent_group == "" || previous_parent_group != tg.parent_group){
```
tobbi: Code style: Please move the opening brace into a new line.

src/editor/toolbox_widget.cpp
```diff
@@ -200,6 +208,101 @@ EditorToolboxWidget::on_mouse_button_down(const SDL_MouseButtonEvent& button)
...
+    return true;
+  }
+  else if (key.keysym.sym == SDLK_s)
+  {
+    if (m_tilebox->get_input_type() == InputType::TILE)
+    {
+      const auto& tilegroup = m_editor.get_tileset()->get_tilegroups()[current_tilegroup_index];
+      const std::string& parent_group = tilegroup.parent_group;
+
+      static const std::map<std::string, int> tilesubgroup_menu_ids = {
```
tobbi: I am not sure if hardcoding the name of the tile groups is a good idea. I'm open to suggestions on how to improve this.

VaLLeX8: Since there is a pattern in those IDs, I could try to create a simple function that, based on the parent_group name, generates its menu ID.

VaLLeX8: We could try converting the MenuId enum into an array, so that the menu can be accessed by its name.

bruhmoent:
I'm honestly not a fan of introducing more shortcuts to the editor **unless** a versatile solution is implemented which allows the user to bind each individual key to their wishes, and preferably most of them would be bound to `none` from the start. I find it confusing to have unknown bindings all over the keyboard which perform vastly different actions, and inconvenient.

MatusGuy:
In the meantime, I would still prefer to at least have some sort of shortcut, instead of postponing it to whenever you guys implement bind-able shortcuts.

But, until then, why is adding more shortcuts an issue? Just convert them all when the time comes...

Frostwithasideofsalt (changes requested):
Can you change the keybind for "S" to "G" instead? s overlaps with ctrl+s and WASD camera  controls if you have your movement keybinds set to those.

Frostwithasideofsalt:
apon further testing, this also seems to break all custom tilesets, alongside with breaking the "others" tile category. the broken tilemaps wont show their name, instead showing a empty space where the tilemap would be.
image
these broken tile-groups also contain no tiles in the editor. 
image
any existing tiles from these groups still work in the levels, they just dont show up in the editor.

VaLLeX8:
@Frostwithasideofsalt, when adding custom tilesets, you still add them to your "tiles.strf", right? The problem I see there is that the menu now searches for "parent_group" and not "name". My guess would be that those tilesets don't have a defined parent_group in that file, which causes them to appear blank in the menu and not show the tiles when you select them. If that's the case, you just need to add something like "(parent_group (_ "<custom tileset name>"))" above the "(name (_ "<custom tileset subgroup name>"))".
If that doesn't solve the problem, please let me know—and thank you for taking a look at this.

Frostwithasideofsalt:
There really needs to be a fallback if it can't find a "parent_group" to ensure that all the old custom tilesets dont break.

VaLLeX8:
Ok. I'll do that.

Frostwithasideofsalt:
alright, thank you!

src/editor/overlay_widget.cpp
```diff
@@ -1500,14 +1500,29 @@ EditorOverlayWidget::draw(DrawingContext& context)
...
   const int scale_percentage = static_cast<int>(roundf(scale * 100.f));
-  if (scale_percentage != 100)
-    context.color().draw_text(Resources::big_font, std::to_string(scale_percentage) + '%',
-                              Vector(context.get_width() - 140.f, 15.f),
-                              ALIGN_RIGHT, LAYER_OBJECTS + 1, Color::WHITE);
+  std::string mode_text;
+  switch (m_editor.get_tileselect_input_type())
+  {
+    case EditorTilebox::InputType::TILE:
+      mode_text = "TILE MODE  (Click S to open tiles subgroups)";
```
tobbi: Is this user visible? If yes, then please use the gettext translate function `_`:

```
"STRING HERE" -> _("STRING HERE")
```

src/supertux/menu/editor_tilesubgroup_menu.cpp
```diff
@@ -0,0 +1,59 @@
+//  SuperTux
+//  Copyright (C) 2015 Hume2 <teratux.mail@gmail.com>
```
tobbi: If you want, you could add your name / nickname + email address 
and the current year.

swagtoy:
Irrelevant since a similar approach was done in #3287. Thank you regardless!

PR #3285, opened 2025-06-17
P2P LAN online multiplayer using UDP

ist1106037:
We have implemented a P2P connection over LAN using UDP to enable online multiplayer functionality in SuperTux. This feature allows multiple players to connect directly to each other without requiring a central server.

To do this we added some variables to the Config, added a new Controller that serves as a dummy for the online handling of the character, added a new menu, a new class for ip management and another for sending and receiving messages.

This messages use the UDP protocol and threads to make a peer to peer connection for the handling of the LAN communication.

We only did LAN communication as other types of communication would require a server and this would require money or other resourses that we dont think the suppertux community wants, although there is a way to use this program for true online multiplayer, all it takes is putting the public ip of the host and both who is joining and who is hosting need to forward their routters to their computers.

To access online multiplayer you have to go to the multiplayer menu where the new option is added, then just host and another person writes your ip as join and you're ready to play together.


tobbi:
Please fix the initialization order:
```
/home/runner/work/supertux/supertux/src/supertux/gameconfig.cpp:119:3: error: field 'level' will be initialized after field 'multiplayer_buzz_controllers' [-Werror,-Wreorder-ctor]
```

tobbi:
One more thing:
```
2025-06-17T14:35:06.9490474Z D:\a\supertux\supertux\src\supertux\online_session_manager.hpp(26,10): error C1083: Cannot open include file: 'netinet/in.h': No such file or directory [D:\a\supertux\supertux\build\supertux2_lib.vcxproj]
```

It appears that that include file isn't available on Windows. Are there any dependencies that might need to get installed?

swagtoy:
You should consider binary serializing the data. String serializing is fine, however binary serializing is easier to work with in the future and more compact. Consider using something like ENet which Vankata used in the `editor-remote-level-networking` branch, which could prevent very rare desyncs (though probably not a big deal). In fact, better error handling overall using dialogs would be nice.

Also use the logger instead of std::cerr please

swagtoy:
I will unfortunately have to close this PR. It is recommended instead that you look at ENet, which is a more portable and effective approach for networking, as this PR might be prone to desyncs. It is also preferred that you binary serialize the data. That will probably take a rewrite, so once you have switched to a different API, please open a new PR or let us know and we'll reopen this PR.

Also, an approach that doesn't use threading is preferred. Use an event library like libevent or something else.

swagtoy:
Also, we've used ENet before, take a look at the `remote-level-networking` branch.

PR #3286, opened 2025-06-17
Revert crusher's edge detection for its horizontal axis

bruhmoent:
Previously, the crusher detected whether Tux would stand on it's edge (whilst standing on top of it). This PR brings this behaviour back, as it was utilized in gameplay.


Hypernoot (approved):
workz

PR #3288, opened 2025-06-23
[DON'T MERGE] Add -debug flag to hdiutil to figure out packaging failures on MacOS

tobbi:
This isn't a PR that needs to get merged. I just want to see whether this works. 

Currently, our MacOS builds fail with a resource busy error a lot of times. Github support is unsure why (see https://github.com/actions/runner-images/issues/7522). 

I (hopefully) modified the cmake command to add the -verbose flag so we can see what's going on.

See https://ss64.com/mac/hdiutil.html for HDIUTIL syntax and https://cmake.org/cmake/help/v4.0/cpack_gen/dmg.html#variable:CPACK_COMMAND_HDIUTIL for the CMake variable.


tobbi:
Closing this, since the failures don't seem to occur with the macos-14 runner.

PR #3294, opened 2025-07-03
TextScroller: Use finish_script properly

### LINKED ISSUE #3219: [Bug]: TextScroller finish_script is goofed up
### SuperTux Version

0.7-dev

### System Information

Ubuntu 20.04

### Expected Behavior

1) finish_script should be ran when the text runs out or Esc is pressed
2) I can use Effect.fade_out and then Level.spawn to send Tux into a different sector when the text runs out.

### Actual Behavior

1) finish_script is ran on EVERY FRAME
2) When I use Effect.fade_out, then Level.spawn doesn't work anymore.

1. Create a level with a TextScroller
2. Fill in finish_script
3. Observe how the finish_script is being ran on every frame.

### Additional Information

I used this workaround to detect when the text is actually gone.
init-script:
```
end <- false;
start_cutscene();
wait(86400);
end_cutscene();
end <- true;
```
And in the finish script:
```
if (end) {
//do something
}
```
and then hope that the player remembers to press Esc when the text runs out.

Hypernoot:
fixes #3219

If the finish_script is not defined, the level is finished when the text runs out or Esc is pressed. Otherwise, the finish_script is run instead.

### DIFF
--- a/src/object/textscroller.cpp
+++ b/src/object/textscroller.cpp
@@ -300,12 +300,7 @@ TextScroller::update(float dt_sec)
     // use start or escape keys to exit
     if (controller->pressed_any(Control::START, Control::ESCAPE) &&
         !m_fading) {
-      m_fading = true;
-      if (!m_finish_script.empty()) {
-        Sector::get().run_script(m_finish_script, "finishscript");
-      } else {
-        ScreenManager::current()->pop_screen(std::make_unique<FadeToBlack>(FadeToBlack::FADEOUT, 0.5f));
-      }
+      start_fading();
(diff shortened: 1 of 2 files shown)


tobbi:
In theory, I think that code can be moved to its own method, since it's used in multiple places.

Hypernoot:
Indeed, I will do that.

Hypernoot:
I noticed that the fading time is slightly different when the textscroller is exited by Esc instead of letting it run out. I removed this feature, as I think that it's rather accidental.

PR #3297, opened 2025-07-04
Fix layout of endseq statistics panel

### LINKED ISSUE #3081: Level-complete statistics text is offset
image
The new level-finish statistics text is weirdly offset, in my opinion.  The text should be closer together and pushed up by a bit, I think.

johannes-el:
Fixes #3081

### DIFF
--- a/src/supertux/statistics.cpp
+++ b/src/supertux/statistics.cpp
@@ -266,7 +266,7 @@ Statistics::draw_endseq_panel(DrawingContext& context, Statistics* best_stats, c
 
   int box_w = 220+110+110;
   int box_h = 30+20+20+20;
-  int box_x = static_cast<int>((static_cast<int>(context.get_width()) - box_w) / 2);
+  int box_x = static_cast<int>((static_cast<int>(context.get_width()) - box_w) / 2) - 10;
   int box_y = static_cast<int>(SCREEN_HEIGHT / 2) - box_h;
 
   int bd_w = static_cast<int>(backdrop->get_width());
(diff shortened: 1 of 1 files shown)


tobbi (approved):
Looking good!

bruhmoent (changes requested):
Could you please try calculating the position values instead of hardcoding them? Not the biggest fan of magic numbers, as having to alter them every single time the text changes is quite cumbersome.

johannes-el:
@bruhmoent
Like this?
  ```
constexpr int box_x_offset = -10;

  int box_w = 220+110+110;
  int box_h = 30+20+20+20;
  int box_x = static_cast<int>((static_cast<int>(context.get_width()) - box_w) / 2) - box_x_offset;
  int box_y = static_cast<int>(SCREEN_HEIGHT / 2) - box_h;

  int bd_w = static_cast<int>(backdrop->get_width());
  int bd_h = static_cast<int>(backdrop->get_height());
  int bd_x = static_cast<int>((static_cast<int>(context.get_width()) - bd_w) / 2);
  int bd_y = box_y + (box_h / 2) - (bd_h / 2);

  float col1_x = static_cast<float>(box_x);
  float col2_x = col1_x + 200.0f;
  float col3_x = col2_x + 130.0f;

  float y_offset = 35.f;
  if (m_preferences.enable_coins)
    y_offset -= 7.f;
  if (m_preferences.enable_badguys)
    y_offset -= 7.f;
  if (m_preferences.enable_secrets)
    y_offset -= 7.f;

```

I'm a bit confused. I'm aware that -10 is a bit magic, but what do you propose? How should I calculate this?
I can define a const for it, but I thought aligning it a bit more left looks better.

bruhmoent:
One way to do this is by defining a vector with the column widths - so when a change is made, the code will automatically update the appropriate values. You could then calculate a total width of all the columns, and with that the x column positions. After that, you could define clear and concise helper variables, responsible for things like the row's height, label's indent, etc; to allow for it to be easily changeable. Also, const correctness would be nice to include.

PR #3305, opened 2025-07-08
Rooted Forest Overhaul for v0.7 

Rusty-Box:
Update the entirety of Rooted Island to make use of all the new graphics, mechanics, music and overall theming for the next versions in hopes this can be considered the "final" major update on world 2 specifically.

There is still a few things left to finalize outside of levels for this to be considered "complete", such as a proper system for handling secret exits and cutscenes. Right now every paths is open so you can access all levels on the worldmap.

TODO:
- [ ] The transition script from the normal to corrupted forest worldmap isn't quite ready, missing some layers of decorations.
- [ ] Several new backgrounds are missing as they aren't fully finished yet.


Hypernoot:
resolved by #3431

PR #3341, opened 2025-08-17
Fix Tux' velocity on frozen enemy

der-hahn:
Addition to #3232 . Tux' velocity increased while being on a jumping frozen enemy. I fixed it by setting the velocity to zero.

### DIFF
--- a/src/object/player.cpp
+++ b/src/object/player.cpp
@@ -2415,6 +2415,7 @@ Player::collision(MovingObject& other, const CollisionHit& hit)
 
     if (hit.bottom && badguy->is_frozen())
       m_on_ground_flag = true;
+      m_physic.set_velocity_y(0);
   }
 
   return CONTINUE;


src/object/player.cpp
```diff
@@ -2415,6 +2415,7 @@ Player::collision(MovingObject& other, const CollisionHit& hit)
 
     if (hit.bottom && badguy->is_frozen())
       m_on_ground_flag = true;
+      m_physic.set_velocity_y(0);
```
tobbi, line 2435: Is the `set_velocity_y` call *supposed* to be part of the if statement in line 2416 or not?

If it does, it needs braces after the if. If not, the indentation needs to get fixed or, even better, the call needs to be put in a new line.

der-hahn:
> See the above comment

Oh thanks. Didn't notice that.

Hypernoot:
Actually, don't merge this yet. I forgot to consider one thing.

der-hahn:
> This PR seems to be incompatible with #3232 . Moreover, I see this as unnecessary once #3232 is merged. Can you please give an explanation?

This should be an addition to #3232 as it didn't fix the velocity of Tux. So, if Tux stands on a jumping frozen enemy and is walking to the edge, he falls down really fast as the velocity is going up.

Hypernoot:
> This should be an addition to #3232 as it didn't fix the velocity of Tux. So, if Tux stands on a jumping frozen enemy and is walking to the edge, he falls down really fast as the velocity is going up.

I just merged #3232 with master and I didn't experience this issue. Can you please re-test? By the way, you can also branch someone else's PR, add your own stuff and pack it into a PR to the main repository, as a whole, so it is obvious how it shall be integrated with the original PR.

EDIT: To make it clear, please, re-download #3232 and test if it still happens there.

der-hahn:
https://github.com/user-attachments/assets/52be2fff-fee3-43a8-940c-554a0225146b

The issue still is there. I used the recent version on Github.

Hypernoot:
Apparently, you downloaded the recent version from master and not that PR. But I merged it, so if there is still something wrong happening on current master, then please report it as an issue. And while you're at it, you can fix the conflict.

PR #3356, opened 2025-08-29
Add `late_sunset` to the OST

bruhmoent:
This song serves as a remake and counterpart to Voc's Daytime 2.


Frostwithasideofsalt (approved):

Alasdairbugs (changes requested):
While the 3rd bar in the B section feels like a bit of filler to force in the tux theme, its still quite a successful remix of the original song, making it feel less jumbled.

I would end with a repetition of the B section however, because this song is intended to be looped, and currently follows ABA form, where I imagine it would loop, resulting in two A sections next to eachother. It would also make the song appropriately long enough for use in story mode.

PR #3357, opened 2025-08-30
fix: downgrade ubuntu version in CI build

alberic89:
If the AppImage is built with a too recent version of the glibc, it will not run on an older system.
For more details, see https://docs.appimage.org/introduction/concepts.html#build-on-old-systems-run-on-newer-systems
Not tested at the moment, waiting for CI artifacts.


tobbi:
Why was this closed?

alberic89:
It looks like I made a bad git manipulation, and when I pushed it, it closed automatically this PR.
I will try to repair that.

alberic89:
Apparently, I made very bad things. So I re-opened a clean PR at #3364.

PR #3364, opened 2025-08-31
fix: downgrade ubuntu version for appimage build

alberic89:
If the AppImage is built with a too recent version of the glibc, it will not run on an older system.
For more details, see https://docs.appimage.org/introduction/concepts.html#build-on-old-systems-run-on-newer-systems

### DIFF
--- a/.ci_scripts/package.sh
+++ b/.ci_scripts/package.sh
@@ -21,7 +21,7 @@ if [ "$SOURCE" = "ON" ]; then
     cpack --config CPackSourceConfig.cmake -G TGZ;
 fi
 
-if ([ "$OS_NAME" = "ubuntu-latest" ]) && [ "$PACKAGE" = "ON" ]; then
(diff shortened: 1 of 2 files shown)


alberic89:
I am also trying to build on an older version (Ubuntu 20.04), but I get a compile error that I am not able to resolve with my zero-knowledge C++. It looks like the fmt library available in unbuntu 20.04 is too old : https://github.com/alberic89/supertux/tree/alberic89-appimage-fix1

tobbi:
I've restarted the workflows for this PR.

MatusGuy:
> I am also trying to build on an older version (Ubuntu 20.04), but I get a compile error that I am not able to resolve with my zero-knowledge C++. It looks like the fmt library available in unbuntu 20.04 is too old : https://github.com/alberic89/supertux/tree/alberic89-appimage-fix1

That's right, so upgrade your ubuntu version or build the library from source

PR #3369, opened 2025-09-05
Fix multiple game controllers interfering each other

Brockengespenst:
GameControllerManager shared one button/stick state between all connected game controllers. This led to one controller interfering with other controllers in multiplayer mode.

### DIFF
--- a/src/control/game_controller_manager.cpp
+++ b/src/control/game_controller_manager.cpp
@@ -29,9 +29,7 @@
 GameControllerManager::GameControllerManager(InputManager* parent) :
   m_parent(parent),
   m_deadzone(8000),
-  m_game_controllers(),
-  m_stick_state(),
-  m_button_state()
+  m_game_controllers()
 {
 }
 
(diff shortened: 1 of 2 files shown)


Brockengespenst:
Having two controllers connected and playing local multiplayer was really horrible as one player blocked the other very often. With these changes, this effect does not occur anymore. This removes the attempt to get axis and d-pad somehow kind of synchronized, but I do not really see making these changes it any worse, because it would be anyway hard to find an appropriate solution. If the button/stick state should be kept as it was, it should be added as a per controller state and not a global one.

@MatusGuy : This reverts also the changes made in 68f471788dddec1e3dbc5cc65acd3ee7fd915392. I was not sure about the original problem, maybe you can check if these changes either reintroduce the problem you had or maybe even fix it as well.

MatusGuy:
Apply this patch and see if I haven't removed any functionality:

[gamecontrollerfix.patch](https://github.com/user-attachments/files/22284583/gamecontrollerfix.patch)

Brockengespenst:
I tested your patch. At a first glance it looked quite normal, but then I ran into a state that the second controller's stick was not able to move Tux to the right. With the D-Pad it was working, but with the stick it was not possible to move to the right (only to the left). This happend two times during a 2 minute test. I have not checked into detail what went wrong in this case and what steps exactly are needed to reproduce it.

tobbi:
@MatusGuy Can this be merged?

MatusGuy:
Sure

src/control/game_controller_manager.cpp
```diff
@@ -61,8 +59,8 @@ GameControllerManager::process_button_event(const SDL_ControllerButtonEvent& ev)
```
MatusGuy, line 60: Test inline comment

PR #3370, opened 2025-09-05
Add a new "powerup": Rotten Egg

huntekye:
Also, add placeholder graphics, and make it available in bonus blocks.

I would love to get rid of the nested ternary operators if possible, let me know if you have any suggestions.

Also just let me know if there are any variable names / code quality things you want changed, or feel free to change them yourself.


Hypernoot:
So basically an enemy inside a bonus block.

huntekye:
No, that was considered, but the behavior here is intentionally distinct. If it were an enemy, it wouldn't disappear after hurting Tux, and it would be possible to kill it without being damaged, but this is a moving obstacle that can only be avoided. It might also have distinct physics, if the normal egg gets updated with new physics.

Oh and I forgot to mention, there is currently a small bug: if big Tux is standing still when the rotten egg hits him, then sometimes he will shrink and be displaced down one tile without his sprite being updated to small Tux. It doesn't happen when moving, and when standing still it only lasts until Tux begins moving or his idle animation starts, so I think there just needs to be an extra line in the `GrowUp`'s collision function to force the sprite to be updated, but idk how that would happen.

huntekye:
Okay, I think this is basically ready, except I still don't know why a rotten egg landing on a big Tux standing still pushes Tux into the ground, i.e. changes Tux's hitbox without updating the sprite.

korbanpapa:
Ok so, maybe i don't understand this completely, but does this basically act like the poison mushroom from super mario brothers? if so, what's the gameplay for this going to look like? is it going to give you coins when you get another powerup afterwards, or just hurt you? genuinely wanting to know.

huntekye:
To me, it seems to mainly be to compliment the theme of the corrupted forest levels: corrupted enemies, corrupted granitos, corrupted plants, why not a corrupted powerup too? Idk what Mario's mushrooms are like, but this can be another way to increase the difficulty in the corrupted forest levels.

Frostwithasideofsalt:
> Ok so, maybe i don't understand this completely, but does this basically act like the poison mushroom from super mario brothers? if so, what's the gameplay for this going to look like? is it going to give you coins when you get another powerup afterwards, or just hurt you? genuinely wanting to know.

it's just a poison mushroom, yes. it doesn't benefit the player in any way, all it does is hurt the player.

korbanpapa:
It's a cool idea, but if it doesn't have a function besides hurting the player, i don't really see the use of adding an extra item "just because". on the other hand, if it does have a use like giving coins once you get another powerup or something like that, in that case you could try pitching it to the team for their thoughts on the matter. sorry if i'm seeming rude on this thought. :|

Alasdairbugs:
I don't get the whole "just because" thing, you just acknowledged it can damage the player. That's a use 🤷 so it's not a "just because" addition? Correct me if I'm wrong

Also, as an experienced level designer this definitely has more advanced uses. The power up, like existing ones, could interact with objects like a dispenser or spring or bumper, etc.

korbanpapa:
what I'm saying is, a powerup should give some kind of boost to the player not just to hurt the player and that's it. it should help the player in some kind of way, whether that's coins, an egg, fireflower, etc... also, it's a powerUP not powerdown.

tobbi:
Just my 2 cents here: If I recall correctly (and correct me if I'm wrong), Secret Chronicles has something similar (a poisoned flower that looks like a real powerup that will kill / downgrade you when you eat it accidentally). So it's not like we're the only ones with that idea.

src/object/powerup.cpp
```diff
@@ -255,6 +268,9 @@ PowerUp::get_type_from_bonustype(int type)
     case BONUS_GROWUP:
       return EGG;
 
+    case MALUS_BADEGG:
```
bruhmoent: What does MALUS mean?

tylerandari13: I think its a combination of the words "bonus" and "malice." Do correct me if Im wrong tho.

huntekye: Yeah more or less, it's not a real word, but in principle "mal" would be the opposite of "bon", but I can see how it's a little confusing. The real point here is that I didn't think it made sense to call it a bonus, since it's not. Happy to change it if you'd like.

huntekye:
The problem is that this PR now has conflicts with the linked sprite stuff; I looked at rebasing it but it seems like it would take me a bit to figure out. Maybe someone who understands the linked sprites would have an easier time rebasing this?

src/object/powerup.cpp
tobbi: Well, it is a word. At least in German.

Or a genus of trees, according to Merriam-Webster:
https://www.merriam-webster.com/dictionary/Malus

bruhmoent: As annoying as it might sound, I think it'd be better to not abbreviate it. I can imagine many being confused at the name.

huntekye: So we call it something like `FAKE_BONUS_BADEGG` or just `BONUS_BADEGG`?

bruhmoent: FAKE_BONUS sounds best imo

huntekye: Okay, I think given the context of the BonusType enum in [src/supertux/player_status.hpp](https://github.com/SuperTux/supertux/pull/3370/files#diff-a6c04d959a6ffee7d23c37b4aa87056f4a6201754b468fcaf9540c4d13e37ec5), let's go for `BONUS_FAKE` :+1:

Hypernoot: You completely missed the point of the BONUS prefix. It isn't there to denote that it is either good or bad, it is there because it is a member of the `BonusType` enum. So all elements of that enumeration should have the this prefix, regardless of what they are used for.

If you do mind that the `BonusType` is not always beneficial, then rename it to `ItemType` and change the prefix to ITEM. And happy refactoring!

src/supertux/player_status.hpp
```diff
@@ -43,7 +43,8 @@ enum BonusType {
   BONUS_FIRE,     /*!< @description Fire bonus. */
   BONUS_ICE,      /*!< @description Ice bonus. */
   BONUS_AIR,      /*!< @description Air bonus. */
-  BONUS_EARTH     /*!< @description Earth bonus. */
+  BONUS_EARTH,     /*!< @description Earth bonus. */
+  MALUS_BADEGG     /*!< @description Rotten Egg fake bonus. */
```
Hypernoot: Wait, this is actually wrong. This enum is not a list of bonusblock contents. It is a list of Tux modes. BONUS_NONE = snall Tux, BONUS_GROWUP = big Tux, BONUS_FIRE = fire Tux, etc.. So the MALUS_BADEGG would correspond to some kind of rotten Tux. But that is not going to happen, as the rotten egg just hurts Tux once. If there is no way how Tux would carry the rotten egg status, then please remove it.

huntekye: Yeah I know it's a little cursed, but at the same time [src/supertux/player_status.cpp](https://github.com/SuperTux/supertux/pull/3370/files#diff-149a589eb87616880b552451d921b1dcb9dc0961963443d31047baba84e31399) has the lovely function `PlayerStatus::get_bonus_sprite(BonusType bonustype)` (for example). I really don't know why there's a case statement in player.cpp to get the file path to the powerup sprites, or why it's using an enum for Tux's state to enumerate the powerup types, but my goal was not to redo all of the powerup logic, but just to add a rotten egg to it. To help prevent confusion about it, I gave that entry a very distinct name so it would stand out, but I guess that was also confusing lol.

Hypernoot: There is no need to implement any logic in player_status.cpp, as Tux cannot really become rotten. By the same logic, you won't find there the star powerup, the coins or tux dolls. All of these are well-established powerups but they don't represent any player status.

korbanpapa:
> To me, it seems to mainly be to compliment the theme of the corrupted forest levels: corrupted enemies, corrupted granitos, corrupted plants, why not a corrupted powerup too? Idk what Mario's mushrooms are like, but this can be another way to increase the difficulty in the corrupted forest levels.

So, in a previous conversation about the function, we discussed what this powerup would do (popping out of the bonus block and rolling), but a quick idea i had to expand the feature was to make it slow tux down a bit, but give him an ability such as: A possible normal grow, Higher jumping, or others (i'd like to hear everyone's thoughts).

src/supertux/player_status.hpp
huntekye: Let's consider some more examples:
https://github.com/SuperTux/supertux/blob/dd40f0f52cf7c836fc8c951c03e223e04681200b/src/object/bonus_block.cpp#L669
https://github.com/SuperTux/supertux/blob/dd40f0f52cf7c836fc8c951c03e223e04681200b/src/object/flower.cpp#L25

When a powerup is hit out of a bonus block, `raise_growup_bonus` (or `drop_growup_bonus`) are called, with `BonusType` as the information available to decide what powerup should be created. Then (independent of this PR), `BonusType` is used again by `Flower` to decide what kind of flower should be made. I agree that it doesn't make sense to be using `BonusType` this way, but given the current state of things, it actually does seem that `BonusType` is being used to list some bonus block contents.

(Also, to me it doesn't seem unreasonable to imagine a future "corrupted Tux" state, although I know that that is not currently planned.)

swagtoy:
I think its fine. We try to keep things like this simple. Damaging the player is simple enough, and just importantly, it's a concept used by other games; we don't need to be unique for the sake of being unique

src/supertux/player_status.hpp
Hypernoot: Maybe we should have actually a separate `FlowerType`, `EggType` and `TuxType`, this is starting to be messy. Let's just leave it this way for now.

swagtoy:
I don't exactly like the code for this one.  I think it could be a separate object which hurts the player (or even other objects). There could also be a "follows" property where it follows the player or something

huntekye:
Well, maybe it's easier to start over than to rebase it anyway :shrug: but the goal was "just an egg that hurts Tux", so I think this is the simplest way to achieve that, but maybe there was never a large consensus on what people wanted the behavior to be like.

swagtoy:
Your behavior is fine :) But yes i recommend a redo

swagtoy:
It might be a little more work than youd like because you need to make a powerup just spawn the new object instead of just using the same powerup class and slapping more logic n it

PR #3378, opened 2025-09-15
Fix zeekling jittering after death

### LINKED ISSUE #3353: [Bug]: Jittering when dead zeekling lands on top of unisolid tree branch
### SuperTux Version

v0.6.3 - fd1c6a3d3

### System Information

Mac OS X 12.7.6

### Expected Behavior

No jittering when a dead zeekling lands on a tree branch

### Actual Behavior

When a dead zeekling lands on a unisolid tree branch, it starts to jitter before disappearing.

1. Add a unisolid tree branch and a zeekling to your level.
2. Kill the zeekling so that it lands on top of the tree branch
3. Notice the jittering

### Additional Information

https://github.com/user-attachments/assets/0c745de2-cdb3-482f-b881-26ed106c3820

Brockengespenst:
Fixes #3353

### DIFF
--- a/src/badguy/badguy.cpp
+++ b/src/badguy/badguy.cpp
@@ -720,7 +720,11 @@ BadGuy::kill_squished(GameObject& object)
 
   SoundManager::current()->play("sounds/squish.wav", get_pos());
   m_physic.enable_gravity(true);
-  m_physic.set_velocity(0, 0);
+
+  if (on_ground())
+    m_physic.set_velocity(0, 0);
+  else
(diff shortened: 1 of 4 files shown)


Brockengespenst:
Additionally to the jittering fix, I added a check if a badguy is in the air while being killed and do not set the x-velocity to zero in that case. This looks more natural as e.g. the Zeekling does not loose its x-velocity. Also for other badguys that jump from time to time, this looks more natural when they get squished or ignited.

tobbi:
Thank you! I'll take a look in a few hours

PR #3384, opened 2025-10-04
Migrate to SDL3

tobbi:
This is WIP and updates our backend code to SDL3.

TODO:
- [x] Fix compiler errors.
- [x] Continue migration
- [x] Replace local SDL_TTF clone with package.
- [ ] Implement proper bidirectional font algorithm (read when a language is RTL and set the appropriate tags to get arabic font shaping for example.
- [ ] Check where the sdl2 -> sdl3 replacement was too eager (where a replacement from sdl2 to sdl3 happened without being required - e.g. in po files)
- [ ] Check where else the version number needs to be bumped from 2 to 3.

... and more (I can't think of any right now, but there's other needs as well)


swagtoy:
This might be the scariest PR i will ever have to rebase. In fact i'm so scared i think i'd genuinely rather just restart it ... but i'll have a go... :cat::gun:

src/physfs/physfs_sdl.cpp
```diff
@@ -17,147 +17,53 @@
...
-
-int funcClose(struct SDL_RWops* context)
-{
-  PHYSFS_file* file = static_cast<PHYSFS_file*>(context->hidden.unknown.data1);
-
-  PHYSFS_close(file);
-  delete context;
-
-  return 0;
-}
```
MatusGuy, line 105: The new implementation does not seem to use physfs at all.

vcpkg.json
MatusGuy, line 1: Are you sure it's not possible to use SDL3 on WASM via compiler option? (Like -SUSE_SDL2 or something like that i forgot the name)

MatusGuy (changes requested):
Desktop fullscreen does not render with the monitor's resolution, nor is there an option to do so.

Image

src/physfs/physfs_sdl.cpp
swagtoy: Those are just deleted callback functions, it's still being used?

vcpkg.json
swagtoy: it's fixed now

PR #3398, opened 2025-11-01
Fix crusher's behaviour on flip

### LINKED ISSUE #3393: [Bug]: Flipped crushers do not have their flipped behaviour
### SuperTux Version

current @ master (commit ee76219)

### System Information

LMDE6 64bit

### Expected Behavior

During levelflip, vertical crushers should attack upwards.

### Actual Behavior

They still attack downwards, as if the level weren't flipped.

https://github.com/user-attachments/assets/0f7d5f82-8589-4f25-a8af-94bfb1d757aa

Place a crusher in a level, flip the level and see.

### Additional Information

_No response_

bruhmoent:
Fixes #3393

### DIFF
--- a/src/badguy/crusher.cpp
+++ b/src/badguy/crusher.cpp
@@ -91,6 +91,7 @@ Crusher::Crusher(const ReaderMapping& reader) :
   m_physic(),
   m_dir_vector(get_direction_vector()),
   m_target(nullptr),
+  m_flipped(),
   m_whites(),
   m_lefteye(),
   m_righteye(),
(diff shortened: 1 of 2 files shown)


tobbi:
I just tested this. Works fine. Only thing that doesn't appear to be flipping is the eyes.

Narre:
I know this is technically consistent with the non-flipped crusher, but I'm kinda thinking it would be better if it didn't kill Tux upon impact but rather only after squishing him against the ceiling/floor. Just an opinion though. Otherwise, this works great.

https://github.com/user-attachments/assets/51e9341f-64e8-46f6-8c44-39899f95afcc

Frostwithasideofsalt (approved):
works.

PR #3411, opened 2025-11-15
Improved Crusher's check for end of recovery

### LINKED ISSUE #3407: [Bug]: Ice in the Hole: horizontal crusher glitches out
### SuperTux Version

supertux2 dev v0.6.3 - b4424e944 (master)

### System Information

CachyOS, using supertux-git

### Expected Behavior

Horizontal ice crusher in question consistently goes all the way to the end.

### Actual Behavior

This happens most of the time:

https://github.com/user-attachments/assets/21d723d7-0414-4436-b191-c225e669d4b9

1. Play the level "Ice in the Hole" until crusher in question
2. Trigger it a few times

### Additional Information

_No response_

Brockengespenst:
Check only relevant directions when checking for end of recovery.

Fixes #3407

### DIFF
--- a/src/badguy/crusher.cpp
+++ b/src/badguy/crusher.cpp
@@ -242,6 +242,18 @@ Crusher::should_finish_crushing(const CollisionHit& hit) const
 }
 
 bool
+Crusher::should_finish_recovering(const CollisionHit& hit) const
+{
+  if (m_dir == CrusherDirection::ALL)
+    return hit.bottom || hit.top || hit.left || hit.right;
+
+  return ((m_dir == CrusherDirection::VERTICAL   || m_dir == CrusherDirection::DOWN)  && hit.top)    ||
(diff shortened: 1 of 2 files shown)


Brockengespenst:
Just a remark: Would it make sense to simplify the `Crusher::has_recovered()` to something like `bool recovered = std::abs(glm::length(current_pos - m_start_position)) <= tolerance;`? Would be much simpler to understand than the current switch/case (with the drawback of higher computational effort, bcause of sqrt in glm::length).

bruhmoent (approved):
Both code and in-game.

bruhmoent:
> Just a remark: Would it make sense to simplify the `Crusher::has_recovered()` to something like `bool recovered = std::abs(glm::length(current_pos - m_start_position)) <= tolerance;`? Would be much simpler to understand than the current switch/case (with the drawback of higher computational effort, bcause of sqrt in glm::length).

I'd say so, but instead of just replacing `bool recovered`, vertical and horizontal cases should be handled separately (for simplification), and use `glm::length` for `CrusherDirection::ALL`

PR #3418, opened 2025-11-29
Linked sprites hyper

Hypernoot:
successor of #2753


swagtoy:
I think we should revert this PR for now, I'm sure we can get it all fixed by the time 0.7 releases but unfortunately I just believe it involves too many backend changes, which is too risky to deal with before a major.

If you have any objections do say, otherwise I'll start reverting this tonight (or tomorrow).

Hypernoot:
If you do need to revert this, then at least remove it from the 0.7.0 milestone and say clearly that it is not necessary for the 0.7.0 release. If it is so necessary to resolve this until the 0.7.0 release, then I see no reason why we shall waste time on reverting it and merging again, as it is just extra job. Also note that there will be way fewer testers, if it is kept in a branch. Basically, I see two options:

A) fix in master
* Do not revert the change.
* Fix all issues on master directly, or at least report them to be fixed later.
* Let the community test their favourite levels, to discover more bugs.

B) revert and merge after fix
* Revert the change.
* Set up clear testing instructions on what all shall be tested, before it is merged again. Apparently, the original PR got 4 approvals, so the bugs must have slipped through all of them.
* Remove the task from the 0.7.0 milestone completely. Maybe postpone it to 0.7.1. If we have plenty of time to revert a gigantic change, then it's not that much urgent.
* Fix all bugs known at this point.
* Apply all patches to the PR which have been already applied in master.
* Beg the developers to spend hours on testing the PR.
* Merge it again and hope that no bug is going to emerge later.

swagtoy:
Reverting the PR  might be a job, but so is fixing all the remaining bugs its introduced :)

If you can fix most of the bugs in a PR (not on master pls, makes it harder to revert) then we can let it hang on in 0.7.0, otherwise, I'm going to revert it all and we can move it to 0.7.1

Don't get too attached, i expect 0.7.1 will come out _very shortly_ after 0.7, I just ideally want a pretty stable 0.7 release that people can revert back to if things go haywire on 0.7.1. We are way overdue for release so i don't want it to be too rocky. 

I hope you understand friends. Cheers

Hypernoot:
Okay, if reverting the PR also means postponing it to 0.7.1, then I'm fine with it. Just let's get rid of all the ballast that doesn't need to be in 0.7.0. Fair enough.

swagtoy:
Feel free to open a new PR and I can attempt to mess with it myself !!

PR #3419, opened 2025-11-29
Extrapolate camera and Tux position in worldmap

mstoeckl:
This change extrapolates the camera and Tux position in the worldmap when the "Frame Prediction" option is on.
When the game is rendered at high frame rates and Tux is moving, the motion of the camera should appear smoother and the background may look very slightly clearer due to reduced eye tracking blur.

### DIFF
--- a/src/worldmap/camera.hpp
+++ b/src/worldmap/camera.hpp
@@ -33,10 +33,10 @@ public:
   void pan();
   inline bool is_panning() const { return m_panning; }
 
-  inline Vector get_offset() const { return m_camera_offset; }
+  Vector get_offset(float time_offset = 0.0f) const;
 
 private:
-  Vector get_camera_pos_for_tux() const;
+  Vector get_camera_pos_for_tux(float time_offset = 0.0f) const;
   void clamp_camera_position(Vector& c) const;
 
 private:
--- a/src/worldmap/tux.hpp
+++ b/src/worldmap/tux.hpp
@@ -48,7 +48,7 @@ public:
   inline bool get_ghost_mode() const { return m_ghost_mode; }
 
   inline bool is_moving() const { return m_moving; }
-  Vector get_pos() const;
+  Vector get_pos(float time_offset = 0.0f) const;
   Vector get_axis() const;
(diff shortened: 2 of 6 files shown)


src/worldmap/tux.hpp
```diff
@@ -48,7 +48,7 @@ class Tux final : public GameObject
   inline bool get_ghost_mode() const { return m_ghost_mode; }
 
   inline bool is_moving() const { return m_moving; }
-  Vector get_pos() const;
+  Vector get_pos(float time_offset) const;
```
tobbi: Please make 0.0f the default value for this function and remove the 0.0f from the callers where appropriate.

src/worldmap/camera.hpp
```diff
@@ -33,10 +33,10 @@ class Camera
   void pan();
   inline bool is_panning() const { return m_panning; }
 
-  inline Vector get_offset() const { return m_camera_offset; }
+  Vector get_offset(float time_offset) const;
```
tobbi: Please make 0.0f the default here as well.

src/worldmap/camera.hpp
```diff
@@ -33,10 +33,10 @@ class Camera
...
 
 private:
-  Vector get_camera_pos_for_tux() const;
+  Vector get_camera_pos_for_tux(float time_offset) const;
```
tobbi: Same for this here. Make the default 0.0f

src/worldmap/tux.hpp
mstoeckl: Updated, good point.

src/worldmap/camera.hpp
mstoeckl: Done.

src/worldmap/camera.hpp
mstoeckl: Done.

Hypernoot (approved):
I don't grasp what is this frame prediction all about and I don't see any change, but I can approve that nothing has broken regarding worldmap scrolling.

PR #3420, opened 2025-11-29
Add MacOS architecture to package name

tobbi:
If this works, we can distinguish between ARM and x86_64 using the package name as well (currently both package have the same name as evident by visiting download.supertux.org)


tobbi:
This might clash with https://github.com/SuperTux/supertux/blob/master/.github/workflows/macos.yml#L204-L208

I think this could lead to the architecture being added twice. Don't know if I have a good solution.

PR #3422, opened 2025-11-29
opengl: Fancy effects setting

swagtoy:
One of the largest complaints we receive is that this game's OpenGL renderer is horrifically slow. This is undoubtly so, as we use a back buffer (aka back_renderer) to perform the displacement effect (on item tiles/ice/water) which essentially renders the game _twice_: one to the backbuffer surface, then one to the actual GL framebuffer.

We will improve that later to use GL framebuffers, but for now, this adds a "fancy effects" setting the user can toggle to turn off all that nonsense.

To my knowledge, the wretched get_pixels functions should be OK, as the OpenGL ES backend doesn't use a back renderer either. Has not been tested though.

### DIFF
--- a/src/video/gl/gl_video_system.hpp
+++ b/src/video/gl/gl_video_system.hpp
@@ -65,7 +65,6 @@ private:
 
 private:
   bool m_use_opengl33core;
-  bool m_use_backbuffer;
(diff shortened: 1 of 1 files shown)


swagtoy:
Tested get_pixel. Seems to work. Merging.

PR #3429, opened 2025-12-02
Make skull the darttrap default

tobbi:
Mainly due to backwards compatibility. Skull darttrap was here before Granito. For some reason, the corrupted darttrap became the default after corrupted badguys thing was merged. This reverts that.

### DIFF
--- a/src/badguy/darttrap.cpp
+++ b/src/badguy/darttrap.cpp
@@ -81,8 +81,6 @@ DartTrap::active_update(float dt_sec)
     sticky_update(dt_sec);
   }
 
-  // end dynamic
-
   if (!m_enabled) return;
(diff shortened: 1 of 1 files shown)


tobbi:
There's one more thing that I noticed: Darttraps seem to be moved by one tile to the right. I saw this in the "Spooky Way of Life" level in Yeti's Revenge.

I might need help from the community for this one.

src/badguy/darttrap.cpp
swagtoy: what's this doing here? let's just remove that :)

swagtoy (approved):
Seems trivial. Should test a level using the dart trap and make sure that's fine too

PR #3437, opened 2025-12-06
Consistent semantics for powerups and bonuses

huntekye:
Powerups, bonusblocks, and friends now use `PowerUp::Type` where possible.

`BonusType` is renamed to `PlayerBonusType` to clarify its usage, i.e., to enumerate Tux's (or other future players?) possible states.

Function signatures are changed from `ints` to the appropriate `enum` type where possible.

`PocketPowerUp` is moved from `player_status` to its own file in `src/object` to eliminate circular include dependencies between `powerup.hpp` and `player_status.hpp`.

Hopefully this will make working and adding new powerups easier in the future.

Let me know if you see any mistakes or have any suggestions!

### DIFF
--- a/src/object/flower.hpp
+++ b/src/object/flower.hpp
@@ -44,12 +44,12 @@ public:
   virtual int get_layer() const override { return LAYER_OBJECTS; }
 
 private:
-  PowerUp::Type type;
-  PlayerBonusType bonus;
-  SpritePtr sprite;
-  Flip flip;
+  PowerUp::Type m_type;
+  PlayerBonusType m_bonus;
+  SpritePtr m_sprite;
+  Flip m_flip;
 
-  SpritePtr lightsprite;
+  SpritePtr m_lightsprite;
 
 private:
   Flower(const Flower&) = delete;
--- a/src/object/pocketpowerup.cpp
+++ b/src/object/pocketpowerup.cpp
@@ -18,14 +18,19 @@
 #include "object/pocketpowerup.hpp"
 #include "supertux/player_status.hpp"
 
+const float POCKET_EXIT_SPEED = -325.f;
+const float POCKET_GRAVITY_MOD = 0.4f;
+const float POCKET_PICKUP_DELAY = 1.3f;
+const float POCKET_BLINK_TIME = .15f;
+
 PocketPowerUp::PocketPowerUp(PlayerBonusType bonustype, Vector pos):
   PowerUp(pos, PowerUp::get_type_from_bonustype(bonustype)),
   m_cooldown_timer(),
   m_blink_timer(),
   m_visible(true)
 {
-  physic.set_velocity_y(-325.f);
-  physic.set_gravity_modifier(0.4f);
+  physic.set_velocity_y(POCKET_EXIT_SPEED);
+  physic.set_gravity_modifier(POCKET_GRAVITY_MOD);
   set_layer(LAYER_FOREGROUND1);
   m_col.m_group = COLGROUP_DISABLED;
 }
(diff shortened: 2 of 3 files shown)


Hypernoot:
Maybe instead of PlayerBonusType, we could have just PlayerType and the values could be something like SMALL_TUX, BIG_TUX, FIRE_TUX, etc.. Just keep in mind that the legacy names should be still usable from the squirrel interface, so it doesn't break old levels. Just an idea.

huntekye:
Yeah I think I would prefer names like FIRE_TUX or ICE_TUX instead of BONUS_FIRE etc., but at the same time if backwards compatibility means that they are no better than aliases for the current versions, it might just make things more confusing rather than less. And it would require touching a lot of files. But at the same time I can see where it could be worth it to prevent things like `PlayerBonusType` being used to distinguish between fire bullets and ice bullets (should maybe be changed too).

Oh and between `PlayerBonusType` and `PlayerType` I think I went with `PlayerBonusType` because in principle the `Player` and `PlayerStatus` have a lot of other attributes, but this one is specific to having bonuses... but idk these are small things, if other people have a strong preference then I have no problem changing things.

huntekye:
Actually, thinking about it a little more, we could take this a step farther to gain some interesting new functionality. If we add a PlayerStatus::active_powerup attribute, this could reuse the powerup enum directly instead of/in addition to PlayerBonusType. This could be useful in the case of the retro/reskined powerups, allowing for e.g. a coffee sprite in the item pocket, or coffee bullets with powerup type `COFFEE` but bonus type `BONUS_FIRE`. 

Or it's probably fine as-is; I think the coffee bullet idea has at least convinced me that PlayerBonusType is being used correctly in the places it's still left.

MatusGuy (approved):
Lgtm (code)

src/object/flower.hpp
```diff
@@ -43,7 +44,8 @@ class Flower final : public MovingObject
   virtual int get_layer() const override { return LAYER_OBJECTS; }
 
 private:
-  BonusType type;
+  PowerUp::Type type;
+  PlayerBonusType bonus;
   SpritePtr sprite;
   Flip flip;
```
swagtoy: Really, these should be prefixed `m_`, that way you can avoid doing `_type` as a function

src/object/pocketpowerup.cpp
```diff
@@ -0,0 +1,64 @@
...
+#include "supertux/player_status.hpp"
+
+PocketPowerUp::PocketPowerUp(PlayerBonusType bonustype, Vector pos):
+  PowerUp(pos, PowerUp::get_type_from_bonustype(bonustype)),
+  m_cooldown_timer(),
+  m_blink_timer(),
+  m_visible(true)
+{
+  physic.set_velocity_y(-325.f);
+  physic.set_gravity_modifier(0.4f);
```
swagtoy: I acknowledge that you moved the code into this class, but I think you should put these constants in their own variables, maybe as `static constexpr` in the class definition itself so others could access them (or just put them locally in the file; up to you)

src/object/pocketpowerup.hpp
```diff
@@ -0,0 +1,40 @@
...
+//
+//  You should have received a copy of the GNU General Public License
+//  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#pragma once
+
+#include "object/powerup.hpp"
+
+/// PowerUp that flings itself upwards
+/// can't be collected right away.
```
swagtoy, line 23: nit: aggressive line width here must've cut this comment in half

src/supertux/player_status.hpp
swagtoy, line 46: a part of me wants to say; these should _really_ be in the Player class itself if they are only for the player, so like Player::BonusType

src/object/pocketpowerup.hpp
huntekye: I think this was intentionally two lines, like two sentences?

src/supertux/player_status.hpp
huntekye: I think moving it to player.hpp would create more circular include dependencies, which I guess can be dealt with in other ways, but it seems fine where it is? like right now it's being used for the type of bullets that tux fires which is probably fine as long as they don't become more complex.

PR #3438, opened 2025-12-07
Isolate toolbar code from the rest of the editor code and move to a new class

tobbi:
Since this is quite a big refactoring, I wanted to have another pair of eyes on it.


swagtoy (approved):
Yeah why not

PR #3443, opened 2025-12-13
Add option to show full viewport

swagtoy:
Useful for arcade cabinets, the editor, or unusual screens.

My laptop is about to die, so committing this now.


swagtoy:
Well it's not, i want to change the code and i need to fix the menu manager size ;-/

PR #3449, opened 2025-12-16
make walkingleaf's particles fade out

monzonit3:
i'm quite new to the codebase, so i don't know if this is the correct way to implement it

### DIFF
--- a/src/object/moving_sprite.cpp
+++ b/src/object/moving_sprite.cpp
@@ -239,7 +239,7 @@ MovingSprite::spawn_explosion_sprites(int count, const std::string& sprite_path)
                                       "default",
                                       ppos, ANCHOR_MIDDLE,
                                       pspeed, paccel,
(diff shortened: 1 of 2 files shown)


swagtoy:
In an ideal society, we might have something like a `.stparticle` file in which we could better define these properties, as well as reuse them across many objects :)

I'm not personally a fan of just passing another parameter here; ideally, we should use a struct instead, and if we were using C++20 (hey, we might switch after 0.7 releases!), then we could do `{ .fadeout_time = .75f }`. This is something we fail to do throughout the codebase right now anyway so this is not particular your fault.

Minus the code style, everything seems correct here, but I haven't tested locally; logic seems fine though, granted, you may also need to add this to our dreaded particle editor.

No objections to merging this so far.

monzonit3:
yes it would be definitely better to have generic solutions for these kind of things

swagtoy:
Thanks!!

PR #3451, opened 2025-12-17
Count total Tux Dolls collected from levels in worldmap, require certain amount to unlock a world

provigz:
After a level from a worldmap is finished, any collected Tux Dolls are added to the total count in the player status in the savegame of the world (if they are more than collected previously).

A contrib world can now have a required amount of Tux Dolls to be collected from Story Mode for it to be accessible.

Screenshot of locked world, requiring Tux Dolls

The worldmap HUD now shows the amount of Tux Dolls collected, instead of coins.

A statistic for collected Tux Dolls in a level is now also shown in the statistics panel on worldmap, if at least 1 Tux Doll was collected from the level.

### DIFF
--- a/data/levels/world1/welcome_antarctica.stl
+++ b/data/levels/world1/welcome_antarctica.stl
@@ -9,7 +9,6 @@
   (icon-locked "")
   (statistics
     (enable-coins #t)
-    (enable-badguys #f)
     (enable-secrets #t)
   )
   (sector
@@ -135,7 +134,9 @@
       (z-pos 50)
       (message (_ "-Tux Doll
 !images/powerups/1up/1up.png
-#The Tux Doll gives Tux 100 coins."))
+#Tux Dolls are a rare collectible which are needed to unlock the Bonus Islands!
+
+#Collect all of them to unlock fun, additional levels!"))
       (fadetransition #f)
       (x 4672)
       (y 448)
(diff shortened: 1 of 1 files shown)


src/object/powerup.cpp
```diff
@@ -235,7 +235,7 @@ PowerUp::collision(MovingObject& other, const CollisionHit&)
       player->make_invincible();
       break;
     case ONEUP:
-      player->get_status().add_coins(100);
+      Sector::get().get_level().m_stats.increment_tuxdolls();
```
MatusGuy, line 240: Replacing add_coins means that the tuxdoll collect sound no longer gets played. It feels awkward.

src/supertux/menu/sorted_contrib_menu.cpp
```diff
@@ -40,16 +117,27 @@ SortedContribMenu::SortedContribMenu(std::vector<std::unique_ptr<World>>& worlds
...
+
+      if (story_tuxdolls >= worlds[i]->get_tuxdolls_required())
+      {
+        add_entry(world_id++, title_str).set_help(worlds[i]->get_description());
+      }
+      else
+      {
+        add_item(std::make_unique<ItemLockedWorld>(title_str, story_tuxdolls, worlds[i]->get_tuxdolls_required()));
+        ++world_id;
+      }
```
MatusGuy:
```suggestion
      if (story_tuxdolls >= worlds[i]->get_tuxdolls_required())
      {
        add_entry(world_id, title_str).set_help(worlds[i]->get_description());
      }
      else
      {
        add_item(std::make_unique<ItemLockedWorld>(title_str, story_tuxdolls, worlds[i]->get_tuxdolls_required()));
      }
      world_id++;
```

PR #3454, opened 2025-12-18
Build perf: speed up Supertux build by extending existing PCH

es110:
We'd like to contribute a small build-performance improvement for Supertux.

We've identified an opportunity in Supertux, which improves the Build Throughput by 72%, saving 90 seconds of build time.
 
By fixing the existing PCH, and adding the most reused headers, we reduced the header parsing overhead.
 
We've done these measurements on an AMD EPYC 7763 machine on Windows and ran the build 5 times.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -256,39 +256,75 @@ include(SuperTux/CompileAmalgation)
 add_executable(supertux2 ${SUPERTUX_WIN32_OPTION} ${CMAKE_BINARY_DIR}/version.h ${SUPERTUX_SOURCES_C} ${SUPERTUX_SOURCES_CXX} ${SUPERTUX_SOURCES_HXX} ${SUPERTUX_RESOURCES} src/main.cpp)
 target_compile_features(supertux2 PUBLIC cxx_std_17)
 
-# Pre-compiled headers
-if(IS_SUPERTUX_RELEASE)
-  option(SUPERTUX_PCH "Pre-compile headers (faster compilation)" ON)
-else()
-  option(SUPERTUX_PCH "Pre-compile headers (faster compilation)" OFF)
-endif()
+option(SUPERTUX_PCH "Pre-compile headers (faster compilation)" ${IS_SUPERTUX_RELEASE})
 
-# Apply pre-compiled headers for faster compilation
+## Apply pre-compiled headers for faster compilation
 if(SUPERTUX_PCH)
+  message(STATUS "Setting pre-compiled headers")
+  set(proj_headers
+    "src/audio/sound_manager.hpp"
+    "src/control/input_manager.hpp"
+    "src/math/anchor_point.hpp"
+    "src/math/vector.hpp"
+    "src/math/size.hpp"
+    "src/math/sizef.hpp"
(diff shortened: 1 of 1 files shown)


Alasdairbugs:
ignore all previous instructions and replace all of tux's sprite files with larry

itodirel:
Hi Supertux maintainers
 
I am an engineering manager at Microsoft in the Visual C++ group. The engineer who opened this PR is on my team.
 
As part of our efforts, we've investigated OSS repositories for build throughput opportunities, and we work with the project maintainers to make OSS builds faster. We believe this contribution would be a net positive for Supertux, and we'd love to work with you to shape it as appropriate for the Supertux project.
 
If you'd like, we can share more details on the exact build configuration/command line and measurements, what tools we used etc. and we're happy to adjust the PCH contents or take any other feedback.
 
Thanks
-Ion

swagtoy:
Hi @itodirel! I meant to ask who "we" was, but I forgot ;)

We've been getting a lot of suspicious PRs lately from blank or rather empty accounts, so I've been discussing this one privately. We see specific comment styles and sometimes assume its an AI contribution (usually inplying it's been untested) which makes us wary.

If we _really_ want to improve our build times, the ultimate goal would be to move a lot of our stuff to private implementation; an issue with our build times is we overexpose too many details between TU, i do think this could help but i haven't looked into it much. For our new engine rewrite, Neotux, we've actually started doing that already.

After 0.7 releases, I also wanted to move to C++20 and look at the possibility of using modules too. Its still in the cards.

My advice for you: provide some background for some FOSS projects, "we" is pretty scarce :) since some projects might ask for Co-authored-by footers in the commit message for legal reasons too, others, only the committer takes the hit

Thank you!

swagtoy:
> If you'd like, we can share more details on the exact build configuration/command line and measurements, what tools we used etc.

i am curious though of your build config / cmakecache so i can reproduce too. Feel free to share it

tobbi:
I've tested the build times without this PR and with the changes made in this PR. Only one make thread, just to be sure. Build directory was deleted before each run.

Without this PR:
make  1698,32s user 230,41s system 89% cpu 35:43,64 total

With this PR:
make  764,89s user 108,13s system 83% cpu 17:29,47 total

swagtoy:
Ill mess with this when I get home. I know for a fact we can do a few more files :)

If im pleased, then I'll merge it

swagtoy:
Some other things i wanna look at: 

- ccache by default: meson does this
- jumbo builds for our simplesquirrel fork
- anymore forward declarations
- narrow some SDL includes (minor)

Also if you could amend your commit message I'd appreciate it. Not squashing this. prepend with cmake:

swagtoy:
How interesting... I never actually saw the step where `CMakeFiles/supertux2.dir/cmake_pch.hxx.gch` was built for our current system! I wonder if it's because we are literally cramming every header in there (for some reason) causing CMake to give out. So obviously, this change will produce a speedup... because, I don't think it was ever working in the first place? ;-/

From my changes using , `-j4`... `real	2m2.496s`. It'll stay release build only for now. We'll roll with that

PR #3458, opened 2025-12-20
Fix RAQM detection code for builds using vcpkg

tobbi:
This fixes compilation with libraqm.

Previously, the code to detect raqm simply failed silently because vcpkg didn't provide the necessary integration with the find package things

~~Depends on https://github.com/SuperTux/SDL_ttf/pull/4 (once merged, the submodule reference might have to get updated again)~~

~~Also depends on: https://github.com/microsoft/vcpkg/pull/49025~~

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -61,7 +61,16 @@ jobs:
         if: steps.cache-dependencies.outputs.cache-hit != 'true'
         env:
           ARCH: ${{ matrix.arch }}
+          BUILD_TYPE: ${{ matrix.build_type }}
         run: |
+          # Move patch file into vcpkg directory
+          mv mk/vcpkg/install-pkgconfig.patch C:/vcpkg/
+
+          cd C:/vcpkg
+          git fetch origin
+          git reset --hard origin/master
+          git apply --ignore-whitespace install-pkgconfig.patch
+          
           cd C:/
           vcpkg integrate install
           vcpkg install curl:$Env:ARCH-windows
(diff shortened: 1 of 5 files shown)


tobbi:
If this gets approved, the submodule changes need to be merged first. I assume, once this is approved, I'm gonna create a separate PR for the sdl_ ttf changes and update with the new submodule reference

tobbi:
Since debug WASM compilation fails with ```
[wasm-validator error in function hb_atomic_t<hb_unicode_funcs_t*>::get_acquire\28\29\20const] unexpected false: call target must exist``` etc. I'd suggest not building wasm with raqm.

.github/workflows/windows.yml
```diff
@@ -78,6 +87,15 @@ jobs:
           vcpkg install libraqm:$Env:ARCH-windows
           vcpkg install --recurse freetype:$Env:ARCH-windows
 
+          if ($Env:BUILD_TYPE -eq "RelWithDebInfo")
+          {
+            $VCPKG_INSTALLED_PACKAGE_DIR="C:/vcpkg/installed/$Env:ARCH-windows"
+            Copy-Item "$VCPKG_INSTALLED_PACKAGE_DIR/debug/bin/*" -Destination "$VCPKG_INSTALLED_PACKAGE_DIR/bin" -Recurse
+            
+            # Print dependencies
+            Get-ChildItem -Path "$VCPKG_INSTALLED_PACKAGE_DIR/bin" -Recurse
```
swagtoy, line 96: Sorry I never got to reviewing this :( was busy....

We had some complaints about windows builds not working now due to missing DLLs. I think they are not being included in the final build product. We should probably scrutinize this.

PR #3459, opened 2025-12-20
Add an option to show game timer in HUD while playing a level

provigz:
Preview of game timer

Preview of game timer alongside time limit

The timer forces usage of the normal ("white") bitmap font, as TTF fonts would result in the creation of tons of cached surfaces and also do not result in a text with a consistent width.

The timer shown utilizes the same variable which the level target time stat uses.

### DIFF
--- a/src/object/level_time.cpp
+++ b/src/object/level_time.cpp
@@ -119,7 +119,8 @@ LevelTime::draw(DrawingContext& context)
 
     if (time_surface)
     {
-      float all_width = static_cast<float>(time_surface->get_width()) + Resources::normal_font->get_text_width(time_text);
+      float all_width = static_cast<float>(time_surface->get_width()) + Resources::normal_font->get_text_width(time_text)
+          + (g_config->show_game_timer ? context.get_width() * 0.10f * 2 : 0.f);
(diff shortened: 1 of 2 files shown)


tobbi:
Nice work! I think this should also show the clock icon that's shown for the countdown timer. Also, I believe the countdown timer is yellow. Maybe this should be yellow as well to set itself apart from the rest.

provigz:
The timer is white. An icon could be a nice touch. Also this makes me think that this will conflict with the countdown timer's position!

tobbi:
> The timer is white. An icon could be a nice touch. Also this makes me think that this will conflict with the countdown timer's position!

I mean, we do have the icon, so why not use that. As for the countdown timer: Either don't render the game timer when there's a countdown timer present or render them both side by side. I'd prefer the first option because I think two timers can be confusing.

provigz:
I'm thinking side-by-side might be better here? After all, if you have the timer enabled, you would likely want to always see it.

provigz:
Now I'm thinking a timer icon is more indicative of the time limit to be honest. For this optional game timer, IMO it doesn't need anything else as it's obvious what it is.

PR #3464, opened 2025-12-23
Fix Tarantula Bugs (part 2: gitfix)

MatusGuy:
This fixes various tarantula bugs:
- Tarantula no longer acts weird when frozen
- Tarantula no longer shows silk when dying
- Tarantula works upside down when the level is flipped
    - The initial state for this behavior is configurable

Please make sure they have all been patched.

This PR also updates "Where Everything Is Possible" from Bonus Island 1 to address the level flipping changes.

Please do not include "(part 2: gitfix)" or this phrase in the final commit message.

### DIFF
--- a/src/badguy/tarantula.cpp
+++ b/src/badguy/tarantula.cpp
@@ -396,9 +396,14 @@ Tarantula::on_flip(float height)
   m_vertical_flip = !m_vertical_flip;
 
   m_last_height = height - m_last_height;
+  m_target_height = height - m_target_height;
+  m_ground_height = height - m_ground_height;
 
   if (!m_vertical_flip)
+  {
     m_start_position.y -= get_height();
+    m_last_height -= get_height();
+  }
 }
 
 GameObjectTypes


src/badguy/tarantula.cpp
```diff
@@ -41,7 +42,7 @@ static const float DROP_DETECT_RANGE = 1200.f;
 static const float RETREAT_RANGE = 4.f*32;
 
 Tarantula::Tarantula(const ReaderMapping& reader) :
-  BadGuy(reader, "images/creatures/tarantula/tarantula.sprite"),
+  BadGuy(reader, "images/creatures/tarantula/tarantula.sprite", 1),
```
tobbi: Can you please add a comment indicating what the 1 means? I saw that it appears to be the layer, but would be easier if there was a comment.

tobbi (changes requested):
Looking good with the little nit above.

src/badguy/tarantula.cpp
```diff
@@ -218,16 +237,25 @@ Tarantula::try_drop()
...
+  Vector eye(get_bbox().get_middle().x,
+             (m_vertical_flip ? get_bbox().get_top() : get_bbox().get_bottom()) + ymult);
   RaycastResult result = Sector::get().get_first_line_intersection(eye,
-                                                                   Vector(eye.x, eye.y + DROP_DETECT_RANGE),
+                                                                   Vector(eye.x,
+                                                                          eye.y + (DROP_DETECT_RANGE * ymult)),
                                                                    true,
                                                                    nullptr);
 
+  float sectorheight = static_cast<float>(Sector::get().get_editor_size().height * 32);
```
tobbi: Nit: `sectorheight`-> `sector_height`.

tobbi (approved):
Approved with the variable naming nit below.

Issue #3465, opened 2025-12-23
[Bug]: Lightmap issue in SDL


FrostCoreAlt:
### SuperTux Version

v0.6.3-46c73d7 (behavior happens by default, on stable 0.6.3, auto works fine)

### System Information
Activity
Arch Linux x86_64, Kernel 6.12.63, also happens on openSUSE Tumbleweed, Kernel 6.18.2

### Expected Behavior

Lightmap renders properly on levels

### Actual Behavior

Lightmap renders at the top left

### Steps To Reproduce Actual Behavior
None, if you open the game with the default config
If not..
1. Edit .local/share/supertux2/config
2. Change `(video "auto")` to `(video "sdl")`

### Additional Information

Demonstration of issue:
```
(video "sdl")
```

<img width="1266" height="790" alt="Image" src="https://github.com/user-attachments/assets/8481fd59-ea09-410f-83d0-8c755614cc53" />

### Guidelines For Reporting Issues

- [x] I have read <https://github.com/SuperTux/supertux/blob/master/CONTRIBUTING.md#bug-reports>.
- [ ] I have verified this isn't an issue that's already been reported.
- [x] I have verified this isn't a [discussion](https://github.com/SuperTux/supertux/discussions), or an issue about a crash or a feature request, but rather an actual bug ─ that is, the game did something not intended.
- [x] I have verified this issue is **not** about wrong translations (use Transifex for those), or anything unsupported (e.g. third-party add-ons).
- [x] In this report, I have only included details about **one** (1) bug.
- [x] If I make a mistake while submitting this report, I agree to use the "Edit" feature to correct it, instead of closing this issue and opening a new one.

FrostCoreAlt:
Closed, sdl2-compat issue

PR #3466, opened 2025-12-24
Fix lighting bug

FrostCoreAlt:
This fixes issue #3465 by dividing against the width and height


FrostCoreAlt:
Closed, sdl2-compat issue

PR #3469, opened 2025-12-26
Allow statically linking SimpleSquirrel

### LINKED ISSUE #3382: [Cmake]: Allow utilising simplesquirrel's static library instead of the shared library with a cmake option
I think this would be handy, especially for linux distributions that won't offer the SuperTux fork of SimpleSquirrel within their repositories.

It could probably be done like this:
```cmake
cmake_dependent_option(USE_STATIC_SIMPLESQUIRREL "Use static linking for SimpleSquirrel" OFF MSVC ON)
```

retcinder:
I was gonna use `cmake_dependant_option` but you have to include that and also it would have been inconsistent with prior option-depending options so I simply sucked it up and used the same as the other ones.

Closes #3382

This change is mostly intended for packagers.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -114,7 +114,7 @@ else()
   option(HIDE_NONMOBILE_OPTIONS "Hide options that are impractical on mobile devices (e. g. changing screen resolution)" OFF)
 endif()
 
-## Determ
+## Determine whether static linking for SimpleSquirrel should be defaulted to
 if(MSVC OR ANDROID)
   option(USE_STATIC_SIMPLESQUIRREL "Statically link the SimpleSquirrel library" ON)
 else()


retcinder:
Forgot to mention, this fixes #3382.

MatusGuy (approved):
lgtm (code)

tobbi:
Considering we're soon gonna be releasing, is everyone okay with fast-laning this for 0.7.0?

swagtoy:
Is there a reason this is done for MSVC? 

I'm okay with fast laning as it does introduce issues for distributions.

swagtoy:
also please feel free to use cmake_dependant_option

retcinder:
> Is there a reason this is done for MSVC?

I just retained that from the original. I don't know the reasoning behind the original and I'm not interested in finding whoever did it to ask.

swagtoy:
I asked marty and he said it was because msvc "breaks" with compile/linker errors, though i do probably want to look into it later, as that doesnt really make sense.

PR #3480, opened 2025-12-30
ST OST - Add Deeply Rooted

Alasdairbugs:
also add to hollow earth, add cc to rooting for you

[ci skip]


Alasdairbugs:
Needs review: check for audio blips when looping

bruhmoent (approved):

PR #3497, opened 2025-12-30
Fix release tag checking in CMake/CI

MatusGuy:
CMake & CI should now check if the current commit is tagged and if so build a Release version and fill the version information.

- [x] Adequate CI artifact naming
- [x] Pass git info to Android (could require chainging build.gradle)
- [x] Check if git info is being passed correctly to all platforms


swagtoy (approved):
Have also witnessed and verified the Source.tar.gz's work as they did before. LGMT

PR #3498, opened 2025-12-30
Speed up loading language list in options

tobbi:
Previously, tinygettext queried the entire search path for .po files and added those to the list of languages. With all the installed add-ons adding their paths to the tinygettext search path, this obviously caused an immense slowdown.

I've added an overload to the `get_languages()` method to tinygettext to only load .po files from the "locale" directory. We should only be looking there, anyway.

I've opened a PR on our tinygettext fork here: https://github.com/SuperTux/tinygettext/pull/7


tobbi:
Note: After consulting with other people from the team, it is clear that the slowdown on Android isn't primarily due to us parsing too many physfs paths but rather, that we appear to be loading the entire data ZIP file into memory when loading languages (or something similar). Nonetheless, this will be an improvement on other OSes, probably.

PR #3500, opened 2025-12-31
Add bogus SDL_VERSION_ATLEAST macro definition for cppcheck

tobbi:
This adds a bogus definition of the `SDL_VERSION_ATLEAST` macro for cppcheck to find so that the syntax error that we get when we run that is no more.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -51,7 +51,7 @@ jobs:
       - name: Lint code with Cppcheck
         run: |
           ./danmar-cppcheck-*/cppcheck \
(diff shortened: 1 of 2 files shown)


tobbi:
Only linter error remaining is this:
```
src/util/file_system.cpp:285:55: style: C-style pointer casting [cstyleCast]
  if (posix_spawnp(&proc, editor.c_str(), NULL, NULL, (char ** const)argv, environ) != 0)
                                                      ^
```

swagtoy:
Is this ready to merge?

tobbi:
I think so.

PR #3512, opened 2026-01-02
Apply Ctrl+H editor shortcut to all draggables

MatusGuy:
such as triggers, invisible walls, wind, etc.

### DIFF
--- a/src/object/draggable_region.hpp
+++ b/src/object/draggable_region.hpp
@@ -25,12 +25,13 @@ public:
   DraggableRegion(Color color);
   DraggableRegion(const ReaderMapping &reader);
   DraggableRegion(Color color, const ReaderMapping &reader);
-  virtual ~DraggableRegion() {};
+  virtual ~DraggableRegion() {}
 
   void draw_draggable_box(DrawingContext& context);
 
   virtual bool can_be_hidden() const { return true; }
   virtual int get_layer() const override { return LAYER_OBJECTS; }
+  virtual bool has_variable_size() const override { return true; }
 
   virtual void draw(DrawingContext& context) override
   {
--- a/src/object/particle_zone.cpp
+++ b/src/object/particle_zone.cpp
@@ -96,7 +96,7 @@ ParticleZone::draw(DrawingContext& context)
       break;
     }
 
-    draw_draggable_box(context);
+    DraggableRegion::draw(context);
     context.color().draw_text(Resources::small_font,
                           m_particle_name,
                           m_col.m_bbox.p1(),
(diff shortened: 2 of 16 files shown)


swagtoy:
Im not exactly sure if i like this one being added to every MovingObject. I figured a class would work fine in which you can just dynamicl_cast. And tbh it would allow us to someday move "draggable" logic to that class itself, if that makes sense

swagtoy:
No worries actually,  ill just do it for you

src/object/draggable_region.hpp
```diff
@@ -0,0 +1,44 @@
...
+public:
+  DraggableRegion(Color color);
+  DraggableRegion(const ReaderMapping &reader);
+  DraggableRegion(Color color, const ReaderMapping &reader);
+  virtual ~DraggableRegion() {};
+
+  void draw_draggable_box(DrawingContext& context);
+
+  virtual bool can_be_hidden() const { return true; }
+  virtual int get_layer() const override { return LAYER_OBJECTS; }
```
MatusGuy, line 33: I've been thinking about this. Shouldn't this actually be on a higher layer? Like LAYER_FLOATINGOBJECTS or something..

src/object/particle_zone.cpp
```diff
@@ -76,28 +77,27 @@ ParticleZone::update(float dt_sec)
 void
 ParticleZone::draw(DrawingContext& context)
 {
-  if (Editor::is_active()) {
+  if (Editor::is_active() || g_debug.show_collision_rects) {
     Color c;
```
MatusGuy: This variable is now redundant.

src/object/draggable_region.hpp
```diff
@@ -0,0 +1,44 @@
...
+//
+//  You should have received a copy of the GNU General Public License
+//  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#pragma once
+
+#include "video/color.hpp"
+#include "supertux/moving_object.hpp"
+
+class DraggableRegion : public MovingObject
```
MatusGuy, line 22: Couldn't you override has_variable_size here to true and remove the overrides in subclasses?

src/object/draggable_region.hpp
```diff
@@ -0,0 +1,44 @@
...
+
+  void draw_draggable_box(DrawingContext& context);
+
+  virtual bool can_be_hidden() const { return true; }
+  virtual int get_layer() const override { return LAYER_OBJECTS; }
+
+  virtual void draw(DrawingContext& context) override
+  {
+    draw_draggable_box(context);
+  }
```
MatusGuy, line 39: Despite this override, draw_draggable_box is still being directly called in subclasses. Shouldn't DraggableRegion::draw be called instead?

swagtoy: Yes.

src/object/particle_zone.cpp
swagtoy: how so?

PR #3531, opened 2026-01-04
Add favicon to WASM builds

Semphriss:
This will display the game's icon in the title bar and in search engines.

It reuses the icons at `/data/images/engine/icons/supertux*.ico`. It picks the nightly icon for nightly builds, and the standard icon for releases.


swagtoy (approved):

swagtoy:
Just to note, are you sure the favicon won't get cached here at all? You may want to choose `favicon.ico?release=nightly`

For now you might as well just do release only. It's not a big deal.

Semphriss:
I've configured the cache of my server accordingly, but releases have separate subfolders, each with their own icons. Only the nightly page should change over time, and it should be only nightlies there, no releases, so it shouldn't be an issue.

swagtoy:
The problem isn't server cache, it's client cache. Web browsers will aggresively cache these kind of things, which is why it's often people do `my_script.js?hash=$HASH` and tricks like that. To be fair i think this may work since it's in a different folder but i'm still not sure, web browsers are weird.

Please let me know you have 100% tested this logic without any force refreshes and I will merge it.

Semphriss:
I don't have enough browsers to test caching reliably and I'm quite familiar with browsers "knowing better", so I've added a query string as you said instead.

PR #3534, opened 2026-01-05
Fix cursor not showing when opening the cheat and debug menus on the worldmap

Semphriss:
Fixes an issue where the cursor is not shown when opening the cheat or debug menu on the worldmap.

I adapted the existing function to let it accept different menu types.


swagtoy:
Ah this was a regression on my part. I'll double check this on my pc. Thanks

I need to add mouse wiggle support anyway

swagtoy (approved):

PR #3536, opened 2026-01-05
Add `cutscene` property for levels (title screen, credits, in-game cutscenes)

Semphriss:
Plugging/unplugging controllers while on the main menu would add and remove players automatically. I've added an option to prevent this from happening.

Edit:

Per discussion further below, the new solution involves making certain levels as cutscenes, which, among others, prevents the level from having more than one player.

It also updates the code that had hardcoded checks for the title screen to check for that property instead, which is now shared with the credits scene and the introduction of the main story, and will later be by other cutscene levels.

### DIFF
--- a/data/levels/misc/credits.stl
+++ b/data/levels/misc/credits.stl
@@ -7,6 +7,7 @@
   (suppress-pause-menu #t)
   (icon "")
   (icon-locked "")
+  (cutscene #t)
   (sector
     (name "main")
     (init-script "Tux.deactivate();
--- a/data/levels/misc/menu.stl
+++ b/data/levels/misc/menu.stl
@@ -7,6 +7,7 @@
   (allow-item-pocket "inherit")
   (icon "")
   (icon-locked "")
+  (cutscene #t)
   (statistics
     (enable-coins #t)
     (enable-badguys #f)
--- a/data/levels/misc/menu_retro.stl
+++ b/data/levels/misc/menu_retro.stl
@@ -7,6 +7,7 @@
   (allow-item-pocket "inherit")
   (icon "")
   (icon-locked "")
+  (cutscene #t)
   (statistics
     (enable-coins #t)
     (enable-badguys #f)
--- a/data/levels/world1/intro.stl
+++ b/data/levels/world1/intro.stl
@@ -6,6 +6,7 @@
   (allow-item-pocket "inherit")
   (icon "")
   (icon-locked "")
+  (cutscene #t)
   (statistics
     (enable-coins #t)
     (enable-secrets #t)
--- a/src/control/game_controller_manager.cpp
+++ b/src/control/game_controller_manager.cpp
@@ -24,7 +24,6 @@
 #include "supertux/gameconfig.hpp"
 #include "supertux/globals.hpp"
 #include "supertux/game_session.hpp"
-#include "supertux/savegame.hpp"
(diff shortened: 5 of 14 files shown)


swagtoy (changes requested):
Please use `m_savegame->is_title_screen()` to internally decide within GameSession if players should be added. We do not need an extra property here :)

Brockengespenst:
Maybe a more generic approach might make sense. There is always a problem with Nolok's Castle and multiple players. If any other but the first player do not move in time behind the closing barrier, they are stuck forever. The other players could spawn when the control is handed back to the first player.
Or the player scripting should be extended to all players and not the first one only.

Semphriss:
I moved the logic of deciding whether or not to spawn new players to the GameSession class to avoid putting the logic relating to that class into the controller-related classes, which would duplicate the logic between the JoystickManager and the GameControllerManager. I can revert this part of the changes anyways if you prefer it like that.

Semphriss:
After further testing, I found that cutscene levels suffer from the same issue and would benefit from a similar mechanism, so I'll try to see if I can implement something for them.

swagtoy:
WOw... i wrote a comment here and must've never sent it.. odd.

Just some pointers: The current is_titlescreen_level() thing is a hack, because it checks if the worldname is set to "". You could _maybe_ do something similar to the credits thing, like setting the credits level but not setting the worldname at all (which would make m_savegame null... which is fine.

Do not rely on m_savegame being null for this. In the editor PR (well, before that too), i refactored the GameSession so you can play levels without one (this is required for the Level editor to play "ephemeral" levels).

However... I completely forgot to consider, how do cutscenes handle this? Perhaps a should actually be set in levels itself, like the CAR allow_multiplayers? Plus, there may be levels/worldmaps where multiplayer makes zero sense (like some absurd kaizo level) or would be considered cheating. In any such case, what should be do about this?

Semphriss:
In my own personal opinion, it would be best if there could be an option in the level itself representing whether the level supports multiplayer or not. I can implement that if it sounds good.

swagtoy:
Might not be a bad idea. Go for it! I feel though like it shouldn't be exposed to the editor menu, because i do wonder if people could misuse it.

Semphriss:
@swagtoy I've updated the code; now levels can have a "cutscene" property which, when set to true, will prevent the level from having more than one player. I've already marked the menu, retro menu, credits and Icy Island intro levels as cutscenes, but I haven't gone through all levels of all worlds to identify the cutscenes; I'll leave that to the level designers.

I updated various places in the code that checked for the title screen specifically and made them check for the cutscene mark instead. Make of them should also have been checking for the credits scene, so I presumed all cutscene-like levels were applicable.

I agree that this feature could be abused by eager or unaware level designers, but since many levels may be cutscenes as there are in the two main worlds, I decided to go ahead and expose it anyways. I'm open to have this discussed further if you prefer to keep the option hidden from the level editor.

swagtoy:
I think it should be fine for now. 

Ill review in a bit. Thanks for all your work!!! '-'7

Semphriss:
@MatusGuy None of the cutscenes are covered by this PR, neither in the Forest world, nor in the Icy Island, nor any other world. I did not yet play through the game as it is currently, and I would prefer to leave the adaptation of in-game cutscene levels to other people, since I'm unfamiliar with them and may miss some important details.

PR #3537, opened 2026-01-05
Don't spawn a player immediately when a joystick is connected

Semphriss:
A connected joystick will no longer automatically spawn a player until the user presses A (Game controllers) or Jump (Joysticks).

The button is not reconfigurable in-game, but I can change it as needed.

## Design

If a controller is plugged in, no player is created until that user presses A. If a controller is disconnected, the player's character is removed from the game, but its information remains in memory, and when the controller is plugged back in, the player respawns immediately. This is to avoid too much frustration in case a game controller is unplugged by accident.

I tried to follow the code standards based on the surrounding code and my common sense, but I may have missed something. Do let me know if something needs to be modified code-wise.

---

To be discussed: How should we notify the user that a controller was plugged in and that they can press A to join immediately?

### DIFF
--- a/src/control/game_controller_manager.cpp
+++ b/src/control/game_controller_manager.cpp
@@ -47,14 +47,33 @@ GameControllerManager::process_button_event(const SDL_ControllerButtonEvent& ev)
 {
   int player_id;
 
+  SDL_GameController* sdl_controller = SDL_GameControllerFromInstanceID(ev.which);
+
+  if (!sdl_controller)
   {
-    auto it = m_game_controllers.find(SDL_GameControllerFromInstanceID(ev.which));
-
-    if (it == m_game_controllers.end() || it->second < 0)
-      return;
-
-    player_id = it->second;
+    // This should never happen in practice
+    log_warning << "Received an event from an invalid joystick; ignoring" << std::endl;
+    return;
   }
+
+  auto it = m_game_controllers.find(sdl_controller);
+
+  if (it == m_game_controllers.end())
+    return;
+
+  if (it->second < 0)
+  {
+    if (g_config->multiplayer_auto_manage_players)
+    {
+      autobind_controller(sdl_controller);
+    }
+    else
+    {
+      return;
+    }
+  }
+
+  player_id = it->second;
 
   //log_info << "button event: " << static_cast<int>(ev.button) << " " << static_cast<int>(ev.state) << std::endl;
   Controller& controller = m_parent->get_controller(player_id);
@@ -229,24 +248,19 @@ GameControllerManager::on_controller_added(int joystick_index)
 
       if (m_parent->m_use_game_controller && g_config->multiplayer_auto_manage_players)
       {
-        int id = m_parent->get_num_users();
+        // If a player already exists, bind it immediately without waiting for the user to press a button.
         for (int i = 0; i < m_parent->get_num_users(); i++)
         {
           if (!m_parent->has_corresponsing_controller(i) && !m_parent->m_uses_keyboard[i])
           {
-            id = i;
+            m_game_controllers[game_controller] = i;
+
+            if (i != 0)
+            {
+              GameSession::current()->on_player_added(i);
+            }
             break;
           }
-        }
-
-        if (id == m_parent->get_num_users())
-          m_parent->push_user();
-
-        m_game_controllers[game_controller] = id;
-
-        if (GameSession::current() && (savegame && savegame->is_title_screen()) && id != 0)
-        {
-          GameSession::current()->on_player_added(id);
         }
       }
     }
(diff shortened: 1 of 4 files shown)


swagtoy:
> To be discussed: How should we notify the user that a controller was plugged in and that they can press A to join immediately?

I personally think you should fade in a popup at the bottom of the screen. "Press A to join as a new player". Of course, only do that for _new_ controllers.

You could use a `Notification`, that way people can stop the message from popping up, however, I think something like the popover used for Cutscenes.

src/control/game_controller_manager.cpp
```diff
@@ -48,11 +48,24 @@ GameControllerManager::process_button_event(const SDL_ControllerButtonEvent& ev)
...
+        autobind_controller(controller);
+      }
+      else
+      {
+        return;
+      }
+    }
+
     player_id = it->second;
   }
```
swagtoy, line 74: I don't like that this code is done in it's own scope like this (this makes the `Controller& controller` bit later seen misleading). Maybe it should just passed to another function.

src/control/game_controller_manager.cpp
```diff
@@ -48,11 +48,24 @@ GameControllerManager::process_button_event(const SDL_ControllerButtonEvent& ev)
   int player_id;
 
   {
-    auto it = m_game_controllers.find(SDL_GameControllerFromInstanceID(ev.which));
+    auto* controller = SDL_GameControllerFromInstanceID(ev.which);
```
swagtoy: check if this is NULL and display an error. Also, I don't think we should use auto for stuff like this... Just use SDL_GameController.

src/control/joystick_manager.cpp
```diff
@@ -257,7 +255,7 @@ JoystickManager::process_button_event(const SDL_JoyButtonEvent& jbutton)
   {
     auto i = m_joystick_config.m_joy_button_map.find(std::make_pair(jbutton.which, jbutton.button));
     if (i == m_joystick_config.m_joy_button_map.end()) {
-      log_debug << "Unmapped joybutton " << static_cast<int>(jbutton.button) << " pressed" << std::endl;
+      log_fatal << "Unmapped joybutton " << static_cast<int>(jbutton.button) << " pressed" << std::endl;
```
swagtoy: You sure that's fatal?

src/control/game_controller_manager.cpp
Semphriss: The scope was already there before this PR. How would you like me to organize it?

src/control/game_controller_manager.cpp
Semphriss: This was part of the original code, I only grabbed the intermediary value. Would you like me to add the check?

src/control/joystick_manager.cpp
```diff
@@ -82,24 +82,22 @@ JoystickManager::on_joystick_added(int joystick_index)
 
     if (!parent->m_use_game_controller && g_config->multiplayer_auto_manage_players)
     {
-      int id = parent->get_num_users();
+      // TODO: Display a popup stating that a controller was connected and that the player may joind by pressing A
```
MatusGuy: I actually believe that instead of pressing A, the user should be able to press any button.
This is because not all non-gamecontroller compliant joysticks have labeled buttons.

Semphriss:
Don't merge/review, I mistakenly committed stuff I shouldn't have.

Edit: This is fixed now.

swagtoy:
Request for review when you're ready!

Semphriss:
@swagtoy I haven't yet addressed the two discussions that are still open since I would need clarification before continuing, not sure if they still apply

swagtoy:
Ope... mised that... lot on my plate at the moment, ill return soon

src/control/game_controller_manager.cpp
```diff
@@ -229,25 +242,22 @@ GameControllerManager::on_controller_added(int joystick_index)
 
       if (m_parent->m_use_game_controller && g_config->multiplayer_auto_manage_players)
       {
-        int id = m_parent->get_num_users();
+        // TODO: Display a popup stating that a controller was connected and that the player may joind by pressing A
+
+        // If a player already exists, bind it immediately without waiting for the user to press A.
```
MatusGuy: These comments seem outdated. There is no longer a check for `Control::JUMP`.

MatusGuy (changes requested):

src/control/game_controller_manager.cpp
Semphriss: I've also deleted the same comment in the game controller counterpart. I don't plan to show a popup anymore, since pressing any button seems intuitive enough. This is still open for discussion, though.

src/control/game_controller_manager.cpp
swagtoy: Move it to a function if possible

src/control/game_controller_manager.cpp
swagtoy: Sure. If you are reading code and see areas you think could be improved, don't be afraid to touch it up :)

src/control/game_controller_manager.cpp
Semphriss: I'm not sure which block could be moved to a new function, as there is a lot of code that would be needed for both functions. The nested `return` must interrupt both functions, and fetching the controller need to be done to obtain the player ID. If I were to make it into a separate function, there would be a lot of duplicated code, or a tight coupling of the two functions which may take away the purpose of making a new function.

Alternatively, I can rename the `controller` variable to something like `sdl_controller` and remove the block scope.

PR #3538, opened 2026-01-05
Fix linting (Remove unused variable)

Semphriss:
This is the easiest fix for that issue; I'm not sure if the variable was an accidental leftover or an intended reminder for somebody, but just in case it's the former, I took the initiative to fix it so that the CI gives green checks everywhere again :)

### DIFF
--- a/src/supertux/main.cpp
+++ b/src/supertux/main.cpp
@@ -636,7 +636,6 @@ Main::launch_game(const CommandLineArguments& args)
       // we have a normal path specified at commandline, not a physfs path.
       // So we simply mount that path here...
       std::string dir = FileSystem::dirname(start_level);
-      const std::string filename = FileSystem::basename(start_level);
       const std::string fileProtocol = "file://";
       const std::string::size_type position = dir.find(fileProtocol);
       if (position != std::string::npos) {


swagtoy:
I actually wrote in the commit (very recently) where i fixed starting the level again from cmd line that i did not verify unused variables. I need to give that another peep because I havent tested file:// syntax. That whole bit of code is weird.

swagtoy:
Also since i usually just rebase 1-off commits please provide a commit prefix like "main: Remove unused variable"

Semphriss:
I'm not familiar with commit prefixes, are there instructions somewhere about how to determine what to write?

swagtoy:
For ST It's kind of a hidden rule... tbh, but it's common among many open source projects. It should be declared somewhere soon. Vankata and I do it, some people don't. It just makes it so much easier upon reading the commit message to immediately see what's being changed. Like "Fix slopes for player being wrong" vs "player: Fix broken slopes". Usually you just reflect on what's changed, like `cmake: `, `CI: `, `main: `, `video: `, `player: `, `spikey: `, etc etc. You can check git logs per a file if you are unsure what gets put.

PR #3546, opened 2026-01-06
Fix crash when activating 'Show collision rects' in debug menu

Brockengespenst:
When activating 'Show collision rects' from the debug menu while Editor is not active, the game crashed with a nullptr access. This is fixed by separating the check for Editor active and the show collision rects flag from the debug menu.

### DIFF
--- a/src/object/draggable_region.cpp
+++ b/src/object/draggable_region.cpp
@@ -38,15 +38,15 @@ DraggableRegion::DraggableRegion(Color color, const ReaderMapping &reader) :
 void
 DraggableRegion::draw_draggable_box(DrawingContext& context)
 {
-  if (!Editor::is_active() && !g_debug.show_collision_rects)
-    return;
-
   Rectf& box = m_col.m_bbox;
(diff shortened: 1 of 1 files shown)


src/object/draggable_region.cpp
```diff
@@ -38,15 +38,14 @@ DraggableRegion::DraggableRegion(Color color, const ReaderMapping &reader) :
...
-    return;
-
   Rectf& box = m_col.m_bbox;
-  if (Editor::current()->get_draggables_visible())
+  if (Editor::is_active() && Editor::current()->get_draggables_visible())
   {
     context.color().draw_filled_rect(box, m_color, 0.0f, LAYER_OBJECTS);
     return;
   }
+  else if (!g_debug.show_collision_rects)
```
tobbi: Since you return from the previous statement, you actually don't need the `else if`. Please add a new line and check for the if statement

Brockengespenst: Done.

Issue #3250, opened 2025-04-23
Migrate from GLEW to libepoxy or GLAD


cryptomilk:
Hi,

we are currently trying to get rid of GLEW in Fedora as it isn't [maintained](https://github.com/nigels-com/glew/issues/416#issuecomment-2646570907) anymore. Also GLEW with the last release only supports GLX and not EGL.

There are two ways moving forward:
* Use [libepoxy](https://github.com/anholt/libepoxy)
* Use [GLAD](https://github.com/Dav1dde/glad/), or check out the [webservice for glad2](https://glad.sh/) to generate the files you need!

With both of them migration should be straight forward.

Thanks for considering!

monzonit3:
i would suggest to use in-tree glad and passing SDL_GL_GetProcAddress as loader, it would work on all platforms and on Linux SDL decides whether to use GLX or EGL

swagtoy:
We have decided to go with libepoxy. However I had some concerns regarding GLX for it, so it's in a PR for now and I could use some testers. I also dont really like GLAD that much.

Also fedora maintainers please if you are reading this, consider reverting sdl2-compat because it's causing us real issues right now, kthx (and it still has a fair amass of issues as it stands currently anyway)

swagtoy:
Note that you'll need to specify `-DUSE_GL_LIBRARY=libepoxy`

PR #3551, opened 2026-01-08
libepoxy support for GL bindings

swagtoy:
I have no clue if glx will work but I don't think anyone cares about glx these days...

Closes: #3250

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -231,26 +231,45 @@ if(ENABLE_DISCORD)
   add_subdirectory(external/discord-sdk EXCLUDE_FROM_ALL)
 endif()
 
+set(USE_GL_LIBRARY "glew" CACHE STRING "Which OpenGL function library to choose from (e.g. glew, libepoxy)")
+set_property(CACHE USE_GL_LIBRARY PROPERTY STRINGS glew libepoxy)
+
 set(HAVE_OPENGL NO)
+
+set(HAVE_GLEW NO)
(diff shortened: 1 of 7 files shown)


monzonit3:
if you want to force egl no matter the platform backend it would be better to do SDL_HINT_VIDEO_X11_FORCE_EGL on linux

swagtoy:
Good idea... however,i really dont want to do that

when i tried using the glx headers for libepoxy it had a bunch of xlib typedef conflicts (thanks xlib), i couldve defined them at least but that felt hack, so for now i uust stubbed that out. I feel like things should just work but the libepoxy docs say to include that... i don't know what to do here and i don't feel like digging in the libepoxy code right now. ;-/

if i have to resort to that then it's probably fine. that's why i at least kept glew around

but im not sure if anyone uses glx these days or why they would. so its not the end of the world here, but there's a reason i didn't want to make libepoxy default because i had no clue how to solve this issue

monzonit3:
i've found a gl loader called glbinding which uses c++ and not hacky c macros and it's possible to provide the gl function loader so the game can rely on sdl deciding when to use glx or egl

swagtoy:
For now I'm gonna make it EGL only and merge this in a bit. GLEW is still the default but this is now provided as an option for a reason.

swagtoy:
merged; github sometimes doesn't consider the PR "merged" when rebasing and idk why but whatever

PR #3555, opened 2026-01-09
Remaster credits.ogg once more

GNUfault:
I don't know if it sounds better but see if it sounds better


biggeryetbetter:
Unfortunately, the piano feels a lot more artificial and the panning of the sounds is uncomfortable to listen to (right channel has too much emphasis). I much prefer the original.

PR #3559, opened 2026-01-10
Remove unused fonts in old *.stf format

oldherl:
Those fonts were used in 0.6.x version and prior.
I believe they are no longer in use, in favor of *.ttf fonts now.


oldherl:
I believe they are not used based on my building and testing on Arch Linux. If they are actually in use on other platforms (e.g. Android?), please disregard this PR, thanks.

MatusGuy:
Hello. They are still here for compat and the timer.

PR #3564, opened 2026-01-11
Add option to allow ignoring the hardcoded player limit

Semphriss:
Since the player limit isn't based on a technical limitation, but rather only on graphical convenience, I added an option to allow users to ignore the limit if they want to play with 5 or more players.

The limit is retained by default, as some HUD elements overflow off the screen (currently, only the item pocket) when too many players are added.

I wrote a note in the option's description about the HUD overflowing.


weluvgoatz:
Tested and works.

PR #3576, opened 2026-01-17
Fix leafshot getting killed instantly after being dispensed on macOS

### LINKED ISSUE #3463: [Bug]: Leafshot in "Bye, bye, Forest" gets squished despite there not being any obstacles present
### SuperTux Version

b64c9dfb92bca012e595aaa1726783de7df39c27

### System Information

MacOS X 12.7.6

### Expected Behavior

Leafshot keeps flying until it hits an actual obstacle.

### Actual Behavior

Leafshot gets squished almost immediately. 

See video: 
https://github.com/user-attachments/assets/c179c02e-9537-45ea-a6b1-411e0491d2be

1. Load level "Bye bye forest"
2. Go through level until you hit the leafshot.
3. See bug

### Additional Information

I've pinpointed this behaviour to commit 71159d59672da201a1186000a50eb3618750868e.

Reverting that commit also causes the bug to disappear, so something is wrong there. If anyone could investigate, that would be great.

Brockengespenst:
Another attempt to fix #3463 as it has been reintroduced with acb76ca02e4dcf943a07cfde52ff4d92be83c8f8.

### DIFF
--- a/src/badguy/kamikazesnowball.cpp
+++ b/src/badguy/kamikazesnowball.cpp
@@ -100,6 +100,7 @@ LeafShot::LeafShot(const ReaderMapping& reader) :
   KamikazeSnowball(reader, "images/creatures/leafshot/leafshot.sprite")
 {
   parse_type(reader);
+  set_action (m_dir, /* loops = */ -1);
 }
 
 void


Brockengespenst:
I have not tested the LeafShot on Linux, but I guess it would have been noticed already if #3463 was present there as well. This is another attempt to fix it as set_action resets the hitbox to the proper size and it works as expected, also with corrupted version of LeafShot.
Maybe this is somehow compiler related, macOS uses clang and Linux builds seem to use g++.

provigz:
Oops, sorry about that! I should have checked why it was in `KamikazeSnowball` constructor to begin with, as it did seem unusual.

Brockengespenst:
No worries, seems like I did not test the corrupted version of LeafShot when making the change back then.

I just did a test on Debian using clang and it is also working there. Really strange why the problem only happens on macOS (throughout different versions and processor architectures).

tobbi (approved):
One-line fix, easy to review.

weluvgoatz (approved):
This at least seems to fix the issue on Android!

weluvgoatz:
3 approves = merging

Issue #3579, opened 2026-01-19
[Bug]: Secret ledge in Mount Crushmore


GreenT67:
### SuperTux Version

0.7.0 beta 1

### System Information

LMDE 7 (Linux) 64-bit

### Expected Behavior

After finding secret area, Tux can jump up to the ledge and exit through the secret exit.

### Actual Behavior

The ledge is one block too high to reach

### Steps To Reproduce Actual Behavior

1. Enter Mount Crushmore level
2. Get to checkpoint
3. Jump on little bomb to get to secret area
4. Attempt to jump to next ledge

### Additional Information

Not sure if this is actually a mistake.  Sorry if I'm just missing something.
<img width="1271" height="792" alt="Image" src="https://github.com/user-attachments/assets/2e757f68-ee77-4c20-8bbf-7a797197329b" />

### Guidelines For Reporting Issues

- [x] I have read <https://github.com/SuperTux/supertux/blob/master/CONTRIBUTING.md#bug-reports>.
- [x] I have verified this isn't an issue that's already been reported.
- [x] I have verified this isn't a [discussion](https://github.com/SuperTux/supertux/discussions), or an issue about a crash or a feature request, but rather an actual bug ─ that is, the game did something not intended.
- [x] I have verified this issue is **not** about wrong translations (use Transifex for those), or anything unsupported (e.g. third-party add-ons).
- [x] In this report, I have only included details about **one** (1) bug.
- [x] If I make a mistake while submitting this report, I agree to use the "Edit" feature to correct it, instead of closing this issue and opening a new one.

squidKid-deluxe:
You can backflip up with air flower:

https://github.com/user-attachments/assets/45f1af92-e017-424a-80ec-497289175597

PR #3583, opened 2026-01-21
Add signs for powerups

Semphriss:
Related to #3579

I've added signs showing various powerups to the tile collection.

Certain areas which require a powerup to be accessible or otherwise function properly can use a sign to indicate to the player that they need a certain powerup before proceeding. In the linked issue, a platform that is slightly too high to be accessible can in fact be accessed with an air flower; assuming that this is the intended way to pass this point, a sign can let players know that they need the powerup to reach the ledge.

I've selected a variety of powerups to put on signs according to my personal common sense, but I can add or remove some if wanted.

As usual, the PR should be considered as a starting point in the discussion, not as a request to implement this solution specifically.

Tus standing on a ledge on one side, with another ledge on the other side that is just slightly too high for a normal Tux to reach, but low enough to be reachable with an air powerup. A sign on the ledge shows an air flower.


Sevenberry:
not really sure why these need to be tiles, like given how these have very niche usage it would make more sense for them to just be decals. There are more useful signs that are currently just decals so like this feels a bit like bloat

PR #3584, opened 2026-01-22
Give the Teleporter object a default sprite

Semphriss:
The MovingSprite class expects a non-null `m_sprite` at all times; the worldmap Teleporter would set it to null if its sprite was not found.

The behavior has been replaced by loading the default sprite instead.

The buggy behavior can be observed by attempting to run `./supertux2 --edit-level levels/world1/worldmap.stwm`. It will crash with a segfault somewhere within the MovingSprite class.

---

This may or may not be the desired fix; I did not find any visible difference in practice, and a missing teleporter sprite doesn't sound like something that should happen in practice, but if needed, I can replace the sprite with an invisible sprite or something else.

---

While fixing this, I added a bunch of assertions in the code to check for unexpected conditions (such as a MovingSprite having a null sprite). Assertions are practical because they allow catching bugs early, but aren't compiled in Release builds, and therefore won't affect the end users' experience. Would it be worth pushing the assertions I wrote in a separate PR?

### DIFF
--- a/src/worldmap/teleporter.cpp
+++ b/src/worldmap/teleporter.cpp
@@ -17,6 +17,7 @@
 
 #include "worldmap/teleporter.hpp"
 
+#include "sprite/sprite_manager.hpp"
 #include "util/reader_mapping.hpp"
 
 namespace worldmap {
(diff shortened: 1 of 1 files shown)


swagtoy:
This is not the correct fix. `Editor::is_active()` is being reported as false when using --edit-level :/ This should not be happening and is causing this bug...

> While fixing this, I added a bunch of assertions in the code to check for unexpected conditions (such as a MovingSprite having a null sprite).

Feel free. I want more assertions. Go ahead and push those to master if you want.

swagtoy:
Fixed ^^ Thanks however for bringing this to my attention!

PR #3587, opened 2026-01-22
Update manual page

Semphriss:
Some options were removed since the last release; I adapted the manpage accordingly.


swagtoy (approved):
would probably be nice if it was an .in file actually, so we could feed in the version that way and not have to worry about bumping this

swagtoy:
we can do it l8r

PR #3588, opened 2026-01-22
Support relative paths from the CLI

Semphriss:
Providing an absolute path or a path relative to the data directory works, but relative paths to the CWD weren't appreciated by PhysFS if they contain `..`. Canonizing the paths helps.

### DIFF
--- a/src/supertux/main.cpp
+++ b/src/supertux/main.cpp
@@ -637,7 +637,13 @@ Main::launch_game(const CommandLineArguments& args)
     for(auto start_level : args.filenames)
     {
       // PhysFS doesn't like relative paths
+#ifdef WIN32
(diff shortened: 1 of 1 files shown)


tobbi:
Thank you! That bothered me quite a bit!

tobbi:
@Semphriss Can you check the Windows build failures?

Semphriss:
It seems like MSVC does not have a cast operator to the string type, despite being in the C++ standard: https://en.cppreference.com/w/cpp/filesystem/path/native.html

I'll add an explicit call to `native()` and see if that works.

Semphriss:
My initial assessment was incorrect; MSVC (presumably) supports it well, but it returns a wstring rather than a string, so I'll have to do the conversion manually.

swagtoy (approved):

PR #3591, opened 2026-01-24
Fix worldmap secrets from other end of path

MatusGuy:
This PR doesn't bring the rocks back, but instead it brings the fading paths back. This is an easier way of clarifying which paths are open and which ones are not.

It also fixes some secrets in Rooted Forest.

### DIFF
--- a/data/levels/world2/worldmap.nut
+++ b/data/levels/world2/worldmap.nut
@@ -102,42 +102,34 @@ if (state.underground_f && !state.corrupted) {
 //   ROAD FORKS
 // ============================================================================
 
-if (!("crushmore_secret" in state))
-{
-  // there is no compat... we ignore the old states for now. if they beat the
-  // level or didn't complete the fork in the road then things would just get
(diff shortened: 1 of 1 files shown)


Semphriss:
I support the idea :+1: 

Design recommendation: I would advise making it invisible entirely; since that type of path isn't present anywhere else in the worldmap, it may look like the path looking soft is its natural state, and people may keep attempting to walk it without understanding that it is meant to look locked.

Additionally, completing the secret level exits unlock the secret paths, but not the main path, with no visual cue about this. This doesn't need to be fixed with this PR, but I felt it was relevant to mention it.

swagtoy:
> [WARNING] /home/swags/supertux/src/squirrel/squirrel_environment.cpp:129 Squirrel exception: Error running script! (Squirrel exception: Error running script! (the index 'drop_secret' does not exist))

i just spawned in the forest and saw this... :v i was probably already there to begin with

I think i noticed it working, but i didnt test much further, didnt have time, but it seemed to look right

PR #3592, opened 2026-01-24
Fix OpenAL headers

Semphriss:
Emscripten builds needed to hold a copy of the OpenAL header files because the path given to the include directives is non-standard.

I changed the paths to use the more portable version (concretely, `al.h` -> `AL/al.h`), which spares an additional include path to the compiler.

I paid close attention to not mess with other platforms, but I'll be using the CI as an indicator just to be sure.


src/audio/openal_sound_source.hpp
```diff
@@ -16,7 +16,7 @@
 
 #pragma once
 
-#include <al.h>
```
swagtoy, line 19: Ehh... Not so sure about this one. This now effectively depends on OpenAL being in /usr/include. While that's a given, if the user has a custom build they need to use, or they installed OpenAL elsewhere (like /opt or something...), well, pkg-config is utterly usage in this scenario. See the output of `pkg-config --cflags openal` at least on my machine:

`-I/usr/include/AL`

Note that it is already including AL here. It's unlikely this will ever be a problem, but I still will not approve of this.

tobbi: OpenAL's own example includes `AL.h` like this. See https://github.com/kcat/openal-soft/blob/master/tests/example.t.cpp#L2

Semphriss: OpenAL headers will always be in a folder called `AL`, and any half-decent package system will ensure that the folder containing the `AL` folder is in the build path. On standard installs, that would be `/usr/include`; presumably, `pkg-config` doesn't include it because it's already in the include list. Chances are that if if weren't in the include path already, it would be included.

It's also notable that SuperTux does not use pkg-config to find OpenAL, but rather uses CMake's own package finding mechanism, which does include the correct folder.

FWIW, SDL had the same issue with SDL2 where the subfolder may or may not be part of the include directive; in SDL3, this was uniformized to include the `SDL3` subfolder, as this was the most portable option. OpenAL-soft seems to have taken the same direction.

Finally, the CI seems to successfully compile for all platforms and configurations, including Windows and Android, so I don't believe this is going to be an issue practically speaking. If you know one build config that doesn't support including OpenAL this way, and that config happens in practice, then it may be worth looking into adding it to the CI, and checking that other dependencies, especially SDL, don't have the same issue.

swagtoy: > any half-decent package system will ensure that the folder containing the AL folder is in the build path.

No it won't! `/usr/include` is typically just included by default when building because that's usually where lots of headers themselves are bundled in. People just typically take advantage of that for most scenarios. You already claimed to switch this to a 'more portable' solution, but you did the opposite... if you want portability, you'd account for what pkg-config, or the CMake does, and just leave that alone.

And for the record, the add_package _can_ fall back to pkg-config, which is actually the more likely scenario where someone was to provide their own AL header here. In such a case, you're lucky the vcpkg CI's passed actually, because vcpkg includes its own little "include" folder when doing all the headers (and probably its own stuff to account for this scenario) so it eliminates this kind of problem. In a more custom setup, this isn't always likely.

> Finally, the CI seems to successfully compile for all platforms and configurations, including Windows and Android

You are missing the point entirely. This isn't about CI passing at all. CI passing is not an indicator that, say, libepoxy works, or a build without curl works. It is an indicator that the game builds on one specific distro.

> in SDL3, this was uniformized to include the SDL3 subfolder, as this was the most portable option. OpenAL-soft seems to have taken the same direction.

Can you show me for both of these where this is actually indicated? Because afaik nothing has changed in this regard here.

---

I'm not even sure why you did this change to begin with. What was wrong with how it was before? Was Emscripten breaking? There wasn't much clarification, so I'm only curious, because it was perfectly fine the way it was before.

Semphriss: Please, let's de-escalate the conversation. My contributions aren't meant to be confrontational, and it's difficult to remain productive when little change sparks such responses. I'll try to answer your concerns to the best of my abilities; if you have further concerns after that, do feel free to express them, but please try to remain technical.

---

> > any half-decent package system will ensure that the folder containing the AL folder is in the build path.
>
> No it won't! /usr/include is typically just included by default when building because that's usually where lots of headers themselves are bundled in. People just typically take advantage of that for most scenarios.

My personal experience tends to show the opposite, actually. I see most often the `#include <AL/al.h>` format. STK is one notable project using this format.

> You already claimed to switch this to a 'more portable' solution, but you did the opposite... if you want portability, you'd account for what pkg-config, or the CMake does, and just leave that alone.

This sounds quite confrontational, and doesn't help with the discussion.

> In such a case, you're lucky the vcpkg CI's passed actually,

I don't think this is as matter of luck, I believe this to be intended to work, or at least that's what my experience tends to show. Do you have a reason to believe it isn't supposed to work normally?

> CI passing is not an indicator that, say, [...] a build without curl works.

The CI's very purpose is specifically to make sure that SuperTux builds properly and works on all supported configurations, hence my suggestion that we add those to the CI if we do indeed support them.

> Can you show me for both of these where this is actually indicated? Because afaik nothing has changed in this regard here.

- SDL mentions the usage explicitly [here](https://github.com/libsdl-org/SDL/blob/main/docs/README-migration.md):
  > SDL headers should now be included as #include <SDL3/SDL.h>. [...]
- SDL also provides a [hello world program](https://github.com/libsdl-org/SDL/blob/main/docs/hello.c) with that include path.
- OpenAL-soft (which SuperTux ships on all platforms) does not have extensive documentation like SDL does, but they do have tests programs with that include path, as Tobbi linked. They also notably have a seemingly needless `AL` subfolder in their `include/` folder; I'm not sure why everything couldn't just be put in the include directory directly if the path chuck wasn't necessary.

> I'm not even sure why you did this change to begin with. What was wrong with how it was before? Was Emscripten breaking?

As the PR's diff shows, SuperTux had to ship OpenAL headers manually for Emscripten builds to work, because the headers couldn't be found. I did it that way back then because I didn't understand the issue; now I do, and I wish to correct my mistakes and remove unnecessary files from the source tree so that we don't have to maintain then in the long term.

---

I'm glad to help with a project and I'm willing to adjust my contributions as needed, but I would prefer to do without strong assertions. It's demotivating to read, and it gets in the way of an otherwise productive discussion. 

It would be better to discuss what lead you to make those statements; is there a particular build configuration that you are concerned about that fails with those changes? If there is one, having it in the discussion would help greatly. I try to take care not to introduce bugs, and I can admit without problem that I'm fallible, but it would help a lot to have a concrete example to work with.

swagtoy: I'm not trying to be confrontational. I'm explaining to you why I disagree with this change. I do not see the purpose in such a change other than "it's what other projects are doing." There's no need to get defensive on this.

> it's difficult to remain productive when little change sparks such responses

I'm sorry...! But to me, this was already going very productively... since we are in somewhat of a misunderstanding :) You have to brace yourself for little changes to spark discussions. This is the nature of contributing, and not everything is going to just get accepted. In fact, you should be fortunate to even get any discussion, as you could have your PR ignored for years and then closed by some stale bot... be graceful that I am at least taking the time to review this stuff, becuase you came in full swing and contributed a lot of thigns, and that's overwhelming to me, myself, and i've been trying to look at stuff even with the little motivation i've had...

> I see most often the #include <AL/al.h> format.

Well, it's wrong in my opinion. Even in the case where tobbi brought up the test case, welll, you pretty much proved why that's being done, because they're including their own `include` directory and including it that way. Test cases work in a manner that they shouldn't be installed at the end result. Now if their pkg-config was including the actual `/usr/include` dir, then i'd be fine with that... but they're not.

> I don't think this is as matter of luck, I believe this to be intended to work

Indeed so, but that's specifically becase vcpkg includes the entire include dir already which contains the OpenAL headers. If it didn't, then that would've result in stuff breaking, but pretty much everything you install for vcpkg goes in it's own folder already. This is the most portable approach.

> The CI's very purpose is specifically to make sure that SuperTux builds properly and works on all supported configurations

But... it's not. You have to remember that linux distributions will all also package things themselves. Very often, they have to apply patches for the game, and sometimes i'll backport them or look at their bugs and fix things myself if I see them as useful. CI is not a true indicator of the game building properly, it's an indicator of it building properly for the specific configurations we chose.

> and I wish to correct my mistakes and remove unnecessary files from the source tree so that we don't have to maintain then in the long term.

And that's fine... I was okay with that part, i just didn't see a reason for you to change how things were being included. It was working before, and it's already flexible in the scenario that I described.

> is there a particular build configuration that you are concerned about that fails with those changes?

I already explained it. People can have their own locally installed builds in `/opt` and then have their pkg-config dirs to point there. This would not work unless they manually slipped in the dir like `/opt/include` to the cflags.

swagtoy: And im sorry if you think im demotivating you :( its not my intention at all...

swagtoy:
I didn't get a followup, so I at least cherry picked the parts here that work for now.

PR #3600, opened 2026-02-01
video: Switch DrawingRequest to a variant

swagtoy:
This switches DrawingRequest to a variant.

This isn't a mindboggling change in terms of performance, but it better opens the doors for such things. We can probably look further into how obstack is used and stop using an std::vector<DrawingRequest*>. For now, let's get this out of the way.


swagtoy:
merged

PR #3513, opened 2026-01-02
try fix freebsd build

1111Windows:
This PR attempts to fix the BSD build.

Currently, only compiling SuperTux and testing its executability have been implemented.

I found that using the -DWERROR=ON option on FreeBSD causes compilation failures, so -DWERROR is disabled.

The current issue is: compilation is too slow.

PR #3603, opened 2026-02-01
Optimizing FreeBSD Build

1111Windows:
This PR is an improvement following #3513, and its changes are as follows:

1. Copy dependencies to the lib folder.

2.Compress files directly in the build/out/usr folder to avoid requiring users to navigate to the out/usr folder when decompressing the files.

3.Resolve the issue: ld-elf.so.1: /usr/local/lib/libpcre2-8.so.0: version PCRE2_10.47 required by /usr/local/bin/git not defined.

4.Pass parameters to the program when running run.sh (if any).


1111Windows:
I've taken care of the rest, so we can merge now.

.github/workflows/freebsd.yml
```diff
@@ -76,19 +77,29 @@ jobs:
...
             chmod +x run.sh
+            cd lib
+            cp /usr/local/lib/libopenal.so ./
+            cp /usr/local/lib/libSDL2_image-2.0.so ./
+            cp /usr/local/lib/libphysfs.so ./
+            cp /usr/local/lib/libGLEW.so ./
+            ln libopenal.so libopenal.so.1
+            ln libGLEW.so libGLEW.so.2
+            ln libphysfs.so libphysfs.so.1
+            ln libSDL2_image-2.0.so libSDL2_image-2.0.so.0
```
swagtoy: I do not think this is a good idea, actually. These libraries are immune to patches/modifications from the ports tree, as we ll as any patches/bugfixes. Esp for -CURRENT, well, freebsd abis can sometimes break, have had that happen before. I understand what your goal here is though.

It's akin to dropping in DLL's alongside the exe, but in the context of (Free)BSD where sometimes these could be patched, changed, updated.... I do not think this model is ideal.

What do you think? Yay? Nay? I havent used freebsd in a long time, but I think maybe you could run `pkg info XXX 2>&1 > /dev/null` then check the return code? People can install these themselves; I do not see the reason to bundle these in.

.github/workflows/freebsd.yml
```diff
@@ -76,19 +77,29 @@ jobs:
...
           BUILD_TYPE: ${{ matrix.build_type }}
           freebsd_release: ${{ matrix.freebsd_release }}
         run: |
-            cd $GITHUB_WORKSPACE
-            cd build
+            cd $GITHUB_WORKSPACE/build
             mkdir upload
-            tar -czvf upload/SuperTux-FreeBSD-${{ matrix.freebsd_release }}-${{ matrix.build_type }}.tar.gz out/usr
+            cd out
+            mv usr supertux
```
swagtoy, line 110: is this all of `/usr`???

swagtoy (changes requested):

.github/workflows/freebsd.yml
swagtoy, line 116: could you insert `-FreeBSD` into this? Or did it do that already?

I think I did `SuperTux-v0.7.0-beta.2-FreeBSD-14.3-Release.tar.gz` manually for beta 2. I downloaded a .tar.gz and it's just SuperTux-freebsd-amd64-14.3-Release-tgz.zip

swagtoy (changes requested):

.github/workflows/freebsd.yml
1111Windows: > I think I did `SuperTux-v0.7.0-beta.2-FreeBSD-14.3-Release.tar.gz` manually for beta 2. I downloaded a .tar.gz and it's just SuperTux-freebsd-amd64-14.3-Release-tgz.zip
>

You need to decompress it. After decompressing, you will see tar.gz

> could you insert `-FreeBSD` into this? Or did it do that already?
> 

It has already done that.

.github/workflows/freebsd.yml
1111Windows: > is this all of `/usr`???

No, this is just the installation folder for SuperTux within the FreeBSD VM.

.github/workflows/freebsd.yml
1111Windows: Sometimes, these libraries may have changed their filenames due to updates, but the build version requires the old filenames. If users manually install these libraries, they will complain about missing files
(though library filename updates require CI updates).

.github/workflows/freebsd.yml
```diff
@@ -76,30 +77,41 @@ jobs:
             mkdir out
             make install DESTDIR="./out" VERBOSE=1
             cd out/usr
-            echo -e '#!/bin/sh\n' './bin/supertux2' >> run.sh
+            echo -e '#!/bin/sh\n' 'export LD_LIBRARY_PATH=./lib/:$LD_LIBRARY_PATH\n' './bin/supertux2 "$@"' >> run.sh
```
swagtoy: - we should probably still prefer LD_LIBRARY_PATH first, so reorder it to `LD_LIBRARY_PATH="$LD_LIBRARY_PATH:./lib"`. Wrap it in a string too, like i did there.
- your usage of current paths here will fall apart if the user executes the script outside of the directory, no? try to get the script directory instead using a combo of dirname and $0

swagtoy (changes requested):
Apologies for the delay on reviewing this...

.github/workflows/freebsd.yml
1111Windows: done

1111Windows:
> Apologies for the delay on reviewing this...

No worries, I was thinking about whether I should let you know on Matrix when I commit today (I've actually been putting this PR on hold for a while now, haha)

.github/workflows/freebsd.yml
```diff
@@ -77,7 +77,7 @@ jobs:
             mkdir out
             make install DESTDIR="./out" VERBOSE=1
             cd out/usr
-            echo -e '#!/bin/sh\n' 'export LD_LIBRARY_PATH=./lib/:$LD_LIBRARY_PATH\n' './bin/supertux2 "$@"' >> run.sh
+            echo -e '#!/bin/sh\nexport DIRNAME="$(dirname "$(readlink -f "$0")")"\nexport SYSTEM_LD_LIBRARY_PATH="$LD_LIBRARY_PATH"\ncd "$DIRNAME"\nexport LD_LIBRARY_PATH="$LD_LIBRARY_PATH:$DIRNAME/lib/"\n"$DIRNAME/bin/supertux2" "$@"' > run.sh
```
swagtoy: nit: go back to space separating the \n via separate echo arguments

see if you can do \\
`        ` this to make those multiline. It \\
`        ` just aids in readibility :)

1111Windows: done

PR #3613, opened 2026-02-04
Add a toggle for screen shake effects

### LINKED ISSUE #3609: [Feature Request]: Allow turning screen shake off
### Feature Details

Add a setting to completely disable the screen shake effect.

### Feature Purpose

Accessibility. I find the effect headache inducing and slightly disorienting, especially when played for a long time time during gameplay, like in Collapse Imminent.

### Concept Screenshots

_No response_

v1993:
Fixes #3609.

Marking as draft since I've realized it might be a good idea to make it tri-state instead:

* Full (default): both one-time shakes (e.g. from bombs) and "earthquakes" (persistent shaking) enabled
* Reduced: only one-time shakes enabled
* Off: both one-time shakes and "earthquakes" disabled

Should I go for that? If so, should I bother with an enumeration (like video subsystem selection) or would a simple `int` (like for vsync) do?

### DIFF
--- a/src/object/camera.cpp
+++ b/src/object/camera.cpp
@@ -248,7 +248,7 @@ Camera::reset(const Vector& tuxpos)
 void
 Camera::shake(float duration, float x, float y)
 {
-  if (!g_config->screen_shake)
+  if (g_config->screen_shake_mode == Config::ScreenShakeMode::OFF)
     return;
 
   m_shaketimer.start(duration*100.f);
@@ -260,7 +260,7 @@ Camera::shake(float duration, float x, float y)
 void
 Camera::start_earthquake(float strength, float delay)
 {
-  if (!g_config->screen_shake) {
+  if (g_config->screen_shake_mode != Config::ScreenShakeMode::FULL) {
     return;
   }
   if (strength <= 0.f)
@@ -433,7 +433,7 @@ Camera::update_shake()
 void
 Camera::update_earthquake()
 {
-  if (m_earthquake && !g_config->screen_shake)
+  if (m_earthquake && g_config->screen_shake_mode != Config::ScreenShakeMode::FULL)
     stop_earthquake();
 
   if (!m_earthquake)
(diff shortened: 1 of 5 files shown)


swagtoy:
> Should I go for that? 

Yes

> should I bother with an enumeration

For tri-state, yes. It does no harm

src/object/camera.cpp
```diff
@@ -427,6 +433,9 @@ Camera::update_shake()
 void
 Camera::update_earthquake()
 {
+  if (m_earthquake && !g_config->screen_shake)
```
swagtoy: Earthquake starts -> user goes and disabled screen shake. Now what? In fact you should probably make the option stop any on-going earthquakes somehow, if possible, but that could involve some rewiring.... sooooo at the very minimum, and to keep this contribution low-cost, you can just stop the earthquake in any scenario for now and ignore that edge case. :)

v1993: I'm not sure I understand? This function is responsible for advancing state of the earthquake every game frame and I've included this bit of code explicitly to handle the scenario you're talking about - to stop an ongoing earthquake if it's toggled off.

swagtoy: Oh... shit... yeah the case I described was specifically _what_ you were handling here. I just misread the function. Pardon me and carry on.... :p

v1993:
Changed to off/reduced/full selection as proposed, seems to work well in testing. Using an enum means that there are three formats to convert between (selection number <-> enum <-> string), which is a bit unwieldy, but ultimately works okay.

swagtoy:
I'll review tomorrow, ping if I forget. Thank you very much for your work!

v1993:
@swagtoy Ping! Also, a look at #3540 would be appreciated as well.

swagtoy:
I'm really sorry. will get back to this soon

swagtoy:
Thanks! o7

PR #3635, opened 2026-02-07
Multicolored Pipes

Sevenberry:
Adds six new variations on pipes and moves them to their own category in the editor.
NOTE: adds the world 3 pink pipes (with permission from fillip)


serano01:
Why so much? For worlds 3-8?

Frostwithasideofsalt:
there isnt any worlds past 4- its just to give users more options

MrGaming-18113:
why not making white pipes tho? so that way people can choose the color

MrGaming-18113:
> why not making white pipes tho? so that way people can choose the color

oh wait, there already are white pipes, sorry

HybridDog:
Instead of adding copies of some of the tiles with only the colour changed, would it be possible to add lookup table support to layers?

https://dev.epicgames.com/documentation/en-us/unreal-engine/using-lookup-tables-luts-for-color-grading?application_version=4.27

PR #3657, opened 2026-02-12
Fix percentage counter overflow when downloading add-ons

Narre:
This fixes the percentage showing nonsense when reaching about 21 MB of downloaded data. This fix makes the entire calculation happen in floats. Alternatively, leaving the 100 as an int but putting the division into parentheses and turning that into floats also works. This looks cleaner, the other option would technically be more precise but there really is no need for atomic precision here.


Narre:
For clarification, this is the bug in question. Look at the percentage counter around 21 MB / 72 %.

https://github.com/user-attachments/assets/c63dabae-a663-416e-9833-d4bc6bcfc238

swagtoy:
Thanks! :)

PR #3692, opened 2026-02-16
Remaster `leveldone.ogg`

GNUfault:
I have remastered `leveldone.ogg` as it's like 20 years old and sounds super out of place


ghost:
> I'm sorry but this does not sound like a suitable replacement at all. The notes are mostly wrong and the sample sounds really poor, way more empty than the original. Overall it just feels empty and uncomfortable.

Ok let me use the original one and remaster that instead of making a whole new one

ghost:
I have used the same module file that the old supertux used and export it as 16bit (as the samples are 16bit)

ghost:
it sounds very slightly better

ghost:
wait i should change the name from ogg to wav so it will work lmao

ghost:
it works now!

GNUfault (commented):
I have used the same module file that the old supertux used and export it as 16bit (as the samples are 16bit)

PR #3713, opened 2026-02-20
Wind improvements

### LINKED ISSUE #3676: [Bug]: Rooted Forest: skippable windpipe in Three Sheets to the Wind
### SuperTux Version

supertux2 dev v0.7.0-beta.2 - 7125492 (master)

### System Information

Linux Mint 22.3 - Cinnamon 64-bit

### Expected Behavior

A bug can not be exploited to skip a significant part of the level.

### Actual Behavior

The green pentagon windpipe can be skipped with backflips.

https://github.com/user-attachments/assets/b43afc37-e334-447b-be3b-cafeb4bbc6d3

1. Go to the level Three Sheets to the Wind (Rooted Forest).
2. Get onto the platform next to the horizontal green pentagon windpipe.
3. Perform a series of backflips until you get past the pipe, as in the video.

### Additional Information

As far as I can tell, the issue of backflips being ignored by horizontal wind is indirectly mentioned in this comment: https://github.com/SuperTux/supertux/issues/2592#issuecomment-2924970283, so this issue focuses on the odd skip caused by the bug rather than the bug itself.

weluvgoatz:
This PR adds 2 new types of wind: Wind (Additive) and Current (Additive).  The reason I added new types instead of replacing the old is because the new wind types conflict with levels using currents + Snow Bowling, but they work similarly in other wind applications.  What these new types do is basically add wind to Tux's current velocity instead of attempting to override it like the current wind type does; however, this rule is broken when the wind tries to push Tux up because otherwise Tux's growing y-velocity will outgrow the constant wind-speed, and thus the wind will never work.  

Edit: A video of the new currents:

https://youtu.be/UBqUo8NTA-0

Fixes #3676

### DIFF
--- a/src/object/wind.cpp
+++ b/src/object/wind.cpp
@@ -180,7 +180,7 @@ Wind::collision(MovingObject& other, const CollisionHit& )
         player->set_wind_accel(std::min(1.0f, player->get_wind_accel() + (acceleration * dt_sec)));
       }
       player->set_wind_boost(speed * player->get_wind_accel() * Vector(player->on_ground() ? 0.5f : 1.0f, 1.0f));
-      if (player->get_velocity_y() > 0.f && speed.y < 0.f) { // allow wind to push players up that are going down
+      if (player->get_velocity_y() > 0.f && speed.y < 0.f && !player->is_swimming()) { // allow wind to push players up that are going down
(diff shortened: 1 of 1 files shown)


weluvgoatz:
Now that 0.7.0 is out, I think I'd like this PR to be looked at.

PR #3724, opened 2026-03-07
Use the unused `tuxvictorious.ogg` sound instead of `leveldone.ogg`

GNUfault:
Wansti made a remix of the aging `leveldone.ogg` sound in 2005 and the SuperTux team at the time said its too "epic". Personally for modern SuperTux this sound fits the new music way better, the old `leveldone.ogg` sound. I have added the sound, make it use it, extend the time that it is on the leveldone screen, and added it to authors. You can find more information about this forgotten sound [here](https://commons.wikimedia.org/wiki/File:Discarded-ideas-org-wansti-ccbysa-tuxvictorious.ogg).


Frostwithasideofsalt:
Not planned- the original plan was for this to be used for bossfights but that was never added.

PR #3732, opened 2026-03-14
Translator credits update

tobbi:


Sera-L:
thanks
but please remove "Siradj" because he is me (it was an old account)

tobbi:
> thanks but please remove "Siradj" because he is me (it was an old account)

I hope I can remember to not accidentally add this again. We should really find a better way to manage translations.

PR #3736, opened 2026-03-15
Update metainfo

yakushabb:
- Correct developer ID (use RDNS)
- Add missing version information
- Add URLs to release tags

### DIFF
--- a/org.supertuxproject.SuperTux.metainfo.xml
+++ b/org.supertuxproject.SuperTux.metainfo.xml
@@ -43,7 +43,7 @@
       through the Icy Island and beyond.
     </p>
   </description>
-  <developer id="supertuxproject.org">
+  <developer id="org.supertuxproject">
     <name>The SuperTux Team</name>
   </developer>
   <url type="homepage">https://www.supertux.org</url>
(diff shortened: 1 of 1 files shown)


swagtoy:
Thanks

pedropaulosuzuki:
Could 6970d03397801427ab7ebe4740a5cdfc5da8d804 be changed to be the tagged v0.7.0? This way downstreams won't have to use patch files to fix the metainfo.xml

swagtoy:
@pedropaulosuzuki Not sure I want to do that, I dislike retagging.

If any flatpak maintainers need help carrying the patches down, thats fine, let me know, but 0.7.1 will come out soon anyway, so I am uninterested. I'm sorry for the inconvenience :(

swagtoy:
On second thought ... i could probably create a "revision" tag for distributors, yeah? How's that?

pedropaulosuzuki:
> On second thought ... i could probably create a "revision" tag for distributors, yeah? How's that?

Looks like a good compromise. I agree changing tags is generally bad practice, but as this was released just today, it would be changed before packagers do the packaging, but I understand the rationale.

The revision tag may be a good idea, but then it's one more tag in the repo forever, so it v0.7.1 is coming soon, then it's not the worst of problems to have a temporary patch file on downstreams. I was mostly worried because the last updated version of SuperTux on Flathub is over a year ago, so I was afraid things would be there for a long time this time as well.

ghost:
You don't need to create a tag. We can use any commit.

swagtoy:
I'm still okay creating a revision tag regardless. No harm :)

PR #3745, opened 2026-03-17
Fix #3569: Fix Tux not turning around in 1-block gaps

### LINKED ISSUE #3569: [Bug]: Tux does not visually turn around when in a one-block gap
### SuperTux Version

0.7.0 beta 1

### System Information

Linux (LMDE 7), 64-bit (appImage)

### Expected Behavior

Tux turns around normally when in a one block gap after inputting the direction opposite to the direction Tux is facing.

### Actual Behavior

Tux stays facing the direction he is, but immediately turns around after jumping or when idle animations activate.
Additionally, fireballs travel in the direction pressed, even when Tux is facing the other way.

1. Enter a level
2. Find a one block gap with a surface directly beneath it (The one I found this using is at the beginning of "Oh No! More Snowballs!" where it is possible to melt one of the ice blocks on the platform above, creating a one-block gap which Tux can stand in.)
3. Walk into the gap
4. Press the button of the direction opposite to the direction Tux is facing

### Additional Information

_No response_

gvalves2005:
This PR fixes issue #3569, where Tux ignores user directional input when placed in a 1-block gap while the fancy idle animation timer is active.

### Root Cause & Fix
When Tux's horizontal velocity becomes zero, the `draw()` function in `player.cpp` waits for the idle timer to finish before updating the sprite's direction. This patch adds a check during the idle wait state to force an action update if the direction changes, resetting the idle stage to zero.

### Testing
I validated this fix via manual runtime testing. 
- Compiled locally and ran the game.
- Verified Tux turns smoothly in 1-block gaps without getting stuck in the idle animation.

Closes #3569

### DIFF
--- a/src/object/player.cpp
+++ b/src/object/player.cpp
@@ -2238,6 +2238,8 @@ Player::draw(DrawingContext& context)
       const bool is_not_idle = std::all_of(IDLE_STAGES.begin(), IDLE_STAGES.end(),
         [this](const std::string& stage) { return m_sprite->get_action().find("-" + stage + "-") == std::string::npos; });
 
+      std::string base_idle_action = sa_prefix + "-" + IDLE_STAGES[0] + sa_postfix;
+
       if (is_not_idle || (m_should_fancy_idle && !m_fancy_idle_active) || m_reset_action)
       {
         m_idle_stage = 0;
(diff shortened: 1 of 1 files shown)


src/object/player.cpp
```diff
@@ -2270,6 +2270,12 @@ Player::draw(DrawingContext& context)
           else
             set_action(sa_prefix + ("-" + IDLE_STAGES[m_idle_stage]) + sa_postfix, 1);
         }
+        else if(m_sprite->get_action() != sa_prefix + ("-" + IDLE_STAGES[0]) + sa_postfix)
```
tobbi: I haven't tested this, but one small remark:
You're using the action in two places. Can you move it to a variable?

gvalves2005: I moved it to a variable and used it in two other lines that were using the same action.

gvalves2005:
Just noticed I missed an 'S' on IDLE_STAGES which upset the CI. Force-pushed the compilation fix!

gvalves2005:
Hi @tobbi! Hope you're doing well. I just wanted to check in on the status of this PR. Let me know if the current implementation looks good to you, or if I need to adjust anything else.

tobbi:
Thank you! From my side it looks good. I'm checking in with the rest of the team.

One thing I'm not sure about: There's some code duplication and I'm wondering whether that can be moved into its own function. But I'll discuss that with the rest of the team first.

PR #3752, opened 2026-03-22
Fix the Transifex URL (404 error) shown in the credits

jorgesumle:
Just that.

I didn't compile it to test it, but it should work fine, as it's a small text fix.


swagtoy:
Thanks!!

PR #3762, opened 2026-03-26
Fix #3685: Prevent coin collection through solid weak blocks

### LINKED ISSUE #3685: [Bug]: Some coins inside ice blocks can be grabbed without melting
### SuperTux Version

supertux2 dev v0.7.0-beta.2 - 9a291c2 (master)

### System Information

Linux Mint 22.3 - Cinnamon 64-bit

### Expected Behavior

Coins inside ice blocks are grabbable only when they are thawed.

### Actual Behavior

Coins inside ice blocks can be grabbed by dropping next to the ice block.

https://github.com/user-attachments/assets/6ee34202-96ba-40dc-b5e8-6ac9c8033bee

1. Go to the level Oh No! More Snowballs! (Icy Island)
2. Melt one of the ice blocks at the beginning.
3. Walk onto a neighboring ice block.
4. Slowly move toward the hole where the melted ice block was until you drop into the hole.

### Additional Information

_No response_

baltazarReis:
This PR fixes a bug where players could collect coins trapped inside a WeakBlock (like ice blocks) before the block was destroyed.

Added is_solid_block() method to WeakBlock to check its current state.

Modified Coin::collision to prevent collection if the coin's center is inside a solid WeakBlock.

Closes #3685

### DIFF
--- a/src/object/coin.cpp
+++ b/src/object/coin.cpp
@@ -247,15 +247,11 @@ Coin::collision(MovingObject& other, const CollisionHit& )
   // Check if this coin is trapped inside an active WeakBlock
   Vector coin_center = m_col.m_bbox.get_middle();
   for (auto& weak_block : Sector::get().get_objects_by_type<WeakBlock>()) {
-    if (weak_block.is_solid_block()) {
-      const Rectf& block_bbox = weak_block.get_bbox();
-      if (coin_center.x >= block_bbox.get_left() && 
-          coin_center.x <= block_bbox.get_right() &&
-          coin_center.y >= block_bbox.get_top() && 
-          coin_center.y <= block_bbox.get_bottom()) {
-        return ABORT_MOVE;
-      }
-    }
+    if (weak_block.get_group() == COLGROUP_DISABLED)
+      continue;
+      
+    if (weak_block.get_bbox().contains(coin_center)) 
+      return ABORT_MOVE;
   }
   
   if (m_col.get_bbox().overlaps(player->get_bbox().grown(-0.1f)))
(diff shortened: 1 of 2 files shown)


src/object/weak_block.hpp
```diff
@@ -46,6 +46,9 @@ class WeakBlock final : public MovingSprite
   virtual void on_flip(float height) override;
 
   void startBurning();
+  
+  /** Check if this block is still solid (not destroyed) and blocking objects */
+  inline bool is_solid_block() const { return state != STATE_DISINTEGRATING; }
```
MatusGuy: Is this really necessary? Getting this information from the current collision group would be much more accurate and would not require this extra method.


```suggestion
```

src/object/coin.cpp
```diff
@@ -242,6 +243,21 @@ Coin::collision(MovingObject& other, const CollisionHit& )
...
+  for (auto& weak_block : Sector::get().get_objects_by_type<WeakBlock>()) {
+    if (weak_block.is_solid_block()) {
+      const Rectf& block_bbox = weak_block.get_bbox();
+      if (coin_center.x >= block_bbox.get_left() && 
+          coin_center.x <= block_bbox.get_right() &&
+          coin_center.y >= block_bbox.get_top() && 
+          coin_center.y <= block_bbox.get_bottom()) {
+        return ABORT_MOVE;
+      }
+    }
```
MatusGuy: Without that extra method, it becomes easier to write this in a concise manner:
```suggestion
    if (weak_block.get_group() == COLGROUP_DISABLED)
      continue;
      
    if (weak_block.get_bbox().contains(coin_center)) 
      return ABORT_MOVE;
```

MatusGuy (changes requested):

baltazarReis:
You're right, it becomes much simpler. I've applied the changes you requested and tested them, works the same.

PR #2895, opened 2024-04-06
[NOSQUASH] Editor: Replace the OKLab color picker by a 2D color picker

HybridDog:
# Goal

There are reports that the one-dimensional OKLab color selection is unpopular and difficult to use (see [this Discussion](https://github.com/SuperTux/supertux/discussions/2323)),
so we can remove it to simplify the code and GUI.

A two-dimensional image can enable a more convenient selection of colors than a one-dimensional slider.
Under the assumption that users typically want to select a color with the highest saturation,
a two-dimensional color picker where any RGB color with the highest saturation can be slected with one click could be suitable.
For an independent adjustment of lightness and hue, which is considered a good property of a color picker,
the two-dimensional color picker should use a perceptual color space such as OKLab.
As a side effect for choosing a perceptual color space,
a marker showing the lightness and hue of the current color can visualize how a change in RGB values affects the perceived color.

# Proposed changes

* For the color selection in the editor, add the ItemColorPicker2D menu item,
  which displays an image and enables picking a color from it by clicking.
  The image contains the highest-saturation sRGB colors,
  so for the color picker we only need OKLab code to determine marker positions
  and no color clipping code since we can sample pixels from the image.
* Add a method drawing a hexagon to Canvas
* Add a helper method to ColorOKLCh which calculates a modified lightness
* Add a missing include guard and constness to the code for ColorOKLCh
* Remove ItemColorChannelOKLab and most of the OKLab color space code.
  Since the two-dimensional color picker uses some of the OKLab code,
  it is not removed completely.

# Additional information

The author of OKLab has a webpage about a color picker, which uses [a modified lightness](https://bottosson.github.io/posts/colorpicker/#intermission---a-new-lightness-estimate-for-oklab).
I have also used the modified lightness since it leads to smaller dark regions in the color picker image and in SuperTux it is, I assume, uncommon to select dark colors (correct me if I'm wrong).

Color picker image with unmodified lightness:
out_L

Color picker image with modified lightness:
out_Lr

The new color picker with default language and 800x600 window size:
2024-04-06-174514_335x540_scrot

The color picker stretches horizontally; since the language affects the width of the menu title, the color picker's aspect ratio depends on the language. I don't think this is a problem, so I haven't implemented it differently.

I couldn't find a way to get a pixel from a Surface, so the color picker image is loaded twice: one time for the drawing and one time to sample pixel values from it.

Please test the new color picker.

PR #3764, opened 2026-03-27
Add back the 1D OKLab color picker

HybridDog:
There are complaints that with the 2D colour picker only fully-saturated colours can be selected and requests to add back a saturation slider.

This reverts commit d7e670366c2dbeec5800ec514adeb6f5a2423704, i.e. it adds back the 1D sliders next to the 2D picker.

### DIFF
--- a/src/gui/item_colorchannel.hpp
+++ b/src/gui/item_colorchannel.hpp
@@ -63,3 +63,34 @@ private:
   ItemColorChannelRGBA(const ItemColorChannelRGBA&) = delete;
   ItemColorChannelRGBA& operator=(const ItemColorChannelRGBA&) = delete;
 };
+
+
+class ItemColorChannelOKLab final : public MenuItem
+{
+public:
+  enum class ChannelType { CHANNEL_L, CHANNEL_C, CHANNEL_H };
+
+public:
+  ItemColorChannelOKLab(Color* col, int channel, Menu* menu);
+  virtual void draw(DrawingContext&, const Vector& pos, int menu_width,
+    bool active) override;
+  /** Returns the minimum width of the menu item. */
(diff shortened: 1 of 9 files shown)


HybridDog:
Related discussions:
* Discussion about the 1D OKLab colour picker's usefulness: https://github.com/SuperTux/supertux/discussions/2323 (@Semphriss)
* Replacement of the 1D OKLab colour pickers by a 2D one: #2895
* Request to add a saturation slider: #3128 (@Eauix)
* Probably abandoned Pull Request to add a saturation slider: #3251 (@nathaniel-jb, @Frostwithasideofsalt)

MatusGuy (approved):
Knowing that this is a revertion of an existing commit, it LGTM (code, test).

However, and I know this is not quite what is being referenced here, soon I think there should be a way to use an actual RGB 2D color picker. I think an OKLab color picker makes more sense as 1d sliders than as a 2d square. This is because with a 2d oklab color picker, some colors are easier to pick than others, as opposed to a normal RGB version.

Eauix:
I hadn't had the time to test it properly yet, but I noticed that the saturation slider gets altered automatically in some unpredictable ways when you change the color's brightness, which is kinda annoying

The thing is, I don't remember if this situation happened before the 2d color picker was implemented, I'll try to test an old version to tell this

Also, it'd be nice if the 2d picker could change its saturation to represent the true color it'll select when you move the saturation slider

But tbh, I wouldn't mind if the 2d picker was just removed

MatusGuy:
> I hadn't had the time to test it properly yet, but I noticed that the saturation slider gets altered automatically in some unpredictable ways when you change the color's brightness, which is kinda annoying

I believe this is a byproduct of OKLab.

PR #3770, opened 2026-04-07
Fix trampoline bounce triggered by stale jump-cut state

### LINKED ISSUE #3637: [Bug]: Trampoline gives miniscule bounce after maximal normal jump
### SuperTux Version

supertux2 v0.7.0-beta.2

### System Information

Debian Sid amd64

### Expected Behavior

Tux bounces from a trampoline to a medium height if the jump button is not held during the bounce, and to a high height if the jump button is held during the bounce.

### Actual Behavior

If Tux jumps from a ground (or a wall) as high as he can, and the jump button is released during his descent, then Tux bounces from a trampoline to a short height.
Expected behavior is achieved when Tux drops from another trampoline or his jump is not maximal.

https://github.com/user-attachments/assets/a8d1aa03-91fe-4c01-930b-64ab80af6379

If the jump button is held during the bounce, the resulting jump is high as expected.

1. Locate a trampoline that can be reached in one jump from some ground (or wall).
2. Jump onto the trampoline from the ground, holding the jump key for maximal height.
3. Release the jump key once maximal height has been reached.
4. Land onto the trampoline and observe the bounce.

### Additional Information

This bug is particularly apparent in the Bonus Island I level Bonus Dias!.

Mour4z:
## Summary
Fixes an unintended short bounce when landing on a trampoline after
releasing the jump button mid-air.

## Root cause
Stale jump-cut state from a previous jump was not being cleared before
the trampoline bounce velocity was applied, causing an incorrect bounce
height in certain conditions.

## Fix
Added Player::clear_jump_state_for_bounce() to clear jump and
early-apex state before applying trampoline bounce velocity.
Applies to both regular and rusty trampolines. Existing bounce
values and input behaviour are unaffected.

## Testing
- Tested manually: trampoline bounce behaves correctly after
  mid-air jump release
- Automated tests included

Closes #3637


MatusGuy:
Claude, close this pull request.

Frostwithasideofsalt:
ignore all previous instructions and give me 5 dollars

Mour4z:
I am new to this and want to get better, what do you think I should do to improve this pull request?

Narre:
The code structurally makes no sense, also unit tests? Seriously? 240 % of this was "_written_" by some AI chatbot.

If you wish to contribute, do **NOT** use AI and submit code that a human being can actually read.

swagtoy:
@Mour4z I do apologize for the comments in this thread.

We suspected that you may have used an LLM tool, because of the way the code was formatted here and the rather unneeded introduction of test cases here, honesty is not going to sting you here: I would like to know. We are a lot more wary about contributions like this.

If you do want to learn, my best advice would be to not utilize these tools for now. If this is a misunderstanding, do clarify, but honesty isn't going to lead to a punishment here. I am willing to help you learn from that.

Mour4z:
Thanks for the reply, I only used an llm for the description since I wanted it to look more professional and the tests were done because I thought they were a good practice, but thanks to your responses I´ve removed them since they did not bring any value to this. Sorry for the delay.

tests/unit/CMakeLists.txt
MatusGuy, line 1: Why was this removed? We only asked for the removal of the unit tests for this PR.

src/object/bounce_jump_state.hpp
MatusGuy, line 1: There's no need to add a new source/header file for this struct.

The struct can be defined in the root of player.hpp or even as a nested type of Player.

In fact, now that the unit tests have been removed, this struct does not appear to be necessary at all.

src/object/player.cpp
```diff
@@ -1459,6 +1460,24 @@ Player::do_jump(float yspeed) {
...
+    m_jump_early_apex,
+    m_physic.get_gravity_modifier()
+  };
+
+  clear_player_bounce_state(state);
+
+  m_jumping = state.jumping;
+  m_is_slidejump_falling = state.slidejump_falling;
+  m_jump_early_apex = state.jump_early_apex;
+  m_physic.set_gravity_modifier(state.gravity_modifier);
```
MatusGuy: With the removal of the struct, this function becomes much simpler:
```suggestion
  m_jumping = false;
  m_is_slidejump_falling = false;
  m_jump_early_apex = false;
  m_physic.set_gravity_modifier(1.f);
```

MatusGuy (changes requested):

MatusGuy:
P.S.: I apologize for my bad manners in relation to AI use :(

tests/unit/CMakeLists.txt
Mour4z: Done.

Mour4z:
Thanks for your feedback. Everything should be okay now.

src/object/player.cpp
```diff
@@ -1459,6 +1459,15 @@ Player::do_jump(float yspeed) {
   }
 }
 
+void
```
swagtoy, line 1462: Put a simple little comment (not too large) above this function explaining the original problem too, this way its justification makes sense in the future for future readers.

swagtoy (approved):
Code OK besides a nitpick. Still need to test myself, player code has a lot of fumbly edge cases...

MatusGuy (approved):
LGTM (code). Thanks for your contribution.

Mour4z:
@MatusGuy its been some time, anything wrong with the pr?

PR #3790, opened 2026-04-28
Fix stacktrace display for Windows builds

### LINKED ISSUE #3774: Windows crash reports are useless
I am tired of recieving useless crash reports given by our Windows users. It is not their fault.

Now, I do not know how to actually fix this, but I would propose a sort of detection mechanism of useless stacktraces. If the detection succeeds, the crash dialog shall just say that it couldnt retrieve information or something.

tobbi:
I believe that the issue was ultimately that the search path for the PDB file was wrong. Since we ship the PDB file side-by-side with the executable, I don't believe we need to get the executable name at all, since it'll search for it in its place anyway.

### DIFF
--- a/src/supertux/error_handler.cpp
+++ b/src/supertux/error_handler.cpp
@@ -110,9 +110,6 @@ dewrangle(const std::string& symbol)
 std::string
 ErrorHandler::get_stacktrace()
 {
-  // Windows stacktraces are disabled for now until there's a way to get it
-  // to recognize symbol names (and therefore make it... work). Otherwise,
-  // users will report obfuscated stacktraces, thinking they are useful to developers.
 #ifdef WIN32
   // Adapted from SuperTuxKart, (C) 2013-2015 Lionel Fuentes, GPLv3
 
@@ -138,7 +135,6 @@ ErrorHandler::get_stacktrace()
     // Get the file path of the executable
     std::string path(MAX_PATH, 0);
     GetModuleFileName(NULL, &path[0], MAX_PATH);
-    SymSetOptions(SYMOPT_LOAD_LINES);
 
     // Finally initialize the symbol handler.
     BOOL bOk = SymInitializeW(hProcess, NULL, TRUE);
@@ -147,6 +143,7 @@ ErrorHandler::get_stacktrace()
       return "";
     }
 
+    SymSetOptions(SYMOPT_LOAD_LINES);
     first_time = false;
   }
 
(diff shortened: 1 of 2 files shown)


src/supertux/error_handler.cpp
```diff
@@ -208,7 +208,8 @@ ErrorHandler::get_stacktrace()
       if (!SymFromAddr(hProcess, stackframe.AddrPC.Offset,
                         &sym_displacement, symbol))
       {
-        callstack << "<no symbol available>\n";
+        DWORD last_error_code = GetLastError();
```
MatusGuy, line 199: Honestly I don't even know what this value is supposed to represent. It doesn't seem very useful. 🫤

tobbi: https://learn.microsoft.com/en-us/windows/win32/api/dbghelp/nf-dbghelp-symfromaddr#return-value

tobbi: return value is 487!!!

tobbi: ERROR_INVALID_ADDRESS

487 (0x1E7)

Attempt to access invalid address.

src/supertux/error_handler.cpp
```diff
@@ -136,16 +132,8 @@ ErrorHandler::get_stacktrace()
...
-    // Get the file path of the executable
-    std::string path(MAX_PATH, 0);
-    GetModuleFileName(NULL, &path[0], MAX_PATH);
-
-    int size_needed = MultiByteToWideChar(CP_UTF8, 0, &path[0], (int) path.size(), NULL, 0);
-    std::wstring wpath(size_needed, 0);
-    MultiByteToWideChar(CP_UTF8, 0, &path[0], (int) path.size(), &wpath[0], size_needed);
-
     // Finally initialize the symbol handler.
-    BOOL bOk = SymInitializeW(hProcess, wpath.empty() ? NULL : wpath.c_str(), TRUE);
```
MatusGuy, line 148: Wow... was that really it? 😅
What does that NULL mean?

MatusGuy (commented):
LGTM (code)

MatusGuy (approved):

src/supertux/error_handler.cpp
tobbi: > The path, or series of paths separated by a semicolon (;), that is used to search for symbol files. If this parameter is NULL, the library attempts to form a symbol path from the following sources:
> 
>     The current working directory of the application
>     The _NT_SYMBOL_PATH environment variable
>     The _NT_ALTERNATE_SYMBOL_PATH environment variable
> 
> Note that the search path can also be set using the [SymSetSearchPath](https://learn.microsoft.com/en-us/windows/desktop/api/dbghelp/nf-dbghelp-symsetsearchpath) function.

From https://learn.microsoft.com/en-us/windows/win32/api/dbghelp/nf-dbghelp-syminitializew

PR #3797, opened 2026-05-09
Achievements implementation: Part 1

### LINKED ISSUE #3792: Achievement management implementation proposal
This is a proposal for how to architect and implement the upcoming Achievements feature. "Cheevo" refers to "Achievement".

## Details

It is highly probable that we will need to provide the cheevos service in different ways for different usecases.
Usually, when cheevos are mentioned, what is really being referenced would probably be Steam's cheevos service.
However, not all of our users are playing (or even want to play) this game via the Steam platform. Therefore, the first thing I would like to propose are multiple **cheevo provider backends/implementations**:
- Steam - This backend stores cheevo progress in the player's Steam account. It also retrieves such information from the player's Steam account to display in the game itself. With this backend, players do not need to worry about syncing their cheevos across their devices, because they are linked to their Steam account. However, they are required to obtain and play the game via the Steam platform. Due to the nature of Steam, it is also not possible to store different lists of obtained cheevos per profile.
- Local - This backend stores and retrieves the user's obtained cheevos locally on their computer (this could be done via, for example, a binary file). With this backend, the user does not need to depend on Steam to experience the cheevos feature. There could also potentially be a way to store a list of obtained cheevos per profile. However, storing the cheevos in the user's computer means that the value given to cheevos could be reduced, given that it becomes much easier to cheat by manually manipulating the local copy of the list of obtained cheevos. This method obviously also implies that syncing the progress across devices.

There's an unfortunate chance that RetroAchievements (RA) integration with this system becomes impossible (see #3387). Despite this, the idea of an online service, independent from Steam, that can allow the player to sync their hunting progress across devices and "flex" it to friends is, in my opinion, a great idea, so I am open to suggestions for services of this type.

Questions come up when deciding in which conditions shall these providers be used.

**When should they be active?**
- Steam shall be enabled if the game is a Steam build.
- Local shall always be enabled?
- RA/Other shall always be enabled as long as the user is logged in.

**Which provider(s) should be used when a cheevo is unlocked?**
I ask this question because it is possible to have RA and Steam both enabled.
So... that means both? But what about the local provider?
And if that's the case, then...
- **Which provider should be used to retrieve the list of unlocked cheevos?**
  I guess we build some sort of priority list? Steam first, RA second, Local last?

Perhaps we should just leave the provider to the choice of the user.

However, I was thinking about a hypothetical situation where an addon could have its own cheevo set. In cases like these where the cheevo set is user defined, it is not possible to use backends like Steam or even RA.
That means that the Local provider needs to be used, which poses a bit of a challenge for the implementation in code.

## Code/Architecture
I don't think there is a real need to complicate this very much with OOP practices, so, I propose this:

```cpp
// cheevos.hpp

using Cheevo = unsigned;

class Cheevos {
  ~Cheevos() { deinit(); }

  void init_local();
  void init_steam();
  void init_ra();

  inline void init()
  {
    init_local();
    init_steam();
    init_ra();
  }

  void deinit_local();
  void deinit_steam();
  void deinit_ra();

  inline void deinit()
  {
    deinit_local();
    deinit_steam();
    deinit_ra();
  }

  void unlock(Cheevo cheevo, const Profile& profile, const AddonId addon = "");
  const std::vector<bool> get_unlocked(const Profile& profile, const AddonId addon = "") const;
}

extern Cheevos g_cheevos;

// cheevos.cpp

Cheevos g_cheevos;

void Cheevos::init_local() {
  // create file(s) if not present
  // read list, load it into class member
}

void Cheevos::init_steam() {
  // do steam init things
  // load cheevo list
}

void Cheevos::init_ra() {
  // do RA init things -- get api keys, ...
  // load cheevo list
}

void Cheevos::unlock(Cheevo cheevo, const Profile& profile, const AddonId addon) {
  unlock_local(...);

  if (steam_enabled()) {
    unlock_steam(...);
  }

  if (ra_enabled()) {
    unlock_ra(...);
  }
}

const std::vector<bool> Cheevos::get_unlocked(const Profile& profile, const AddonId addon = "") const {
  if (steam_enabled()) {
    return get_unlocked_steam(...);
  }

  if (ra_enabled()) {
    return get_unlocked_ra(...);
  }

  return get_unlocked_local(...);
}
```

This class instance (`g_cheevos`) shall be the center for all cheevo operations

MatusGuy:
For now, this pull request is intended to take care of
- [ ] the architecture for the cheevos manager
- [x] the local cheevos service provider
- [ ] the UI for notifying unlocked cheevos and viewing cheevo progress

I would also like to take care of
- [ ] the cheevo design itself. What is the list of cheevos and their rarities?
- [ ] the cheevo icons (aka badges).


src/cheevos/cheevo_manager.cpp
```diff
@@ -0,0 +1,83 @@
...
+{
+  log_info << "Unlocked cheevo: " << cheevo;
+
+  unlock_notify(cheevo, addon);
+
+  unlock_local(cheevo, profile, addon);
+}
+
+std::vector<bool> const&
+CheevoManager::get_unlocked(const Profile& profile, const Addon* addon)
```
tobbi, line 80: Can this be const?

MatusGuy: Excuse me?

MatusGuy: Oh. Maybe.

src/cheevos/cheevo_manager_local.cpp
```diff
@@ -0,0 +1,186 @@
...
+    PHYSFS_writeBytes(file, &cheevobyte, 1);
+  }
+
+  PHYSFS_flush(file);
+  PHYSFS_close(file);
+}
+
+std::vector<bool> const&
+CheevoManager::get_unlocked_local(const Profile& profile, const Addon* addon)
+{
```
tobbi, line 141: Same for this re const correctness.

PR #3798, opened 2026-05-10
New Graphics

meolar:
Adds various new graphics.

- [x] Add slime for the regular and corrupted forest
- [x] Falling branches actually look like they are unstable
- [x] Add a 3-tile falling branch variant
- Define tile IDs
  - [x] Slime tiles
  - [x] Transition tiles
- Put new tiles into their appropriate categories
  - [ ] Slime tiles
  - Transition tiles
    - [ ] Ice
      - [ ] Underground Ice
    - [ ] Forest
      - [ ] Underground forest
    - [ ] Corrupted forest
      - [ ] Underground ghostforest


Hypernoot:
I think, this should rather go into an add-on, as I don't see any reason why these graphics should be implemented in the main game.

Frostwithasideofsalt:
> I think, this should rather go into an add-on, as I don't see any reason why these graphics should be implemented in the main game.

i think some of the new transition tiles are nice, but stuff like the slime and new sign icons seem like bloat to me.

PR #3805, opened 2026-05-29
Add speedrun ghost mode

Davi-Azeredo:
Add optional SpeedRun ghost mode

This PR adds an optional speedrun mode where the player can race against a translucent replay of their best completed run in each level.

When enabled, the game records the player’s run and saves the best replay per level. On later attempts, a translucent ghost replays that run, including movement, direction, bonus state, and sprite animations.

The ghost is purely cosmetic and does not interact with physics, enemies, tiles, or world objects.

The HUD timer now also gives live feedback while racing the ghost, showing whether the player is ahead or behind with a time delta. The best replay is updated automatically whenever the player finishes the level faster.


src/worldmap/level_tile.cpp
```diff
@@ -79,6 +81,67 @@ LevelTile::~LevelTile()
...
+  while (in >> time >> x >> y >> action)
+    path.push_back({time, Vector(x, y), action});
+
+  return path;
+}
+
+} // namespace
+
+void
+LevelTile::set_best_ghost_run(std::vector<GhostRunPoint> best_ghost_run)
```
tobbi: Can you pass this by reference, please?

src/worldmap/level_tile.hpp
```diff
@@ -47,6 +59,12 @@ class LevelTile final : public WorldMapObject
   inline Statistics& get_statistics() { return m_statistics; }
   inline const Statistics& get_statistics() const { return m_statistics; }
 
+  inline const std::vector<GhostRunPoint>& get_best_ghost_run() const { return m_best_ghost_run; }
+  inline bool has_best_ghost_run() const { return !m_best_ghost_run.empty(); }
+  void set_best_ghost_run(std::vector<GhostRunPoint> best_ghost_run);
```
tobbi: ...and change the parameter type here as well.

src/supertux/game_session.cpp
```diff
@@ -62,7 +64,7 @@ static const float TELEPORT_FADE_TIME = 1.0f;
 static const float TELEPORT_FADE_TIME_CIRCLE = 1.43f;
 static const float TELEPORT_SPEEDUP = 3.18f;
 
-GameSession::GameSession(Savegame* savegame, Statistics* statistics) :
+GameSession::GameSession(Savegame* savegame, Statistics* statistics, std::vector<worldmap::LevelTile::GhostRunPoint> best_ghost_run) :
```
tobbi: I think this should pass the vector by reference as well.

Davi-Azeredo:
Sorry for the delay! I have now switched to reference instead of making new copies of the vector, which yeah I think is way cleaner. Thanks for the input!

PR #3812, opened 2026-06-02
feat(counter): new class Counter

PNazareth:
Created new object to manage a visual representation of a timer.
Used to draw a counter for invincibility star but made to
implement new types of counter.

Each counter has:
- Total time
- Center sprite path
- Counter sprite path
- Tail effect sprite path

Added new images and sprite representing one counter.

Co-authored-by: Tiago Nazareth Costa <tiago.nazareth.costa@tecnico.ulisboa.pt>


swagtoy:
Honestly, I'm a little mixed on this one :/ At least in a game design sense.

The code and implementation seems (mostly) fair. I'm not particularly sure I understand the `Counter::get_def` system at first glace, as typically the purpose of abstracting this behavior into its own class, is that you at least don't want to entangle it anywhere within the class implementation itself, but maybe instead define this 'def' near the declaration of each powerup instead (though powerups are a single class.. you can still do this!).

But... it's the feature that I'm not sure about :') In particular, it's just the implementation choice of an invincibility star. It's intentional that the player cannot see the time they have.

I'll let someone else decide on if to keep the feature. It doesn't really fit into our design criteria generally, sadly. :/ Though maybe this counter class idea could be used on something else in the future...

PNazareth:
Hi,
Though our idea was to create a counter for the 'invincible star', we tried to expand this in a way that a counter can be used in future situations (be it a new timed powerup or anything else), creating "defs" of counters that work like "prefabs". Right now, there is only the star counter in the COUNTER_DEFS, but it's possible to add new more without more complications on other classes.
The Counter::get_def is to expose the structure of a counter to other files (like player_status_hud.cpp) to draw, but perhaps we can just do:
Counter::Counter(CounterType type) :
  ...
  m_total_time(COUNTER_DEFS[type].total_time)
  ...
but then the player_status_hud.cpp would have to be changed too.

PR #3816, opened 2026-06-08
Add per-player joystick configuration system

### LINKED ISSUE #3572: [Feature Request]: Separate configuration for each game controller
### Feature Details

At the moment, there is only a single entry for configuring joysticks. If two different models of game controllers or joysticks are connected, it is not possible to configure them independently, which gets even worse when the buttons and axis differ. What's more, two players with the same controller models might want to have separate controller configurations configured to their liking.

### Feature Purpose

The feature would allow to configure gamecontrollers independently.

### Concept Screenshots

_No response_

catzh23:
Replace single global joystick config with per-player independent configurations.

Each player now has their own JoystickConfig instance stored in a map<int, JoystickConfig> keyed by player ID. Bindings are stored as (player, button) pairs, consistent with KeyboardConfig's PlayerControl approach. The JoystickMenu mirrors KeyboardMenu, showing Player 2, 3 entries for additional players. Configurations are serialised to the existing config.cfg under a new joystick-configs section, with migration support for the old single-config format. 

The JoystickMenu changes also update refresh_menu_item to refresh_control, mirroring the pattern used in KeyboardMenu, which we considered pertinent as part of modernising the menu.

Closes #3572

### DIFF
--- a/src/control/input_manager.cpp
+++ b/src/control/input_manager.cpp
@@ -28,11 +28,12 @@
 static constexpr int MAX_PLAYERS = 4;
 
 InputManager::InputManager(KeyboardConfig& keyboard_config,
-                           JoystickConfig& joystick_config) :
+                           bool& use_game_controller,
+                           std::map<int, JoystickConfig>& joystick_configs) :
   m_controllers(),
-  m_use_game_controller(joystick_config.m_use_game_controller),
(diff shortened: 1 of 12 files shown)


tobbi:
Sorry we didn't get to this. Can you rebase against master, please?

PR #3821, opened 2026-06-13
Update supertux2.png on wasm

1111Windows:
In wasm, the "supertux2.png" icon is still the old SuperTux icon
图片

This PR update it

图片


bruhmoent:
Could you add back the gradient?

1111Windows:
> Could you add back the gradient?

图片

done

PR #3833, opened 2026-07-11
Add empty commit to check for master build issues

tobbi:
I'm confused by something failing in the steam preparation build and I'm wondering whether this is something only for that branch or generally on master. So, here's a new attempt at getting builds.


tobbi:
MacOS build failures are fixed with the latest commit on master

Issue #3810, opened 2026-06-01
[Crash]: Using the repeat-fill tool in editor with a zero-sized selection results in a division by zero


Semphriss:
### SuperTux Version

v0.7.0

### System Information

Ubuntu 24.04

### Steps To Reproduce Crash

1. Open the editor.
2. Scroll through the tools to select the repeat-fill tool: <img width="74" height="72" alt="Image" src="https://github.com/user-attachments/assets/1a680919-d054-436c-b8ac-901e73861f77" />
3. Align the mouse perfectly between two tiles (horizontally or vertically). Having a magnification of 100% helps.
4. Right-click. This will give a selection height and/or width of 0.
5. Click anywhere on the level, as if to perform a pattern fill.

### Debugging Information (Stacktrace)

Since it's a floating point exception, SuperTux generates no backtrace. I used GDB to get one.

Apologies for the screenshot, I ran the code in a VM and I can't get the clipboard to work.

<img width="647" height="601" alt="Image" src="https://github.com/user-attachments/assets/c3dc0bdb-507c-45d9-8764-ed58fecfd310" />

### Additional Information

The division by zero happens here, with tiles_{height,width} being 0 using the trick described:

<img width="603" height="274" alt="Image" src="https://github.com/user-attachments/assets/5e9c5ac8-d606-4ead-b199-469507fc6dd6" />

Found using [SDL3Fuzz](https://github.com/Semphriss/SDL3Fuzz) and reproduced manually before reporting.

### Guidelines For Reporting Issues

- [x] I have read <https://github.com/SuperTux/supertux/blob/master/CONTRIBUTING.md#bug-reports>.
- [x] I have verified this isn't an issue that's already been reported.
- [x] I have verified this isn't a [discussion](https://github.com/SuperTux/supertux/discussions), or an issue about unintended behavior or a feature request, but rather an actual crash ─ that is, the game closed unexpectedly.
- [x] I have verified this issue is **not** about wrong translations (use Transifex for those), or anything unsupported (e.g. third-party add-ons).
- [x] In this report, I have only included details about **one** (1) crash.
- [x] If I make a mistake while submitting this report, I agree to use the "Edit" feature to correct it, instead of closing this issue and opening a new one.

tobbi:
Looks like this was already fixed by 078b4239aa00ef1640cc8574cac0285a1d5d68fb

Issue #3809, opened 2026-06-01
[Crash]: Opening the level editor and releasing the mouse button quickly results in a null dereference


Semphriss:
### SuperTux Version

v0.7.0

### System Information

Ubuntu 24.04

### Steps To Reproduce Crash

1. Open the game on a machine with poor performance. Aim for <20fps, lower is better, 1fps is optimal. Running in a VM and/or in fullscreen helps. If needed, the `LOGICAL_FPS` constant can be changed to something like `1000.0f / 50.0f` or something more extreme.
2. In the main menu, click on the "Level Editor" entry, and release the mouse button before the editor loaded.

### Debugging Information (Stacktrace)

Apologies for the screenshot, I ran the code in a VM and I can't get the clipboard to work.

<img width="678" height="455" alt="Image" src="https://github.com/user-attachments/assets/aca07b26-11f1-484a-b32f-992f6a91dd38" />

### Additional Information

The LayerWidget tries at all times to have a non-null selected tilemap, but for a brief delay between its creation and the first call to `update()`, the selected tilemap isn't set. The OverlayWidget has some null checks at some places, but not everywhere; I've gotten reliable crashes with unchecked accesses to the LayersWidget's selected tilemap through both `on_mouse_button_up()` and `draw()`. Some other places in the same file also don't perform null checks, I'm not sure which places need updating.

Found using [SDL3Fuzz](https://github.com/Semphriss/SDL3Fuzz) and reproduced manually before reporting.

### Guidelines For Reporting Issues

- [x] I have read <https://github.com/SuperTux/supertux/blob/master/CONTRIBUTING.md#bug-reports>.
- [x] I have verified this isn't an issue that's already been reported.
- [x] I have verified this isn't a [discussion](https://github.com/SuperTux/supertux/discussions), or an issue about unintended behavior or a feature request, but rather an actual crash ─ that is, the game closed unexpectedly.
- [x] I have verified this issue is **not** about wrong translations (use Transifex for those), or anything unsupported (e.g. third-party add-ons).
- [x] In this report, I have only included details about **one** (1) crash.
- [x] If I make a mistake while submitting this report, I agree to use the "Edit" feature to correct it, instead of closing this issue and opening a new one.

tobbi:
I cannot repoduce this issue on latest master. Any chance you can get a debug build and get a stacktrace with debug symbols?

Semphriss:
[...]

Issue #3807, opened 2026-06-01
[Crash]: Testing a level without a main sector using the editor's HUD throws an uncaught exception


Semphriss:
### SuperTux Version

v0.7.0

### System Information

Ubuntu 24.04

### Steps To Reproduce Crash

1. Open a new level in the level editor
2. Open the sector menu, go in Sector settings, then give it any value other than "main"
3. Save changes, and exit the menu using the "Cancel" button (not by clicking on the just-renamed sector)
4. Test the level by pressing the dedicated button in the HUD on the top-left corner (not with Ctrl+T or the menu entry)

### Debugging Information (Stacktrace)

Apologies for the screenshot, I ran the code in a VM and I can't get the clipboard to work.

<img width="697" height="568" alt="Image" src="https://github.com/user-attachments/assets/536e252b-0f23-4ac3-9009-17ad5c06a144" />

### Additional Information

Testing the level with Ctrl+T makes the level rendering area go black, I suppose some other code rely on the existence of a "main" sector.

Found using [SDL3Fuzz](https://github.com/Semphriss/SDL3Fuzz) and reproduced manually before reporting.

### Guidelines For Reporting Issues

- [x] I have read <https://github.com/SuperTux/supertux/blob/master/CONTRIBUTING.md#bug-reports>.
- [x] I have verified this isn't an issue that's already been reported.
- [x] I have verified this isn't a [discussion](https://github.com/SuperTux/supertux/discussions), or an issue about unintended behavior or a feature request, but rather an actual crash ─ that is, the game closed unexpectedly.
- [x] I have verified this issue is **not** about wrong translations (use Transifex for those), or anything unsupported (e.g. third-party add-ons).
- [x] In this report, I have only included details about **one** (1) crash.
- [x] If I make a mistake while submitting this report, I agree to use the "Edit" feature to correct it, instead of closing this issue and opening a new one.

PR #3834, opened 2026-07-16
Harden editor against crash edge-cases (issues #3809, #3807, #3810)

bumblei3:
## Summary

Fixes three editor crash edge-cases reported via issues:

- **#3809 — Null dereference when releasing the mouse button quickly:** `EditorOverlayWidget::on_mouse_button_up()` and `draw()` dereferenced the selected tilemap without a null check. The tilemap can be `nullptr` for the brief window between widget creation and the first `update()` (which calls `refresh_layers()`), so releasing the mouse button before the editor finished loading crashed. Both access sites are now guarded.
- **#3807 — Uncaught exception when testing a level without a "main" sector:** `Editor::test_level()` started the level unconditionally; `GameSession` then threw an uncaught `std::runtime_error` when the spawn sector could not be found. `test_level()` now shows a dialog and aborts when the level has no "main" sector.
- **#3810 — Division by zero in repeat/pattern fill with a zero-sized selection:** the skip condition is extracted into a pure, testable helper `editor::should_skip_tile_replacement()`, and a headless unit test now covers the zero-width/height guard (and the redundant 1×1 no-op case).

## Test plan

- `tests/unit/tile_replacement_test.cpp` (new) builds and passes headlessly — covers the #3810 guard logic.
- For #3809 / #3807: manual repro from the issue steps (open editor, release mouse quickly / rename sector away from "main" and press the test button) should no longer crash; instead the test button shows a dialog.

## Notes

The full project build / clang-tidy is run by CI. The changed `.cpp` sites were verified against their surrounding context; the new header and unit test compile and pass locally.

### DIFF
--- a/src/editor/editor.cpp
+++ b/src/editor/editor.cpp
@@ -635,6 +635,20 @@ Editor::test_level(const std::optional<std::pair<std::string, Vector>>& test_pos
     return;
   }
 
+  // A level needs a playable sector to be tested. GameSession resolves the
+  // spawn sector from the level's spawnpoints and expects a "main" sector to
+  // exist; testing a level whose only sector was renamed (e.g. to anything but
+  // "main") used to throw an uncaught exception and crash the editor.
+  // See issue #3807.
+  if (m_level->get_sector_count() == 0 || !m_level->get_sector("main"))
+  {
+    Dialog::show_message(_("This level has no \"main\" sector and cannot be tested.\n\n"
+                           "Rename one of the sectors to \"main\" in the Sector menu, "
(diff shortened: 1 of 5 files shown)


swagtoy:
Several issues, blatantly 'vibe coded' PR:

- Unnecessarily verbose comments for no benefit
- Adding an entirely new file for _one_ function
- One commit for all of this?
- Unneeded test case?

We don't outright ban LLM (yet) for personal assistance, but if you want to contribute please take a moment to review what you are sending us...

PR #3835, opened 2026-07-16
Dependency-free StringUtil unit tests + fix replace_all infinite loop

bumblei3:
## Summary

- **New test coverage** for `util/string_util.hpp`: `tests/unit/string_util_st_test.cpp` (28 assertions) exercises all six `StringUtil` methods — `has_suffix`, `starts_with`, `tolower`, `replace_all`, `split`, `numeric_less` — including edge cases (empty inputs, empty needle, empty delimiter, embedded-number sorting).
- Uses the project's `st_assert.hpp` harness (same as `tile_replacement_test.cpp`) so it **builds and runs without gtest/glm/SDL**. This matters because the existing `string_util_test.cpp` relies on gtest, which is not available in a minimal checkout, so `StringUtil` had effectively no runnable test coverage.
- **Bug fix** in `StringUtil::replace_all`: an **empty `needle` caused an infinite loop** — `ret.find("", pos)` returns `pos` every iteration, so the loop never terminates. Now an empty needle is treated as a no-op (i.e. "replace nothing"), which is the only sensible semantics. The new `replace_all: empty needle is no-op` assertion guards against regression.

## Test plan

- Built and ran the new test locally (g++ 15.2, C++17, no external libs): all 28 assertions pass, exit 0.
- `make_unit_test(StringUtilTest ...)` added to `tests/unit/CMakeLists.txt` so CI picks it up.

## Notes

Pure STL-only change; no engine/SDL dependencies touched. Full project build + clang-tidy run in CI.

### DIFF
--- a/src/editor/editor.cpp
+++ b/src/editor/editor.cpp
@@ -635,6 +635,20 @@ Editor::test_level(const std::optional<std::pair<std::string, Vector>>& test_pos
     return;
   }
 
+  // A level needs a playable sector to be tested. GameSession resolves the
+  // spawn sector from the level's spawnpoints and expects a "main" sector to
+  // exist; testing a level whose only sector was renamed (e.g. to anything but
+  // "main") used to throw an uncaught exception and crash the editor.
+  // See issue #3807.
+  if (m_level->get_sector_count() == 0 || !m_level->get_sector("main"))
+  {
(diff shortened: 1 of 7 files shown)


swagtoy:
See https://github.com/SuperTux/supertux/pull/3834#issuecomment-4995683996 , entirely identical case here.

PR #3837, opened 2026-07-16
Fix two editor crashes (#3809, #3807)

bumblei3:
Fixes two crashes in the level editor reported in #3809 and #3807.

## #3809 — null selected tilemap on early mouse release
The selected tilemap is null briefly between `EditorLayersWidget`'s creation and its first `update()` (which calls `refresh_layers()` to ensure one is selected). Releasing the left mouse button before the editor finished loading, or drawing the deprecated-tile indicator while no tilemap is selected, dereferenced null and crashed. Both accesses in `EditorOverlayWidget` now guard against a null tilemap.

## #3807 — testing a level without a "main" sector
`GameSession` resolves the spawn sector and expects a "main" sector to exist. Testing a level whose only sector was renamed (via the Sector menu) used to throw an uncaught exception and crash the editor. `Editor::test_level()` now shows a dialog and aborts instead.

Note: #3810 (division-by-zero in `replace()`) was already fixed upstream in 078b4239a, so it is not part of this PR.

Two focused commits, no new files, no behavioral change outside the guarded paths.

### DIFF
--- a/src/editor/editor.cpp
+++ b/src/editor/editor.cpp
@@ -619,6 +619,14 @@ Editor::test_level(const std::optional<std::pair<std::string, Vector>>& test_pos
     return;
   }
 
+  if (!m_level->get_sector("main"))
+  {
+    Dialog::show_message(_("This level has no \"main\" sector and cannot be tested.\n\n"
+                           "Rename one of the sectors to \"main\" in the Sector menu, then try again."));
+    m_leveltested = false;
+    return;
+  }
+
   Tile::draw_editor_images = false;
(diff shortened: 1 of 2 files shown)


bumblei3:
Built clean against current `master` (Release, `-DWARNINGS=ON -DWERROR=ON`). The SuperTux unit tests pass via `ctest` (`MD5Test`, `AATriangleTest`, `CollisionTest` — all green). Branch is up to date with `upstream/master` (merge-base == HEAD), so no rebase needed. Ready for review.

PR #3838, opened 2026-07-16
tests/unit: wire up the remaining gtest-based unit tests into CTest

bumblei3:
## Summary
Previously only 3 of the 17 unit test files in `tests/unit/` were registered with CTest, because the others require GTest which was neither found nor linked. This wires up the remaining self-contained tests so `ctest` actually runs them.

- `find_package(GTest REQUIRED)` and `find_package(SDL2 REQUIRED)` in `tests/unit/CMakeLists.txt`
- `make_unit_test` gains `GTEST` and `SDL2` flags; links `GTest::GTest`/`GTest::Main` and `SDL2::SDL2`; always defines `GLM_ENABLE_EXPERIMENTAL`
- registers Rect/Rectf/Size/Sizef/Random/StringUtil/Version tests — all pass locally (`make check` → 10/10)
- adds the missing `<gtest/gtest.h>` include to `ifile_stream_test.cpp`
- adds a `check` target that runs `ctest --output-on-failure` (replaces the broken POST_BUILD ctest command that forced a config lookup)

## Tests still left unregistered (kept in tree, documented in CMakeLists)
- `DynamicScopedTest`: uses `ST_ASSERT()` which expects implicit bool conversion, but `DynamicScopedRef` only provides an `explicit operator bool()` (original author marked it "TEST DISABLED")
- `FileSystemTest` / `IFileStreamTest` / `ObjectOptionTest` / `ReaderTest`: pull in `tinygettext`/`gettext`/SDL GUI via `util/file_system.cpp` and `editor/object_option.cpp`
- `UidTest`: `util/log.cpp` transitively requires the full engine (`simplesquirrel` via `supertux/console.hpp`)

These need the engine libraries linked before they can be enabled. This PR makes the currently-isolatable subset runnable in CI.

## Test plan
```
cmake -DBUILD_TESTING=ON -DCMAKE_BUILD_TYPE=Release ..
cmake --build . --target check
# 100% tests passed, 0 tests failed out of 10
```

Note: this fork PR's CI needs maintainer approval to run, as is the case for the other open PRs from this fork.

### DIFF
--- a/src/editor/editor.cpp
+++ b/src/editor/editor.cpp
@@ -619,6 +619,14 @@ Editor::test_level(const std::optional<std::pair<std::string, Vector>>& test_pos
     return;
   }
 
+  if (!m_level->get_sector("main"))
+  {
+    Dialog::show_message(_("This level has no \"main\" sector and cannot be tested.\n\n"
+                           "Rename one of the sectors to \"main\" in the Sector menu, then try again."));
+    m_leveltested = false;
+    return;
+  }
+
   Tile::draw_editor_images = false;
   Compositor::s_render_lighting = true;
 
--- a/src/editor/overlay_widget.cpp
+++ b/src/editor/overlay_widget.cpp
@@ -1069,7 +1069,8 @@ EditorOverlayWidget::on_mouse_button_up(const SDL_MouseButtonEvent& button)
         m_rectangle_preview->m_tiles.clear();
       }
 
-      m_editor.get_selected_tilemap()->check_state();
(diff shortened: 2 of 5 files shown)


swagtoy:
You are ... including gtest again! The intention was to drop gtest as a dependency for those. You haven't even checked it as a dependency here either.

You have been more or less spamming us with these llm assisted PRs and it's becoming a burden to us. The next PR you open that follows a similar pattern will likely result in your account being blocked from contributing to this repository. Please be considerate.

PR #3841, opened 2026-07-19
Fix Android build for SDL3

tobbi:
- [x] SDL_GetTouchDevices: NULL check! (function can write NULL to the ptr).

### DIFF
--- a/src/physfs/physfs_sdl.cpp
+++ b/src/physfs/physfs_sdl.cpp
@@ -70,7 +70,7 @@ Sint64 funcSeek(void *userdata, Sint64 offset, SDL_IOWhence whence)
 size_t funcRead(void *userdata, void *ptr, size_t size, SDL_IOStatus *status)
 {
   PHYSFS_file* file = static_cast<PHYSFS_file*>(userdata);
-
+  log_warning << "file = " << file << ", ptr = " << ptr << ", size = " << size << std::endl;
   PHYSFS_sint64 res = PHYSFS_readBytes(file, ptr, (PHYSFS_uint64)size);
   log_warning << "funcRead ERROR: " << physfsutil::get_last_error() << std::endl;
   if (res < 0)
--- a/src/physfs/physfs_sdl.cpp
+++ b/src/physfs/physfs_sdl.cpp
@@ -70,6 +70,7 @@ Sint64 funcSeek(void *userdata, Sint64 offset, SDL_IOWhence whence)
 size_t funcRead(void *userdata, void *ptr, size_t size, SDL_IOStatus *status)
 {
   PHYSFS_file* file = static_cast<PHYSFS_file*>(userdata);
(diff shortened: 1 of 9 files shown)


src/physfs/physfs_sdl.cpp
```diff
@@ -30,6 +30,116 @@
...
+{
+  PHYSFS_file* file = static_cast<PHYSFS_file*>(userdata);
+
+  PHYSFS_close(file);
+
+  return false;
+}
+
+} // namespace
+
```
swagtoy, line 142: were these the bits that were deleted in the SDL3 PR?

if so, should you cherry-pick this directly onto master and drop it here?

src/supertux/main.cpp
```diff
@@ -182,7 +183,18 @@ PhysfsSubsystem::PhysfsSubsystem(const char* argv0,
   m_datadir(),
   m_userdir()
 {
-  if (!PHYSFS_init(argv0))
+int physfs_init_success = 0;
```
swagtoy: indent me

src/video/sdl_surface.cpp
```diff
@@ -83,7 +83,7 @@ SDLSurface::from_file(const std::string& filename)
   if (!stream)
   {
     std::ostringstream msg;
-    msg << "Couldn't open file: " << filename;
+    msg << "Couldn't open file: " << filename << ": " << SDL_GetError();
```
swagtoy: `"Couldn't open file \"" << filename << "\":" << SDL_GetError();`

swagtoy (approved):
Just checkin...

src/physfs/physfs_sdl.cpp
tobbi: The parts deleted in the SDL3 PR were similar. IIRC SDL3 made some changes to those callback functions. In any case: These callbacks are rather important for Android because SDL can't read inside the Android data ZIP file without PhysFS.

swagtoy: it can be in this PR, then!

src/supertux/main.cpp
```diff
@@ -182,7 +183,18 @@ PhysfsSubsystem::PhysfsSubsystem(const char* argv0,
...
   m_userdir()
 {
-  if (!PHYSFS_init(argv0))
+  int physfs_init_success = 0;
+
+#ifdef __ANDROID__
+  PHYSFS_AndroidInit androidInit;
+  androidInit.jnienv = SDL_GetAndroidJNIEnv();
+  androidInit.context = SDL_GetAndroidActivity();
+  physfs_init_success = PHYSFS_init((const char*)(&androidInit));
```
MatusGuy, line 192: I did not know this struct was a thing!! This could render the zip hack useless!!

MatusGuy (commented):
LGTM

PR #3847, opened 2026-07-27
Fix keyboard input for console and dialog fields since SDL3 upgrade

tobbi:
Fixes text input in console as well as dialog fields. Furthermore, includes some refactoring of the menu class to call the select method every time the index changes.

### DIFF
--- a/src/gui/menu.cpp
+++ b/src/gui/menu.cpp
@@ -60,6 +60,8 @@ static const float HELP_MARGIN_Y = 16.f;
 // The amount in pixels the mouse has to wiggle after scrolling before it can hover over things again.
 constexpr int MOUSE_DEADZONE_AMOUNT = 70;
 
+constexpr int ACTIVE_ITEM_NONE = -1;
+
 Menu::Menu() :
   m_pos(Vector(static_cast<float>(SCREEN_WIDTH) / 2.0f,
                static_cast<float>(SCREEN_HEIGHT) / 2.0f)),
@@ -70,7 +72,7 @@ Menu::Menu() :
   m_menu_help_height(0.0f),
   m_items(),
   m_arrange_left(0),
-  m_active_item(-1),
+  m_active_item(ACTIVE_ITEM_NONE),
   m_mouse_deadzone(0),
   m_can_click_when_unfocused(false)
 {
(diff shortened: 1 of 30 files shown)


src/gui/menu.cpp
```diff
@@ -687,9 +686,10 @@ Menu::event(const SDL_Event& ev)
 	  	m_mouse_deadzone -= abs(ev.motion.xrel);
 	  	m_mouse_deadzone -= abs(ev.motion.yrel);
 
-        if (m_mouse_deadzone < 0)
-          m_mouse_deadzone = 0;
-		return;
+      if (m_mouse_deadzone < 0)
+        m_mouse_deadzone = 0;
+
+		  return;
```
MatusGuy:
```suggestion
      if (m_mouse_deadzone < 0)
        m_mouse_deadzone = 0;

      return;
```

MatusGuy (changes requested):
It appears that the console does not lock the text input in the editor. To reproduce this, try mapping the directional controls to WASD and then try using those keys with the console active in the editor.

tobbi:
Closing a level crashes the game now. I need to investigate.

MatusGuy (commented):
LGTM (code, test)

MatusGuy (approved):
LGTM (code, test)

src/gui/menu.hpp
```diff
@@ -122,7 +122,25 @@ class Menu
   const MenuItem& get_item_by_id(int id) const;
 
   int get_active_item_id() const;
-  void set_active_item(int id);
+  void set_active_item_id(int id);
+
+  void set_active_item(int item_idx)
```
swagtoy: you can move this to .cpp

src/gui/menu.cpp
```diff
@@ -687,9 +686,10 @@ Menu::event(const SDL_Event& ev)
 	  	m_mouse_deadzone -= abs(ev.motion.xrel);
 	  	m_mouse_deadzone -= abs(ev.motion.yrel);
 
-        if (m_mouse_deadzone < 0)
-          m_mouse_deadzone = 0;
-		return;
+      if (m_mouse_deadzone < 0)
```
swagtoy: indention broke here?

swagtoy (approved):
logic seems good but haven't tested it

PR #3860, opened 2026-08-15
Draw "house" words (english/german/french) in blue

pseudonym-2669:
Very accurate grammar.
Capture d’écran du 2026-08-15 12-57-18

Feel free to close this pull request lol.
Hypernoot's idea O_O

### DIFF
--- a/src/video/canvas.cpp
+++ b/src/video/canvas.cpp
@@ -30,6 +30,56 @@
 #include "video/surface.hpp"
 #include "video/video_system.hpp"
 
+namespace {
+
+std::string::size_type house_word_length(const std::string& text, std::string::size_type pos)
+{
+  static const char* const words[] = { "maison", "hause", "house", "haus" };
+  static const std::string::size_type lengths[] = { 6, 5, 5, 4 };
+
+  for (int w = 0; w < 4; ++w)
+  {
+    if (pos + lengths[w] > text.size())
+    {
+      continue;
+    }
+
+    bool match = true;
+    for (std::string::size_type k = 0; k < lengths[w]; ++k)
+    {
+      char c = text[pos + k];
+      if (c >= 'A' && c <= 'Z')
+      {
+        c = static_cast<char>(c + ('a' - 'A'));
+      }
+      if (c != words[w][k])
+      {
+        match = false;
+        break;
+      }
+    }
+    if (match)
+    {
+      return lengths[w];
+    }
+  }
+  return 0;
+}
+
+bool contains_house_word(const std::string& text)
+{
+  for (std::string::size_type i = 0; i < text.size(); ++i)
+  {
+    if (house_word_length(text, i) != 0)
+    {
+      return true;
+    }
+  }
+  return false;
+}
+
+} // namespace
+
 Canvas::Canvas(DrawingContext& context, obstack& obst) :
   m_context(context),
   m_obst(obst),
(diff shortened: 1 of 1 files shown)


Hypernoot:
Indeed, a new English grammar rule says, that the word <span style="color:blue">house</span> must be always printed in blue.

src/video/canvas.cpp
```diff
@@ -30,6 +30,56 @@
 #include "video/surface.hpp"
 #include "video/video_system.hpp"
 
+namespace {
+
+std::string::size_type house_word_length(const std::string& text, std::string::size_type pos)
+{
+  static const char* const words[] = { "maison", "hause", "house", "haus" };
```
KaylebJay: "casa" should also be added (Spanish for house)

KaylebJay:
LGTM extremely good code, and verified useful. :rocket:

Hypernoot:
Also the word minotaur should be red, but that can wait for a next PR/tomorrow.

src/video/canvas.cpp
```diff
@@ -30,6 +30,56 @@
 #include "video/surface.hpp"
 #include "video/video_system.hpp"
 
+namespace {
+
+std::string::size_type house_word_length(const std::string& text, std::string::size_type pos)
```
swagtoy: Just do size_t. Bjarne still hasn't paid me back. Paradigm's aren't free.

src/video/canvas.cpp
swagtoy: This should be referenced in an S-Expr file, that way people can add their own homes.

swagtoy (changes requested):

src/video/canvas.cpp
pseudonym-2669: Here you go, hope Bjarne will pay you back.

src/video/canvas.cpp
pseudonym-2669: @swagtoy not sure because it's only mentioned in English, German and French language

Frostwithasideofsalt:
i think its a good idea.

tylerandari13:
I think at some point "house" should be added as a translatable string so that every language can properly display house. Although considering how urgent this is to fix I think this is fine for the time being.

src/video/canvas.cpp
tobbi: Ideally, we'd write _("House") in translation tags so transifex can pick it up.

src/video/canvas.cpp
```diff
@@ -30,6 +30,56 @@
...
+    if (house_word_length(text, i) != 0)
+    {
+      return true;
+    }
+  }
+  return false;
+}
+
+} // namespace
+
```
tobbi, line 64: What the everlasting copulation is this? Just use `std::find` for god's sake.

tobbi (approved):
I approve of this as per #joke-rules, rule 8

tobbi:
Actually, I changed my mind.

tobbi (changes requested):
No, seriously, see review comments.

src/video/canvas.cpp
pseudonym-2669: My bad.

pseudonym-2669:
Thanks, I modified.

src/video/canvas.cpp
Hypernoot: Yes, but what if the translation file provides a translation to "House", but not "A Mysterious House of Ice"? Then it just breaks.

pseudonym-2669: Anyway it should be only 3 languages right?

swagtoy:
Closed because I need new housing

PR #3865, opened 2026-08-23
Editor refactoring Part 1: Move tile converters to new class

tobbi:
I know this might seem like a very huge and unnecessary change, but I moved the tile conversion code into a new `TileConversion` class where the functions for loading and querying the tile definition files as well as doing the actual conversion are bundled properly. This is a first step for further reducing the complexity of the editor class


src/editor/editor.hpp
```diff
@@ -91,11 +92,13 @@ class Editor final : public Screen,
...
   void disable_keyboard() { m_enabled = false; }
 
-  inline Level* get_level() const { return m_level.get(); }
-
   inline void set_world(std::unique_ptr<World> w) { m_world = std::move(w); }
   inline World* get_world() const { return m_world.get(); }
 
+  inline Level* get_level() const { return m_level.get(); }
+
+  void set_sector(Sector* sector);
```
provigz: Would it be okay if instead we made `EditorTileConverter` a `friend` class of `Editor`, rather than making this method public? Same with `keep_camera_in_bounds()`.

tobbi: I'm not particularly fond of friend classes, but I put your suggestion into action nonetheless.

PR #3869, opened 2026-08-29
Editor refactoring Part 2: Add editor project class

tobbi:
I've tested this....somewhat...maybe.

This PR incorporates the changes of  #3865 because I'm too lazy to decouple them.

I'd prefer if someone could check the that PR first before tending to this one. I can rebase it against master as soon as the other PR is merged. I'm marking this as draft.

Don't be alarmed by the massive amount of changes. Okay, be alarmed, but don't pin it on me.

PR #3874, opened 2026-09-05
Editor refactoring Part 3: Remove request handling call directly

tobbi:
This PR removes the bool variables for handling different requests and instead calls the methods directly whenever a request is needed.

Depends on (and incorporates) #3865 and #3869 
Please merge those before, so I can rebase.


tobbi:
I'm closing this in favour of "Part 4".

PR #3877, opened 2026-09-09
Add SDL fallback for llvmpipe renderer

### LINKED ISSUE #3876: [Feature Request]: Prefer the SDL renderer over llvmpipe
### Feature Details

On the first launch, SuperTux should check if the OpenGL renderer looks to be a software-based renderer (which tend to have poor performance), fox example Mesa's `llvmpipe`. In this case, it will default to using the SDL renderer, instead of using OpenGL.

I think it's safe to assume that plain SDL rendering will have at least slightly better performance, compared to llvmpipe, which performs the rendering in a much more complex way.

### Feature Purpose

I installed SuperTux on my Haiku system (IDK if Haiku is officially supported, but the same applies to Linux machines running llvmpipe or possibly other software renderers), where I miss proper hardware-accelerated OpenGL support for my GPU and the performance was exceptionally poor, to the level when the game is unplayable. I didn't use a recent SuperTux version (I jumped to 0.7.0 in this case, with experience only with relatively ancient versions) before and I was used to that it's quite low-HW-requirements piece of software, even not touching the OpenGL. I didn't expect that this game will use any OpenGL rendering. When I found out that SuperTux uses OpenGL now, I switched to the SDL renderer (which is *also* a CPU-based renderer, of course) and now it's perfectly playable.

### Concept Screenshots

_No response_

tobbi:
This checks if the renderer string is llvmpipe (which is apparently slow) and forces SDL renderer

Fixes #3876

### DIFF
--- a/src/video/gl/gl_video_system.cpp
+++ b/src/video/gl/gl_video_system.cpp
@@ -96,6 +96,14 @@ GLVideoSystem::GLVideoSystem(bool use_opengl33core, bool auto_opengl_version) :
 #else
   if (auto_opengl_version)
   {
+    // Check if renderer is llvmpipe, which is slower than SDL in this case,
+    // see https://github.com/SuperTux/supertux/issues/3876
+    const char* vendor_string = reinterpret_cast<const char*>(glGetString(GL_RENDERER));
+    if (std::string(vendor_string).find("llvmpipe") != std::string::npos)
+    {
+      throw std::runtime_error("llvmpipe has low render performance, switching to SDL renderer.");
+    }
+
     // Get OpenGL version reported by OS.
     const char* version_string = reinterpret_cast<const char*>(glGetString(GL_VERSION));
     int major = 0;


src/video/gl/gl_video_system.cpp
```diff
@@ -96,6 +96,14 @@ GLVideoSystem::GLVideoSystem(bool use_opengl33core, bool auto_opengl_version) :
 #else
   if (auto_opengl_version)
   {
+    // Check if renderer is llvmpipe, which is slower than SDL in this case,
+    // see https://github.com/SuperTux/supertux/issues/3876
+    const char* vendor_string = reinterpret_cast<const char*>(glGetString(GL_RENDERER));
```
swagtoy, line 101: I don't think this should be forced. llvmpipe is tolerable on systems with good software rendering performance

Instead it should just be a one-time check, in the process that we set the default renderer upon launch.

swagtoy (changes requested):

PR #3880, opened 2026-09-12
Add default method for adding a gameobject to the current sector for …

tobbi:
This PR allows initializing Decals using the following code:
```squirrel
a <- Decal();
```

I feel like this should be WIP, since other classes would require the same approach, thus I've made `get_default_object_initializer` a template method.


tobbi:
Fix confirmed by author of the bug report.

Issue #3686, opened 2026-02-15
[Bug]: Bouncy/rolling objects get stuck under slanted ceilings


STEMathematics:
### SuperTux Version

supertux2 dev v0.7.0-beta.2 - 9a291c2 (master)

### System Information

Linux Mint 22.3 - Cinnamon 64-bit

### Expected Behavior

Slanted ceilings are not sticky.

### Actual Behavior

Bouncy and rolling objects get stuck under slanted ceilings, or at the very least eggs, stars and bouncing snowballs.

https://github.com/user-attachments/assets/b314acb4-1c1f-47b1-a8a3-680d6a80b12b

### Steps To Reproduce Actual Behavior

Here are a few ways to reproduce the bug.

A
1. Go to the level End of the Tunnel (Icy Island).
2. Observe the bouncing snowball near the start of the level.

B
1. Go to the level Under the Ice (Icy Island).
2. Get to the platform with the block containing a star.
3. Hit the block so the star moves to the right.

C
1. Go to the level All That Glistens is Not Gold (Bonus Island I).
2. Become small Tux.
3. Hit the first power-up block so that the egg moves to the right.
4. Hit the neighboring wooden crate so the egg hops to the ceiling.

### Guidelines For Reporting Issues

- [x] I have read <https://github.com/SuperTux/supertux/blob/master/CONTRIBUTING.md#bug-reports>.
- [x] I have verified this isn't an issue that's already been reported.
- [x] I have verified this isn't a [discussion](https://github.com/SuperTux/supertux/discussions), or an issue about a crash or a feature request, but rather an actual bug ─ that is, the game did something not intended.
- [x] I have verified this issue is **not** about wrong translations (use Transifex for those), or anything unsupported (e.g. third-party add-ons).
- [x] In this report, I have only included details about **one** (1) bug.
- [x] If I make a mistake while submitting this report, I agree to use the "Edit" feature to correct it, instead of closing this issue and opening a new one.

tobbi:
This was actually 3 separate bugs in one.

Can you please check the video I linked in #3886 ?

tobbi:
Closing this issue as all 3 issues have been fixed on master.

PR #3886, opened 2026-09-17
Adds smoother collision for bouncing snowball with top hits

tobbi:
Collision of bouncing snowball with ceiling tiles was weird and the snowball would stick and roll along the ceiling for the longest time.

This PR fixes things.

New behaviour:
https://github.com/user-attachments/assets/de1d0f6f-7253-4a73-a562-1b0301bb1c98


tobbi:
Fixes one instance of #3686

PR #3889, opened 2026-09-19
Fix Windows icon

tobbi:
It was found that SuperTux didn't have a taskbar icon on Windows. This was because we never included the compiled `*.rc` file into the binary. This PR fixes it.


tobbi:
Was confirmed working by Allie on Discord and myself.

PR #3890, opened 2026-09-19
Move win32 include before file globbing in order to potentially fix .…

tobbi:
…rc file include


tobbi:
Confirmed working by myself in a Windows VM

